PR #99, opened 2026-02-20
Tool calling fixes + test script

rossturk:
Streaming tool call filter, concurrency fixes, tool calling test suite:
- Add StreamingToolCallFilter to suppress tool-call markers from SSE content chunks while accumulating raw text for end-of-stream parsing
- Add index field to streaming tool_calls delta (OpenAI spec compliance)
- Fix ctrl_sub mutex deadlock in streaming.rs with timeout-based polling
- Fix InferenceService deadlock: spawn_local instead of await for continuations, LocalSet wrapper for single-threaded runtime
- Use sleep(Duration::ZERO) between tokens to drive tokio reactor
- Add call_tools.py test suite (15/16 pass, 3 xfail for tool_choice)

### DIFF
--- a/crates/hyprstream-rpc/src/streaming.rs
+++ b/crates/hyprstream-rpc/src/streaming.rs
@@ -1376,9 +1376,30 @@ impl StreamChannel {
         let tokens = Arc::clone(&self.cancel_tokens);
         tokio::spawn(async move {
             loop {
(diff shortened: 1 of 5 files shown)


ewindisch:
superseded, thanks for the PR

Issue #134, opened 2026-06-03
M2: Streaming plane — StreamService = moq-lite relay node


ewindisch:
Part of #131. (Supersedes the old 'in-process publisher' framing.)

A moq relay is just an `Origin` fed by upstream + served downstream — so StreamService becomes a **moq-lite relay node** built on moq-net the *library* (with our trust-store auth, NOT moq-relay's coarse `?jwt=` path-gate). This consolidates CDN-egress (#139), sidecar (#140), and cross-Origin-mesh (#142).

## M2a — in-process Origin + serve downstream
- `StreamService`/`EventService` hold a `moq_net::Origin`; in-process publishers append via Rust API (`authorize_signer` gates); external subscribers consume on `moql`.
- Port §7.5 chained-HMAC tokenstream (1:1 Track/Group/Frame); delete custom late-join rejoin (`streaming.rs:1553-1620`); collapse `notification.rs:436` pre-registration.
- PR #147 landed the wire layer (`IrohMoqProtocolHandler` + `OriginShared`); service refactor remaining.

## M2b — relay duplex + peering
- Enable the announce/subscribe duplex (the subscribe slot left `None` in #147) + optional upstream `Client`. This is the moq-relay protocol → wire-interchangeable with moq-relay + Cloudflare.
- **Consolidates**: #139 (egress = upstream peering), #140 (sidecar = deployment mode), #142 (mesh = relay peering).

## DH-topic = relay-mode-only
Direct peer↔peer streams get privacy from the transport (Track path can be plaintext); the DH-derived unguessable topic is needed only when egressing through an untrusted relay/CDN. PQ for that path = #153 (ML-KEM).

## CDN-portability invariants
moq-lite primitives only; standard ALPN; opaque hierarchical track names `{tenant}/{service}/{topic}/{instance}`; opaque frame payloads; respect upstream Group cache consts (#141).

Exit: tokenstream end-to-end (publisher → Origin → external subscriber); late-join native; authorize_signer enforced; (M2b) two StreamServices peer as a relay mesh.

ewindisch:
Security follow-ups from the whole-stack review: #161 (StreamRegister Classical-vs-Hybrid downgrade) and #163 (consumer Group-ordering + terminal-block enforcement) apply to this milestone's M2b consumer + control-plane signing.

ewindisch:
[...]

Issue #131, opened 2026-06-03
[EPIC] Replace ZMQ: moq-net streaming + pluggable key-dialed transports


ewindisch:
**Cross-repo epic.** Backend = this repo (GitHub). Frontend = `cyberdione/www-cyberdione-ai` (GitLab, MR !17 `amp` merged 2026-06-04).
**Master plan**: Obsidian `tasks/Epic - ZMQ to moq-net + iroh.md` (full architecture + rationale).

## Reframe (2026-06-04) — read first

The original plan made iroh the substrate. Investigation + spikes (#94, #95, #150) + an auth deep-review reframed it:

1. **Identity = key-dialing, PQ-layered.** Every peer is an Ed25519 key (onion addr / iroh NodeId / did:key / JWK = same 32 bytes). Rule: **dial Ed25519 (routing), trust composite Ed25519+ML-DSA (authz), encrypt ML-KEM-768 (confidentiality).** iroh is the clearnet sibling of the onion/Arti work.
2. **moq-lite is streams-only** → both planes (moq streaming + Cap'n Proto RPC) ride **any reliable-stream `web_transport_trait::Session`**: quinn, iroh, onion (Tor stream via qmux), WebRTC, WebSocket. Not QUIC-bound.
3. **iroh demoted from substrate to optional transport.** Its only unique value is clearnet NAT-traversal. **Default quinn (clearnet+browser), onion (key-rooted/anonymous), iroh optional (clearnet-behind-NAT).**
4. **UDP/TCP wall**: no moq/QUIC over Tor; run moq-lite over the Tor *stream* (qmux) instead.
5. **StreamService = a moq-lite relay node** (consolidates egress/sidecar/mesh).
6. **Browser can't use iroh** (spike #150: iroh = RFC 7250 RPK, browsers = X.509). Browser uses `@moq/net` → WebTransport → **quinn** (permanent endpoint).
7. **Auth**: OAuth(JOSE)+COSE composite-PQ — but COSE migration (#64) never wired and `pq-hybrid` is off by default. Fix = finish COSE as composite multi-sig + enable PQ (one task).

## Milestones

**M0 — Substrate + RPC over iroh (DONE / in-PR)**
- [x] #132 Phase-1 substrate — PR #143
- [x] #133 Phase-2 RPC plane (canary PolicyService) — PRs #144 / #145 / #146

**M1 — Transport abstraction (generalize off iroh)**
[...]

Issue #157, opened 2026-06-05
M3 deploy-blocker: mesh ML-DSA peer attestation (bind) + persistent mesh signing key


ewindisch:
Blocks production deploy of #152 (NOT the commit). From the M3 security re-review.

Under default Hybrid the mesh is **fail-closed to the point of non-functional**: `main.rs` installs an EMPTY `KeyedPqTrustStore`, no attestation layer `bind()`s peer Ed25519→ML-DSA mappings, and `LocalSigner::new` mints an **ephemeral per-process** ML-DSA key never published to peers. So every cross-node peer is an 'unknown anchor' → rejected (secure, but mesh dead unless `HYPRSTREAM_ENVELOPE_POLICY=classical`).

- Persist the mesh ML-DSA signing key alongside node identity (stable across restarts).
- Build the peer-attestation path that populates `KeyedPqTrustStore` (Ed25519 signer → ML-DSA verifying key), e.g. via the did:web/JWKS key bundle (#154) or discovery.
- Then default Hybrid is functional for multi-node.

Fail-closed today (safe), so commit is fine; this gates enabling Hybrid in a real cluster.

ewindisch:
Acceptance criteria surfaced by the #161 fix (StreamRegister now signs Hybrid like RPC):

1. **Wire the *single* node ML-DSA identity into BOTH `LocalSigner` (RPC plane) and `StreamChannel` (streaming plane) via `with_pq_key()`, and anchor its public key in the shared PQ trust store.** Today both planes independently auto-generate *ephemeral* ML-DSA keys in their respective `::new`, and `with_pq_key` has zero production callers — so under the daemon's default Hybrid policy both planes fail-closed (empty trust store). #157 must install the *same* persistent key into both; installing into only one silently fail-closes the other.

2. **Add a daemon-level integration test** that registers a stream and publishes through it under the production Hybrid policy. Current coverage (`streaming::tests::stream_register_hybrid_verifies_only_when_pq_anchored`, `envelope::tests::m3_*`) exercises the crypto in isolation; nothing yet proves the daemon installs a key the StreamService verifier accepts end-to-end across both planes.

ewindisch:
Dependency surfaced in PR #190 review: we currently have **two trust stores keyed by the same Ed25519 `VerifyingKey`** that can drift —

- `hyprstream_service`'s `TrustStore` (`VerifyingKey → Attestation` + scope) — consulted by the streaming **publish gate** and scope authz.
[...]

PR #190, opened 2026-06-10
M2a #134: moq-net Origin streaming plane for StreamService

ewindisch:
moq-net `Origin` streaming plane wired for StreamService (M2a). The substrate for the streaming-plane cutover.

📚 Stacked PR — **base: `ewindisch/moq-166`**. Part of epic #131.

### DIFF
--- a/crates/hyprstream/src/services/factories.rs
+++ b/crates/hyprstream/src/services/factories.rs
@@ -403,6 +403,26 @@ fn create_streams_service(ctx: &ServiceContext) -> anyhow::Result<Box<dyn Spawna
     // PULL backend - publishers connect via PUSH
     let pull_transport = global_registry().endpoint("streams", SocketKind::Push);
 
+    // In-process publish gate: only accept signers attested in the trust store.
+    // Shared by the ZMQ StreamRegister path and the moq in-process plane (M2a).
+    let gate = |pubkey: &[u8; 32]| -> bool {
+        use ed25519_dalek::VerifyingKey;
+        let Ok(vk) = VerifyingKey::from_bytes(pubkey) else {
+            return false;
+        };
+        hyprstream_service::global_trust_store().get(&vk).is_some()
+    };
+
+    // moq-lite streaming origin (epic #134 M2a). A standalone origin is built
+    // here so StreamService holds it and in-process publishers can append to it.
+    // When the iroh substrate is wired into bootstrap (M2b), this origin will be
+    // the substrate's shared `IrohMoqProtocolHandler` origin so external moq
+    // subscribers consume the same tree over the `moql` ALPN.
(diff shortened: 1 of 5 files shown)


crates/hyprstream/src/services/factories.rs
```diff
@@ -403,20 +403,35 @@ fn create_streams_service(ctx: &ServiceContext) -> anyhow::Result<Box<dyn Spawna
     // PULL backend - publishers connect via PUSH
     let pull_transport = global_registry().endpoint("streams", SocketKind::Push);
 
+    // In-process publish gate: only accept signers attested in the trust store.
+    // Shared by the ZMQ StreamRegister path and the moq in-process plane (M2a).
+    let gate = |pubkey: &[u8; 32]| -> bool {
+        use ed25519_dalek::VerifyingKey;
```
ewindisch, line 409: how does this fit in with our PQC hybrid designs?

ewindisch: Good question — they're **complementary layers, not the same check**, and the relationship is worth making explicit.

This `gate` is **identity/authorization**: "is this Ed25519 publisher *attested* and allowed to publish on this scope?" It consults `global_trust_store()` — a `TrustStore` keyed by Ed25519 `VerifyingKey → Attestation` (scope-based). It is **Ed25519-only**; it carries no ML-DSA material.

The **PQC hybrid** check lives at a different layer: the StreamRegister *envelope* is signed with the nested-COSE composite (EdDSA + ML-DSA-65) and verified in `process_request` against the **separate** `KeyedPqTrustStore` (`global_pq_store()`), which under Hybrid fail-closes if there's no kid-anchored ML-DSA key for the signer. So on the ZMQ StreamRegister wire path, a publisher must pass **both**: PQC envelope verify (authentic + post-quantum-bound) *and* this gate (attested identity authorized for the scope). For the **in-process** moq plane there's no envelope on the wire, so the gate is the sole admission — which is consistent with our in-process trust model (co-located = same trust domain, same reasoning as the inproc RPC dial).

The real smell your question surfaces: **we have two trust stores keyed the same way (Ed25519 vk) that can drift** — `TrustStore` (attestation+scope) and `KeyedPqTrustStore` (Ed25519→ML-DSA). Today fail-closed envelope verify protects correctness, but the dual registry is exactly what **#157** (mesh attestation + persistent ML-DSA) should unify: one attestation record binding Ed25519 ↔ ML-DSA ↔ scope, so the gate and the PQC verify resolve from the same source. I'll note this dependency on #157.

ewindisch: are you suggesting that rpc is only curve, and streams are hybrid or classical? We must also extend PQ to rpc... is this well covered by our other issues / PRs?

Issue #161, opened 2026-06-05
[CRITICAL][sec] StreamRegister signed Classical but verified Hybrid — streaming auth downgrade/break


ewindisch:
From the whole-stack security review (M2a↔M3 seam). **Blocker.**

`build_stream_register_envelope()` (streaming.rs:1882) signs with `SignedEnvelope::new_signed` = **Classical Ed25519-only**, but StreamService verifies under the global policy via `verify_any_signer_with(.., global_verify_policy())` (service/streaming.rs:298-302) = **Hybrid** in the daemon. Outcome: either (a) stream registration is **permanently broken** (Hybrid requires a kid-anchored ML-DSA outer the Classical sig lacks → fail-closed reject), or (b) the streaming control plane is a strict **downgrade** vs RPC (no SNS path for StreamRegister exists).

Fix: route StreamRegister through `new_signed_hybrid` (envelope.rs:1032) using the node's ML-DSA key, mirroring the RPC signer, so streaming key-exchange auth gets the same SNS enforcement as RPC.
Affects M2a (#134) + M3 (#152).

ewindisch:
Fixed on `ewindisch/moq-m3` (commit 795d0eb61).

**Fix:** `build_stream_register_envelope()` no longer hardcodes `new_signed` (Classical). `StreamChannel` now carries an ML-DSA-65 key (auto-generated in `::new`, mirroring `LocalSigner` on the RPC plane; `with_pq_key()` builder for the persistent identity), and StreamRegister signs **Hybrid** (EdDSA + ML-DSA SNS composite) when a PQ key is present — the same policy predicate as `rpc_client.rs`. Added a unit test proving a Hybrid register verifies only when the signer's ML-DSA key is anchored in the PQ trust store, and fail-closes otherwise.

**Code+security review (subagent):** CONFIRMED-GOOD on correctness (kid conventions match sign↔verify, SNS composite structurally equivalent to the RPC path), RPC-consistency, and completeness — the only other `*_signed` sites in streaming-adjacent code are `ResponseEnvelope` (server-response path, separate type/scope); cancel uses HMAC; subscribe is raw ZMQ SUB. No new regression: the RPC `LocalSigner` *also* auto-generates an ephemeral ML-DSA key today and `with_pq_key` has zero production callers, so **both** planes equally fail-closed under the daemon's Hybrid default until #157. This brings streaming to parity rather than introducing a streaming-specific break.

[...]

Issue #160, opened 2026-06-05
[CRITICAL][sec] global verify config fail-OPEN when uninstalled (re-opens M3 fail-open)


ewindisch:
From the whole-stack security review. **Blocker.**

`global_verify_policy()` (envelope.rs:846-851) returns **Classical** when `VERIFY_CONFIG` is unset — even though the intended default is Hybrid. Any subprocess-spawner service (systemd mode = separate processes), test, or early-init path that reaches a verify site before `install_verify_config` (main.rs:~2015) runs verifies **EdDSA-only**. This re-opens the M3 fail-open (#152) via a side door.

Fix: make `global_verify_policy()` return **Hybrid when unset** (fail-closed), OR hard-error if uninstalled in non-test builds. Audit every subprocess-spawner service path to confirm it calls `install_verify_config`. Add a `verify_config_installed()` assertion at verify sites in non-test builds.
Affects M3 (#152).

ewindisch:
Fixed on branch `ewindisch/moq-m3` (commit cdd7b5a2e).

**Fix:** `global_verify_policy()` now fail-closes to **Hybrid** when `VERIFY_CONFIG` is uninstalled in non-test builds (was Classical/fail-open). `cfg(test)` retains the Classical default since in-process unit tests share one `OnceLock` and use per-call `UnwrapOptions` overrides.

**Subprocess audit (the second half of this issue):** found a real gap — the standalone worker entrypoint (`main.rs:1098`) constructs an `InprocManager` and serves RPC but never installed the config. Extracted `install_envelope_verify_config()` and now call it at **both** RPC-serving entrypoints (inproc daemon + worker spawn). Other `InprocManager::new()` sites are `#[cfg(test)]`; iroh/QUIC `*ProtocolHandler` serving sites are test-only and reach verification through the same global policy.

Integration test `iroh_rpc_e2e` opts into Classical explicitly (it validates the classical wire path; compiled non-test it now sees the fail-closed default).

Code+security review (subagent): confirmed correct & complete, no production path under `cfg(test)`, both entrypoints covered. Per-request `warn!` gated behind `Once` to avoid log-flooding a misconfigured node.

Pending: push (needs security-key touch) as part of the M3 PR.

ewindisch:
[...]

Issue #152, opened 2026-06-04
M3: Hybrid envelope signing — nested-COSE SNS composite (EdDSA+ML-DSA), runtime policy


ewindisch:
Part of #131. Security-critical. **Re-scoped 2026-06-04** against IETF PQUIP (RFC 9794 + draft-ietf-pquip-hybrid-signature-spectrums) after a security review found the first cut was fail-open.

## Why re-scoped (PQUIP cross-reference)

The first implementation used **Concatenation** (two independent COSE_Sign entries) = **Weak Non-Separability (WNS)** per the PQUIP spectrums draft — security depends on *every* verifier enforcing "require both," which is "a policy requirement, not a cryptographic assurance." The security review found exactly the predicted failure: clients dual-signed but servers verified EdDSA-only (no `PqTrustStore` wired) → **strippable, fail-open**.

Key PQUIP findings applied:
- **SNS ⊥ backwards-compatibility**: a verifier that only checks one component can't get strong non-separability. But the only verifiers that *can't* do PQC are **external OAuth/JOSE JWT consumers** — our mesh peers and the browser all run **our WASM** (ml-dsa compiles to wasm32), so they ARE full hybrid verifiers. So we can take SNS on every surface we control with no compat loss.
- **Key-reuse restriction**: never publish the same key as both a standalone classical JWK and a hybrid component.

## Target design (per-surface)

| Surface | Scheme | Non-separability |
|---|---|---|
| **Mesh RPC envelopes, streaming envelopes, browser-via-our-WASM** | **Nested COSE**: EdDSA inner, ML-DSA-65 outer over `payload ‖ EdDSA-sig` | **SNS** — not strippable |
| **OAuth JWT → external JOSE verifiers only** | Concatenation + dual-JWK + RFC 7517 skip-unknown | WNS (backwards-compat, separate keys) |

- **Stay composite** (keep EdDSA component) — hedge against an ML-DSA *classical* break; don't go pure-PQ.
- **Runtime `CryptoPolicy {Classical|Hybrid}`**, default **Hybrid ENFORCED** (not compile-time feature).
- **kid-anchored**: ML-DSA key resolved from `PqTrustStore` (wire from `key_rotation::global_ml_dsa_verifying_keys`), never self-asserted. Fail-closed: Hybrid + no anchor ⇒ reject.
- **Separate keys** across mesh-hybrid vs external-classical-JWK.
[...]

PR #191, opened 2026-06-10
M3: nested-COSE hybrid envelope (#152) + fail-closed Hybrid (#160) + StreamRegister hybrid (#161)

ewindisch:
- **#152** nested-COSE hybrid envelope signing (EdDSA + ML-DSA-65) + enforced policy. **Relabeled SNS→WNS** per `draft-ietf-pquip-hybrid-signature-spectrums` §1.3.4 — downgrade resistance is policy-enforced (require_pq + kid-anchoring), not cryptographic non-separability. Follow-ups (AAD alg-id bind, redundant `sig` removal) tracked on #152.
- **#160** fail-closed global verify policy when uninstalled (closes the M3 fail-open).
- **#161** StreamRegister signs under Hybrid like RPC.

All reviewed (code + security). 📚 Stacked PR — **base: `ewindisch/moq-m2a`**. Part of epic #131.

### DIFF
--- a/crates/hyprstream-rpc/src/crypto/mod.rs
+++ b/crates/hyprstream-rpc/src/crypto/mod.rs
@@ -28,14 +28,41 @@
 
 pub mod backend;
 pub mod cose_sign1;
+pub mod cose_sign;
 pub mod envelope_crypto;
 pub mod event_crypto;
 pub mod hmac;
 pub mod key_exchange;
 pub mod notification;
-#[cfg(feature = "pq-hybrid")]
 pub mod pq;
 pub mod signing;
+
+/// Runtime crypto mode selecting how envelopes/tokens are signed and verified.
+///
+/// This REPLACES the old compile-time `pq-hybrid` cargo feature. PQ primitives
+/// (ML-DSA-65, ML-KEM-768) are always compiled; the policy chooses whether they
+/// are used at runtime.
+///
+/// - `Hybrid` (default): sign with a COSE composite (EdDSA + ML-DSA-65) and
+///   REQUIRE both components to verify.
+/// - `Classical`: sign with EdDSA only and accept EdDSA-only signatures. A
+///   `Classical` verifier still accepts a `Hybrid`-signed item via its EdDSA
+///   component (RFC 7517 skip-unknown interop).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum CryptoPolicy {
+    /// EdDSA only (classical).
+    Classical,
+    /// EdDSA + ML-DSA-65 composite (post-quantum hybrid). Default.
+    #[default]
+    Hybrid,
+}
+
+impl CryptoPolicy {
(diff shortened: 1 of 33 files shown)


crates/hyprstream-rpc/src/crypto/mod.rs
```diff
@@ -28,15 +28,42 @@
...
+/// (ML-DSA-65, ML-KEM-768) are always compiled; the policy chooses whether they
+/// are used at runtime.
+///
+/// - `Hybrid` (default): sign with a COSE composite (EdDSA + ML-DSA-65) and
+///   REQUIRE both components to verify.
+/// - `Classical`: sign with EdDSA only and accept EdDSA-only signatures. A
+///   `Classical` verifier still accepts a `Hybrid`-signed item via its EdDSA
+///   component (RFC 7517 skip-unknown interop).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum CryptoPolicy {
```
ewindisch, line 52: should we implement Cipher Suites rather than "Crypto Policy"? Give strengths, weaknesses, and recommendation.

ewindisch: Worth thinking through. Short version: **keep `CryptoPolicy` now; evolve toward an *authenticated suite identifier* (not TLS-style negotiated cipher suites) when algorithm agility becomes real.**

**Cipher suites (TLS-style registry of {sig, KEM, AEAD, hash} combos)**
- *Strengths*: real algorithm agility — add ML-DSA-87 / Falcon / swap ML-KEM params without touching policy code; a single wire-encodable id names a vetted combo; per-peer capability.
- *Weaknesses*: **negotiation is a downgrade-attack surface** (TLS's entire cipher-suite history is downgrade CVEs — FREAK, Logjam, ...); versioning + negotiation state machine; large test matrix; **YAGNI** for a 2-point space today.

**`CryptoPolicy` enum (current)**
- *Strengths*: simple; **fail-closed default**; **not negotiated** → no downgrade surface (policy is enforced at the verifier, peers don't haggle); fits the actual need (Classical transitional → Hybrid).
- *Weaknesses*: not algorithm-agile — a third primitive set means a new variant + code; it conflates "mode" with "which algorithms."

**Recommendation.** Don't build negotiated cipher suites — that imports TLS's downgrade problems for no current benefit. When we *do* need agility, introduce a **suite id that is bound into the signed material** (COSE protected header / the `external_aad`), so the chosen suite is *authenticated*, not negotiated — a verifier under fail-closed policy rejects anything not matching its required suite, with no handshake to downgrade. That's actually the same mechanism as the **#152 follow-up** (binding the hybrid-composite alg-id into the AAD): that hybrid-alg-id *is* the seed of a suite identifier. So the migration path is: `CryptoPolicy` (now) → authenticated suite-id in AAD (when we add a 2nd hybrid combo) → a small suite registry if it ever grows. Keep enforcement fail-closed at every step.

Issue #148, opened 2026-06-04
moq Phase 6.5: Extract transport-neutral survivors from zmtp_quic.rs


ewindisch:
Part of #131. **Blocks #138.** Blocked by #133 (so LocalServiceBridge re-points its import).

`zmtp_quic.rs` (2,417 LoC) entangles ZMQ with non-ZMQ code. Before Phase 7 can delete the ZMQ parts, extract the transport-neutral survivors into their own modules — pure moves, no behavior change.

## Extract

- **`process_request`** → `crates/hyprstream-rpc/src/service/dispatch.rs`. Envelope verify + JWT/DPoP + Casbin dispatch; transport-agnostic. Already imported by `iroh_rpc::LocalServiceBridge` (Phase 2) and by `handle_wt_stream` (browser). Update both import paths.
- **`WebTransportServer`** + `handle_webtransport_session` + `handle_wt_stream` → `crates/hyprstream-rpc/src/transport/web_transport_server.rs`. The live browser MCP/RPC surface (quinn H3).
- **`handle_http_request`** `.well-known/oauth-protected-resource`, `tui-cert-hash`, `/health` → same module.
- **TLS helpers** `webtransport_tls_config`, `cert_hash`, self-signed/pinned → `crates/hyprstream-rpc/src/transport/tls.rs`.

## Exit criterion

- All survivors moved; `zmtp_quic.rs` now contains only ZMQ/ZMTP code (ready for #138 deletion).
- `cargo build --release -p hyprstream` + `-p hyprstream-rpc --tests` clean.
- No behavior change: browser MCP + iroh RPC plane tests still pass.

ewindisch:
**RESEQUENCED EARLY (consolidated plan).** `process_request` + WebTransport/TLS survivors are the **shared RPC-dispatch core the production cutover needs** — extract them up front, not at teardown. New blocker edges: **#148 blocks #136 (cutover) and #138 (teardown)**; #148 is blocked by **N1 (rename)** only. Also de-iroh the generic core during the move (`IrohRequestProcessor`→`RequestProcessor`, `iroh_rpc`→`dispatch`).

ewindisch:
**Part 1 done** on `ewindisch/moq-148` (commit 829e774fc, stacked on moq-178).

[...]

PR #194, opened 2026-06-10
#148 part 1: extract process_request → service/dispatch.rs (transport-neutral dispatch core)

ewindisch:
Extract `process_request` into `service/dispatch.rs` — the transport-neutral envelope-verify → JWT/DPoP → Casbin → handle_request → signed-response core shared by every front-end. All callers repointed; `service/zmq.rs` drops its last `zmtp_quic` dep.

Note: the rest of #148 (WebTransportServer + TLS helpers) is FE-2-gated (handle_wt_stream still uses ZMTP framing) and deferred. 261 rpc tests green.

📚 Stacked PR — **base: `ewindisch/moq-178`**. Part of epic #131.

### DIFF
--- a/crates/hyprstream-rpc/src/service/dispatch.rs
+++ b/crates/hyprstream-rpc/src/service/dispatch.rs
@@ -1,0 +1,132 @@
+//! Transport-neutral request dispatch core (#148).
+//!
+//! `process_request` is the single envelope-verify → JWT/DPoP → Casbin →
+//! `handle_request` → signed-response pipeline shared by every front-end:
+//! the ZMQ `RequestLoop`, the WebTransport server, and the generic RPC plane's
+//! `LocalServiceBridge`. It contains no transport-specific code (no ZMQ, no
+//! quinn) — extracted out of `transport/zmtp_quic.rs` so the ZMQ-specific
+//! remainder of that file can be deleted in #138 without touching this.
+
+use anyhow::Result;
+use tracing::{debug, error, warn};
+
+/// Envelope signer verification mode (re-exported from `crate::envelope`).
+///
+/// - **FixedSigner**: internal service-to-service — envelope signer must match
+///   a known `server_pubkey` for mutual authentication. Peers pre-share keys.
+/// - **AnySigner**: external clients (e.g. browser over WebTransport) — any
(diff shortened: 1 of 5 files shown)


crates/hyprstream-rpc/src/service/dispatch.rs
```diff
@@ -0,0 +1,132 @@
...
+    raw_bytes: &[u8],
+    service: &S,
+    verification: EnvelopeVerification<'_>,
+    signing_key: &ed25519_dalek::SigningKey,
+    nonce_cache: &crate::envelope::InMemoryNonceCache,
+) -> Result<(Vec<u8>, Option<crate::service::Continuation>)>
+where
+    S: crate::service::RequestService,
+{
+    use crate::ToCapnp;
```
ewindisch: is this the appropriate place for 'use' -- why is crate even allowing this in clippy?

ewindisch: Two parts:

**Why clippy "allows" it:** there is no lint for function-scoped `use` — it's legal, idiomatic Rust (sometimes used deliberately to scope an import to one fn). The import-related clippy lints we deny (`wildcard_imports`, etc.) only fire on glob `use foo::*`; none of these are globs, so nothing flags them. So it's not clippy being lax — there's simply no rule here to enforce.

**Is it the appropriate place:** not really — these four (`ToCapnp`, `ResponseEnvelope`, `Builder`, `serialize`) are general and used throughout the fn, so module-top would be cleaner and more discoverable. They're function-local only because #148 part 1 extracted `process_request` **verbatim** out of `zmtp_quic.rs` to minimize churn in that commit. Hoisting them to module scope is a trivial, zero-risk cleanup — I'll do it on this branch unless you'd rather keep the extraction diff minimal.

ewindisch: move it, and create an issue to add the clippy rule.

ewindisch: Done — hoisted the four `use` statements (`ToCapnp`, `ResponseEnvelope`, capnp `Builder`/`serialize`) to module scope on `moq-148`. Pure cleanup, no behaviour change; the children (#195/#196) were restacked onto it.

Issue #186, opened 2026-06-10
M1: Kill Continuation — streaming bootstrap via RPC-reply StreamInfo + moq-subscribe (prereq for #136)


ewindisch:
Split out of #136 per the 2026-06-10 review (it's a trait-signature change across the ~21 streaming call-sites — independently reviewable, should not be buried in the cutover).

`RequestService::handle_request` returns `(Vec<u8>, Option<Continuation>)` (zmq.rs:300); the ZMQ `RequestLoop` runs the continuation *after* sending the REP to avoid a stream_id race. The new dispatch core (`service::dispatch::process_request`, and `IrohRequestProcessor` which returns only `Bytes`) **cannot carry a Continuation**. So the RPC cutover (#136) breaks all streaming-initiating calls unless this lands first.

**Replace** the continuation pattern with: the RPC reply returns a signed `StreamInfo` (DH-topic + keys); the client then subscribes on the moq plane; moq announce/retention covers 'subscribe slightly late'. Remove `Continuation` from the trait signature + the 21 sites. **#136 blocked-by this.** Note: #148 just extracted `dispatch.rs` *with* the Continuation return type still present — this ticket removes it.

ewindisch:
**Implemented (M1 scope) on local branch `moq-186` (off `moq-148`); reviewed + fixed.**

`process_request` now returns `Result<Vec<u8>>` and spawns the streaming pump itself (`streaming::spawn_stream_pump`), so all four transport front-ends — ZMQ `RequestLoop`, both WebTransport sites, and `LocalServiceBridge` — are uniform bytes-in/bytes-out. **The generic RPC plane no longer rejects streaming**, which is the actual `#136`-cutover blocker this ticket exists to clear.

Scope decision (M1 vs M2): this removes `Continuation` from the **dispatch boundary + every transport front-end**. `handle_request` keeps `Option<Continuation>` as the internal service→core contract, so the 16 service impls + the proc-macro codegen are **untouched** (low blast radius). The `Continuation` *type* itself dies in M2, when the pump's transport (`StreamChannel`) moves to a service-owned moq broadcast (#134) and the per-RPC spawn goes away entirely.

Code+security review (ship-with-fixes), all applied:
[...]

Issue #151, opened 2026-06-04
M1: Transport abstraction — planes over web_transport_trait::Session + quinn transport


ewindisch:
Part of #131. Generalize the RPC + streaming planes off iroh-specifically so quinn/onion/webrtc are first-class transports.

moq-lite is **streams-only** (proven: moq-rs ships a WebSocket transport; only data type is `Group` over `open_uni`) and the Cap'n Proto RPC plane is reliable-bidi-only — so both ride **any** `web_transport_trait::Session`.

- Re-parameterize `IrohRpcProtocolHandler`/`IrohTransport` + moq Origin wiring over `web_transport_trait::Session` (not `iroh::endpoint::Connection`).
- Add a **quinn `Session` transport** as the clearnet+browser baseline (reuses existing quinn/WebTransport infra; the endpoint that survives ZMQ teardown).
- iroh becomes an *optional* `Session` impl (clearnet-behind-NAT niche).
- (Follow) **onion `Session`**: Arti TCP stream via a qmux multiplexer — key-rooted/anonymous; carries moq-lite+capnp as reliable streams (no QUIC-over-Tor; UDP/TCP wall).

Exit: planes run over quinn AND iroh through the same trait; one integration test per transport.

ewindisch:
Scope note: iroh is **kept** as a retained optional transport (decision on #131). So #151 = generalize the planes over `web_transport_trait::Session` and add **quinn** as the default clearnet+browser transport, with iroh re-slotted as one optional impl (not removed) and onion to follow.

ewindisch:
M1 also requires: transport `Session` impls expose **drop-detection + re-dial-by-key**, and the RPC client owns a **retry policy** (ZMQ's transparent reconnect becomes explicit). Tracked as #.

ewindisch:
**Expanded (consolidated plan):** this ticket also **absorbs the `TransportConfig` overhaul** — replace `{Inproc, Ipc, Quic(ZMTP), SystemdFd}` with a transport-selection model over `web_transport_trait::Session`: **Quic (baseline) + optional Iroh + optional Tor + in-process**. 

[...]

Issue #162, opened 2026-06-05
[HIGH][sec] M1 transport: no per-server connection cap + oversized MAX_FRAME_BYTES


ewindisch:
Whole-stack security review (M1 #151). Neither `QuinnRpcServer::run` nor the iroh Router bounds concurrent **connections** (only streams-per-server). Attacker opens N connections → CPU (TLS handshakes) + memory; each accepted stream can buffer up to `MAX_FRAME_BYTES`=64 MiB in `read_to_cap`'s growing Vec. Fix: add a connection semaphore in the accept loop; lower `MAX_FRAME_BYTES` to a realistic control-plane ceiling (~1-4 MiB) — large blobs belong on the streaming ALPN per the module docs. Couples with #159.

ewindisch:
Scope additions surfaced by the #159 security review (all connection-level DoS, complementary to the stream-level slowloris fixed in #159):

1. **Unbounded WebTransport handshake serializes the accept loop (MEDIUM).** `QuinnRpcServer::run` resolves `request.ok().await` *inline* in the accept `select!` (quinn_transport.rs:~179), not in a spawned task. A single peer stalling the WebTransport CONNECT (after the QUIC handshake) blocks the accept loop from accepting any further connections — connection-level head-of-line DoS. Fix: wrap in a `HANDSHAKE_TIMEOUT` and/or move `ok().await` into the spawned task.
2. **Audit quinn endpoint `TransportConfig` for `max_idle_timeout`** so an established-but-silent connection holding zero streams is reaped — the per-stream read timeout (#159) does not cover an idle connection with no open streams.
3. **Connection cap (already this issue's core):** `DEFAULT_STREAM_LIMIT` bounds total in-flight *streams* server-wide, but nothing caps concurrent *connections*; a peer opening many idle connections (no streams) consumes fd/memory unbounded. This is the primary ask of #162.

These belong together as connection-level transport hardening.

ewindisch:
Core fixed on `ewindisch/moq-m1` (commit 51df7f1c9).

[...]

PR #195, opened 2026-06-10
#151(a): in-process transport + inproc registry + dial() factory

ewindisch:
The dialing layer (#151 follow-on): `dial(&TransportConfig, signer, vk) -> Arc<dyn RpcClient>` — the single place transport choice is made, erasing the non-object-safe `Transport` behind `RpcClient`. Adds `InMemoryTransport` (RPC over `IrohRequestProcessor`, reusing the exact sign/verify path — no second security model) and a process-local `Weak` inproc registry.

Inproc arm implemented; lazy quinn/iroh/moq transports + codegen de-leak are follow-on increments. Code+security reviewed (ship-with-fixes, applied: Weak registry, warn-on-overwrite, exhaustive match, corrected None-key doc). 9 new tests, 269 rpc green.

📚 Stacked PR — **base: `ewindisch/moq-148`** (sibling to #186). Part of epic #131.

### DIFF
--- a/crates/hyprstream-rpc/src/transport/in_memory.rs
+++ b/crates/hyprstream-rpc/src/transport/in_memory.rs
@@ -43,6 +43,12 @@ use crate::transport_traits::Transport;
 /// Default ceiling on a single in-process request, mirroring the networked
 /// RPC transport. A wedged co-located handler should surface as a timeout
 /// rather than hanging the caller indefinitely.
+///
+/// Note this deadline also covers *enqueue* latency: `LocalServiceBridge`
+/// forwards over a bounded mpsc, so a saturated bridge can make `send()` block
+/// on the channel and eventually trip this timeout. The error therefore
+/// conflates "handler wedged" with "bridge backpressure" — acceptable for the
+/// transport surface, but not a substitute for a real backpressure signal.
 const DEFAULT_TIMEOUT: Duration = Duration::from_secs(30);
 
 /// In-process RPC transport over an [`IrohRequestProcessor`].
--- a/crates/hyprstream-rpc/src/transport/in_memory.rs
+++ b/crates/hyprstream-rpc/src/transport/in_memory.rs
@@ -1,0 +1,133 @@
+//! In-process RPC transport for co-located services.
+//!
+//! Dials a service running in the *same process* without a network socket.
+//! This is the in-memory dial target the [`crate::dial`] factory selects for
+//! `inproc://` endpoints — the replacement for ZMQ's `inproc://` transport.
+//!
+//! # No second security model
+//!
+//! [`InMemoryTransport::send`] forwards the canonical request bytes to an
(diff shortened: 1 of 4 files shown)


crates/hyprstream-rpc/src/transport/in_memory.rs
```diff
@@ -0,0 +1,139 @@
...
+/// Default ceiling on a single in-process request, mirroring the networked
+/// RPC transport. A wedged co-located handler should surface as a timeout
+/// rather than hanging the caller indefinitely.
+///
+/// Note this deadline also covers *enqueue* latency: `LocalServiceBridge`
+/// forwards over a bounded mpsc, so a saturated bridge can make `send()` block
+/// on the channel and eventually trip this timeout. The error therefore
+/// conflates "handler wedged" with "bridge backpressure" — acceptable for the
+/// transport surface, but not a substitute for a real backpressure signal.
+const DEFAULT_TIMEOUT: Duration = Duration::from_secs(30);
```
ewindisch, line 52: do we have a better way of managing rpc configuration? what about Config modules/crates?

ewindisch: Agreed this shouldn't be a bare `const` long-term — same theme you raised on #196, which I've now wired: per-service stream cap moved to `ServerConfig.max_concurrent_streams_per_service` + a startup `install_*` (mirroring `install_verify_config`).

For RPC transport tunables specifically there are currently **two idioms** in the tree: builder-override (`QuinnRpcServer::with_stream_limit/with_connection_limit`, #162) and process-global install (verify policy, stream cap). And several bare consts: this `DEFAULT_TIMEOUT`, plus `MAX_FRAME_BYTES` / `DEFAULT_STREAM_LIMIT` / `DEFAULT_CONNECTION_LIMIT` in `rpc_session.rs`.

**Recommendation:** consolidate them into one `RpcConfig`/transport-tunables struct sourced from `ServerConfig`, threaded through the server/`dial()` builders — reserving the process-global `install_*` only for values read deep in `process_request` (verify policy, stream cap) where there's no builder to thread. For *this* const specifically: it's the in-process dial timeout, so the natural home is the dial/`TransportConfig` layer, which is **#151(a)** (the dial factory, currently paused on its lazy-transport increment). I'll fold per-transport timeouts into that when #151(a) resumes rather than leave it a const.

Want me to file a tracking ticket for "unify RPC transport tunables into config" so it's not buried in this thread?

ewindisch: file the ticket.

ewindisch: Filed as #197 ("Unify RPC transport tunables into config"). Per-transport timeouts (incl. this in-process `DEFAULT_TIMEOUT`) are scoped there to land with #151(a)'s dial/TransportConfig layer; server-side limits can migrate to ServerConfig independently.

Issue #136, opened 2026-06-03
M1: RPC production cutover — replace RequestLoop/ZMQ in Spawnable::run with the generic plane (quinn baseline; iroh/tor optional)


ewindisch:
> [!warning] **SUPERSEDED / RE-SCOPED 2026-06-10** — see the consolidated plan (`docs/plans/2026-06-05-zmq-replacement-consolidated.md`, milestone M1) and the review comments below.
>
> This ticket is no longer "port 4 services to `IrohRpcTransport` one PR each." It is the **RPC production cutover**: replace `RequestLoop`/ZMQ ROUTER in `Spawnable::run` with the generic dispatch plane for **all** spawn-path services at once, over the QUIC baseline. The iroh framing assumption is dropped (transport is pluggable; iroh is one optional dial).
>
> **Hard prereqs:** #166 (rename, done) · #148 (extract `process_request` → `service/dispatch.rs`, part 1 done) · #151 (transport abstraction + dial) · security #159/#160/#162. **Kill `Continuation`** (#186) — streaming bootstrap moves to RPC-reply `StreamInfo` + moq-subscribe; without it the cutover breaks all 21 streaming call-sites. Cross-milestone dep on #154 (DID addressing layer the cutover dials).
>
> ---
> *(original body retained below for history)*

Part of #131. Blocked by #133.

After PolicyClient canary (#133), port the remaining 4 services to `IrohRpcTransport`, one PR each, same shape as Phase 2:

- [ ] `registry`
- [ ] `model`
- [ ] `worker`
- [ ] `mcp`

Each retains the same `ZmqService` trait, gets the iroh wire as an alternative transport. Feature-flag remains until Phase 7 removes ZMTP.

Parallelizable across engineers — each service is independent.

ewindisch:
**RE-SCOPED (consolidated plan).** This is the **RPC production cutover**, not a service-by-service iroh port. All spawn-path services share one chokepoint (`Spawnable::run → RequestLoop → ZMQ ROUTER`, spawnable.rs:73-104), so the cutover is a **single mechanical switch** to the generic plane (quinn baseline; iroh/tor optional), not 'one PR per service'.

Scope: replace RequestLoop/ZMQ in `Spawnable::run`; **kill `Continuation`** → streaming bootstrap becomes 'RPC reply carries signed StreamInfo → client subscribes on the moq plane' (else this breaks all 21 streaming call-sites). 
[...]

PR #196, opened 2026-06-10
#186: spawn streaming pump in dispatch core, not transport front-ends

ewindisch:
`process_request` now returns `Result<Vec<u8>>` and spawns the streaming pump itself (`streaming::spawn_stream_pump`, per-service bounded), so all four transport front-ends are uniform bytes-in/bytes-out and **the generic RPC plane no longer rejects streaming** — the #136-cutover blocker this clears.

M1 scope: removes `Continuation` from the dispatch boundary + every front-end. `handle_request` keeps `Option<Continuation>` as the internal service→core contract, so the 16 service impls + proc-macro codegen are untouched; the type dies in M2 when the pump's transport moves to a service-owned moq broadcast (#134). Code+security reviewed (ship-with-fixes; per-service cap fix applied; LocalSet invariant verified). 261 rpc green; full workspace compiles.

📚 Stacked PR — **base: `ewindisch/moq-148`** (sibling to #151a). Part of epic #131.

### DIFF
--- a/crates/hyprstream-rpc/src/service/dispatch.rs
+++ b/crates/hyprstream-rpc/src/service/dispatch.rs
@@ -38,15 +38,15 @@ pub use crate::envelope::EnvelopeVerification;
 ///
 /// # Streaming
 ///
-/// A streaming handler returns a `Continuation` (the server-side pump that runs
-/// after the reply). As of #186 that pump is spawned **here**, via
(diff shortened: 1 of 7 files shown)


ewindisch:
Superseded by **#214** (reconciled onto post-#136-cutover `epic-moq`; this PR forked off moq-148, pre-cutover). The dispatch-core pump spawn + per-service cap landed via #214. Closing.

PR #239, opened 2026-06-15
cleanup(#175): delete LoadBalancerService (ROUTER/DEALER, never instantiated)

ewindisch:
Part of epic #131, addresses #175.

Stacked on #238.

## Summary
- Deletes `LoadBalancerService` struct + impl from `spawner/service.rs` (~155 lines)
- Removes `LoadBalancerService` from `spawner/mod.rs` pub use and `hyprstream-service/lib.rs` re-exports
- This was the only ROUTER/DEALER ZMQ proxy; it was never instantiated in production (only `ProxyService` was used)
- Dead code eliminated rather than migrated — no functional change

## Test plan
- [ ] `cargo build` passes
- [ ] `grep -r LoadBalancerService` returns no results outside deleted code

Issue #211, opened 2026-06-12
[#136 follow-up] dial-layer hardening: from_endpoint fail-closed + lazy inproc dial + remove dead rpc_client


ewindisch:
Latent / robustness items surfaced by the #136 code review (none are active bugs
in current call paths; filing so they don't rot).

1. **`TransportConfig::from_endpoint` silently defaults unknown schemes to
   Inproc** (transport/mod.rs ~415). `from_endpoint("quic://host:port")` →
   `Inproc{"quic://host:port"}` → `dial()` → `lookup_inproc` miss → cryptic "no
   in-process service registered". No current caller passes a networked string to
   the generated `for_endpoint(&str)` (they use `from_resolver` which carries the
   full TransportConfig), so it's latent — but it should fail closed (return/parse
   an explicit error or handle quic/iroh/tcp) rather than silently misroute.

2. **Inproc dial is eager (lazy→eager regression vs ZmqConnection).** `dial()`'s
   Inproc arm calls `lookup_inproc` at construction and captures the processor;
   the old `ZmqConnection` was lazy (connected on first send). So a generated
   `for_service()` to a not-yet-registered dependency now fails at construction
   instead of on first RPC. Mitigated today by dependency-ordered startup stages
   (e.g. policy before inference), so it works — but it's less forgiving of
   ordering than before. Consider making `InMemoryTransport` look up
   `register_inproc` per-send (lazy) to restore the old tolerance, or add a brief
   retry in `dial`'s inproc arm.

3. **Dead code**: `ServiceContext::rpc_client` (factory.rs ~631) still builds a
   raw `ZmqConnection` client; it has no callers (confirmed) and would mismatch a
   bridged service if used. Delete it (or route through `dial()`).

Part of Epic #131; all post-#136.

ewindisch:
Two of the three items implemented on branch `ewindisch/moq-211` (commit `c4133e8c1`):

1. ✅ Deleted `ServiceContext::rpc_client()` — confirmed no callers
2. ✅ `TransportConfig::from_endpoint`: unknown `://` schemes now panic with a descriptive message instead of silently defaulting to Inproc

Item 3 (lazy inproc dial retry in the Inproc arm) deferred — currently mitigated by dependency-ordered startup. Awaiting push.

ewindisch:
Implemented in #240.

ewindisch:
[...]

PR #240, opened 2026-06-15
fix(#211): dial-layer hardening — delete dead rpc_client + from_endpoint fail-closed

ewindisch:
Part of epic #131, addresses #211. Follow-up to #136 code review.

Stacked on #239.

## Summary
- Deletes `ServiceContext::rpc_client()` — dead method with no callers, used `ZmqConnection` internally
- `TransportConfig::from_endpoint` now panics on unknown `://` schemes instead of silently defaulting to inproc (e.g., `from_endpoint("quic://host:port")` previously produced an inproc config)
- Bare names (no scheme) still default to inproc as intended
- Item 3 (lazy inproc dial retry) deferred — mitigated by ordered startup; low priority without evidence of real race

## Test plan
- [ ] `cargo build` passes
- [ ] `grep -r rpc_client` returns no results
- [ ] `from_endpoint("quic://host:port")` panics with a clear message
- [ ] `from_endpoint("my-service")` still produces `EndpointType::Inproc`

### DIFF
--- a/crates/hyprstream-rpc/src/transport/mod.rs
+++ b/crates/hyprstream-rpc/src/transport/mod.rs
@@ -421,8 +421,17 @@ impl TransportConfig {
             EndpointType::Ipc {
                 path: PathBuf::from(path),
             }
+        } else if endpoint.contains("://") {
+            // Unknown scheme — fail-closed: misrouting to inproc would yield a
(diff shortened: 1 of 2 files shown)


ewindisch:

PR #245, opened 2026-06-15
sec(#207): restrict UDS socket to owner-only + SO_PEERCRED peer-uid enforcement

### LINKED ISSUE #207: [sec] UDS transport: bind socket perms + SO_PEERCRED peer-uid check for the ipc plane
Follow-up from the moq-151g security review of `UdsSession`
(`crates/hyprstream-rpc/src/transport/uds_session.rs`), the
`web_transport_trait::Session` over a Unix-domain socket that carries the
same-host **ipc** plane (RPC + moq streaming) after ZMQ removal.

## What's already handled in moq-151g
- Concurrent-substream cap (`MAX_CONCURRENT_STREAMS`, yamux `set_max_num_streams`).
- Per-substream tag-read timeout (`TAG_READ_TIMEOUT`) — closes the
  pre-dispatch slowloris that evaded rpc_session's `REQUEST_READ_TIMEOUT`.
- Fail-closed plane-selector validation in `accept_uds`.
- App-layer identity is unchanged: every request is a COSE `SignedEnvelope`
  verified regardless of transport — this is the real authn.

## What's deferred to the daemon / #136 layer (this ticket)
The transport currently grants **implicit same-host trust**: any process that
can `connect()` to the listener path is accepted at the channel layer. Two
defense-in-depth gaps, both owned by whoever binds the `UnixListener` (the
daemon bootstrap in #136), not by the Session impl:

1. **Socket file permissions.** The bind path must be created with
   owner-only (`0o700`) or owner+group (`0o770`) mode and live in a
   non-world-writable directory (NOT bare `/tmp`). A world-accessible socket
   makes "same-host" a false guarantee. Decide the path + mode at the daemon
   bind site and document it.
2. **`SO_PEERCRED` peer-uid/gid check.** Read peer credentials on the accepted
   `UnixStream` and enforce an allow policy (e.g. same-uid as the daemon, or a
   configured service gid) before handing the connection to the RPC dispatcher
   / moq server. `accept_uds` can be extended to surface `(uid, gid, pid)` as
   metadata; enforcement is a daemon-policy decision.

## Why deferred, not done now
moq-151g is the transport primitive; it has no listener and makes no bind/policy
decisions. Socket provisioning + peer-cred policy belong to the daemon's
listener bootstrap (#136's "daemon provisions prod transport listeners"). The
SignedEnvelope app-layer check already authenticates callers, so this is
defense-in-depth / least-privilege, not a correctness blocker.

Part of Epic #131. Coordinate with #136 (daemon listener provisioning).

ewindisch:
## Summary
- Sets UDS socket file permissions to 0o600 (owner-only) on bind.
- Adds `SO_PEERCRED` peer-UID check on accepted connections: rejects connections from UIDs other than the server process owner.

Closes #207

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9489,6 +9489,7 @@ dependencies = [
  "bitflags 2.11.0",
  "cfg-if 1.0.3",
  "libc",
+ "memoffset 0.9.1",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -149,7 +149,7 @@ sqlparser = "0.39.0"
 json-threat-protection = "0.1.1"
 
 # System utilities
-nix = { version = "0.27", features = ["fs", "process", "signal", "user"] }
+nix = { version = "0.27", features = ["fs", "process", "signal", "socket", "user"] }
 users = "0.11"
 rlimit = "0.10"
 clap = { version = "4.4", features = ["derive", "env"] }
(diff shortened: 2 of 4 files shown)


ewindisch:

PR #249, opened 2026-06-15
sec(#179): atomic 0600 writes for OAuth/JWT rotation key persistence

### LINKED ISSUE #179: [sec] OAuth/rotation signing keys written as raw bytes with default perms (atomic 0600)
**At-rest exposure**, independent of the full KeyProvider/HSM work (which is deferred — see the KeyProvider plan + #129).

`crates/hyprstream/src/auth/key_rotation.rs` persists key-slot private material with plain `std::fs::write`, **raw key bytes, no explicit permissions, non-atomic**:
- `:178` / `:180` — Ed25519 slot key bytes + meta JSON
- `:424` / `:426` — (ES256/second store) key bytes + meta
- `:620` / `:622` — ML-DSA seed + meta

`std::fs::write` creates files at `0666 & ~umask` (typically world-readable 0644). These are OAuth/JWT + ES256 + ML-DSA **signing** keys → at-rest disclosure risk, plus non-atomic writes can leave a torn key file on crash.

**Fix (small, standalone — no trait/HSM dependency):**
- Atomic write: temp file in the same dir + `fsync` + `rename`.
- `0600` perms (set on the temp fd before rename; `OpenOptions::mode(0o600)` on unix).
- Zeroize transient key buffers after write.
- Apply to all six sites (and any sibling save paths).

Relates to **#129** (LocalFileProvider) and the deferred **KeyProvider** plan, which should subsume this storage path. Doing the hardening now closes the exposure regardless of the larger abstraction's timeline.

ewindisch:
## Summary
- Replaces raw `std::fs::write` for rotation signing keys with `auth::credentials::write_secret()` — does tempfile+rename (atomic) and sets 0600 perms.
- Fixes #179: key bytes were written non-atomically with default perms (readable by group/world on some systems).

Closes #179

### DIFF
--- a/crates/hyprstream/src/auth/key_rotation.rs
+++ b/crates/hyprstream/src/auth/key_rotation.rs
@@ -174,10 +174,11 @@ fn load_slot(secrets_dir: &Path, name: &str) -> Option<KeySlot> {
 }
 
 fn persist_slot(secrets_dir: &Path, name: &str, slot: &KeySlot) -> anyhow::Result<()> {
-    let (key_path, meta_path) = slot_paths(secrets_dir, name);
-    std::fs::write(&key_path, slot.key.to_bytes())?;
+    // Atomic write + 0600 perms (#179) — replaces bare std::fs::write which
+    // used 0644 (world-readable) and was non-atomic (torn file on crash).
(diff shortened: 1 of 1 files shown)


ewindisch:

Issue #164, opened 2026-06-05
[sec][low/med] M-stack hardening batch: broadcasts leak, key zeroization, env-downgrade, identity cross-bind


ewindisch:
Whole-stack security review — lower-severity items batched:

- **MEDIUM** `OriginInner.broadcasts` (moq_stream.rs:124,198) is an unbounded Vec that only grows — high-churn in-process publisher leaks memory for process lifetime. Prune dropped/finished broadcasts or key by topic with replace semantics.
- **LOW** No zeroization on quinn cert private key (quinn_transport.rs:262) or `LocalSigner` keys (signer.rs:19-20) — add Zeroize/zeroize-on-drop + custom Debug. (No key bytes in logs today.)
- **LOW** `HYPRSTREAM_ENVELOPE_POLICY=classical` (main.rs:1997) silently disables PQ process-wide, warn-only. Operator-controlled (not remote), but make the downgrade louder / require explicit opt-in.
- **LOW** Identity coherence: quinn cert-hash pin, COSE Ed25519 cnf, and ML-DSA anchor are not cross-bound. Confirm `authorize_signer` + `PqTrustStore` derive from the same trust source (else a transport-authenticated peer could present a different COSE signer).

ewindisch:
Partial fix in `ewindisch/moq-227` (commit 95447e9f5):

**MEDIUM — broadcast leak fixed**: `OriginInner.broadcasts` changed from `Vec<BroadcastProducer>` to `HashMap<String, BroadcastProducer>` (replace semantics keyed by broadcast path). Announcing the same topic twice now drops the old producer instead of accumulating. Memory is bounded to one producer per active broadcast.

Remaining LOW items:
- **Key zeroization** (`LocalSigner`, quinn cert key): `ed25519_dalek::SigningKey` already has zeroize enabled; a `ZeroizeOnDrop` derive on `LocalSigner` + manual drop for the quinn rcgen cert material would complete this.
- **Env-downgrade loudness**: existing `tracing::warn!` at startup when `HYPRSTREAM_ENVELOPE_POLICY=classical` is set is acceptable; could be escalated to a structured event with a countdown warning, or require an additional acknowledgment env var.
- **Identity coherence**: design/verification task — confirm that quinn cert-hash pin, COSE cnf, and ML-DSA anchor all derive from the same trust root via `authorize_signer` + `PqTrustStore`.

ewindisch:
Partial progress on `ewindisch/moq-227`:

[...]

PR #251, opened 2026-06-15
sec(#164): LocalSigner Debug hides key bytes; classical-policy downgrade is now error-level

ewindisch:
## Summary
- Adds custom `Debug` impl on `LocalSigner` that prints only pubkey hex (not key bytes).
- Changes `HYPRSTREAM_ENVELOPE_POLICY=classical` from `warn` to `tracing::error!` + `eprintln!` to make the PQ downgrade visible.

Partial close of #164


ewindisch:

PR #257, opened 2026-06-15
sec(#158): WASM PqTrustStore — enforce SNS outer for registered signers

### LINKED ISSUE #158: M3 deploy-blocker: browser WASM PqTrustStore — enforce SNS outer (replace .classical())
Blocks production deploy of #152 (NOT the commit). From the M3 security re-review — the **one remaining true downgrade hole**.

`wasm_api.rs` verifies envelopes with `.classical()`, so the browser accepts a Hybrid envelope via inner EdDSA only and does NOT enforce the outer ML-DSA. A post-quantum attacker who forges/steals EdDSA can send an inner-only/stripped envelope and the browser accepts it. The browser independently trusts envelopes (WASM verify is the trust boundary), so despite the mesh being SNS, the browser endpoint is strippable.

Design intent was browser-gets-SNS-via-WASM (ml-dsa compiles to wasm32). Fix:
- Add a WASM-side `PqTrustStore` (peer Ed25519→ML-DSA, from the same key bundle the browser already fetches) and verify with Hybrid policy, enforcing the nested outer.
- Until then, the browser path is documented WNS/downgradable.

Fail-OPEN on the browser path relative to a quantum adversary — must fix before claiming browser PQ protection.

ewindisch:
## Summary
- Closes the browser envelope-stripping hole: `verify_signed_envelope` now enforces Hybrid policy (outer ML-DSA-65 required) when the signer's ML-DSA key has been registered via `register_pq_trust()`.
- Unregistered signers continue under Classical (EdDSA-only) — documented WNS until the browser key bundle is provisioned with the peer's ML-DSA key.

New WASM-bindgen exports:
- `register_pq_trust(ed25519_pubkey, ml_dsa_vk)` — bind signer Ed25519 → ML-DSA-65
- `unregister_pq_trust(ed25519_pubkey)` — remove binding
- `clear_pq_trust()` — clear all bindings

Thread-local `WASM_PQ_BINDINGS` maps Ed25519 → ML-DSA-65 vk bytes. On each `verify_signed_envelope` call a temporary `KeyedPqTrustStore` is built; if the specific signer is registered, `opts.with_pq_store()` is used (Hybrid) rather than `opts.classical()`.

Closes #158

### DIFF
--- a/crates/hyprstream-rpc/src/wasm_api.rs
+++ b/crates/hyprstream-rpc/src/wasm_api.rs
@@ -38,6 +38,12 @@ thread_local! {
     static STREAM_HMAC_TABLE: std::cell::RefCell<std::collections::HashMap<u32, crate::crypto::StreamHmacState>> =
         std::cell::RefCell::new(std::collections::HashMap::new());
     static NEXT_HMAC_HANDLE: std::cell::Cell<u32> = std::cell::Cell::new(1);
+
+    // WASM-side PQ trust store: maps Ed25519 pubkey (32 bytes) → ML-DSA-65 vk
+    // bytes. Populated by register_pq_trust(). When a signer is present here,
+    // verify_signed_envelope enforces Hybrid (outer SNS layer required). (#158)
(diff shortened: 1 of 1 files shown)


ewindisch:

PR #258, opened 2026-06-15
fix(#182): auto-grant self-ownership + viewer fallback for JIT federated users

### LINKED ISSUE #182: [authz] No owner-self primitive in Casbin model — new users get zero rules
Verified 2026-06-10. The Casbin model (`auth/policy_manager.rs:127-144`) is `r = sub, dom, obj, act` with no `owner` field; the matcher uses only `g()`/keyMatch/dom/act — **no `${sub} == ${owner}` primitive**. Default is deny-by-default (`:154`), so a new user gets **zero capabilities** unless an operator applies a template or writes explicit rules; there is no notion of "the resources I own."

Three approaches were sketched in the Obsidian follow-up (owner-self). Needs a design decision then impl. Bumps in priority once federated/JIT users arrive (Plan B Phase 4). Tracking here so it's not lost.

ewindisch:
## Summary
- Closes zero-capabilities-on-first-login for JIT/OIDC federated users.
- Adds self-ownership rule (`user:{sub}:*`) always on first login.
- When provider supplies no `default_scopes`, falls back to viewer-level rules (`query:model:*`, `query:registry:*`).

Approach 2 (registration-time auto-grant) — no Casbin model changes.

Closes #182

### DIFF
--- a/crates/hyprstream/src/services/oauth/oidc_callback.rs
+++ b/crates/hyprstream/src/services/oauth/oidc_callback.rs
@@ -557,7 +557,26 @@ async fn provision_federated_user(
         tracing::warn!(provider = %provider_slug, "PolicyManager not available — default_scopes not applied");
         return;
     };
-    for scope in default_scopes {
+
+    // Self-ownership rules: always grant access to the user's own namespace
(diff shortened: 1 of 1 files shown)


ewindisch:

Issue #184, opened 2026-06-10
[test] OAuth integration fixture: server-mediated registration + least-privilege policy


ewindisch:
Verified 2026-06-10. Two fixture issues in `tests/fixtures/hyprstream.ts`:
1. **Pre-start registration** (:262-273): the fixture runs `wizard --non-interactive` to populate UserStore+policy **before** server spawn (:293); registration is not server-mediated (the OAuth flow at :359 only authenticates the already-created user). Should exercise the server-mediated registration path.
2. **Least-privilege** (:264): the wizard adds the local user as `admin` → `vec![("*","*")]` (`bootstrap_manager.rs:465`) → `p, <user>, *, *, *, allow`. The fixture policy models nothing; tighten to a representative least-privilege policy so authz regressions are actually caught.

ewindisch:
## Partial fix: commit `aaa947621`, branch `ewindisch/moq-184`

**Addressed: least-privilege test fixture (#184.2)**

Added `--initial-user-role <ROLE>` to `hyprstream wizard --non-interactive` (default: `admin`, unchanged for production use). Tests pass `--initial-user-role operator` so the test user gets `infer/query/serve on model:*` instead of wildcard admin — authz regressions now cause test failures.

`tests/fixtures/hyprstream.ts` in `www-cyberdione-ai` updated to:
```
wizard --non-interactive --initial-user-role operator
```

**Deferred: server-mediated registration (#184.1)**

The fixture still uses `wizard --non-interactive` (pre-start, not server-mediated). Switching to server-mediated registration requires the fixture to:
1. Start the server with no pre-existing users
2. Drive the OAuth device flow or admin registration API during fixture setup
3. Use that server-minted token for subsequent test connections

That's a larger fixture refactor; tracked here but not in this PR.

ewindisch:
Landed in `main` via the 0.5.0 epic merge (#308). Closing as completed.

PR #259, opened 2026-06-15
feat(#184): --initial-user-role for wizard --non-interactive (least-privilege tests)

ewindisch:
## Summary
- Adds `--initial-user-role <ROLE>` (default: admin) to `hyprstream wizard --non-interactive`.
- Tests can pass `--initial-user-role operator` to avoid the wildcard admin policy, so authz regressions are caught.
- Companion change: `tests/fixtures/hyprstream.ts` in www-cyberdione-ai updated to use `--initial-user-role operator`.

Partial close of #184 (server-mediated registration path deferred)


ewindisch:

PR #262, opened 2026-06-15
perf(#201): prefill LM-head slice — run norm+matmul over last position only

### LINKED ISSUE #201: perf(inference): prefill projects entire sequence through LM head, then discards all but the last token
## Summary

During generation, the cached forward path runs the final RMSNorm **and the LM-head projection over every position in the sequence**, then immediately throws away all but the last token's logits. For models with a large vocabulary (Qwen3.5 ≈ 152K), the LM-head matmul is the single largest matmul in the model, so doing it `seq_len` times during prefill instead of once is a substantial, avoidable cost — in both compute and peak memory.

## Where

`crates/hyprstream/src/runtime/architectures/qwen3_5.rs` — the `ModelOperations` cached wrappers:

```rust
fn forward_with_cache(&self, input: &Tensor, start_pos: usize) -> Result<Tensor> {
    let hidden = self.forward_inner(Some(input), None, start_pos, None)?; // [1, seq, hidden]
    let normed = self.norm.forward(&hidden)?;                             // norm over ALL positions
    self.lm_head_apply(&normed)                                           // [1, seq, vocab]  ← expensive
}
```

`lm_head_apply` reshapes to `[batch*seq, hidden]` and does `flat.matmul(lm_head.tr())` → `[batch*seq, vocab]`.

The consumer in `crates/hyprstream/src/runtime/torch_engine.rs` (`forward_cached` / `forward_with_delta_cached`) then keeps only the last row:

```rust
let last_token_logits = logits.narrow(1, seq_len - 1, 1).squeeze_dim(1);
```

The same pattern exists in `forward_with_cache_and_delta`, `forward_from_embeddings`, and the Llama implementation (`architectures/llama.rs`, `forward_with_cache_inner`).

## Cost

For a prompt of length `S` with hidden size `H` and vocab `V`:

- **Current prefill LM head:** `S × H × V` multiply-adds, plus it materializes a `[1, S, V]` logits tensor.
- **Needed:** `1 × H × V` — only the last position is required to sample the first generated token.

At `S = 2048`, `V ≈ 151936`: the transient logits tensor alone is ~1.2B elements (~2.5 GB BF16 / ~5 GB FP32), allocated and freed on every prefill. This is both a throughput cost and a real OOM source for long-context / long-prompt requests (e.g. agent calls with large context).

## Why it's safe to fix

- **Training is unaffected.** `compute_loss_and_backward` → `forward_training` → `model.forward()` uses the **non-cached** `forward`, which legitimately needs logits at all positions for the full-sequence cross-entropy. Only the cached generation path is changed.
- **Every cached-path caller wants last-only.** Both call sites in the generation loop feed straight into the `narrow(..., seq_len - 1, ...)` slice.
- **SSM/recurrent correctness is preserved.** In Qwen3.5 the GatedDeltaNet conv/recurrent state depends on the whole prefill sequence. The proposed slice happens **after** `forward_inner` returns, so all positions still flow through the layers and update SSM state — only the final norm + LM head are trimmed.

## Proposed fix

Slice `hidden` to its last position **before** `norm` + `lm_head_apply` in the cached generation wrappers. Minimal, localized; no trait changes:

```rust
fn forward_with_cache(&self, input: &Tensor, start_pos: usize) -> Result<Tensor> {
    let hidden = self.forward_inner(Some(input), None, start_pos, None)?; // [1, seq, H]
    let seq = hidden.size()[1];
    let last = hidden.narrow(1, seq - 1, 1);                              // [1, 1, H]
    let normed = self.norm.forward(&last)?;
    self.lm_head_apply(&normed)                                           // [1, 1, vocab]
}
```

The existing `narrow(1, seq_len - 1, 1)` in `torch_engine` then operates on a length-1 sequence (no-op) and stays correct, so `torch_engine.rs` needs no change.

Apply to the three Qwen3.5 cached wrappers (`forward_with_cache`, `forward_with_cache_and_delta`, `forward_from_embeddings`) and the Llama equivalent, ideally via a small shared `last_position(&Tensor)` helper.

### Optional follow-up (larger blast radius, not required)

Thread an explicit `logits_to_keep: usize` (0 = all, N = last N) through the `ModelOperations` forward signatures (HF calls this `num_logits_to_keep`). This generalizes the optimization and leaves room for speculative decoding, which needs the last N positions rather than just the last one. Deferring this keeps the first PR small and safe.

## Scope / assumptions

- The current generation path is **batch = 1** (`[1, seq]`). Slicing to the last position is correct for that. If batched or ragged-length prefill is added later, the slice must become per-sequence (each sequence's own last position), not a single global `seq - 1`.

## Verification status

The safety argument above — particularly the claim that slicing after `forward_inner` preserves GatedDeltaNet/SSM state — is from **code inspection, not execution**. Nothing here has been compiled or run. The byte-identical greedy check below is the actual merge gate, not a formality; run it before merging rather than relying on the reasoning in this issue.

## Acceptance

- Sampled tokens are byte-identical before/after for greedy decoding (the discarded positions never affected the result).
- Measurable prefill latency reduction and lower peak memory on long prompts, scaling with prompt length.
- Training path (`forward_training` / `compute_loss_and_backward`) unchanged.

ewindisch:
## Summary
- `forward_with_cache`, `forward_from_embeddings`, `forward_with_cache_and_delta` in `qwen3_5.rs` were running the full-sequence RMSNorm + LM-head matmul over all positions even though callers always kept only the last token.
- Fix: narrow to last position before norm+lm_head (`hidden.narrow(1, seq-1, 1)`).
- For Qwen3.5 (vocab ≈ 152K) this reduces prefill LM-head work from O(seq_len) to O(1).

Closes #201

### DIFF
--- a/crates/hyprstream-9p/README.md
+++ b/crates/hyprstream-9p/README.md
@@ -1,0 +1,35 @@
+# hyprstream-9p
+
+Minimal 9P2000.L server for bridging the hyprstream VFS to Wanix via DMA ring buffers.
+
+## What it does
+
+Implements the subset of the 9P2000.L wire protocol needed for read/write filesystem access:
+
+| Message pair | Purpose |
+|---|---|
+| `Tversion` / `Rversion` | Protocol negotiation |
+| `Tattach` / `Rattach` | Attach to root fid |
+| `Twalk` / `Rwalk` | Path traversal (resolve fid) |
+| `Tlopen` / `Rlopen` | Open file (9P2000.L variant) |
+| `Tread` / `Rread` | Read data |
+| `Twrite` / `Rwrite` | Write data |
+| `Treaddir` / `Rreaddir` | List directory |
+| `Tgetattr` / `Rgetattr` | Stat file |
+| `Tclunk` / `Rclunk` | Close fid |
+| `Rlerror` | Error response |
+
+**Wire format**: all messages are 4-byte LE length-prefixed, matching the DMA ring buffer framing used by the Wanix WASM runtime.
+
+**WASM extras** (`target_arch = "wasm32"`): `dma` module (SAB ring-buffer transport) and `wanix_mount` (mount the VFS into the Wanix namespace via DMA).
+
+## Architecture position
+
+```
+hyprstream-vfs           (Mount trait, Namespace, Fid)
+    ↑
+hyprstream-9p            ← you are here
+    ↑
+hyprstream               (9P server factory)
+www-cyberdione-ai        (Wanix WASM: mounts via DMA ring buffer)
+```
--- a/crates/hyprstream-discovery/README.md
+++ b/crates/hyprstream-discovery/README.md
@@ -1,0 +1,40 @@
+# hyprstream-discovery
+
+Service discovery and endpoint resolution for the hyprstream cluster.
+
+## What it does
+
+`DiscoveryService` is the authoritative source for service endpoint lookup. It wraps the `EndpointRegistry` from `hyprstream-rpc` and exposes it as:
+
+1. A local `Resolver` implementation — in-process lookup for services in the same daemon.
+2. A Cap'n Proto RPC service — remote endpoint queries from clients or peer daemons.
+
+### Operations
+
+- `list_services` — enumerate registered services + their endpoint sets (QUIC, iroh, UDS, inproc).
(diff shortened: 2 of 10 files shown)


ewindisch:

Issue #202, opened 2026-06-11
Epic: validate and harden per-layer LoRA rank allocation (ttn_profile)


ewindisch:
## Disposition

Refactor, not scrap or rewrite. The scaffolding (tiered profiles, caching, banding, runtime utilization tracking) is sound and worth keeping. The unproven part is one thing: the assumption that the spectrum of a layer's base weight predicts how much LoRA rank its adaptation needs. That assumption lives behind one statistic (`bond_entropy`) and one mapping (`entropy_to_rank`), both at a clean boundary, so it can be replaced without touching the rest of the system. The work below validates that assumption and, independently, makes correctness depend less on it.

## Context

`crates/hyprstream/src/runtime/ttn_profile.rs` allocates per-layer LoRA rank from the spectrum of each layer's weight matrices: SVD each projection, take the Shannon entropy of the normalized squared singular values ("bond entropy"), map that to a rank band {4,8,16,32}. Resolution is tiered: embedded ablation-derived profiles, then disk cache, then computed-from-weights, then a uniform fallback.

Only one profile is validated: the embedded Qwen3.5-0.8B profile, whose ranks come from ablation perplexity-delta. Every computed (Tier 3) profile uses bond entropy instead, and the two are only equivalent if bond entropy correlates with ablation importance. That correlation has not been checked.

## ⚠️ Status (verified against main, 2026-07-05)

**The epic stalled on missing ground truth, and the checklist below reflects the revised plan.** Verified facts:

- The interim guardrail landed (`2ac4f5b82`): Tier-3 ranks capped at 8 for unrecognized models + `warn!` at the Tier-3 dispatch. Two residual gaps (Tier-2 cache bypasses the cap; `find_embedded_profile` silent on known-type size mismatch) are being fixed in the cleanup PR.
- **The spike's input data is not in the repo**: only 6/24 layers have ablation perplexity-deltas (all 18 GDN layers are `None`); embedded entropies are hardcoded literals, not Tier-3 outputs; `tnn-transformer-1/results/linearization_ablation.json` is external and not located. Recovering or regenerating it is now the explicit gate (#841). This also means the "gold standard" Tier-1 profile has no ground truth for 18/24 layers.
- `online_rank_allocator` confirmed dead (zero production call-sites) with the `|rank − entropy|` unit mismatch intact → resolved by removal (cleanup PR).
[...]

PR #264, opened 2026-06-15
fix(#202): cap Tier 3 entropy-based LoRA rank at 8 for unrecognized models

ewindisch:
## Summary
- Tier 3 entropy→rank mapping (Tier 3) was calibrated only against Qwen3.5-0.8B ablation data. Allocating ranks up to 32 for models with no embedded profile was unvalidated and could cause TTT divergence.
- Interim guardrail: cap `recommended_rank` at 8 (matching the uniform fallback) inside `compute_weight_entropy_profile()`.
- Adds a `warn!` at the Tier 3 dispatch site in `get_layer_profile()` so operators can see when a model is unrecognized.
- The cap will be lifted once the ablation spike completes and results are encoded in an embedded profile for the model.
- The runtime-correction loop (`delta_rank_utilization`) will raise effective rank over time as needed.

Partial close of #202 (ablation spike + validated Tier 3 calibration deferred)

### DIFF
--- a/crates/hyprstream/src/runtime/ttn_profile.rs
+++ b/crates/hyprstream/src/runtime/ttn_profile.rs
@@ -339,7 +339,10 @@ pub fn compute_weight_entropy_profile(
             0.95 // Default: near-uniform → low rank
         };
 
-        let recommended_rank = entropy_to_rank(avg_normalized);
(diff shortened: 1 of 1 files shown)


ewindisch:

Issue #275, opened 2026-06-16
M3: ResponseEnvelope Hybrid signing — COSE SNS (EdDSA+ML-DSA) parity with SignedEnvelope; PQ-ready StreamInfo attestation


ewindisch:
Part of epic #131. Completes the response-side of #152 (Hybrid envelope signing). Same downgrade class as the closed CRITICAL #161, on the un-migrated sibling type.

## Problem
`RequestEnvelope`/`SignedEnvelope` got the full M3 (#152) treatment — CBOR `COSE_Sign` composite (EdDSA inner, ML-DSA-65 outer = SNS), kid-anchored `PqTrustStore`, runtime `CryptoPolicy` default **Hybrid ENFORCED**. The response side, `ResponseEnvelope`, was **never migrated** and is **classical Ed25519 only**:

- `common.capnp` `struct ResponseEnvelope` has only `sig @2 :Data(64)` + `cnf @3 :Data(32)` — no `cose`, no PQ fields.
- `ResponseEnvelope::new_signed`/`verify` (`envelope.rs:1701-1738`) hardcode a single `ed25519_dalek` `verify_strict` over `request_id ‖ payload` — no COSE, no ML-DSA, no policy hook.
- `unwrap_response` (`envelope.rs`) and the server signing sites (`service/dispatch.rs:87/114/135`, `transport/rpc_session.rs`, `transport/iroh_rpc.rs`) all use this classical-only path.

#152's own target-design table lists **"Mesh RPC envelopes, streaming envelopes, browser-via-our-WASM → Nested COSE (SNS)"** and notes *"only the StreamInfo envelope pays"* — i.e. responses were always in-scope; the implementation just didn't land there.

## Impact (security)
`ResponseEnvelope` authenticates **server-asserted data** — most importantly `StreamInfo` (the ephemeral DH `dhPublic` that anchors topic + MAC-key derivation, the QoS contract, and the moq transport paths). Today that attestation is:
- **Not PQ-ready** — a quantum/"harvest-now" adversary that forges an Ed25519 response can forge the DH public → controls the derived topic/MAC keys → breaks streaming E2E auth.
- **Asymmetric** — requests are Hybrid-enforced, responses are classical. The `StreamOpt` schema comments currently *claim* the QoS contract is "Authenticated by the Ed25519 signature over the enclosing SignedEnvelope" — wrong on both counts (it's a `ResponseEnvelope`, classical-only).

## Design decision: do NOT merge the two envelope structs
[...]

PR #289, opened 2026-06-18
feat(#157): persistent mesh ML-DSA signing key + admin-anchored PQ trust store

ewindisch:
Close the M3 deploy-blocker: hybrid COSE response signing (#275) had an
ephemeral ML-DSA-65 key and an empty peer trust store, so Hybrid attestation
had nothing persistent to sign with and nothing to verify against.

Scope 1 — persistent mesh signing key (derive, don't store):
- node_identity.rs: derive_mesh_mldsa_key() = HKDF derive_purpose_key(node
  Ed25519 key, "hyprstream-mesh-mldsa-v1") -> ml_dsa_sk_from_seed; seed zeroized.
  Stable across restarts, no new key file, PQUIP key-separation (distinct label).
- signer.rs / streaming.rs: LocalSigner::new + StreamChannel::new derive the
  mesh key instead of minting an ephemeral one.
- svc.rs: RequestService::pq_signing_key() default is now
  Some(derive_mesh_mldsa_key(signing_key())) instead of None — every service
  signs Hybrid by default. The EdDSA half is signed with the same Ed25519 key
  dispatch already uses, so the kid-anchored binding (Ed25519 signer -> ML-DSA
  vk) is self-consistent; a Classical verifier still accepts via inner EdDSA.

Scope 2 — publish the #mesh-pq Multikey VM:
- oauth/state.rs: OAuthState carries mesh_pq_verifying_key, derived from the
  SAME Ed25519 key in with_signing_key (published VM == signing key).
- did_document.rs: un-gate mldsa65_verification_method; root_did_document emits
  #mesh-pq (ML-DSA-65 Multikey, 0x1211) with auth/assertion refs.

Scope 3 — populate KeyedPqTrustStore (eager, admin-anchored, immutable):
- config: new OAuthConfig.mesh_peers: HashMap<label, MeshPeerConfig{ed25519_
  multibase, mldsa65_multibase}>; inline out-of-band Multikey strings, serde
  default. Distinct from trusted_issuers.
- auth/mesh_trust.rs: build_mesh_pq_trust_store binds each peer's Ed25519 signer
  identity -> trusted ML-DSA-65 key; malformed entries logged + skipped (never
  trusted); codec-checked Multikey decode.
- bin/main.rs: install_envelope_verify_config builds the store EAGERLY before
  install_verify_config; immutable thereafter. Empty/absent mesh_peers => empty
  store => unchanged behavior (Hybrid fails closed for unknown peers). Matches
  the KeyedPqTrustStore "established out-of-band" contract + Tiles admission.

Tests: deterministic+separated mesh key derivation; published #mesh-pq VM ==
derived signing key; trust store verifies a configured peer's ML-DSA signature,
empty config => empty store, malformed entry skipped, wrong-codec rejected.

Issue #157. Part of epic #131.

### DIFF
--- a/crates/hyprstream-rpc/src/crypto/pq.rs
+++ b/crates/hyprstream-rpc/src/crypto/pq.rs
@@ -49,6 +49,14 @@ pub fn ml_dsa_vk_bytes(key: &MlDsaVerifyingKey) -> Vec<u8> {
     key.to_bytes().to_vec()
 }
 
+/// Derive the raw ML-DSA-65 verifying-key bytes (1952 bytes) from a signing key.
+///
+/// Convenience for callers outside this crate that hold a signing key but do not
+/// depend on the `ml_dsa` crate's `Keypair` trait directly.
+pub fn ml_dsa_sk_to_vk_bytes(key: &MlDsaSigningKey) -> Vec<u8> {
+    ml_dsa_vk_bytes(&key.verifying_key())
+}
+
 pub fn ml_dsa_vk_from_bytes(bytes: &[u8]) -> Result<MlDsaVerifyingKey> {
     let encoded = EncodedVerifyingKey::<MlDsa65>::try_from(bytes)
         .map_err(|_| anyhow::anyhow!("invalid ML-DSA-65 verifying key length (expected 1952 bytes)"))?;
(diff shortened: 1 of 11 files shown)


ewindisch:

Issue #278, opened 2026-06-16
[sec] M3: bind composite hybrid alg-id into COSE external_aad — signature-level non-separability (per draft-ietf-pquip-hybrid-signature-spectrums)


ewindisch:
Part of epic #131. Surfaced by checking #275 against `draft-ietf-pquip-hybrid-signature-spectrums-07` (PQUIP WG). Workspace-wide (the shared `cose_sign` layer), so it covers the request side (`SignedEnvelope`) and the response side (`ResponseEnvelope`, #275).

## Problem
Our nested hybrid COSE (`crypto/cose_sign.rs`: inner EdDSA `COSE_Sign1` + outer ML-DSA-65 over message ‖ inner) binds only `[envelope_schema_id, inner_type_id]` into `external_aad`. It does **not** bind a hybrid-composite algorithm identifier. The code documents this (`cose_sign.rs:50-52`):

> "the external_aad does not currently carry a hybrid-composite algorithm identifier, so the inner EdDSA COSE_Sign1 produced in Hybrid mode is byte-identical to one produced in Classical mode."

Consequence vs the spectrums draft: non-separability is achieved **only at the protocol level** (the Hybrid-policy "reject if outer ML-DSA missing" check) — the draft's weaker, verifier-honesty-dependent placement. At the **signature level** the construction is separable: a stripped inner EdDSA is a valid *standalone Classical* signature. Two draft mitigations are unmet at the signature level:
- **Key restriction** — the inner EdDSA uses the *same* Ed25519 identity key as classical-standalone signing (the reuse the draft says to prohibit).
- **Alg-id binding** — absent.

Per `draft-ietf-pquip-hybrid-signature-spectrums-07` (§ Nested Construction / Non-Separability), signature-level non-separability requires binding the `hybridAlgID` into each component's signed input: `sig = Sigma.Sign(hybridAlgID ‖ m)`.

## Scope
- Bind a composite/hybrid algorithm identifier (the EdDSA+ML-DSA-65 component pair) into the COSE `external_aad` (`crypto/cose_sign1::build_external_aad`) — or a COSE protected header — for both the inner and outer layers, request and response.
- Result: the inner EdDSA `COSE_Sign1` is no longer byte-identical between Hybrid and Classical modes → **signature-level WNS** per the draft, robust even if a verifier mis-implements the policy check (independent of the protocol-level enforcement in #277).

## Relation
[...]

PR #290, opened 2026-06-18
feat(#278): bind hybrid-composite alg-id into COSE AAD — signature-level non-separability

ewindisch:
Close the signature-level separability gap in the nested hybrid COSE
construction (cose_sign.rs): previously the inner EdDSA COSE_Sign1 produced in
Hybrid mode was byte-identical to a Classical one, so a stripped inner EdDSA
was a valid standalone Classical signature — non-separability rested on the
verifier's require_pq policy alone (the weaker, verifier-honesty-dependent
placement per draft-ietf-pquip-hybrid-signature-spectrums).

Per the spectrums draft (§ Nested Construction / Non-Separability), bind a
hybrid-composite algorithm identifier into each component's signed input
(sig = Sign(hybridAlgID ‖ m)):

- HYBRID_COMPOSITE_ALG_ID = b"id-MLDSA65-Ed25519" (the LAMPS composite name from
  draft-ietf-lamps-pq-composite-sigs pairing ML-DSA-65 + Ed25519) — a fixed,
  human-auditable label rather than an unallocated IANA COSE code-point.
- build_hybrid_external_aad: deterministic CBOR [base_aad:bstr, algid:bstr].
  Classical AAD = bare [int,int] array; Hybrid AAD = [bstr,bstr] array — the
  two TBS inputs can never collide.
- Bound into BOTH inner + outer layers, Hybrid mode ONLY, via the shared
  cose_sign layer → covers request (SignedEnvelope) and response
  (ResponseEnvelope) paths. Classical mode signs/verifies the bare base AAD →
  byte-for-byte unchanged.
- Mode gating: sync sign_composite keys off pq_sk.is_some(); async/WASM path
  (rpc_client.rs) threads hybrid = signer.pq_pubkey().is_some() into inner_tbs.
- Verifier reconstructs the inner AAD from the composite's SELF-DESCRIBING shape
  (outer present ⇒ inner was Hybrid-signed). A stripped Hybrid composite
  (outer=null) reconstructs the bare AAD and FAILS the inner EdDSA *signature*
  check — independent of require_pq. A genuine Classical inner still verifies
  under a Classical verifier (skip-unknown interop preserved).

Closed the cose_sign.rs gap doc-comment; added a backwards-compat/transition
note (classical sig/cnf retained per WNS; peers agree on policy — a Hybrid→
Classical-only consumer was already require_pq-rejected, so no new interop
surface breaks).

Tests: hybrid-AAD distinctness/determinism + alg-id embedding; inner EdDSA
TBS+signature Hybrid≠Classical; stripped-Hybrid-inner fails standalone-Classical
*signature* verification; Classical round-trip regression; Classical inner does
not verify under hybrid AAD. cose_sign: 22 pass. Native + wasm32 build clean.

Issue #278. Part of epic #131.

### DIFF
--- a/crates/hyprstream-rpc/src/crypto/cose_sign.rs
+++ b/crates/hyprstream-rpc/src/crypto/cose_sign.rs
@@ -4,23 +4,32 @@
 //! ## Where this sits in the PQUIP hybrid-signature taxonomy
 //!
 //! In the taxonomy of `draft-ietf-pquip-hybrid-signature-spectrums` §1.3.4 this
-//! construction is **Weak Non-Separable (WNS)**, *not* Strong Non-Separable
+//! construction is still **Weak Non-Separable (WNS)**, *not* Strong Non-Separable
 //! (SNS). SNS requires *fusion* — the two algorithms' signatures combined into a
 //! single object that cannot be verified by either component alone. We do not do
 //! that: we emit two distinct COSE_Sign1 signatures and bind them by nesting
-//! (the outer signs `payload ‖ inner_sig`). A classical-only verifier can — and,
-//! per [`verify_composite`]'s `require_pq = false` path, deliberately does —
-//! accept the inner EdDSA signature on its own. That standalone-acceptability is
-//! exactly what SNS precludes, so labelling this SNS would overstate the
-//! cryptographic guarantee.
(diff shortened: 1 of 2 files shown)


ewindisch:

Issue #277, opened 2026-06-16
[sec] M3: Response-side Hybrid enforcement — process-global fail-closed default + client PQ trust anchor


ewindisch:
Part of epic #131. Follow-up surfaced by the #275 Opus review. Completes the enforcement side of #275.

## Context
#275 gave `ResponseEnvelope` the COSE composite (EdDSA + ML-DSA-65) signing mechanism — domain-separated (`RESPONSE_ENVELOPE_TYPE_ID`), fail-closed, and **un-downgradable once a client opts in** (`with_response_pq_store`). But enforcement ships **dormant**: no production caller installs the response PQ store, so the client default is `CryptoPolicy::Classical` and server-asserted `StreamInfo` (the `dhPublic` that anchors topic/MAC-key derivation, plus the QoS contract) is **not PQ-attested in practice yet**.

Asymmetry: the request side has `global_verify_policy()` (process-global, fail-closed Hybrid default); the response side has no equivalent — the client default is hardcoded Classical with no "uninstalled → Hybrid" safety net.

## Scope
- Add a response-side process-global `install_verify_config` equivalent (fail-closed Hybrid default mirroring the request side).
- Wire the daemon + WASM clients to anchor the server's ML-DSA-65 verifying key (via peer attestation / DID `#mesh` resolution) so they can enforce Hybrid on responses.

Until this lands, a classical-Ed25519 forgery of a response still defeats real clients (the documented opt-in scope). Relates #157 (mesh attestation), #152, #275.

ewindisch:
Blocked by #157 (persistent mesh ML-DSA key + KeyedPqTrustStore population). The trust anchor #277 needs IS the store #157 populates (bidirectional Ed25519→ML-DSA mapping), so once #157 lands #277 is just: point the response-verify path at the populated KeyedPqTrustStore + flip the response default to fail-closed Hybrid. NOT dependent on the DidWebResolver (that is transport resolution; the PQ key path is #157+#154).

ewindisch:
## WNS verification posture (IETF PQUIP) — design reference

This PR's response-side enforcement, and the shared `verify_cose` path it consults, follow the **Weak Non-Separable (WNS)** posture from the IETF PQUIP working group — **per-identity** enforcement, not blanket fail-closed Hybrid.

### Specs / drafts of record
[...]

PR #291, opened 2026-06-18
feat(#277): response-side Hybrid enforcement — WNS per-identity verify posture

ewindisch:
Complete the enforcement side of #275 with the correct WNS (Weak Non-Separable)
posture from draft-ietf-pquip-hybrid-signature-spectrums: enforce the ML-DSA-65
outer PER-IDENTITY (only for signers anchored out-of-band), with a classical
inner-EdDSA floor for unanchored signers — NOT blanket fail-closed Hybrid.

Response-side process-global config (mirrors the request side):
- envelope.rs: ResponseVerifyConfig + RESPONSE_VERIFY_CONFIG OnceLock;
  install_response_verify_config (first-write-wins), global_response_verify_policy
  (uninstalled default fail-closed Hybrid in prod, Classical under cfg(test)),
  global_response_pq_store. Escape hatch unified via envelope_policy_from_env()
  / ENVELOPE_POLICY_ENV (HYPRSTREAM_ENVELOPE_POLICY=classical downgrades both
  directions in lock-step).
- rpc_client.rs: response_verify_policy/store became Option (None ⇒ consult the
  global default); per-client override still wins. wasm path unchanged (#158).
- main.rs: the SAME admin-anchored mesh PQ store (#157) backs both request and
  response configs, built once.

WNS per-identity verify (the correctness fix — both verify_cose sites):
- require_pq = verify_policy.uses_pq() && anchored_pq.is_some(). Removed the
  "Hybrid policy requires a kid-anchored ML-DSA-65 key, none resolved" bail.
  For an anchored signer require_pq stays true → un-downgradable Hybrid; for an
  unanchored signer, verify the inner EdDSA (classical floor). Safe because
  ed_vk = VerifyingKey::from_bytes(&self.cnf) (PQ-lookup identity == EdDSA-verified
  identity) and verify_with constant-time-pins cnf == expected_pubkey when known.
  PQ is never resolved from the self-asserted COSE entry.
- This fixes the fresh-install break (empty mesh_peers rejected ALL responses,
  incl. a node verifying its own in-process services — inproc_bridged_e2e).

Docs/refs: docs/cryptography-architecture.md gains a "Hybrid (PQ/T) Signatures —
WNS posture" section citing the PQUIP drafts (hybrid-signature-spectrums, RFC
9794, lamps-pq-composite-sigs, cose-mldsa). Doc-comments on verify_with/
verify_cose/unwrap_response_with updated to the WNS per-identity wording.

Tests: reverted the Classical-masking response installs added to the e2e/service
tests (they now exercise the real default). 6 new WNS tests across both paths
(anchored-downgrade-rejected via stripped outer; unanchored classical fallback;
anchored enforced) + a verify_any_signer_with unanchored + forged-cnf case.
inproc_bridged_e2e passes; native + wasm32 build clean; clippy clean;
hyprstream-rpc 340 pass (1 pre-existing moq_event flake). Note: 7 pre-existing
auth::policy_manager wildcard-validation failures exist on the baseline,
unrelated to this change.

Issue #277. Part of epic #131.

### DIFF
--- a/crates/hyprstream-rpc/src/envelope.rs
+++ b/crates/hyprstream-rpc/src/envelope.rs
@@ -894,6 +894,117 @@ pub fn apply_global_verify_config<'a>(
     opts
 }
 
+/// Name of the mid-rollout escape-hatch env var shared by the request and
+/// response verify paths. Setting it to `classical` downgrades BOTH directions
+/// (operators staging the PQ rollout before peer ML-DSA bindings are
+/// provisioned). Any other value (or unset) means Hybrid.
+pub const ENVELOPE_POLICY_ENV: &str = "HYPRSTREAM_ENVELOPE_POLICY";
+
+/// Parse [`ENVELOPE_POLICY_ENV`] into a [`CryptoPolicy`]. Single source of truth
+/// for the escape hatch so the request side (`install_verify_config` in
+/// `main.rs`) and the response side share identical semantics. Defaults to
+/// fail-closed [`CryptoPolicy::Hybrid`]; only the literal `classical` downgrades.
+#[cfg(not(target_arch = "wasm32"))]
+pub fn envelope_policy_from_env() -> crate::crypto::CryptoPolicy {
+    match std::env::var(ENVELOPE_POLICY_ENV).ok().as_deref() {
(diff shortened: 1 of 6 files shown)


ewindisch:

Issue #276, opened 2026-06-16
M2: subscribe-time authz + per-tenant announce scoping on the networked moq plane (metadata defense-in-depth)


ewindisch:
Part of epic #131. Follow-up surfaced by the #274 Opus review.

## Context
#274 made the moq streaming plane network-reachable (served on the `web_transport_quinn` endpoint at `/moq`). That moved the **open-subscribe** surface from same-host UDS to the open network. This is NOT a content-security hole — content is already protected:
- the broadcast path is a 64-hex DH-derived (HKDF-of-ECDH) **capability topic** — unguessable, learned only from a signed `StreamInfo` over authenticated RPC;
- frames carry a chained per-frame HMAC keyed by material only a DH-authorized client derives (and, once AEAD lands, are ciphertext).

So an unauthorized subscriber gets frames it cannot derive keys for or verify. This is metadata **defense-in-depth**, not a fix for an open content hole. Open-subscribe-once-connected is the standard moq model (cf. Cloudflare's coarse `?jwt=` gate).

## Residual (metadata) exposure
- A *connected* peer can announce-enumerate topic names (moq announces by prefix).
- The relay/origin sees both endpoints' network addresses.

This is the "rendezvous like tor, but less anonymously" posture: rendezvous-by-secret-topic + content-blind relay, but single-hop and address-visible.

## Scope
- Add a subscribe-time authorization hook on the `moql` accept path (the `web_transport_quinn` `/moq` handler + the iroh `IrohMoqProtocolHandler`), gating who may subscribe.
- Per-tenant announce scoping so a peer cannot enumerate other tenants' broadcast names.

## Relation to fuller anonymity
Tor-level (no single party sees both ends) needs an oblivious/independent relay (#168) + multi-hop + AEAD (#153). The per-PDS relay deliberately trades that for trust-domain simplicity. Relates #134/#168/#153.

ewindisch:
## Authz model (resolved): reuse the existing identity + policy stack — not a new model

Per maintainer: authz is a **mix of public and private** over the stack we already have.

- **Identity:** atproto DID / OIDC subject / federated peer (existing federation).
- **Policy:** Casbin via PolicyService — subscribe is a `stream:subscribe` check on the stream resource, keyed on the authenticated identity.
- **Public/private:** a per-stream attribute (producer-declared, alongside `StreamOpt`). Public → announced/subscribable freely; private → PolicyService-gated.

Two enforcement points (defense-in-depth):
[...]

Issue #274, opened 2026-06-16
M2: StreamInfo networked moql addressing + networked moq-subscribe client (restore remote streaming)


ewindisch:
Part of epic #131. Completes the client/addressing side of #134 M2a (the server-side serve-downstream wire layer landed in #147; remote subscribers still cannot connect).

## Problem (regression)
The ZMQ→moq migration replaced `StreamInfo`'s network-reachable endpoint with `moqUdsPath` — a local filesystem path. So `StreamInfo` hands subscribers a UDS path, and the only moq-subscribe client knows how to dial UDS. The `StreamService` `Origin` is served over the `moql` ALPN server-side (#147/#134-M2a), but **remote subscribers have no networked address to dial and no networked client** — streaming is local-host-only. For a federated app this is a regression introduced by the migration.

## Scope
- **Schema (`streaming.capnp`), dense id-spin** (capnp forbids ordinal gaps → live fields renumbered, dead fields removed, not reserved). Remove `endpoint @1` (ZMQ XPUB), `moqUdsPath`, `moqBroadcastPath`. Add `broadcastPath` (DH topic / moq track name) + `reach :List(ReachOption)` where `ReachOption { role :Role(direct|relay); transport :TransportConfig }`, transport encoded via the **`service_entry.rs` codec** (one source of truth with the DID doc). `reach` carries **network-routable transports only** — never a UDS path or in-process handle (meaningless off-box). Same id-spin + `reach` shape applied to the parallel `tui.capnp` / `hyprstream-rpc-std/notification.capnp` StreamInfos. Codec generated (depends on the codegen issue).
  - Ordinal-reuse is safe: pre-release (no shipped client), `StreamInfo` is never persisted/cached (built fresh + consumed in-handshake), and the spin lands atomically with codec + all producers/consumers in one PR.
- **`dial_stream(&TransportConfig) -> web_transport_trait::Session`** next to `dial()`, dispatching on `EndpointType` (Quic/WebTransport from the reach list; Ipc/Inproc only from locally-resolved same-host config, never from the wire). Reuses the per-transport session primitives (`IrohSubstrate::connect(.., ALPN_MOQ_LITE)`, `connect_uds(.., PLANE_MOQ)`).
[...]

PR #294, opened 2026-06-18
feat(#276): moq subscribe-time authz + per-tenant announce scoping (metadata defense-in-depth)

ewindisch:
#274 moved the moq open-subscribe surface to the network. Content is already
protected (DH-derived capability topic + per-frame chained HMAC); #276 adds the
metadata defense-in-depth the issue asks for: gate who may subscribe + stop a
connected peer from announce-enumerating other tenants' broadcast names.

Key constraint (reported, not worked around): moq-net 0.1.8 exposes NO per-subscribe
callback — authorization is STRUCTURAL. A session can only enumerate/subscribe to
broadcasts visible through the OriginConsumer it is served; OriginConsumer::scope
narrows that by path prefix. So per-tenant scoping == hand each session a
tenant-scoped consumer, and a subscriber that cannot see a broadcast cannot
subscribe to it.

- moq_authz.rs (new, hyprstream-rpc, native-only): tenant_of / tenant_prefix /
  filter_announces_by_tenant (pure, exact-segment match — no alice/alice2 bleed);
  tenant_scoped_consumer (live OriginConsumer::scope); Visibility / SubscribeDecision
  / PeerIdentity / SubscribeAuthorizer trait + DefaultAuthorizer (public-open,
  private-policy-gated, fail-closed when private with no gate); VisibilityFn/PolicyGate
  aliases for the Casbin gate. 13 unit tests.
- transport/iroh_moq.rs: MoqAuthzConfig + PeerTenantResolver + with_authz; the
  accept path derives PeerIdentity from the authenticated Connection::remote_id()
  (EndpointId) and serves a tenant-scoped consumer (fail-closed if no scope).
  Live integration test over real iroh: a bob-scoped subscriber reads bob's
  broadcast and times out on alice's.
- transport/quinn_transport.rs: with_moq_authz + anonymous-peer scoping in the
  /moq run loop. The WebTransport CONNECT is not mutually authenticated (server
  cert pinned, client anonymous at TLS), so per-peer enforcement requires the iroh
  path; scoping still applies when the resolver yields a tenant from non-identity
  context (e.g. single-tenant endpoint). Documented.

Seams (documented // TODO(#276)): (1) per-track public/private is not enforced
in-band (no moq-net subscribe hook) — SubscribeAuthorizer is a tested policy unit
ready for a future hook; today public/private is expressed via scoping. (2) The
service spawner cannot reach PolicyManager (hyprstream) without a dependency cycle,
so it installs the permissive default and documents where a hyprstream-built
MoqAuthzConfig (Casbin gate via global_policy_manager) threads in. Defaults
preserve the working open same-tenant model. Casbin reuse: PolicyManager::
check_with_domain via global_policy_manager().

Tests: 15 new (public→allowed incl anonymous; private+authorized→allowed;
private+unauthorized→rejected; private+no-gate→fail-closed; tenant isolation via
pure filter + live moq-net scope + live iroh wire; segment-exact no prefix bleed).
Native + wasm32 build clean (native-gated); clippy clean; hyprstream-rpc 390 pass
(1 known moq_event flake); networked /moq integration test still passes.

Issue #276. Part of epic #131.

### DIFF
--- a/crates/hyprstream-rpc/src/lib.rs
+++ b/crates/hyprstream-rpc/src/lib.rs
@@ -152,6 +152,8 @@ pub mod streaming;
 #[cfg(not(target_arch = "wasm32"))]
 pub mod moq_stream;
 #[cfg(not(target_arch = "wasm32"))]
+pub mod moq_authz;
+#[cfg(not(target_arch = "wasm32"))]
 pub mod moq_event;
 #[cfg(not(target_arch = "wasm32"))]
 pub mod transport;
--- a/crates/hyprstream-rpc/src/moq_authz.rs
+++ b/crates/hyprstream-rpc/src/moq_authz.rs
@@ -1,0 +1,490 @@
+//! Subscribe-time authorization + per-tenant announce scoping for the moq
+//! streaming plane (#276).
+//!
+//! ## Why this module exists
+//!
+//! #274 made the moq streaming plane network-reachable (served on
+//! `web_transport_quinn` at `/moq`, and over the iroh `moql` ALPN via
+//! [`crate::transport::iroh_moq::IrohMoqProtocolHandler`]). That moved the
+//! *open-subscribe surface* onto the network. This is **not** a content hole —
+//! stream content is already protected by the §7.5 trust model (a 64-hex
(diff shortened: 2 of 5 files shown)


ewindisch:

Issue #282, opened 2026-06-17
Activate iroh substrate (daemon bind) + pkarr dial-by-Ed25519 discovery — shared P2P transport (Tiles interop)


ewindisch:
Part of epic #131. Was the deferred "PR-E" follow-up; now **also interop-driven** (Tiles is all-in on iroh). See `docs/interop/tiles-alignment.md`.

## Context
Tiles' primary networking is **iroh** (public relays + mDNS) with Ed25519 identities. A shared iroh + pkarr substrate gives direct P2P interop. Today `IrohSubstrate` is test-only (not bound in the daemon); the live planes are quinn/WebTransport (RPC + moq, #274). No pkarr.

## Scope
- Bind `IrohSubstrate` in the daemon (parallel to the quinn bind); serve both ALPNs (`hyprstream-rpc/1` + `moql`).
- Advertise `#iroh` (ed25519 node_id) + `IrohTransport` service entry in `root_did_document()`.
- **pkarr discovery:** publish node addresses/relays keyed by the ed25519 node_id; **dial-by-node_id-alone** (resolve reach at dial time). The node_id ties to the DID `#iroh` VM (same Ed25519 family Tiles uses).
- Implement the `dial_stream` `Iroh` arm (stubbed in #274) → real iroh dial.

## Superset
We speak iroh **and** moq-over-quinn/WebTransport; Tiles speaks iroh. Shared substrate = iroh; we add the rest. (Relay distinction: iroh DERP = packet/NAT relay ≠ moq content relay #134/#168 — they compose.)

## Relates
#131, #137 (the de-iroh gate now has a real iroh plane to gate), #156 (re-dial-by-key), #200 (QUIC RawPublicKey), #274 (`dial_stream` Iroh arm), #279. Tiles interop.

## Acceptance
- Daemon binds iroh; a peer dials our `moql`/RPC over iroh.
- pkarr resolves our node_id → current addresses; `#iroh` VM published in the DID doc.
- Interop verified against an iroh peer (Tiles-style dial-by-Ed25519).

ewindisch:
**Status review (post-0.5.0 merge #308): core implementation complete; remaining = live interop validation + polish.**

**Done (in `main`):**
- `IrohSubstrate` bound + served in the daemon (OAuth service), both ALPNs (`hyprstream-rpc/1` + `moql`) — `build_oauth_iroh_substrate`, gated on `[quic] iroh = true`.
[...]

Issue #281, opened 2026-06-17
Interop: admit + resolve did:key (Ed25519) peers as a subset of our DID model (Tiles)


ewindisch:
Part of epic #131. Interop with Tiles (tiles.run) — see `docs/interop/tiles-alignment.md`.

## Context
Tiles uses `did:key` (Ed25519) for device/account identity (+ ATProto). Our did:web/did:plc model is a **superset** — we interoperate by admitting + resolving `did:key` peers, not by adopting a second identity model.

## Superset principle
`did:key` ⊂ our DID model: a `did:key` is just an Ed25519 key-as-identity (self-certifying), equivalent to one of our `#mesh`/`#iroh`-style Ed25519 VMs. No new model — route it through existing key-binding paths.

## Scope
- **Admission (#137):** accept a `did:key` peer whose Ed25519 channel key matches its `did:key` (the key *is* the identity), under Casbin origin/policy.
- **Resolution (#279):** decode a `did:key` string → Ed25519 verifying key (multicodec `0xed01` + multibase). Note `did:key` carries no transport — reach comes from iroh discovery/linking, not the DID string.
- **Encoding (#280):** our Multikey/multicodec emission makes our keys `did:key`-shaped (reverse direction interop).

## Relates
#137, #279, #280, #131. Tiles interop.

## Acceptance
- A `did:key` (Ed25519) peer is admitted via #137 + Casbin.
- Resolver decodes a `did:key` → Ed25519 key; round-trips our Ed25519 VM ↔ `did:key`.

ewindisch:
Landed in `main` via the 0.5.0 epic merge (#308). Closing as completed.

Issue #280, opened 2026-06-17
DID VMs: emit Multikey (not Ed25519VerificationKey2020) + confirm #mesh ML-DSA uses registered ML-DSA-65 multicodec


ewindisch:
Part of epic #131. Surfaced checking the DID doc against the atproto DID spec (https://atproto.com/specs/did).

## Problem
The atproto spec standardizes `#atproto` as a `Multikey` verificationMethod, and we emit it correctly. But our ed25519 verification methods — including `#mesh` — emit `type: "Ed25519VerificationKey2020"` (`did_document.rs:131-138`, `ed25519_verification_method`), the older/deprecated VM type. This is inconsistent with `#atproto` (`Multikey`) and less modern-conformant. The `publicKeyMultibase` we already emit is the multicodec-prefixed form `Multikey` expects (`ed25519_to_multibase` = `z‖0xed01‖key`), so the switch is near-free.

The `#mesh`/`#iroh` **fragment names** and the `IrohTransport`/`QuicTransport` **service types** stay — they are valid, spec-permitted extensions ("other protocol and application features may make use of verificationMethod keys and service entries") and there is no cross-tool standard for them. This ticket is **VM type + key multicodec only**.

## Scope
1. Emit ed25519 (and other non-`#atproto`) verificationMethods as `type: "Multikey"` (keep `publicKeyMultibase` — already multicodec-prefixed). Covers `#mesh` + the ed25519 VMs in `did_document.rs`.
2. Confirm/ensure the `#mesh`-bound **ML-DSA-65** key (the #154/#157 key bundle) is published as a `Multikey` with the **registered ML-DSA-65 multicodec** so PQ-aware DID tooling can parse it. If currently bespoke, align it.
3. Keep the `JsonWebKey2020` JWK fallback (non-multibase consumers).
4. Decode side (`service_entry.rs`, DidWebResolver #279, #157 trust-store population): accept `Multikey`; tolerate `Ed25519VerificationKey2020` during transition.

## Dependencies / ordering
- Refines #154 (the DID node-key bundle, CLOSED).
- **Should land before/with #157** — #157 populates `KeyedPqTrustStore` from the `#mesh` ML-DSA key, so the multicodec encoding must be settled there.
- Feeds #137 (admission matches peer key ↔ DID VM), #279 (DidWebResolver parses VMs), #185 (cert↔DID).
- Refs: atproto DID spec, W3C DID Core, Multikey / multicodec registry.

## Acceptance
[...]

Issue #137, opened 2026-06-03
M3: Federation gate over DID/key binding (transport-agnostic, de-iroh)


ewindisch:
> [!warning] **SUPERSEDED / RE-SCOPED 2026-06-10** — see the consolidated plan (`docs/plans/2026-06-05-zmq-replacement-consolidated.md`, milestone M3) and the review comments below.
>
> Generalize off iroh: this is the **federation admission gate over any `Session`**, bound to **DID-document key material** (not an iroh-specific node-id check). Two-stage admission still applies, but the trust anchor is the resolved did:web/did:* doc's verification method (#154), and the channel key must bind to that doc VM (#185 cert↔DID binding). `subject_pattern` / RFC 6454 origin semantics unchanged.
>
> **Deps:** #154 (DID identity doc — the thing we admit against) · #185 (cert-to-DID binding).
>
> ---
> *(original body retained below for history)*

Part of #131. Blocked by #132 #133.

§4.2 `subject_pattern` stays unchanged (RFC 6454 origin, wildcards work).

Two-stage admission at iroh transport:
1. **Origin admission** (existing): is `https://hs.partner.org` allowed by Casbin/tuple rule?
2. **Key binding** (new, small): does the connecting peer's NodeId (Ed25519 pubkey) match a JWK published at `<origin>/.well-known/jwks.json` (tagged `use:"fed"` or carrying a known `kid`)?

Reuse the existing `jwks_uri` cache (from CIMD `private_key_jwt` work) — no new fetch infra, no new cache, no new rotation pipeline.

Each instance publishes its own NodeId as one of its JWKs (federation role) and as a `verificationMethod` in its did:web doc.

Fail-closed if either stage fails (§4.4 posture preserved).

**Exit criterion**: two instances on different NATs federate via iroh + pkarr; PolicyService admits via origin rule; transport gate matches connecting NodeId against peer's published JWKS.

ewindisch:
**RE-SCOPED (consolidated plan):** generalize off iroh. Federation/peer trust binds to the **DID/key** (did:web primary; resolve → doc → `verificationMethod`) over **any** transport Session, not iroh NodeId/pkarr specifically. Aligns with #154 (did:web atproto-compatible identity doc).

ewindisch:
[...]

PR #295, opened 2026-06-18
feat(#281): admit + resolve did:key (Ed25519) peers — Tiles interop

ewindisch:
Tiles uses did:key (Ed25519) for device/account identity. Our did:web/did:plc
model is a superset: a did:key is just an Ed25519 key-as-identity (self-certifying),
equivalent to one of our #mesh/#iroh Ed25519 VMs. No second identity model — route
it through the existing #137 key-binding path. did:key carries no transport (reach
comes from iroh discovery, #282).

- did_web.rs: did_key_to_ed25519 (strip "did:key:" + fragment, decode via the
  existing decode_ed25519_multikey — multibase z + multicodec ed25519-pub 0xed01;
  pure, no fetch; rejects non-did:key / wrong codec / bad multibase / bad length);
  ed25519_to_did_key ("did:key:z" + base58btc(0xed01 ‖ key)); is_did_key predicate.
  Multicodec is the SAME source of truth as #280/#137 (no duplicated 0xed01); the
  encode is byte-identical to #280's ed25519_to_multibase payload.
- admission.rs: did:key arm at the top of admit_key_against_did — if is_did_key,
  decode the key from the DID and admit iff it equals the peer's authenticated
  channel key (self-certifying), with NO resolver call / no fetch; mismatch or
  parse error fails closed. Stage 1 (origin/Casbin) still runs first. did:web keeps
  the resolver+JWKS path.

Tests: 8 did_web (known-fixture decode, random round-trip, fragment strip, reject
non-did:key/wrong-codec/bad-length/non-base58btc, did:key ↔ #280 VM Multikey
equivalence); 4 admission (match→admit, mismatch→reject, did:key still enforces
stage-1 origin, invalid→fail-closed; a NeverResolve double proves no fetch on the
did:key path). Native + wasm32 build clean; clippy clean.

Issue #281. Part of epic #131.

### DIFF
--- a/crates/hyprstream-rpc/src/admission.rs
+++ b/crates/hyprstream-rpc/src/admission.rs
@@ -19,12 +19,21 @@
 //!
 //! 2. **Key binding** — does the peer's authenticated channel key (the raw
 //!    Ed25519 public key it proved possession of when establishing the session /
-//!    signing its envelopes) match a `verificationMethod` published in that
-//!    peer's resolved DID document (the `#mesh` / `#iroh` Ed25519 VM), OR an
-//!    Ed25519 key in a federation JWKS the caller supplies? The DID document is
-//!    resolved via the [`crate::did_web::DidWebResolver`] landed in #279 (no new
-//!    fetch/cache infra); the VM extraction is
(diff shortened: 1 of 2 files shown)


ewindisch:

Issue #200, opened 2026-06-11
Implement QuicServerAuth::RawPublicKey (RFC 7250) for the QUIC RPC transport


ewindisch:
**Implement `QuicServerAuth::RawPublicKey` — RFC 7250 raw-public-key identity binding for the QUIC RPC transport.**

The `QuicServerAuth` policy enum (in `transport/mod.rs`) already defines the `RawPublicKey([u8;32])` arm, but `LazyQuinnTransport`/`dial()` currently bail on it. This ticket implements it.

**Why:** cert-hash pinning (`Pinned`) authenticates the TLS *channel* ("a server presenting this cert"), not the peer's DID *identity* — the #185 gap. RFC 7250 binds the handshake to the server's Ed25519 public key, so the peer must prove key possession (this is exactly what iroh does, verified against iroh's source).

**Why it's non-trivial (scoped during the #151(a) server-auth refactor):**
- `web_transport_quinn::ClientBuilder` exposes no raw-key / custom-verifier hook (only WebPKI / cert-hashes / no-verify). So this needs a **custom rustls `ClientConfig`** with a hand-written RFC 7250 verifier (adapt iroh's `verify_tls13_signature_with_raw_key` + SPKI-equality, in `iroh/src/tls/verifier.rs`), a **raw `quinn` connect**, and `web_transport_quinn::Session::raw(conn, ...)` to wrap it.
- It needs a **server-side change**: `QuinnRpcServer` must present a raw public key (RFC 7250 server config + `Session::raw`), separate from its current X.509/WebTransport path.
- It is **browser-incompatible** (WebTransport mandates `serverCertificateHashes`), so it's a native-only mode; browser/public peers keep `Pinned`/`WebPki`.
- It **overlaps iroh**, which is quinn + RFC 7250 + discovery and already provides identity-bound transport. The only delta is a direct identity-bound QUIC dial *without* iroh's discovery/relay machinery.

**Security note:** a hand-written TLS verifier is security-critical — must be reviewed hard, with constant-time key comparison and correct handshake-signature verification.

**Decision (2026-06-10):** deferred — use iroh for identity-bound native transport today. This ticket tracks the eventual non-iroh direct-dial path. Until then `dial()` fails fast on `RawPublicKey`.

ewindisch:
[...]

PR #296, opened 2026-06-18
feat(#282): activate iroh substrate + pkarr discovery; wire #137 gate into iroh accept

ewindisch:
Activate the iroh P2P plane (was test-only) as a shared substrate for Tiles
interop, and give #137's federation gate the real per-connection channel key it
lacked on the quinn path. Config-gated: [quic] iroh (serde default false) — the
quinn/WebTransport-only baseline is unchanged; an operator opts in.

1. Daemon bind (native-only, fail-soft): when [quic] iroh=true the spawner binds
   IrohSubstrate in parallel to the quinn endpoint for each QUIC-enabled service,
   serving both ALPNs (hyprstream-rpc/1 + moql) with the same processor + shared
   global moq Origin, and installs the process-wide iroh client endpoint. OAuth
   (the DID-controller identity) binds its own canonical federation substrate so
   the advertised #iroh node_id has a live listener. Bind failure → quinn continues.
2. dial_stream Iroh arm (replaces the #274 stub): real iroh moql dial via a
   MoqStreamSession{Quinn|Iroh} enum with connect_moq() dispatch. Loopback
   endpoint-to-endpoint moql round-trip tested.
3. DID advertisement: root_did_document advertises the #iroh Ed25519 VM (= node_id
   as Multikey) + an IrohTransport entry accept:[hyprstream-rpc/1, moql], only when
   bound. Verified node_id == the OAuth RPC service signing key (state.signing_key)
   so the advertised endpoint id matches the live listener.
4. #137 in the LIVE iroh accept path: new object-safe IrohPeerAdmission trait in
   hyprstream-rpc (avoids the rpc→hyprstream cycle); both iroh handlers call it with
   remote_id().as_bytes() and DROP the connection fail-closed on reject, before any
   stream is served. The hyprstream adapter (IrohFederationAdmission) maps remote_id()
   → self-certifying did:key (#281) as both the key-binding identity (stage 2,
   self-cert) and the stage-1 federation:register subject, then runs the #137 gate.
   remote_id() is the authenticated Ed25519 (iroh QUIC TLS bound it) — no #200/#185
   seam. The per-request service layer still re-verifies envelope signer == remote_id().
5. pkarr discovery: enabled via iroh presets::N0 (PkarrPublisher + n0-DNS resolver) —
   publish node_id→addresses + dial-by-node_id. Pure node_id↔#iroh/did:key binding
   unit-tested.

Documented seams (infra/schema, not faked): (a) pkarr live DHT round-trip + NAT
traversal need real infra / a second peer — can't be hermetically unit-tested.
(b) wire-reach iroh: reach_to_transport_config's Iroh arm stays None because the
codegen wire reach enum's Iroh is a unit variant with no nodeId/relay payload
(stream.capnp) — a StreamInfo can't yet publish a dialable iroh reach (dial_stream's
local iroh arm works + is tested). (c) non-OAuth spawned services set
iroh_admission=None (accept-open) — threading a PolicyClient through every factory
is a wider change; OAuth (the federation controller) has the full gate. All
mitigated by iroh being default-off.

Tests: dial_stream iroh moql loopback + no-reach fail-fast; iroh accept-path
admit-by-remote_id + reject-fail-closed; IrohPeerAdmission hook open/allow/deny;
federation_admission iroh admit/reject + node_id↔did:key/#iroh binding;
did_document #iroh VM advertised-when-bound/omitted-when-not. Native + wasm32 build
clean (no iroh wasm leak); clippy clean; hyprstream-rpc 410 pass (1 known moq_event
flake, verified on baseline).

Issue #282. Part of epic #131.

### DIFF
--- a/crates/hyprstream-rpc/src/dial.rs
+++ b/crates/hyprstream-rpc/src/dial.rs
@@ -220,6 +220,35 @@ where
 /// streaming plane (default path → RPC). Both server and client use this.
 pub const MOQ_PATH: &str = "/moq";
 
+/// A dialed moq streaming session, erased over the concrete transport.
+///
+/// [`dial_stream`] returns this so a caller can dial the streaming plane over
+/// either quinn/WebTransport (`/moq` path-dispatch) **or** iroh (the `moql`
+/// ALPN) and hand the result straight to `moq_net::Client::connect`, which is
+/// generic over `web_transport_trait::Session`. The two underlying session
+/// types (`web_transport_quinn::Session` / `web_transport_iroh::Session`) do not
+/// share a concrete type, so this enum dispatches at the `connect` call.
+pub enum MoqStreamSession {
+    /// quinn / WebTransport `/moq` session (#274).
+    Quinn(web_transport_quinn::Session),
+    /// iroh `moql`-ALPN session (#282).
+    Iroh(web_transport_iroh::Session),
+}
+
+impl MoqStreamSession {
+    /// Run the moq-lite handshake over this session, returning the live
+    /// `moq_net::Session`. Dispatches to the concrete transport's `connect`.
+    pub async fn connect_moq(
+        self,
+        client: &moq_net::Client,
+    ) -> std::result::Result<moq_net::Session, moq_net::Error> {
+        match self {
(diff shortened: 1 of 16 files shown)


ewindisch:

PR #301, opened 2026-06-18
perf(infer): chunk FP32 attention over query axis to cap prefill VRAM

ewindisch:
Long-prompt prefill materialized the full [batch, heads, q_seq, kv_seq] FP32
score + softmax tensors, making peak VRAM O(q_seq * kv_seq). For an 8K-token
prompt these grow past ~13 GB per full-attention layer and the CUDA caching
allocator retains the high-water mark.

Process the query axis in chunks of 1024 so the score/softmax tensors are
O(ATTN_CHUNK * kv_seq) instead. Causal masking is identical: the per-chunk
mask tril(start) is the exact restriction of the full tril(0) to rows
[start, start+cur). The decode path (q_seq == 1) and short prompts take the
single-shot branch unchanged, so output and numerics are byte-for-byte
preserved there. Backend-agnostic (plain matmul/softmax) — unlike SDPA's
flash/efficient kernels, which panic on this ROCm/libtorch build.

Applied to both the Llama/Qwen path (llama.rs compute_attention) and the
Qwen3.5 hybrid full-attention layers (qwen3_5.rs).

### DIFF
--- a/crates/hyprstream/src/runtime/architectures/llama.rs
+++ b/crates/hyprstream/src/runtime/architectures/llama.rs
@@ -529,21 +529,15 @@ impl LlamaAttention {
             (k_for_attn, v_for_attn)
         };
 
-        // Manual attention implementation
-        // Note: SDPA was tested but:
-        // - efficient_attention panics on this ROCm/libtorch version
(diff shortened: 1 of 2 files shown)


ewindisch:

PR #302, opened 2026-06-18
fix(stream): read moq groups by exact sequence (get_group) to stop tail-group skip

ewindisch:
The inference stream's HMAC is an ordered, gap-fatal chain (mac_n =
HMAC(key, mac_{n-1} || data_n)). The consumer read groups with
TrackConsumer::next_group(), whose cursor returns the first group with
sequence >= its monotonic cursor in *arrival* order. moq serves each Group on
its own QUIC uni-stream, so Groups can arrive out of order; when the small
terminal Group (the always-retained max_sequence) arrived before the
penultimate one, next_group returned it and advanced the cursor past the
penultimate, permanently skipping it — fatally breaking the chain and failing
every inference after the first ('MAC verification failed' / 'timeout waiting
for broadcast'). The first stream worked only because its consumer subscribed
during production and read every Group live, in order.

Switch both networked and UDS stream consumers to a get_group(expected_seq)
loop: get_group waits for each exact sequence (even out-of-order arrivals) and
returns None only at/after final_sequence, guaranteeing in-order, gap-free
delivery. Fixes multi-turn TUI/CLI inference (#145). Validated: 10/10
consecutive inferences (was 1/8).

### DIFF
--- a/crates/hyprstream-rpc/src/moq_stream.rs
+++ b/crates/hyprstream-rpc/src/moq_stream.rs
@@ -714,16 +714,24 @@ async fn moq_stream_handle_task(
             return;
         }
     };
-    let mut track = match bc.subscribe_track(&Track::new(STREAM_TRACK)) {
+    let track = match bc.subscribe_track(&Track::new(STREAM_TRACK)) {
         Ok(t) => t,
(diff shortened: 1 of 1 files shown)


ewindisch:

PR #303, opened 2026-06-18
feat(tui): refresh model list on ModelList open via registry RPC

ewindisch:
The shell model list was captured once at startup and never re-scanned, so a
model cloned out-of-band (the CLI, or another TUI window) did not appear until
the TUI was restarted. The in-TUI clone flow already appended its result, but
external additions were invisible.

Add a RefreshFn to GitOps that re-queries the registry (registry.list() +
model status) — the same authoritative 'name:branch' source the initial list is
built from — and returns it via GitOpProgress::ModelList. ShellApp fires it when
the ModelList view opens; tick() merges the result, preserving the current
selection and the out-of-band git status (ahead/behind/dirty). Standalone mode
(no git ops) falls back to a synchronous discover_models() re-scan.

Merge logic extracted to a pure merge_models() helper with unit tests covering
additions, upstream removals, and selection-follows-moved-index.

### DIFF
--- a/crates/hyprstream-tui/src/main.rs
+++ b/crates/hyprstream-tui/src/main.rs
@@ -44,13 +44,12 @@ fn run_shell(mut args: impl Iterator<Item = String>) {
     };
 
     let (cols, rows) = terminal_size();
-    let models = hyprstream_tui::shell_app::discover_models(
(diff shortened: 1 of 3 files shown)


ewindisch:

PR #305, opened 2026-06-18
ewindisch/policy wildcard fix authz tests

ewindisch:
- **spike(policy): wildcard validation regression breaks runtime authz mutations**
- **fix(policy): drop wildcard ban; add atproto-scoped authz invariant tests**


ewindisch:

Issue #310, opened 2026-06-18
Epic: Multi-GPU & Multi-Host Inference (+ coupled training)


ewindisch:
Tracking epic for multi-GPU + multi-host inference, with coupled TTT and staged distributed training. Plan: `docs/plans/2026-06-18-multi-gpu-multi-host-inference-spike.md`.

**Scope:** ModelService(router) → InferenceService(owns 1..N GPUs) → TorchEngine(per-GPU). Intra-host pipeline (layer split) first, then SPIFFE-style identity + mesh policy, then a cross-host router, then inter-host pipeline for multi-TB models.

**North star:** multi-TB models. **v1 hardware:** 35 GB model on 2× MI210 (homogeneous, single host), then cross-vendor cross-machine (MI210 + RTX 5090).

**Settled decisions:** D1 go straight to 2b (validate vs single-GPU ground truth); D2 device-code OAuth first + pre-auth token for unattended; D3 `pds join/attach` CLI + wizard, semi-open authz for the homelab phase; D4 one SPIFFE trust domain per PDS (enables federation via did:web bundle + #137 gate); D5 param-server delta aggregation first.

## Sub-issues

**Prerequisites (gating)**
- [ ] #311 — Spike: single-process multi-device tch feasibility (gates 2b)
- [ ] #312 — Verify epic is the origin/main base + substrate present + MI210 P2P hop cost
- [ ] #328 — ClusterKeySource fail-unsafe fix (C-IDENT) — prereq of P2 mesh policy
- [ ] #130 — Policy wildcard regression (existing) — prereq of P2 mesh policy

**Phase 1 — intra-host multi-GPU (no networking)**
- [ ] #313 — DevicePool + `RuntimeConfig.devices`
- [ ] #314 — 2b intra-host pipeline: per-layer device_map + stage-aware load
- [ ] #315 — No-fragile-fallbacks hardening (require config.json/index.json, fail-fast)
- [ ] #316 — TTT-on-split: cross-device autograd over the partition
- [ ] #317 — Validate 2b (+2a replication) vs single-GPU ground truth on 2×MI210

**Phase 2 — identity + trust + networked node**
- [ ] #318 — PDS attach (Layer 1): `hyprstream pds join/attach` CLI + wizard
- [ ] #319 — Mesh policy (Layer 2): `mesh:rpc`/`inference:peer-call`, per-host SVID, deny-by-default
- [ ] #320 — InferenceService network endpoint + static reach registry
[...]

PR #330, opened 2026-06-18
perf(rpc): restore zero-copy Cap'n Proto framing on the streaming + control planes

ewindisch:
Restores the original zero-copy Cap'n Proto framing that the M1 (#151) generic RPC plane layered redundant copies over — including on the **streaming plane**, where pipeline activations / tensors flow. Surfaced while planning multi-GPU/host inference (epic #310). Wire-compatible; no format change.

## Why
A research spike confirmed moq-net already yields `bytes::Bytes` at the wire (`moq_stream.rs` `group.read_frame()`), so **every copy on the large/tensor path was in our own code** — no fork of `web-transport-trait` needed. (`web_transport_trait::RecvStream` v0.3.5 exposes only `read(&mut [u8])`, but that path is the 4 MiB control plane only.)

## Streaming plane (large/tensor path) — 3 copies eliminated
- **Borrowing verify:** new `StreamVerifier::verify_parts(topic, capnp, mac)` over slices; `verify(&[Vec<u8>])` is now a thin wrapper (generic-subscriber path + tests unchanged).
- **`verify_moq_frame`** passes slices into the moq `Bytes` frame — no per-frame `.to_vec()`.
- **Incremental MAC:** `keyed_mac_truncated_parts(key, &[prefix, capnp])` — byte-identical to MAC-of-concatenation (blake3 / HMAC are streamed), so it drops the `input` concat copy with **no wire change**. New equivalence test `test_keyed_mac_truncated_parts_equals_concat`.
- **`read_message_from_flat_slice`:** the `StreamBlock` reader borrows the frame instead of copying into an owned message — a tensor payload stays a view into the receive buffer end-to-end.

## Control plane
- **`read_to_cap`** reads straight into the output buffer (drops the scratch + `extend_from_slice` second copy; the one remaining kernel→userspace copy is forced by the external trait). Kept `MAX_FRAME_BYTES = 4 MiB` (#162).
- **`parse_stream_info`** uses `read_message_from_flat_slice`.

## Tests
`hyprstream-rpc` builds standalone (no libtorch). 413 lib tests pass; 18 changed-area tests green (MAC equivalence, tamper/ordering integrity, moq round-trips). Clippy clean. The 2 failures in a full run are pre-existing and unrelated (#148 moq_event "subscriber closed"; an env-parallel `paths` test).

## Follow-up (not in this PR)
Control-plane zero-copy via a `ZeroCopyRecv` extension trait (quinn `read_chunk -> Bytes`) — low value (4 MiB only); deferred.

Part of #310.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6323,7 +6323,7 @@ dependencies = [
 
 [[package]]
 name = "hyprstream"
-version = "0.4.0"
+version = "0.5.0"
 dependencies = [
  "anyhow",
  "async-stream",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -180,7 +180,11 @@ image = { version = "0.25", default-features = false, features = ["png", "jpeg"]
 # All three crates must track together — generated code from capnpc
 # targets specific internal APIs in capnp; mismatched majors break
 # RawStructSchema struct construction.
-capnp = "0.24"
+# `unaligned`: read capnp messages directly from network buffers (moq/quinn `Bytes`)
+# that can start at any byte offset — required for the zero-copy
+# `read_message_from_flat_slice` path on the streaming + RPC planes (otherwise capnp
+# errors with UnalignedSegment). Byte-wise word access; keeps the read zero-copy.
+capnp = { version = "0.24", features = ["unaligned"] }
 capnp-futures = "0.24"
 
 # Utilities
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -183,7 +183,17 @@ image = { version = "0.25", default-features = false, features = ["png", "jpeg"]
 # `unaligned`: read capnp messages directly from network buffers (moq/quinn `Bytes`)
 # that can start at any byte offset — required for the zero-copy
 # `read_message_from_flat_slice` path on the streaming + RPC planes (otherwise capnp
-# errors with UnalignedSegment). Byte-wise word access; keeps the read zero-copy.
+# errors with UnalignedSegment).
+#
+# SAFE in Rust — and NOT the same as C++'s `-DCAPNP_ALLOW_UNALIGNED`. The C++ flag only
+# disables the alignment *check* and leaves genuine UB (unaligned access is UB in C/C++;
+# compilers may emit SIMD / assume a pointer's low bits are zero), which is why capnp's
+# author recommends copying into an aligned buffer instead (StackOverflow Q 75006774).
+# Rust's `unaligned` feature instead changes the access *mechanism*: each word becomes
+# `Raw = [u8; N]` (alignment 1) read via `from_le_bytes(*raw)` (capnp `private/primitive.rs`)
+# — the canonical, fully-defined Rust idiom for unaligned reads, with no UB. On x86_64/ARM64
+# it lowers to a single hardware unaligned load (~free) and does NOT memcpy or sacrifice
+# zero-copy. See PR #330.
 capnp = { version = "0.24", features = ["unaligned"] }
 capnp-futures = "0.24"
 
--- a/crates/hyprstream-rpc/src/crypto/backend.rs
+++ b/crates/hyprstream-rpc/src/crypto/backend.rs
@@ -132,6 +132,44 @@ pub fn keyed_mac_truncated(key: &[u8; 32], data: &[u8]) -> [u8; 16] {
     output
 }
 
+/// Multi-part variant of [`keyed_mac_truncated`] — authenticates the
+/// concatenation of `parts` **without allocating a joined buffer**.
+///
+/// `keyed_mac_truncated_parts(k, &[a, b])` is byte-for-byte identical to
+/// `keyed_mac_truncated(k, &[a, b].concat())` (both Blake3's keyed hasher and
+/// HMAC are defined over the streamed input), so this is a zero-copy drop-in on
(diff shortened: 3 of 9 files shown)


ewindisch:
Update: control-plane zero-copy is now **included** in this PR (commit 12ec73f3), not deferred.

A research spike found the `ZeroCopyRecv` extension trait is unnecessary: `read_chunk(&mut self, max) -> Result<Option<Bytes>>` is already a **defaulted method on `web_transport_trait::RecvStream` v0.3.5** (lib.rs:272), overridden zero-copy by quinn (recv.rs:141 → `Bytes` referencing the QUIC recv buffer) and iroh (recv.rs:105); UDS inherits the one-copy default. The earlier 'only `read(&mut [u8])`' impression came from reading the UDS *impl* (which implements only the 3 required methods and inherits the default).

So `read_to_cap` just switches to `read_chunk`: a single-chunk control message (the common case) is now **fully zero-copy** (no alloc, no copy); multi-chunk pays at most one concatenation. No new trait, no bound, no fork. 4 MiB cap (#162) preserved. 61 RPC/framing tests pass (uds/iroh round-trips, zmtp).

Bonus noted for a follow-up: `SendStream::write_chunk(Bytes)` exists (quinn/iroh zero-copy override) — switching the response writer to it would make the reply path zero-copy too.

ewindisch:
Added the reply/request **write** path too (commit c0f7f4e8): both `SendStream` writes now use `write_chunk(Bytes)` instead of `write_all(&buf)` — the reply envelope (already `Bytes`) is moved not cloned, and the request `Bytes::from(Vec)` is a zero-copy move. `write_chunk` is the defaulted trait method (quinn/iroh zero-copy override; UDS one-copy default). The control plane is now zero-copy on **both read and write**; combined with the streaming-plane commit, the full RPC + streaming path is zero-copy end-to-end. 20 RPC round-trip tests pass (uds/iroh req-resp).

ewindisch:
## Self-review (caught one Critical, now fixed in 0dc84919)

**CRITICAL — alignment (fixed):** `read_message_from_flat_slice` requires 8-byte alignment unless capnp's `unaligned` feature is on; otherwise it returns `UnalignedSegment`. The zero-copy path parses `Bytes` straight from the moq/quinn receive buffer, which can start at **any** offset — so the streaming + RPC parse would have errored on unaligned frames in production. Masked in tests because in-process `Vec`-backed buffers happen to be aligned. Proven with a deliberately-offset-by-1 test (`flat_slice_parses_unaligned_block`: `UnalignedSegment` without the feature, passes with it). Fix: enable `capnp/unaligned` (byte-wise word access — keeps the read zero-copy).

**Verified OK:**
- **MAC equivalence:** `keyed_mac_truncated_parts` is byte-identical to MAC-of-concat (blake3/HMAC streamed) — `test_keyed_mac_truncated_parts_equals_concat`; tamper/ordering integrity tests still pass.
- **`read_to_cap` (read_chunk):** cap enforced via `max = cap-total+1` then `total>cap` bail (off-by-one matches the old `>cap`); single-chunk returns the `Bytes` verbatim (zero-copy); `(Some,Some)` arm genuinely unreachable; empty/FIN-first → `Bytes::new()`.
- **`write_chunk`:** reply `Bytes` moved not cloned; request `Bytes::from(Vec)` is a move; UDS uses the trait's one-copy default.
- **`verify_parts` split:** `verify(&[Vec])` wrapper keeps the generic-subscriber path + tests unchanged; borrow lifetimes end before return (payload copied out).

**Tests:** 415 lib pass; the one failure (`moq_event_pub_sub_roundtrip`) is pre-existing #148, unrelated (passes/fails independent of this diff). Clippy clean.

ewindisch:
### Note: isn't unaligned capnp UB? (re: the well-known C++ SO answer)

A reviewer familiar with [SO #75006774](https://stackoverflow.com/questions/75006774) will recall capnp's author says the fix for unaligned buffers is to **memcpy into an aligned buffer**, because `-DCAPNP_ALLOW_UNALIGNED` only disables the check and leaves **genuine UB** (unaligned access is UB in C/C++; the compiler may emit SIMD or assume a pointer's low bits are zero). So: doesn't enabling `unaligned` here put us in the same UB trap, and shouldn't we copy instead?

**No — that's a C++ constraint that does not carry to Rust.** The C++ flag and the Rust feature are different things:
- **C++ `CAPNP_ALLOW_UNALIGNED`:** disables the assertion, keeps the `reinterpret_cast<word*>` → still UB. Hence Kenton's 'copy instead'.
- **Rust `unaligned` feature:** changes the *access mechanism*. Per `capnp-0.24.1/private/primitive.rs`, a word becomes `Raw = [u8; N]` (alignment **1**) read via `u64::from_le_bytes(*raw)` — the canonical, **fully-defined** Rust idiom for unaligned reads. No `reinterpret_cast`, no UB. `&[u8; 8]` is valid at any address; the read is defined behavior.

So in Rust we get what the C++ asker couldn't: **safe zero-copy from unaligned wire buffers**, no memcpy. On x86_64/ARM64 (our targets) it lowers to a single hardware unaligned load (~free). The copy-into-aligned approach (Kenton's C++ answer) is a strictly-worse fallback here — it would re-add the per-message copy this PR removes.

(Documented at the feature bring-in in `Cargo.toml`. For multi-GB tensors later, framing the bulk *outside* capnp remains the better architecture — to dodge capnp's message-size/traversal limits, not for alignment.)

Issue #393, opened 2026-06-23
F2-b/c: MoQ-native event plane (selective/named-track) + optional firehose adapter — the atproto divergence


ewindisch:
Part of #387 (Phase 2b+c). Plan §3. **Deliberate atproto divergence.**

**MoQ is the native event plane** (demand-driven, O(subscribed), selective/named-track subscription of signed `commit`/`refUpdate` events → feeds the scheduling directory's OID→who-has-it map). **The ATProto firehose is an OPTIONAL consume-only interop adapter** (harvest `sh.tangled.git.refUpdate` as extra untrusted hints) — NOT produced by default, NOT depended on.

Rationale: the firehose's fan-in-everything model (O(whole-network) per relay) is the social-scale centralization the federated-scheduling design rejected; its main virtue (total-order+cursor backfill) is redundant w/ git-CAS. Keeps atproto's data model (DID, signed records, MST); rejects its default event transport for one that federates past social scale. **Stance: MoQ-native, firehose-bridgeable** (pushback for atproto devs).

Load-bearing: MoQ event bus must support selective/named-track subscription (verify); **#145 (gap-fatal late-join) becomes load-bearing** if MoQ is primary — must fix. Builds on #355's moq substrate. Replaces bespoke gossip.

ewindisch:
Decision: A (firehose-backfill-on-first-join). Firehose for cold-start history, MoQ for live. Per-OID track restructure is the forced build gap.

ewindisch:
Merged via #417.

Issue #358, opened 2026-06-19
S3: moq relay (rendezvous) for browsers + fan-out + anonymization


ewindisch:
PR-D. Port the existing event-plane rendezvous (moq_event.rs with_origin ingest-and-re-serve) to the inference plane: a relaying node dial_streams the upstream producer, subscribes to broadcastPath, re-announces into its local origin; advertise Destination Role::Relay; consumer relay-selection. WebTransport-served for browsers (no wasm iroh). Relay stays content-blind via S5 AEAD. Mostly net-new but templated. Topology/authz = human-gated. Part of #355. Plan: docs/plans/2026-06-19-streaming-plane-restoration-epic.md

ewindisch:
Merged via PR #378 into `feature/streaming-plane-restoration`. Restored moq relay rendezvous (PR-D): relay client linking a producer origin UP to a remote relay via `dial_stream` (`serve_origin_to_relay_background`/`run_relay_announce_link`, modeled on event-plane `with_origin`); `producer_reach()` now constructs `Role::Relay` via the shared `service_entry` codec (stream & DID addresses can't drift); `select_reach` does QoS-aware direct-vs-relay ordering (Pipe/Live→direct-first, Job/Log→relay-first) as a **pure stable permutation** of the service-advertised reach (server authority: never invents/drops a reach → a relay-only/anonymized stream stays relay-only); `connect_moq_reach` gives automatic direct↔relay fallback on the same `broadcastPath`; relay server mode via `QuinnRpcServer::with_moq_relay` + `[quic] relay="<uri>"` daemon wiring. **Relay-blind by construction** — AEAD-sealed+HMAC'd at source, relay holds no enc_key/mac_key, forwards opaque Bytes (new `moq_relay_rendezvous` test asserts relay-forwarded bytes fail verification under a wrong key). **No capnp/RPC surface change** (populated existing Role::Relay/Destination/broadcastPath). Verified: native + wasm32 build, new rendezvous test + select_reach/server-authority/relay-codec unit tests pass, clippy clean. (1 lib failure `moq_event_pub_sub_roundtrip` is pre-existing on the branch base in the untouched event plane — #148.)

ewindisch:
## Topology/authz contract — DECIDED (2026-06-25, human sign-off)

Corrects an earlier draft that proposed Casbin-gating relay + per-PDS relays. Both were wrong.

### 1. Relay is content-blind by construction — relay authz is NOT a content control
[...]

Issue #357, opened 2026-06-19
S2: iroh+pkarr direct reach for native peers (TransportConfig.iroh node_id)


ewindisch:
PR-E-direct. Add node_id to TransportConfig.iroh (today Void, cannot advertise an iroh reach); advertise the iroh reach in announcedAt; wire the dial_stream iroh arm to consume it; pkarr dial-by-node_id. Gives native peers NAT + cross-instance + discovery with NO content relay. The multi-gpu mesh reach (#320) depends on this. capnp schema field = human-gated. Part of #355. Plan: docs/plans/2026-06-19-streaming-plane-restoration-epic.md

ewindisch:
Implemented in PR #369 (squash-merged to feature/streaming-plane-restoration). TransportConfig.iroh Void→IrohReach{nodeId @0 :Data fixedSize(32)}; producer_reach() advertises an iroh Destination ahead of Quic when the substrate is bound; reach_to_transport_config resolves it; connect_moq_reach/dial_stream dial-by-node_id-alone via pkarr/N0 discovery (S1's Quic+UDS fallbacks preserved; DH-blind topic untouched; no relay). Unit tests green; full pkarr-discovered 2-node dial needs live validation. Unblocks #320 (multi-gpu mesh reach). Closes #357.

Issue #356, opened 2026-06-19
S1: unify StreamInfo reach across tui/notification; drop dead UDS/endpoint fields (fixes #142 + TUI cross-process timeout)


ewindisch:
PR-C. Migrate tui.capnp (subEndpoint/moqBroadcastPath/moqUdsPath) and notification.capnp (dead streamEndpoint @2 + moqUdsPath @3/moqBroadcastPath @4) onto the networked announcedAt:List(Destination) reach + a single producer_reach() builder; delete dead UDS/endpoint fields. Fixes #142 (cross-instance model-load notifications silently fail) and the TUI cross-process UDS timeout. capnp schema spin = human-gated. Part of #355. Plan: docs/plans/2026-06-19-streaming-plane-restoration-epic.md

ewindisch:
Implemented in PR #368 (squash-merged to feature/streaming-plane-restoration). Unified StreamInfo reach across tui.capnp + notification.capnp onto announcedAt:List(Destination) (no UDS-on-wire); single connect_moq_reach() resolver (networked-first, UDS-local fallback, fail-closed). FIXES #142 (git_handlers consumes networked reach) + the TUI cross-process timeout, with regression tests. Left producer_reach iroh arm for S2. Closes #356.

PR #469, opened 2026-06-26
fix(#148): moq_event subscriber wait for events subtree

### LINKED ISSUE #148: moq Phase 6.5: Extract transport-neutral survivors from zmtp_quic.rs

ewindisch:
## Summary

- **Root cause**: `scope_consumer_for_patterns` did two sequential `scope()` calls. After the first scope to `"local/events"`, the consumer's nodes are keyed with absolute paths, so the second call with a relative source name like `"worker"` found no matching node and returned `None` — causing the subscriber task to abort immediately with "origin has no event scope".
- **Fix**: Build absolute broadcast paths (`"local/events/{source}"`) and call `scope()` only once on the root consumer. A single scope on a root consumer always creates the leaf node via the shared origin tree; the `announced()` loop then waits correctly for the broadcast to appear.
- **Test isolation fix**: The test was using `MoqEventSubscriber::new()` which relies on the process-global `OnceLock`. In a multi-test binary, subsequent `init_global_moq_event_origin` calls are silently ignored, so the test picked up a stale origin. Switched to `new_for_test(origin.consumer())` to match the isolation pattern all other moq_event tests use.
- **Race exercise**: The test now deliberately creates the subscriber before the publisher to verify the fix handles the subscriber-before-publisher race (#148 proper).

## Test plan

- [ ] `cargo test -p hyprstream-rpc moq_event_pub_sub_roundtrip` passes 3×/3× (verified locally)
- [ ] All 10 `moq_event` unit tests pass (no regressions)
- [ ] CI on `ewindisch/355-mergeup` (#467) should unblock once this lands on main

Fixes #148

### DIFF
--- a/crates/hyprstream-rpc/src/moq_event.rs
+++ b/crates/hyprstream-rpc/src/moq_event.rs
@@ -843,14 +843,23 @@ async fn run_subscriber_task(
 
     let patterns = Arc::new(patterns);
 
-    // Scope the consumer to the `local/events` subtree, then further narrow by
-    // source names extracted from the patterns.
+    // Scope the consumer to the relevant `local/events/{source}` broadcast paths.
+    //
+    // #148: `scope_consumer_for_patterns` now does a SINGLE scope() call on the root
+    // consumer using absolute broadcast paths (`"local/events/{source}"`). The scoped
+    // consumer registers at the leaf node of the shared origin tree, so:
+    //   • if the publisher already created the broadcast, `consume_initial` replays it
+    //     into this consumer's pending queue immediately;
+    //   • if the publisher hasn't registered the broadcast yet, the consumer's `announced()`
+    //     loop below will block until the announcement arrives — no race, no abort.
+    //
+    // `scope()` on a root consumer always succeeds (it creates the leaf node), so
+    // `scope_consumer_for_patterns` only returns `None` when the origin is closed.
     match scope_consumer_for_patterns(&consumer, &patterns) {
         Some(c) => consumer = c,
         None => {
-            // Origin has no `local/events` tree yet — using the root consumer would
-            // expose all RPC traffic to the event channel. Abort instead.
-            tracing::warn!("event subscriber: origin has no event scope; subscriber task exiting");
+            // Origin is closed — nothing to subscribe to.
+            tracing::warn!("event subscriber: origin closed; subscriber task exiting");
             return;
         }
     }
@@ -878,21 +887,30 @@ async fn run_subscriber_task(
 
 /// Scope the origin consumer to `local/events/{sources}` based on patterns.
 ///
-/// If any pattern is `""` (subscribe-all), returns `None` and the caller uses
-/// an unscoped consumer rooted at `local/events`.
+/// Returns a consumer scoped to all source-specific broadcast paths that the
+/// given patterns require. If any pattern is `""` (subscribe-all), returns a
+/// consumer scoped to the whole `local/events` subtree.
+///
+/// # Two-level scope fix
+///
+/// The original code did two `scope()` calls: first to `"local/events"`, then a
+/// second relative scope to each source name. The second call failed silently
+/// because after the first `scope()` the consumer's internal nodes are keyed
+/// with absolute paths (e.g. `"local/events"`), not relative ones — so passing
+/// `"worker"` to the second `scope()` finds no matching node and returns `None`.
+///
+/// The fix: build full absolute broadcast paths (`"local/events/{source}"`) and
+/// call `scope()` only ONCE on the original root consumer.
 fn scope_consumer_for_patterns(
     consumer: &OriginConsumer,
     patterns: &[String],
 ) -> Option<OriginConsumer> {
-    // Scope to local/events first
-    let events_consumer = consumer.scope(&[Path::new(EVENT_PREFIX)])?;
-
-    // If subscribe-all pattern present, return events-scoped consumer
+    // If subscribe-all pattern present, scope to the whole local/events subtree.
     if patterns.iter().any(String::is_empty) {
-        return Some(events_consumer);
+        return consumer.scope(&[Path::new(EVENT_PREFIX)]);
     }
 
-    // Extract unique source names from patterns (e.g., "worker." → "worker")
+    // Extract unique source names from patterns (e.g., "worker." → "worker").
     let sources: Vec<String> = patterns
         .iter()
         .filter_map(|p| {
@@ -908,11 +926,20 @@ fn scope_consumer_for_patterns(
         .collect();
 
     if sources.is_empty() {
-        return Some(events_consumer);
+        // No recognizable source prefix in any pattern — subscribe to everything.
+        return consumer.scope(&[Path::new(EVENT_PREFIX)]);
     }
 
-    let path_refs: Vec<Path<'_>> = sources.iter().map(|s| Path::new(s.as_str())).collect();
-    events_consumer.scope(&path_refs)
+    // Build ABSOLUTE broadcast paths (`"local/events/{source}"`) so we can
+    // scope the root consumer in a single call — avoids the two-level-scope bug
+    // where relative names ("worker") don't match the absolute keys held in a
+    // previously-scoped consumer ("local/events").
+    let full_paths: Vec<String> = sources
+        .iter()
+        .map(|s| format!("{}/{}", EVENT_PREFIX, s))
+        .collect();
+    let path_refs: Vec<Path<'_>> = full_paths.iter().map(|s| Path::new(s.as_str())).collect();
+    consumer.scope(&path_refs)
 }
 
 /// Read all groups from a single source broadcast's `events` track and relay
@@ -1131,34 +1158,41 @@ mod tests {
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
     async fn moq_event_pub_sub_roundtrip() -> Result<()> {
         let origin = MoqEventOrigin::new();
-        init_global_moq_event_origin(origin.clone());
 
-        // Publisher
-        let mut pub_ = origin.publisher("worker")?;
-
-        // Subscriber
-        let mut sub = MoqEventSubscriber::new();
+        // Use new_for_test so this test has an isolated origin consumer and does
+        // not race with other tests over the process-global OnceLock (#148).
+        // The OnceLock is single-writer; a second init_global_moq_event_origin call
+        // in the same binary is silently ignored, so any test that relies on ::new()
+        // picking up the right global is inherently order-dependent.
+        let mut sub = MoqEventSubscriber::new_for_test(origin.consumer());
         sub.subscribe("worker.")?;
 
-        // Start subscriber background task (lazy; triggered by recv)
-        // Publish before recv so the broadcast is announced when the task starts.
+        // Publisher created AFTER the subscriber is set up so we can also exercise
+        // the #148 wait-for-subtree path: recv() starts the background task before
+        // the publisher has registered its broadcast.
+        // The subscriber task must wait until the broadcast appears rather than abort.
+        let recv_task = tokio::spawn(async move { sub.recv().await });
+
+        // Small yield to let the subscriber background task start and block on
+        // the wait-for-subtree loop before we create the publisher.
+        tokio::time::sleep(std::time::Duration::from_millis(5)).await;
+
+        // Publisher registers the broadcast (creates `local/events/worker`).
+        let mut pub_ = origin.publisher("worker")?;
         pub_.publish("sandbox123", "started", b"payload")?;
 
-        // Give the origin a moment to propagate the announcement.
-        tokio::time::sleep(std::time::Duration::from_millis(10)).await;
-
-        // Manually start the task by calling ensure_started and check recv.
         let result = tokio::time::timeout(
             std::time::Duration::from_secs(5),
-            sub.recv(),
+            recv_task,
         ).await;
 
         match result {
-            Ok(Ok((topic, payload))) => {
+            Ok(Ok(Ok((topic, payload)))) => {
                 assert_eq!(topic, "worker.sandbox123.started");
                 assert_eq!(payload, b"payload");
             }
-            Ok(Err(e)) => panic!("recv error: {e}"),
+            Ok(Ok(Err(e))) => panic!("recv error: {e}"),
+            Ok(Err(e)) => panic!("task panicked: {e}"),
             Err(_) => panic!("recv timeout"),
         }
 
--- a/crates/hyprstream-rpc/src/moq_event.rs
+++ b/crates/hyprstream-rpc/src/moq_event.rs
@@ -1173,9 +1173,18 @@ mod tests {
         // The subscriber task must wait until the broadcast appears rather than abort.
         let recv_task = tokio::spawn(async move { sub.recv().await });
 
-        // Small yield to let the subscriber background task start and block on
-        // the wait-for-subtree loop before we create the publisher.
-        tokio::time::sleep(std::time::Duration::from_millis(5)).await;
+        // Let the subscriber background task start and block on the #148
+        // wait-for-subtree loop before the publisher registers the broadcast.
+        // A fixed 5ms sleep flaked on loaded CI (the task might not have reached
+        // its wait yet, letting the publisher win the race so the wait path was
+        // never exercised). Fully deterministic ordering would need a
+        // subscriber-readiness signal (a small product hook, flagged separately);
+        // absent that, repeated yields plus a generous bounded settle make the
+        // subscriber-before-publisher ordering reliable.
+        for _ in 0..10 {
+            tokio::task::yield_now().await;
+        }
+        tokio::time::sleep(std::time::Duration::from_millis(50)).await;
 
         // Publisher registers the broadcast (creates `local/events/worker`).
         let mut pub_ = origin.publisher("worker")?;


ewindisch:
## Agentic Code Review

**Overall:** The core #148 fix (single-level absolute-path scope in `scope_consumer_for_patterns`) is correct and well-reasoned. However, this PR is substantially wider than its title: it bundles #148, #356 (networked notification reach), #357 (iroh dial-by-node_id), and #358 (QoS relay selection + moq relay rendezvous + schema spins on 3 capnp files), making it hard to review and scope correctly.

### Findings

**1. [CORRECTNESS] `crates/hyprstream-rpc/src/moq_event.rs` (test) — Flaky timing assumption in #148 regression test**
The test yields `sleep(5ms)` to let the subscriber task start before the publisher registers its broadcast. Under load a 5ms yield is not guaranteed — if the subscriber hasn't entered its `announced()` loop yet, the publisher announces first and the wait-for-subtree path is not exercised. The test can pass vacuously without validating the race it was written to catch.
> Failure scenario: on a loaded CI runner the subscriber task starts AFTER `pub_.publish()`, the event is already in the queue, and the test passes without ever hitting the new wait path — the regression regresses silently.

**2. [CORRECTNESS] `crates/hyprstream-rpc/src/moq_stream.rs:772` — No exponential backoff in relay reconnect loop**
`serve_origin_to_relay_background` reconnects with a fixed 500ms delay on every failure. If the relay is permanently unreachable (wrong address, TLS failure, firewall) the task spins indefinitely at 2 Hz, emitting a debug log on every attempt with no backoff ceiling and no circuit-break.
> Failure scenario: operator misconfigures relay URI; the task busy-loops at 2Hz for the process lifetime, generating log noise and holding a tokio task slot.

**3. [CORRECTNESS] `crates/hyprstream-rpc/src/moq_stream.rs:453` — TOCTOU: `producer_reach()` snapshot excludes iroh if called before iroh binds**
`producer_reach()` reads `global_iroh_node_id()` at call time. In the spawner, QUIC binds first; iroh binds slightly later and then registers via `init_global_iroh_node_id`. Any `StreamInfo` constructed in the window between QUIC bind and iroh bind advertises Quic-only reach; native peers that connect in that window will not attempt iroh dial.
> Failure scenario: a fast-connecting native peer subscribes in the boot window, gets a Quic-only reach, and misses the iroh-direct path for the lifetime of that stream handle.

**4. [CORRECTNESS] `crates/hyprstream-rpc/src/moq_stream.rs:808` — No capability verification that relay is in relay mode**
`run_relay_announce_link` dials the relay and sends `with_origin(producer)`. If the remote endpoint is a subscriber-mode server (no `with_moq_relay`), the handshake proceeds but the relay does not re-serve announced broadcasts. The producer believes its origin is linked up; subscribers dialing the relay find nothing.
> Failure scenario: operator points `[quic] relay` at a plain `/moq` consumer endpoint; the link succeeds silently, `moq relay announce link established` is logged, but subscribers get no events — hard to diagnose.

**5. [CLEANUP] PR scope vs title**
The PR title says `fix(#148): moq_event subscriber wait` but includes #356 (notification schema spin + networked reach), #357 (iroh dial-by-node_id), #358 (QoS relay rendezvous + 3-capnp schema spin + relay announce link + `events/endpoints.rs` deletion). These are independently reviewable changes. The schema spins (#356/#357 notification.capnp, tui.capnp ordinal reuse) require pre-release safety assurance that is easier to verify in a focused PR.

crates/hyprstream-rpc/src/moq_event.rs
```diff
@@ -1131,34 +1158,41 @@ mod tests {
...
+        // the #148 wait-for-subtree path: recv() starts the background task before
+        // the publisher has registered its broadcast.
+        // The subscriber task must wait until the broadcast appears rather than abort.
+        let recv_task = tokio::spawn(async move { sub.recv().await });
 
-        // Give the origin a moment to propagate the announcement.
-        tokio::time::sleep(std::time::Duration::from_millis(10)).await;
+        // Small yield to let the subscriber background task start and block on
+        // the wait-for-subtree loop before we create the publisher.
+        tokio::time::sleep(std::time::Duration::from_millis(5)).await;
```
ewindisch: **Flaky timing (correctness, #148 test).** This relies on a fixed 5ms `sleep` to let the subscriber background task reach the wait-for-subtree loop before the publisher is created. On a loaded CI runner the task may not have started within 5ms, so the test no longer deterministically exercises the subscriber-before-publisher race it claims to guard — it can pass for the wrong reason (publisher wins the race) or, worse, flake. Prefer a deterministic signal that the subscriber task is registered/blocked (e.g. await an `announced()`/ready notification, or a channel/`Notify` the task fires before blocking) rather than a wall-clock sleep. The outer 5s timeout is fine; it's the 5ms ordering assumption that's fragile.

crates/hyprstream/schema/tui.capnp
```diff
@@ -375,25 +375,29 @@ struct ConnectResult {
...
-  moqUdsPath @4 :Text;
-  # #275: network-routable ways to reach the producer's moq plane (mirrors the
-  # inference StreamInfo's `announcedAt`). A cross-process viewer dials the first
-  # reach it supports (see `dial_stream`) and subscribes to `moqBroadcastPath`,
-  # rather than connecting to its OWN local moq UDS plane — which only carries the
-  # producer's broadcast when the producer is co-located in the very same process.
-  # Co-located clients ignore this and use the same-host UDS fast path (`moqUdsPath`).
-  reach @5 :List(Destination);
+  # MAC key for the chained-HMAC verification (32 bytes).
+  macKey @1 :Data;
```
ewindisch, line 390: **🔶 capnp schema spin — ordinal reuse with type change (needs sign-off, see Q2).** `StreamInfo` reuses field ordinals while changing their types: @1 `Text subEndpoint` -> `Data macKey` (this line), @2 `Data macKey` -> `Text broadcastPath`, @3 `Text moqBroadcastPath` -> `List(Destination)` (L400). Reusing an ordinal with a different type is normally a hard wire-compat break: any reader/writer built against the old schema will misread these fields. The commit message argues this is safe because it's pre-release, built-fresh-and-consumed-in-handshake, and not persisted. That reasoning is plausible but is exactly the kind of change that should not land buried in a multi-ticket PR without explicit human confirmation that no shipped/persisted message uses these ordinals. Holding for sign-off.

crates/hyprstream-rpc-std/schema/notification.capnp
```diff
@@ -80,9 +84,15 @@ struct NotificationResponse {
...
-  moqBroadcastPath @4 :Text;   # moq broadcast path for the assigned topic
+  # moq broadcast path for the assigned topic ("{prefix}/{topic}").
+  broadcastPath @2 :Text;
+  # #356: network-routable ways to reach the NotificationService's moq plane,
+  # mirroring the inference StreamInfo's `announcedAt`. A subscriber dials the
+  # first reach it supports (see `dial_stream`) and subscribes to `broadcastPath`,
+  # so cross-instance model-load notifications work (fixes #142). Co-located
+  # subscribers ignore this and use the same-host UDS fast path resolved from
+  # LOCAL config — the UDS path is NEVER advertised on the wire.
+  announcedAt @3 :List(Destination);
```
ewindisch, line 95: **🔶 capnp schema spin — ordinal reuse with type change (needs sign-off, see Q2).** `SubscribeResponse` reuses ordinal @3 (`moqUdsPath :Text` -> `announcedAt :List(Destination)`) — a type change on a live ordinal — and repurposes @2 (`streamEndpoint :Text` -> `broadcastPath :Text`, L88), dropping the old @4 `moqBroadcastPath`. Same wire-compat concern as `tui.capnp`: an old reader decoding @3 as `Text` against new `List(Destination)` bytes is undefined. Please confirm no shipped/persisted consumer of these ordinals exists before this merges (Q2).

crates/hyprstream-rpc/src/moq_stream.rs
```diff
@@ -123,24 +128,172 @@ pub fn global_producer_reach() -> Option<&'static NodeStreamReach> {
...
-        Some(r) => vec![Destination {
+    use crate::stream_info::{IrohReach, QuicReach, Destination, Role, TransportConfig};
+    let mut reach = Vec::new();
+
+    // iroh-direct first (#357): native peers prefer the NAT-traversing,
+    // pkarr-discoverable direct path. Listed ahead of Quic so the shared
+    // resolver (`connect_moq_reach`) tries it before the Quic/UDS fallbacks.
+    // The node_id comes from the iroh substrate bind (== signing_key pubkey),
+    // and is `None` when iroh is disabled/unbound (Quic-only advertisement).
+    if let Some(node_id) = global_iroh_node_id() {
```
ewindisch: **TOCTOU / split-brain in advertised reach (#357/#358).** `producer_reach()` reads `global_iroh_node_id()` here and `global_producer_reach()` (Quic) below, plus `global_relay_reach()` lower down — three independent `OnceLock`s set at different bootstrap points (substrate-bind vs QUIC-bind vs relay-config). Per the #357 commit these are deliberately first-wins and fire at different times, so there is a window where `producer_reach()` is called after one is set but before another, advertising an incomplete reach (e.g. iroh-only, or Quic-only) to a subscriber. Worst case the advertised set omits the only reach the peer can actually use until the next bootstrap step completes. Consider gating advertisement until the substrate set the node expects is bound, or documenting/handling the partial-reach window explicitly.

crates/hyprstream-rpc/src/moq_stream.rs
```diff
@@ -950,6 +1117,231 @@ pub fn dial_transport_to_wire(
...
+) {
+    tokio::spawn(async move {
+        loop {
+            if let Err(e) = run_relay_announce_link(&producer, &relay).await {
+                tracing::debug!(
+                    "moq relay announce link ended: {e}; reconnecting in {:?}",
+                    RELAY_RECONNECT_DELAY
+                );
+            }
+            tokio::time::sleep(RELAY_RECONNECT_DELAY).await;
```
ewindisch, line 1421: **Fixed reconnect delay, no backoff (correctness, #358).** The const at L1278 is named `RELAY_RECONNECT_DELAY` and documented as "Reconnect backoff", but this loop sleeps a flat 500ms forever. If the relay is down or rejecting, every producer hammers it twice a second with no exponential backoff and no jitter — a thundering-herd / self-DoS against the federation anchor (the default relay). Recommend exponential backoff with a cap + jitter, and consider logging above `debug!` after sustained failures so a persistently-unreachable relay is observable.

crates/hyprstream-rpc/src/moq_stream.rs
```diff
@@ -950,6 +1117,231 @@ pub fn dial_transport_to_wire(
...
+) -> Result<()> {
+    use moq_net::Client as MoqClient;
+
+    // Reuse the ONE wire-reach → dial-config resolver so the relay is dialed by
+    // the same code path (Quic / WebTransport / iroh) every direct reach uses.
+    let dest = crate::stream_info::Destination { role: crate::stream_info::Role::Relay, transport: relay.clone() };
+    let cfg = reach_to_transport_config(&dest)
+        .ok_or_else(|| anyhow!("relay reach is not a dialable network transport"))?;
+
+    let stream_session = crate::dial::dial_stream(&cfg).await?;
```
ewindisch, line 1449: **No relay-capability / authorization check before announcing UP (#358).** `run_relay_announce_link` dials the configured relay and immediately `with_origin(producer)` — announcing this node's broadcasts UP to it — without verifying the dialed endpoint is actually an authorized relay (capability/role). The relay is described as "producer-chosen (default = PDS / federation anchor)", but nothing here confirms the peer accepts a relay role or is entitled to ingest this origin's announcements. Frames are AEAD-sealed so confidentiality holds, but a misconfigured or substituted relay endpoint would silently receive the broadcast announcements (metadata: track names / `broadcastPath`, traffic patterns) with no error surfaced. Recommend a capability/role check (or mutual-auth assertion) on the relay session before `with_origin`.

ewindisch (commented):
## 🔶 Decision required (human sign-off — capnp schema + scope)

The core #148 fix in this PR is correct (single `scope()` on the root consumer + `new_for_test` isolation — nicely reasoned). But the PR as it stands is much wider than its title and bundles **#148 + #356 + #357 + #358**, including **capnp schema spins on 3 files** (`notification.capnp`, `tui.capnp` with ordinal reuse, `streaming.capnp`). That bundling blocks clean review of the schema changes and is why I'm asking for explicit sign-off before this merges. Inline comments above flag the specific correctness + schema-spin hunks. No code changes were made by this review.

Please answer:

**Q1 (scope — recommend SPLIT):** Should this be split into focused PRs — the **#148 fix alone** (small, self-contained, mergeable now) vs **#356 / #357 / #358 + the capnp schema spins** as a separate stack? The #148 fix is held hostage by review of the much larger reach/relay work, and the schema changes can't be reviewed cleanly while buried in a 2.6k-line diff. (Recommendation: split — land #148 now, review the rest separately.)

**Q2 (schema sign-off):** The capnp spins on `notification.capnp` (`SubscribeResponse`: ordinal @3 changes type `Text` -> `List(Destination)`) and `tui.capnp` (`StreamInfo`: @1 `Text subEndpoint` -> `Data macKey`, @2 `Data macKey` -> `Text broadcastPath`, @3 `Text moqBroadcastPath` -> `List(Destination)`) are **ordinal reuse with type changes** — normally a wire-compat break. The commit message asserts this is safe (pre-release, built-fresh-and-consumed-in-handshake, not persisted, atomic landing). Please **explicitly confirm** there is no shipped client and no persisted/on-disk message using these ordinals, and approve the schema changes. Do you approve the schema reuse?

**Q3 (correctness — smaller, real):** Before merge, should the producer->relay reconnect get **exponential backoff** (currently a fixed 500ms despite the const being named `*_DELAY`/"backoff"), the relay link get a **relay-capability/authorization check** before announcing the local origin UP to it, and the `producer_reach()` **iroh/Quic OnceLock TOCTOU** be closed? These are independent of Q1/Q2 and can ride with whichever PR carries #358.

**Regardless of the above: the 3 currently-failing CI checks must be green before merge** — see below.

### CI status (3 failing)
- **Clippy — fail (real, code).** `hyprstream-rpc` lib-test: `MutexGuard` held across await + disallowed `std::sync::Mutex` (5 errs); `hyprstream` lib: doc-list-without-indentation + manual `Option::map` (2 errs); `hyprstream` lib-test: ~30 errs — `unwrap()`/`expect()`/`expect_err()` on `Result`, `mem::forget` on a `Drop` type, `to_string()` on `&str` (mostly in `services/registry.rs` test code ~L3000-3253). Must be fixed.
- **WASM (browser client) — fail (real, code).** `error[E0433]: cannot find module or crate tracing` at `crates/hyprstream-rpc/src/node_identity.rs:189` — `tracing` is used on the wasm32 target without being in scope (missing `cfg`-gate or missing wasm dep). Must be fixed.
- **cargo-deny (bans + licenses) — fail (infra/flake, NOT this PR).** `error: no such command: deny`: the workflow's `cargo install cargo-deny --locked 2>/dev/null || true` swallowed the install failure, so `cargo deny check` ran with no binary. This is a CI-environment issue, not a code defect — a re-run or a workflow fix should clear it.

ewindisch:
## 🔧 Fix plan (status: planned — blocked on 2 decisions below)

Grounded against head `bd47a6b46`. The #148 core fix is correct; nearly all failures come from the bundled #356/#357/#358 work + the capnp spins.

### CI failures (must go green regardless of decisions)
1. **WASM** — `error[E0433]: cannot find module or crate \`tracing\`` at `crates/hyprstream-rpc/src/node_identity.rs:189`. The #357 code calls `tracing::` on wasm32 where the crate isn't in scope. Fix: cfg-gate the call (`#[cfg(not(target_arch = "wasm32"))]`) or add `tracing` to the wasm32 dep set / use a cfg'd shim. Mechanical.
2. **Clippy** —
   - `crates/hyprstream-rpc/src/service/svc.rs:1025–1198` — `MutexGuard` held across `.await` + disallowed `std::sync::Mutex`. Fix: use `parking_lot::Mutex` (already a dep) or drop the guard before the await point / restructure. Real (deadlock-risk) — worth fixing properly, not allow-ing.
   - `crates/hyprstream/src/auth/service_jwt.rs:101–134`, `services/factories.rs:1599`, `services/registry.rs:3004–3187` — `unwrap()`/`expect()` lints (the registry ones are largely in test code). Fix: proper error handling in non-test paths; `#[allow(clippy::unwrap_used, clippy::expect_used)]` on the test modules (established pattern) for test-only sites.
3. **cargo-deny** — infra flake (`cargo install cargo-deny --locked || true` swallowed an install failure → `no such command: deny`). Re-run; if it recurs it's a workflow fix, not this PR's code.

### Correctness findings (from review)
4. `moq_event.rs:1178` — replace the flaky 5ms `sleep` with a deterministic readiness signal (`Notify`/channel the subscriber fires before blocking, or await `announced()`). *(travels with #148)*
5. `moq_stream.rs:217` — `producer_reach()` TOCTOU partial-reach window: gate advertisement until the expected substrate set is bound, or document + handle the partial-reach window. *(#357/#358)*
6. `moq_stream.rs:1303` — relay reconnect: replace flat 500ms with exponential backoff + cap + jitter; raise log level above `debug!` after sustained failure. *(#358)*
7. `moq_stream.rs:1331` — add a relay capability/role assertion before `with_origin()` announces UP. *(#358; light authz-posture touch — flagging)*

### Recommended sequencing
Split **#148** (items 1-on-test-side + 4) into its own small PR — it has no schema, and the CI failures above are NOT from #148, so it can go green and merge now. The #356/#357/#358 + schema work (items 1-WASM, 2, 5, 6, 7 + the ordinal spins) becomes the remainder PR carrying the schema sign-off.

### 🔶 Decisions required (HITL — blocking; no code applied pending these)
- **Q1 — scope split:** split #148 out (mergeable now) from #356/#357/#358 + schema? *(recommend: split)*
- **Q2 — capnp schema sign-off:** `tui.capnp:390` and `notification.capnp:95` reuse ordinals @1/@2/@3 with **type changes** — a hard wire-compat break, justified only as pre-release / not-persisted / atomic-landing. This is human-gated; **no schema regeneration or field changes will be made without explicit approval.**

ewindisch:
Closing as superseded — and the one unique change is moot pending a design decision.

- The flaky moq_event test this fixes is already handled in main by `fix(test): stabilize flaky moq_event + lazy_iroh tests`.
- The subscribe-before-announce / wait-for-subtree case is covered by main's `announced()` discovery model + #393 (per-OID tracks + firehose-backfill late-join).
- The only unique change — `scope_consumer_for_patterns` absolute-vs-relative scoping — sits on top of an unresolved federation question: event-track paths use a literal `local/events/{source}` with **no origin/node/DID namespace**, so a moq relay shared across federated origins would collide. That namespacing decision comes first; this tweak isn't worth merging in isolation.

Follow-up: a federation-namespacing ticket for event-track paths (origin/DID-scoped tracks for shared-relay safety).

Issue #465, opened 2026-06-25
Phase 3.7 — WanixMount: wire hyprstream-9p as wasm32 dep of rpc-std


ewindisch:
Part of #460. **Repo: hyprstream**. Depends on Phase 2's wasm transport; NOT on #355. UNBLOCKED for the dep-wiring part.

`crates/hyprstream-9p/src/wanix_mount.rs` — `WanixMount<T: P9Transport>` is entirely `#[cfg(target_arch = "wasm32")]` (browser-only): Wanix VFS 9P ops → `P9Client`/`P9Transport` → `DmaTransport` (SharedArrayBuffer) → capnp RPC (nine.capnp). This is the #409 "browser both-paths" Wanix half.

GAPs:
- `hyprstream-9p` is NOT yet a wasm32 dependency of `hyprstream-rpc-std`, so WanixMount can't be mounted in the browser namespace from rpc-std. Wire the dep (wasm32-gated) so `mount_wanix` (vfs_mount.rs, #409) is buildable in the frontend wasm bundle.
- www frontend has NO Wanix binding — expose the WanixMount 9P filesystem to the browser app (Plan9 namespace: walk/read/write service files over RPC). Rides the SAME wasm transport stack as Phase 2.

Consistent with backend model: 9P is an API FRONTEND to nine.capnp RPC (#391); rpc-over-iroh/moq stays the substrate. DRY 9P↔RPC translation (#391, merged #412) shared with the Kata server-side translator (#408).

Issue #464, opened 2026-06-25
Phase 3.5 — ts_codegen typed discriminated unions (prerequisite to Phase 3)


ewindisch:
Part of #460. **Repo: hyprstream** (`hyprstream-rpc-build/src/ts_codegen/`). UNBLOCKED.

The TS codegen emits a degenerate stub for pure-union envelope structs: `interfaces.rs:47` produces `export type X = { variant: string; data: unknown };`. The PARSERS already emit real `{ variant: 'name', data: <typed> }` payloads at runtime — the interface TYPE just throws the type info away. Upgrade `emit_struct_interface` to emit a typed discriminated union with one arm per variant (mirroring the parser's per-variant data shape) plus an `{ variant: 'unknown'; data: null }` fallback.

This is "upgrade the stub to typed arms," not "add from scratch."

Verify by regenerating `www/src/rpc/generated/*.ts` and `tsc`-checking the frontend. Prerequisite for Phase 3's `reach`/ReachOption regen.

NOTE: touches the RPC TS surface — flag any generated wire-type change that needs human review (capnp/RPC-adjacent gate).

ewindisch:
Hit this regenerating www's streaming.ts against current HEAD (2c35bce7d). The typed-union upgrade is mostly done — scalar/struct/Void arms now emit `{ variant: 'x'; data: <typed> }` correctly. **But there's an unhandled sub-case: inline `group` arms.**

streaming.capnp's `StreamOpt` QoS unions have non-empty inline groups — `Ordering.unordered`, `Delivery.atLeastOnce`, `OverflowPolicy.dropOldest` each carry fields. For those, codegen emits:

```ts
export type OverflowPolicy =
  | { variant: 'block'; data: undefined }
  | { variant: 'dropOldest'; data: Group }   // <-- undefined type name
  | { variant: 'unknown'; data: null };
```

→ 6× `TS2304: Cannot find name 'Group'` in the regenerated streaming.ts (frontend won't compile).

It's **not just a type-alias problem** — the group arms aren't round-tripped at all:
- **Type:** `interfaces.rs` → `union_variant_data_type(f)` resolves a group arm's `type_name` to the bare `Group`; no interface is synthesized.
- **Builder:** `builders.rs:764` `FieldSection::Group => {} // groups are inlined, not handled here` — the group's fields are never written.
[...]

Issue #460, opened 2026-06-25
Epic: Frontend RPC layer — backend phases (Phase 2 iroh/moq→wasm, Phase 3.5 ts_codegen unions, Phase 3.7 WanixMount wasm dep)


ewindisch:
Backend-only sub-phases of the frontend epic. Frontend tracking (Phase 1, 3, 4) moved to GitLab: gitlab.com/cyberdione/www-cyberdione-ai

## Backend phases (hyprstream repo)
- #462 Phase 2 — iroh/moq → wasm32 deps + wasm seam (browser-native Rust transport)
- #464 Phase 3.5 — ts_codegen typed discriminated unions (prerequisite for Phase 3 regen)
- #465 Phase 3.7 — WanixMount: wire hyprstream-9p as wasm32 dep of rpc-std

## Frontend phases (tracked on GitLab)
- Phase 1 — atproto host addressing
- Phase 3 — StreamInfo reshape regen (gated on #355)
- Phase 4 — wire-up + e2e

ewindisch:
Closing — this is a www-cyberdione-ai frontend issue, being moved to GitLab (gitlab.com/cyberdione/www-cyberdione-ai).

Issue #409, opened 2026-06-24
Browser: replace GenericServiceMount shim with WanixMount 9P bridge for real filesystem-walk access


ewindisch:
Browser: wire both WanixMount + standard RPC clients over the StreamInfo/MoQ/WT/Iroh stack

Part of #387. Related to #391 (9P↔RPC DRY), #355 (streaming plane).

## Architecture (settled)
The browser accesses hyprstream services via the same stack as native:

```
StreamInfo (DH handshake, reach list, QoS, AEAD keys)
  over
MoQ (pub/sub tracks, fan-out, retention, late-join)
  over
WebTransport / Iroh (QUIC + NodeId identity + discovery + NAT + PQ)
```

The bespoke Quinn WT path is superseded by Iroh (#355 streaming-plane restoration). The Iroh substrate (`iroh_substrate.rs`) is built with two ALPNs (`moql` + `hyprstream-rpc/1`) but not yet wired into production — wiring it is the transport-consolidation prerequisite.

### Path A — Standard RPC clients (WASM compile target)
`RegistryClient`, `ModelClient`, etc. compiled to WASM, dialing services over WebTransport → Iroh endpoint → capnp RPC (`hyprstream-rpc/1` ALPN). Currently in `build_browser_namespace` via `GenericServiceMount` (method-dispatch). **Improvement:** add a WebTransport arm to `dial()` that wraps a `web_transport_trait::Session` (from the browser native WT API via `web-transport-wasm`) as an `Arc<dyn RpcClient>` via the existing `rpc_session` framing. Then `RemoteModelMount`/`RemoteRegistryMount` work in the browser — real filesystem walk, same as native.

**The gap:** `dial()` has arms for Inproc/Ipc/SystemdFd/Quic/Iroh — no WebTransport variant. `rpc_session.rs` already frames capnp over any `web_transport_trait::Session`. The server already path-dispatches (default path → RPC). One new `dial()` arm closes the gap.

**Constraint:** `RemoteModelMount` uses an embedded tokio runtime + `block_on` — must verify this works in WASM (single-threaded; `current_thread` runtime should work). If not, adapt the async bridge for wasm.

### Path B — WanixMount (Wanix-native 9P/DMA path)
[...]

PR #471, opened 2026-06-26
feat(#464/#465): ts_codegen typed discriminated unions + wasm32 WanixMount seam (Phase 3.5 + 3.7)

ewindisch:
## Summary

Two commits implementing Phase 3.5 and Phase 3.7 of the frontend atproto+iroh/moq epic (#460). Both are backend prerequisites for the frontend Phase 3 StreamInfo reshape regen in www-cyberdione-ai (GitLab #3).

### Phase 3.5 — `8370741de` (`feat(#464)`)

**`crates/hyprstream-rpc-build/src/ts_codegen/interfaces.rs`**: upgrades the TS codegen to emit typed discriminated union arms for pure-union capnp structs.

Previously, `interfaces.rs` emitted `export type X = { variant: string; data: unknown }` for any struct that was a pure union envelope (all fields in the union, no non-union fields). The parsers *already* produced typed `{ variant: '<name>', data: <typed> }` payloads at runtime — the interface just wasn't surfacing it.

Now emits:
```typescript
export type TransportConfig =
  | { variant: 'quic'; data: QuicTransport | null }
  | { variant: 'webTransport'; data: WebTransport | null }
  | { variant: 'iroh'; data: IrohTransport | null }
  | { variant: 'unknown'; data: null };
```

The `data` type per variant mirrors what the parser actually returns: `Void→undefined`, `Text→string`, `Data→Uint8Array`, `List(T)→T[]`, struct pointer→`T|null`, data-section scalar/enum→the TS type.

**`parsers.rs`**: adds `as const` to the List(UnionStruct) element mapper's variant literals so `List(StreamPayload)` types as `StreamPayload[]` rather than widening to `{ variant: string; ... }[]`.

Verified by regenerating www-cyberdione-ai's `src/rpc/generated/*.ts` and running `tsc --noEmit`: zero discriminated-union type errors.

**Guards preserved:**
- `Option<none/some>` special-case keeps precedence (pure-union check is gated on `!type_name.starts_with("Option")`)
- Flat struct fields (non-union) unchanged
- Enum emission unchanged

### Phase 3.7 — `5ee5274c3` (`feat(#465)`)

**`crates/hyprstream-rpc-std/Cargo.toml`**: adds `hyprstream-9p` to the `[target.cfg(wasm32)]` deps. WanixMount and its DmaTransport/P9Client are entirely `#[cfg(target_arch = "wasm32")]`; the dep is browser-gated.

**`vfs_mount.rs`**: adds `mount_wanix(ns, P9Client<T>)` — mounts WanixMount at `/wanix` (9P2000.L over the DMA bridge, #409/#391). This is the rpc-std seam that makes the Plan9 filesystem namespace reachable from the browser; the P9Client is supplied by Phase 2's transport wiring.

Verified: rpc-std builds for wasm32-unknown-unknown (`RUSTFLAGS=--cfg web_sys_unstable_apis`) and native.

## Test plan

- [ ] `cargo clippy -p hyprstream-rpc-build` — no errors
- [ ] `cargo clippy -p hyprstream-rpc-std` — no errors  
- [ ] `cargo check --target wasm32-unknown-unknown -p hyprstream-rpc-std` with `RUSTFLAGS=--cfg=web_sys_unstable_apis` — clean
- [ ] CI green

## Sequencing

Phase 3.5 unblocks: regenerating `src/rpc/generated/*.ts` in www-cyberdione-ai against the new `streaming.capnp` `StreamInfo`/`Destination`/`TransportConfig` shape (GitLab #3).

Phase 3.7 unblocks: connecting Phase 2's wasm transport stack to the browser's Plan9 namespace (`mount_wanix`).

### DIFF
--- a/crates/hyprstream-rpc-build/src/ts_codegen/interfaces.rs
+++ b/crates/hyprstream-rpc-build/src/ts_codegen/interfaces.rs
@@ -1,6 +1,6 @@
 //! Generate TypeScript interfaces from Cap'n Proto struct and enum definitions.
 
-use hyprstream_rpc_build::schema::types::{EnumDef, FieldDef, FieldSection, ParsedSchema, StructDef};
+use hyprstream_rpc_build::schema::types::{EnumDef, FieldSection, ParsedSchema, StructDef};
 use hyprstream_rpc_build::util::to_camel_case;
 
 use super::capnp_to_ts_type;
@@ -55,7 +55,7 @@ fn emit_struct_interface(out: &mut String, s: &StructDef) {
                 out.push_str(&format!(
                     "  | {{ variant: '{}'; data: {} }}\n",
                     f.name,
-                    union_variant_data_type(f)
+                    super::union_variant_data_type(f)
                 ));
             }
             // Fallback arm matching the parser's `default` case.
@@ -88,39 +88,4 @@ fn emit_struct_interface(out: &mut String, s: &StructDef) {
         ));
     }
     out.push_str("}\n\n");
-}
-
-/// Map a union-member field to the TypeScript type carried in the `data` slot.
-///
-/// Mirrors the per-variant `data` expression produced by the parser emitters:
-/// - `Void`              → `undefined`
-/// - scalars / `Bool`    → the TS scalar (`number` / `bigint` / `boolean`)
-/// - `Text`              → `string`
-/// - `Data`              → `Uint8Array`
-/// - `List(Text)`        → `string[]`
-/// - `List(Struct)`      → `Inner[]` (elements are parsed into objects)
-/// - struct              → `TypeName | null` (the parser returns `null` on a null pointer)
-///
-/// Keeping this aligned with the parser guarantees the typed discriminated union
-/// is sound against the runtime values callers actually receive.
-fn union_variant_data_type(f: &FieldDef) -> String {
-    let tn = f.type_name.as_str();
-    if tn == "Void" {
-        return "undefined".to_owned();
-    }
-    if tn.starts_with("List(") {
-        // capnp_to_ts_type already maps List(T) → T[].
-        return capnp_to_ts_type(tn);
-    }
-    if super::is_primitive(tn) || tn == "Text" || tn == "Data" {
-        return capnp_to_ts_type(tn);
-    }
-    // Struct (or enum) reference. Struct pointers can be null at runtime; enums
-    // are data-section and never null, but the only data-section non-scalar union
-    // payloads are enums, which `capnp_to_ts_type` maps to their type name.
-    if matches!(f.section, FieldSection::Pointer) {
-        format!("{} | null", capnp_to_ts_type(tn))
-    } else {
-        capnp_to_ts_type(tn)
-    }
 }
--- a/crates/hyprstream-rpc-build/src/ts_codegen/interfaces.rs
+++ b/crates/hyprstream-rpc-build/src/ts_codegen/interfaces.rs
@@ -1,6 +1,6 @@
 //! Generate TypeScript interfaces from Cap'n Proto struct and enum definitions.
 
-use hyprstream_rpc_build::schema::types::{EnumDef, FieldSection, ParsedSchema, StructDef};
+use hyprstream_rpc_build::schema::types::{EnumDef, FieldDef, FieldSection, ParsedSchema, StructDef};
 use hyprstream_rpc_build::util::to_camel_case;
 
 use super::capnp_to_ts_type;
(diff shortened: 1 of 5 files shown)


ewindisch:
The generated TypeScript output in this PR appears to not match the current schema on main. Before merging, please verify that codegen was run against the actual `streaming.capnp` on main and re-run if not.

**Specific discrepancies to check:**

1. **`TransportConfig` union arms:** The generated output shows three arms (`quic`, `webTransport`, `iroh`). The actual schema in `crates/hyprstream-rpc/schema/streaming.capnp` on main has exactly **two arms**:
```capnp
struct TransportConfig {
  union {
    quic @0 :QuicReach;
    iroh @1 :IrohReach;
  }
}
```
There is no `webTransport` arm. WebTransport is iroh's browser bearer, not a separate `TransportConfig` arm. The generated TS should reflect the two-arm union.

2. **`StreamInfo` field name:** The generated output uses `reach: Destination[]`. The actual schema field is:
```capnp
announcedAt @4 :List(Destination);
```
The generated TS field name should be `announcedAt`, not `reach`.

If codegen was run against a local or stale copy of the schema, please re-run it against the current main schema and update the PR. The generated `src/rpc/generated/*.ts` files must exactly reflect what `ts_codegen` produces from the capnp on main — do not hand-edit generated files to paper over a mismatch.

Phase 3 (StreamInfo reshape and stream-client.ts update) is gated on this PR having correct output, so the discrepancy needs to be resolved before Phase 3 can start.

ewindisch:
## Agentic Code Review

**Overall:** The typed-discriminated-union improvement in interfaces.rs is well-conceived and the as const additions in the list-mapper and inline-struct-pointer paths are correct, but the fix is incomplete: the top-level XResponseResult interface and the standalone union struct result interfaces (generate_struct_parsers) still emit `variant: string; data: unknown`, so callers using those types gain no type narrowing. The WanixMount wiring (Phase 3.7) is mechanically sound and correctly wasm32-gated, but its API change from async + SAB to sync + pre-built P9Client should be flagged as a breaking change. No security issues were found.

### Findings

**1. [CORRECTNESS]** `crates/hyprstream-rpc-build/src/ts_codegen/parsers.rs:45`
Top-level XResponseResult interface still emits `variant: string; data: unknown` instead of the typed discriminated union introduced by this PR — as const is applied only in emit_union_struct_list_field_expr (list elements, ~line 953) and emit_struct_pointer_expr (inline IIFEs, ~line 758), but not in emit_return (called for all top-level and scoped response variants) or in generate_struct_parsers (standalone union struct parsers, ~line 257).
> TypeScript callers that switch on `result.variant` against the generated `XResponseResult` interface still get `variant: string` so narrowing ('quic' | 'iroh' | ...) fails to type-check and the typed discriminated union emitted by interfaces.rs is never assignable to the parser return type, producing tsc errors on any downstream type annotation or exhaustiveness check.\n\n**2. [NIT]** `crates/hyprstream-rpc-build/src/ts_codegen/interfaces.rs:104`\nunion_variant_data_type checks `|| tn == "Text" || tn == "Data"` redundantly because is_primitive_capnp_type already returns true for those types, making those two conditions permanently dead code.\n> Not a runtime failure; the logic is still correct, but the dead conditions obscure the invariant and could mislead a future maintainer into thinking Text/Data would otherwise fall through to the pointer branch.\n\n**3. [CORRECTNESS]** `crates/hyprstream-rpc-build/src/ts_codegen/interfaces.rs:95`\nunion_variant_data_type does not handle Option* struct types (e.g. OptionUint32) — they fall through to the FieldSection::Pointer branch and emit `OptionUint32 | null` instead of the correct `number | undefined` that capnp_to_ts_type already handles.\n> If any pure-union envelope capnp struct ever has an arm typed as an Option* wrapper (e.g. `choice @0 :OptionUint32`), the generated interface arm would be `{ variant: 'choice'; data: OptionUint32 | null }` rather than `{ variant: 'choice'; data: number | undefined }`, causing a type mismatch against the runtime parser value. Currently no such schema exists, so this is latent.\n\n**4. [NIT]** `crates/hyprstream-rpc-std/src/vfs_mount.rs:12`\nModule-level doc comment and the comment at line 513 still say `#409 Path B` after the PR renamed the section header to `#409/#391 (Wanix-native VFS access)`, leaving the two stale references inconsistent.\n> No runtime impact; readers searching for the ticket cross-reference may be confused by the mixed naming.\n\n**5. [CORRECTNESS]** `crates/hyprstream-rpc-std/src/vfs_mount.rs:532`\nmount_wanix signature changed from async fn returning anyhow::Result<()> to a sync fn returning NamespaceError — the PR description does not mark this as a breaking API change, and any out-of-tree caller that previously awaited mount_wanix or matched on anyhow::Error will fail to compile.\n> Any downstream wasm32 crate that previously called `mount_wanix(ns, sab, uname, aname).await` will get a compile error; the function no longer accepts those arguments and is no longer async. Within this repo there are no callers yet, but the API contract change is undocumented.

Issue #167, opened 2026-06-05
M2/N4: EventService XSUB/XPUB → moq (Live preset)


ewindisch:
Part of epic #131. See docs/plans/2026-06-05-zmq-replacement-consolidated.md.

Migrate the event bus (ProxyService XSUB/XPUB, `inproc://hyprstream/events/`) to moq as a `Live` (fan-out, unbounded, at-most-once) stream. **Rewire ALL bus users** (the review found N4's first list under-enumerated): compositor, TUI, WorkerService, WorkflowService + gh_adapter, SecureEvent encrypted-bus variants, CLI SUB consumers (shell/tui/git handlers), and the **system service-ready/-stopping bootstrap signals** (main.rs:2084/2114 — load-bearing for staged startup). Delete the ZMQ ProxyService event path.

ewindisch:
Implementation complete on branch `ewindisch/moq-167` (commit `f2ea295f6`).

**What was done:**
- Added `crates/hyprstream-rpc/src/moq_event.rs`: `MoqEventOrigin`, `MoqEventPublisher`, `MoqEventSubscriber` with a process-global `OnceLock`-backed init
- Each source registers a broadcast under `local/events/{source}` with a single `events` track (Live preset — fan-out, unbounded, at-most-once)
- Subscribers discover source broadcasts via `OriginConsumer::announced()` and filter by topic prefix in-memory (preserving ZMQ prefix-match semantics)
- `events/publisher.rs` + `events/subscriber.rs`: removed ZMQ/tmq, wrap moq types
- `factories.rs`: event factory initializes moq origin + `MoqEventBarrierService` (satisfies `Spawnable` lifecycle, no ZMQ sockets)
- All bus users updated: `bin/main.rs`, `runtime/service.rs`, `workflow/service.rs` — context arg dropped from `EventPublisher::new` / `EventSubscriber::new`
- `SecureEventPublisher` / `SecureEventSubscriber` continue to work unchanged (pure crypto layer)

Awaiting push (requires hardware SK key). Will open PR once pushed.

ewindisch:
Implemented in #238.

PR #474, opened 2026-06-26
fix: allow_any_header() on DID document endpoints (closes #472)

### LINKED ISSUE #472: fix: DID document endpoints must accept arbitrary request headers (W3C DID Core compliance)
## Summary

`CorsConfig::public()` (`crates/hyprstream/src/config/server.rs:71`) sets `permissive_headers: false`, enforcing an explicit header allowlist on all CORS preflights for public endpoints. This causes browser fetches of `/.well-known/did.json` and `/.well-known/atproto-did` to fail preflight whenever the request includes a non-standard header.

This is a **spec-compliance gap**, not a security control.

## Why arbitrary headers must be allowed on DID endpoints

DID documents are public by definition per [W3C DID Core 1.0](https://www.w3.org/TR/did-core/). There is no secret in a DID document — it carries only public keys and service endpoints. The spec requires that any conformant client can fetch a DID document without authentication or header constraints. Rejecting preflights based on request headers violates this requirement.

The issue was surfaced during testing via ngrok, but ngrok is not the root cause — **any** browser client that sends a non-standard request header (browser extension injections, custom SDK headers, `ngrok-skip-browser-warning`, etc.) will hit this. The fix is spec-correctness, not ngrok accommodation.

## Affected routes

- `/.well-known/did.json` (did:web resolution)
- `/.well-known/atproto-did` (atproto handle resolution)

## Fix

Apply `allow_any_header()` (or equivalent `permissive_headers: true`) **only to the DID document routes**, not globally. Authenticated API endpoints must retain their restrictive CORS configuration.

Example approach — in the DID document route handlers, override CORS to allow any header:

```rust
// DID document endpoints are public by W3C DID Core spec — any header allowed
CorsLayer::new()
    .allow_origin(Any)
    .allow_headers(Any)
    .allow_methods([Method::GET])
```

Alternatively, if `CorsConfig` supports per-route override, use that rather than changing the global `public()` constructor (which could affect other endpoints that should remain restrictive).

## What NOT to do

Do **not** set `permissive_headers: true` globally in `CorsConfig::public()` — that would relax headers for all public endpoints, not just DID documents.

## Impact

- Unblocks W3C DID Core-compliant resolution from any browser client
- No security impact: DID documents are unauthenticated, read-only, public data
- Authenticated API endpoints: no change

ewindisch:
## Problem

Closes #472.

The public DID-document routes — `/.well-known/did.json`, `/.well-known/atproto-did`, `/users/:username/did.json`, `/clients/:client_id/did.json` — must be fetchable by any client without CORS preflight failures. Per W3C DID Core a DID document is a public identity anchor that carries no authorization; restricting which request headers may accompany a fetch (browser extensions, SDK-injected headers, reverse-proxy bypass headers such as `ngrok-skip-browser-warning`) has no security value and breaks legitimate clients.

## What changed since the first cut (least-privilege rework)

The initial revision set `permissive_headers: true` in the **global** `CorsConfig::public()` constructor. That relaxed the request-header allowlist for the **entire** public router — `/oauth/token`, `/oauth/register`, `/oauth/par`, `/oauth/revoke`, SCIM discovery, etc. #472 explicitly forbids this ("Do NOT set `permissive_headers: true` globally in `CorsConfig::public()` … apply it only to the DID document routes"). This PR has been reworked to the least-privilege stance the ticket mandates:

- **Reverted the global flip:** `CorsConfig::public()` is back to `permissive_headers: false`. The broad public router keeps its explicit, restrictive request-header allowlist.
- **Scoped the relaxation to DID routes only:** added a `CorsConfig::did_document()` variant (wildcard origin, credentials disabled, permissive headers) applied as a dedicated route-group CORS layer on a separate `did_router` that carries exactly the four public DID GET endpoints. The permissive-header surface is now confined to these secret-free public documents; secret-bearing endpoints never see it.
- **Restored operator visibility:** the permissive-headers log was downgraded to `debug!` in the first cut — it is back to `warn!` so the relaxed posture stays auditable. The stale "development/debugging only" field doc / inline comment now describes the deliberate DID-route use.

## atproto compat / interop

The DID routes remain cross-origin readable with arbitrary request headers (the #472 requirement), so atproto handle→DID resolution and did:web fetches keep working from any origin. The change only stops leaking that relaxation to the rest of the router.

## #501 coordination

`/.well-known/atproto-did` is added by #501 (`ewindisch/500-atproto-did-wellknown`), which has **merged to main**. This branch has been **rebased onto main** so all four DID routes exist and are scoped together in one place (the branch was force-pushed for this rebase). No further coordination is needed.

## Files changed

- `crates/hyprstream/src/config/server.rs` — revert `public()` to `permissive_headers: false`; add `CorsConfig::did_document()`; refresh field docs.
- `crates/hyprstream/src/services/oauth/mod.rs` — move the four DID routes into a dedicated `did_router` with the scoped permissive CORS layer; broad router keeps the restrictive layer.
- `crates/hyprstream/src/server/middleware.rs` — restore `warn!` for permissive headers; update inline comment.

## Verification

```bash
curl -H "X-Custom-Header: foo" https://<host>/.well-known/did.json
# → 200 OK with DID document (no preflight rejection)

# And the broad public router rejects arbitrary headers as before:
curl -X OPTIONS -H "Origin: https://x" -H "Access-Control-Request-Headers: x-custom" \
     https://<host>/oauth/token   # → x-custom NOT in Access-Control-Allow-Headers
```

Claude-Session: https://claude.ai/code/session_01U3qgsHTDWME66JFGZGS3ZA


ewindisch:
## Agentic Code Review

**Overall:** The fix is wider than described: changing CorsConfig::public() to set permissive_headers=true affects every endpoint in the OAuth router (token, register, revoke, introspect, PAR, SCIM) not just DID documents, and the warn!() that would alert operators to this posture has been silenced to debug level. The correct scoped fix would be to add a separate, narrower CORS layer only on the DID document routes rather than broadening the entire OAuth router's header policy.\n\n### Findings\n\n**1. [SECURITY]** `crates/hyprstream/src/config/server.rs:71`\nCorsConfig::public() enables permissive_headers for the entire OAuth router, not just DID document endpoints — the PR title and body both claim the change is scoped to DID routes only, but the single CORS layer covers /oauth/token, /oauth/register, /oauth/par, /oauth/revoke, /oauth/introspect, and all SCIM endpoints as well.\n> A browser attacker can send arbitrary custom request headers (e.g. X-Forwarded-For spoofing, X-Real-IP overrides, custom trace headers) on cross-origin requests to /oauth/token or /oauth/register without preflight rejection. The explicit allowlist that previously forced clients to use only known-safe headers is permanently removed from all OAuth endpoints, not only public identity documents.\n\n**2. [CORRECTNESS]** `crates/hyprstream/src/server/middleware.rs:240`\nThe warn!() that signals permissive header mode is active is downgraded to debug!(), silently hiding the configuration from operators who check logs at info or warn level.\n> An operator deploying to production with permissive_headers=true (now the default for all public OAuth endpoints) sees no log entry at warn/info level confirming this setting is active. Security audits that rely on log evidence to confirm CORS posture will find no trace of the permissive mode.\n\n**3. [NIT]** `crates/hyprstream/src/config/server.rs:33`\nThe struct-level doc comment on the permissive_headers field still says 'NOT recommended for production' and 'development only', contradicting CorsConfig::public() which now sets it to true for production use.\n> A developer reads the field-level doc, concludes permissive_headers=true is a dev-only flag, and reverts CorsConfig::public() to false in a later cleanup PR, reintroducing the original #472 regression.\n\n**4. [NIT]** `crates/hyprstream/src/server/middleware.rs:238`\nThe inline comment '// Permissive mode - allow any header (development/debugging only)' was not updated and now contradicts the deliberately production-enabled intent documented in the CorsConfig::public() doc comment.\n> A code reviewer or auditor sees the comment and the debug!() line, assumes this path is never taken in production, and misses the security posture change introduced by this PR.

ewindisch:
## Agentic Code Review

**Overall:** The PR's actual implementation scope is wider than the description claims: changing CorsConfig::public() to set permissive_headers=true affects the entire OAuth HTTP server (including /oauth/token, /scim/v2/Users, /oauth/introspect) because a single CORS layer wraps all routes. The stated goal of targeting only DID document routes requires per-route CORS differentiation that this PR does not implement. The change is low-risk for the DID document use case specifically, but the mismatch between the stated intent ("specifically on the DID document routes") and the actual blast radius (all OAuth/SCIM routes) is the primary concern and should be resolved before merge.

### Findings

**1. [SECURITY]** `crates/hyprstream/src/config/server.rs:71`
CorsConfig::public() enables permissive_headers for the entire OAuth server, not just DID document routes — the PR description's claim that it targets only DID endpoints is incorrect.
> A cross-origin browser page sends a preflight request with an arbitrary custom header (e.g. X-Injected-Header) to /oauth/token, /oauth/register, /oauth/introspect, /scim/v2/Users, or /oauth/userinfo. With permissive_headers=true the server responds Access-Control-Allow-Headers: * and the browser proceeds with the credentialed request. Any header the token endpoint reads (e.g. Authorization) is already handled, but this unnecessarily broadens the CORS attack surface for all OAuth and SCIM management endpoints that share the same single cors_layer applied in create_app(). The fix should either add per-route CORS differentiation or introduce a separate CorsConfig::did_document() constructor that does not widen the config used by sensitive endpoints.

**2. [CORRECTNESS]** `crates/hyprstream/src/server/middleware.rs:240`
Downgrading the CORS permissive-headers log from warn! to debug! removes the only production-visible signal that wildcard header acceptance is active on a live server.
> An operator deploys with the default OAuthConfig (which now has permissive_headers=true via CorsConfig::public()) and does not realize that all CORS preflights are accepting any header. Previously a warn-level log would surface this at typical log levels; with debug! it is invisible in production unless the operator explicitly enables debug logging for the hyprstream crate. If a misconfiguration or unexpected default is ever introduced for a non-public endpoint, it will go unnoticed.

**3. [NIT]** `crates/hyprstream/src/server/middleware.rs:238`
Inline comment still reads 'development/debugging only' after the PR reframes permissive_headers as appropriate for production public endpoints, creating a misleading contract.
> A future reviewer or auditor reads 'Permissive mode — allow any header (development/debugging only)' alongside a production config where permissive_headers=true is intentional; the comment contradicts the new documented intent and will cause confusion about whether a deployment is mis-configured.

**4. [NIT]** `crates/hyprstream/src/config/server.rs:33`
The CorsConfig struct-level doc comment for permissive_headers was not updated and still describes the field as 'NOT recommended for production', contradicting the new public() constructor that enables it by default.
> The on-disk file (line 33) retains the old doc comment 'Allow all headers (permissive mode for development - NOT recommended for production)' while the PR updates only the pub fn public() doc block. Anyone reading the struct definition to understand the field semantics receives contradictory guidance. (This comment appears in the unchanged file; the diff only updated the pub fn public() doc block.)

ewindisch:
## Agentic Code Review

**Overall:** The PR correctly solves the DID document CORS preflight problem but does so via a blunt instrument: it enables wildcard-header CORS on the entire OAuth server's public CORS policy (CorsConfig::public()), affecting /oauth/token, /oauth/register, /oauth/par, /oauth/device, SCIM discovery, and more — not just DID routes as the PR description claims. The biggest risk is the mismatch between stated and actual scope, combined with silently downgrading the operator warning to debug level, which removes the only production-visible signal that wildcard-header CORS is active. For a low-sensitivity read-only endpoint like did.json this posture is defensible, but applying it server-wide across sensitive OAuth endpoints warrants either a narrow per-route CORS override or at minimum restoring the warn! to maintain auditability.

### Findings

**1. [SECURITY]** `crates/hyprstream/src/config/server.rs:85`
CorsConfig::public() now enables permissive_headers=true for all OAuth public routes, not just DID document endpoints as stated in the PR description
> The PR title claims allow_any_header() is applied 'specifically on the DID document routes', but CorsConfig::public() is the server-wide CORS default for the entire OAuth service (/.well-known/oauth-authorization-server, /oauth/token, /oauth/register, /oauth/par, /oauth/jwks, /oauth/device, /oauth/logout, SCIM discovery, etc.) as defined in config/mod.rs default_oauth_cors(). An attacker or confused browser extension can now attach arbitrary request headers to token exchange, device authorization, and dynamic client registration preflights. While these endpoints are already 'public' in the sense that they require no bearer auth, allowing wildcard headers expands the CORS attack surface slightly — for example, a browser making a credentialless cross-origin POST to /oauth/token with a custom Forwarded or X-Forwarded-For header that bypasses a WAF rule now passes the preflight where it would previously have been rejected.

**2. [CLEANUP]** `crates/hyprstream/src/server/middleware.rs:237`
Operator visibility loss: downgrading warn! to debug! for permissive_headers=true removes the production-time signal that was the only log-level-visible indicator that wildcard header CORS is active
> Before this PR, an operator running at the default log level (info or warn) would see a warning when permissive_headers was enabled, alerting them that the server accepted any CORS header. After this PR, because CorsConfig::public() sets permissive_headers=true and the log line is downgraded to debug, the default-deployed OAuth service silently accepts wildcard headers with no operator-visible indication unless debug logging is explicitly enabled. This makes auditing CORS posture harder and removes a useful canary in production logs.

**3. [NIT]** `crates/hyprstream/src/config/server.rs:70`
Doc comment on CorsConfig::public() still says it is for 'OAuth metadata, DID documents, token endpoint' but the function is also the default CORS for SCIM discovery, device enrollment, and external OIDC callback routes
> A future maintainer reading the doc comment will believe the scope is narrower than it is. The route list in oauth/mod.rs also includes /api/device/challenge, /api/device/enroll, /oauth/callback/:provider, and SCIM discovery routes that are not mentioned in the comment, leading to incorrect reasoning about the security posture of those endpoints.

**4. [NIT]** `crates/hyprstream/src/services/oauth/mod.rs:90`
PR description claims /.well-known/atproto-did requires wildcard header CORS, but that endpoint does not exist in the codebase
> The PR body says 'atproto-did must be fetchable by any client without CORS preflight failures', but grep finds no route or handler for /.well-known/atproto-did anywhere in the Rust source. This either means the endpoint is planned but not yet implemented (and the justification is premature) or it was a copy-paste error. The actual justification is weaker than stated.

crates/hyprstream/src/config/server.rs
```diff
@@ -66,15 +68,19 @@ impl Default for CorsConfig {
...
+    /// browser clients can include arbitrary headers (e.g. `ngrok-skip-browser-warning`
+    /// for dev tunnels) without preflight rejection.
     pub fn public() -> Self {
         Self {
             enabled: true,
             allowed_origins: vec!["*".to_owned()],
             allow_credentials: false,
             max_age: 3600,
-            permissive_headers: false,
+            permissive_headers: true,
```
ewindisch, line 113: This flips `permissive_headers` for the entire `public()` CORS profile, which is applied to the whole public router — `/oauth/token`, `/oauth/register`, `/oauth/par`, `/oauth/revoke`, `/oauth/introspect`, and SCIM — not just the DID document routes. The PR title/body claim a DID-scoped change ("Apply `allow_any_header()` specifically on the DID document routes"), but `public()` is shared, so `Access-Control-Allow-Headers: *` now also applies to the OAuth/SCIM auth surface. If DID-scoping is the intent, consider a dedicated `CorsConfig::did_document()` applied only to the DID routes and leave `public()` restrictive. See Q1 in the decision comment.

crates/hyprstream/src/config/server.rs
```diff
@@ -30,7 +30,9 @@ pub struct CorsConfig {
     #[serde(default = "default_cors_max_age")]
     pub max_age: u64,
 
-    /// Allow all headers (permissive mode for development - NOT recommended for production)
+    /// Allow all headers in CORS preflights (`Access-Control-Allow-Headers: *`).
```
ewindisch, line 33: Field doc was softened from "permissive mode for development - NOT recommended for production" to endorsing it for public endpoints, but `public()` (now `permissive_headers: true`) is the production default for OAuth/SCIM routes. If the server-wide enablement is kept, the doc and the in-code comments should be made consistent (see middleware.rs L238).

crates/hyprstream/src/server/middleware.rs
```diff
@@ -237,7 +237,7 @@ pub fn cors_layer(config: &crate::server::state::CorsConfig) -> tower_http::cors
     if config.permissive_headers {
         // Permissive mode - allow any header (development/debugging only)
         cors = cors.allow_headers(Any);
-        warn!("⚠️  CORS: Allowing ALL headers (permissive mode)");
+        debug!("CORS: allowing all request headers (permissive_headers=true)");
```
ewindisch: Downgrading warn! → debug! removes the only production-visible signal that wildcard-header CORS is active. With `public()` now enabling `permissive_headers` by default, operators have no log evidence that `Access-Control-Allow-Headers: *` is being served on the auth surface. Recommend restoring warn!-level (or info!) for auditability. See Q2 in the decision comment.

crates/hyprstream/src/server/middleware.rs
```diff
@@ -237,7 +237,7 @@ pub fn cors_layer(config: &crate::server::state::CorsConfig) -> tower_http::cors
     if config.permissive_headers {
         // Permissive mode - allow any header (development/debugging only)
```
ewindisch: This inline comment still labels the branch "development/debugging only", which directly contradicts `CorsConfig::public()` now enabling `permissive_headers` as a production default. Either the comment is stale (and should be updated) or the production enablement is unintended — both readings point at the same unresolved decision (Q1).

ewindisch (commented):
Agentic review — inline notes on the auth-surface impact. This is posted as a COMMENT (not approve/request-changes): the change is a security/authz decision that needs human sign-off. NO code fix has been applied. See the separate "Decision required" comment for the explicit questions. Summary: the diff is wider than the title — `CorsConfig::public()` flips `permissive_headers` for the entire public router (OAuth token/register/par/revoke/introspect + SCIM), not just the DID document routes, and the only operator-visible signal of permissive CORS was downgraded from warn! to debug!.

ewindisch:
## 🔶 Decision required (human sign-off — auth surface)

**No code fix has been applied.** This PR changes the CORS posture on an authentication surface, and the right fix direction depends on the answers below. The agentic review found that the change is broader than the PR title suggests:

- `CorsConfig::public()` now sets `permissive_headers = true`. `public()` is the shared profile for the **entire public router** — `/oauth/token`, `/oauth/register`, `/oauth/par`, `/oauth/revoke`, `/oauth/introspect`, and SCIM — not only the DID document routes named in the title/body.
- The only operator-visible signal of permissive CORS was downgraded `warn!` → `debug!`, so the posture is no longer surfaced in production logs.
- The PR body cites `/.well-known/atproto-did` as justification, but **that endpoint does not exist in the codebase**.

### Questions

**Q1 — Scope.** Should the relaxation be narrowed to a new `CorsConfig::did_document()` applied only to the DID routes (`/.well-known/did.json`), leaving `public()` (OAuth/SCIM) with its restrictive header allowlist? Or is server-wide `Access-Control-Allow-Headers: *` across all OAuth/SCIM endpoints intended and acceptable?
_Recommendation: narrow — scope to DID routes only._

**Q2 — Auditability.** Should the `warn!`-level log of permissive mode be restored (vs. the new `debug!`) so operators retain a production-visible signal that wildcard-header CORS is active on the auth surface?
_Recommendation: yes — restore warn! (or info!)._

**Q3 — Justification accuracy.** The PR body cites `/.well-known/atproto-did`, which is not present in the code. Is that endpoint planned (and should be added), or should the justification be corrected to reference only the endpoints that actually exist (`/.well-known/did.json`)?

Once these are answered, the fix can be implemented in the chosen direction. Until then, this stays a decision PR — no changes have been made.

ewindisch:
@claude review linked PRs/issues (closed) can we proceed on this pr?

ewindisch:
**Triage (1:3.0): recommend merge.**

Re-verified the fix is correctly scoped and already addresses all three rounds of agentic review + the 🔶 decision-required comment: `CorsConfig::public()` is untouched (stays restrictive for OAuth/SCIM), the new `CorsConfig::did_document()` is applied only to a dedicated `did_router` carrying the 4 DID GET routes, the `warn!` log is restored, doc comments updated, and `/.well-known/atproto-did` is confirmed to exist (merged via #501). The 23:00 `@claude` review on this PR already reached the same "safe to merge" conclusion.

The failing `build` check (run [28268032684](https://github.com/hyprstream/hyprstream/actions/runs/28268032684)) is **unrelated to this diff** — two pre-existing flaky test failures (`moq_event::tests::moq_event_pub_sub_roundtrip` — matches known issue #167; `transport::lazy_iroh::tests::lazy_iroh_connects_on_first_send_and_caches` — iroh connect flake), neither touched by this PR's CORS-only changeset. Worth a rebase onto current main before merge to pick up any unrelated fixes, but not a blocker on the diff's correctness.

mergeState is BLOCKED only on the above — no outstanding code concerns from me. Recommend merge once rebased/green.

PR #475, opened 2026-06-26
feat(phase2): iroh wasm32 peer + pkarr lookup in browser

ewindisch:
## Summary

Adds iroh as a wasm32 dependency in `hyprstream-rpc`, making the browser a full iroh peer with its own NodeId and pkarr-based address discovery. This is the backend half of Phase 2 — the frontend `transport.ts` becomes a thin binding over these wasm exports instead of a hand-rolled JS WebTransport dialer.

## What lands

- `crates/hyprstream-rpc/src/iroh_peer.rs`: `BrowserIrohPeer` — iroh peer lifecycle, NodeId, `did:key` helpers, pkarr HTTP gateway lookup
- `crates/hyprstream-rpc/src/iroh_exports.rs`: `#[wasm_bindgen]` exports: `IrohPeer` class + `resolve_did_key_pkarr()` free function

## What's deferred (upstream blocker)

`moq-net 0.1.8` requires `Session: Sync`, which is incompatible with browser WebTransport's `!Sync` types. `dial_iroh_reach()` (moq track subscribe/publish over iroh in the browser) is deferred until this is fixed upstream or worked around. See upstream issue (to be filed). The existing `QuicReach` WebTransport path continues to handle browser streaming in the interim.

## CI

`cargo check --target wasm32-unknown-unknown -p hyprstream-rpc` passes (same gate as the CI `wasm` job).

## Sequencing

- Phase 1 (atproto identity, GitLab !18): independent, can merge now
- This PR (Phase 2a): iroh peer + pkarr in browser
- Phase 2b: `dial_iroh_reach()` — gated on moq-net upstream Sync fix
- Phase 3 (StreamInfo regen): gated on #471 merging

Claude-Session: https://claude.ai/code/session_01U3qgsHTDWME66JFGZGS3ZA

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6659,6 +6659,7 @@ dependencies = [
  "ml-dsa",
  "ml-kem",
  "moq-net",
+ "n0-error",
  "nix 0.27.1",
  "once_cell",
  "p256",
--- a/crates/hyprstream-rpc-std/src/iroh_exports.rs
+++ b/crates/hyprstream-rpc-std/src/iroh_exports.rs
@@ -1,0 +1,136 @@
+//! wasm-bindgen exports for iroh peer identity + pkarr discovery (Phase 2).
+//!
+//! Wraps `hyprstream_rpc::iroh_peer` with wasm-bindgen so JavaScript can:
+//!
+//! 1. Generate an ephemeral iroh identity (`IrohPeer::new()`).
+//! 2. Get the browser's NodeId as bytes or as a `did:key` DID.
+//! 3. Resolve a peer's relay URL from the N0 pkarr relay
+//!    (replaces the manual HTTP fetch + DNS-wire parse in `atproto.ts`).
+//!
+//! # Usage (TypeScript)
+//!
+//! ```ts
+//! import { IrohPeer, irohNodeIdFromDidKey, irohDidKeyFromNodeId,
+//!          irohResolvePkarrRelayUrl } from 'hyprstream-rpc-std';
+//!
+//! // Generate a fresh browser identity.
+//! const peer = IrohPeer.new();
+//! console.log("NodeId (z32):", peer.nodeIdZ32());
+//! console.log("did:key:", peer.didKey());
+//!
+//! // Convert a did:key to raw NodeId bytes (for pkarr lookup).
+//! const nodeId = irohNodeIdFromDidKey("did:key:z6Mk...");
+//!
+//! // Resolve a peer's relay URL.
+//! const relayUrl = await irohResolvePkarrRelayUrl(nodeId);
+//! ```
+
+#![cfg(target_arch = "wasm32")]
+
+use wasm_bindgen::prelude::*;
+
+use hyprstream_rpc::iroh_peer::{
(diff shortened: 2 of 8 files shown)


ewindisch:
## Agentic Code Review

**Overall:** The PR correctly adds iroh as a wasm32 dependency and the PkarrRelayClient/SecretKey usage is appropriate for the browser target. The main actionable bug is that the "Ok(None) for unknown peer" contract on resolve_pkarr_relay_url cannot be satisfied because PkarrRelayClient::resolve() returns Err on 404 rather than Option — callers wanting to distinguish "not found" from "network error" will need to inspect the error type instead of matching on None. A secondary issue is the duplicated did:key codec logic that diverges from the native did_web.rs implementation in its handling of DID URL fragments.

### Findings

**1. [CORRECTNESS]** `crates/hyprstream-rpc/src/iroh_peer.rs:151`
PkarrRelayClient::resolve() returns Err on 404 (peer not found), not Option — the Ok(None) branch in resolve_pkarr_relay_url is dead code and callers receive an error instead of null when a peer has no pkarr record.
> JS calls irohResolvePkarrRelayUrl() for a node_id that has never published to pkarr. The pkarr relay returns HTTP 404, iroh maps this to PkarrError::HttpRequest{status: 404}, the ? propagates it as Err, and the wasm export throws a JsError. The documented Ok(None)/null return path is never reached. The function contract (returns null for unknown peers) is violated.

**2. [CORRECTNESS]** `crates/hyprstream-rpc/src/iroh_peer.rs:109`
node_id_from_did_key does not strip DID URL fragments or query parameters, diverging from did_web.rs::did_key_to_ed25519 which transparently handles 'did:key:z6Mk...#z6Mk...' self-referencing DID URLs.
> A caller passes a DID URL with a fragment (e.g., 'did:key:z6MkfooBar#z6MkfooBar', as emitted by DID document verificationMethod references) to irohNodeIdFromDidKey. The base58 decode fails because '#' is not a valid base58btc character, throwing a JsError. The native path (did_key_to_ed25519) accepts the same input without error, creating inconsistent behavior between native and wasm32 for the same conceptual operation.

**3. [CLEANUP]** `crates/hyprstream-rpc/src/iroh_peer.rs:86`
did:key encoding/decoding logic is duplicated from did_web.rs with divergent behavior (no MULTICODEC_ED25519_PUB constant reuse, no fragment stripping), creating two implementations of the same algorithm that can drift apart.
> A future change to the multicodec prefix or encoding rule in did_web.rs (e.g., ML-DSA extension or fragment handling) is not reflected in iroh_peer.rs, causing native and wasm32 to produce or accept different did:key formats for the same key.

**4. [NIT]** `crates/hyprstream-rpc-std/src/iroh_exports.rs:16`
TypeScript usage example in the module doc shows IrohPeer.new() (static call) but #[wasm_bindgen(constructor)] maps to new IrohPeer() (constructor call) in JavaScript/TypeScript.
> A TypeScript consumer copies the documented example and writes IrohPeer.new(), which throws a TypeError at runtime because new() is not a static method on the exported class. The correct call is new IrohPeer().

PR #480, opened 2026-06-26
refactor(#384): de-singletonize GLOBAL_RELAY_REACH — per-server config + per-stream override

### LINKED ISSUE #384: S7: de-singletonize relay reach — per-server config + per-stream override (process-global GLOBAL_RELAY_REACH is wrong granularity)
Part of #355 (Streaming Plane Restoration). Found during the 2026-06-22 reach-alignment review of PR #378.

## Problem

`producer_reach()` sources the relay endpoint from a **process-global `OnceLock` singleton**, `GLOBAL_RELAY_REACH` (`crates/hyprstream-rpc/src/moq_stream.rs:180`), set via `init_global_relay_reach()` (first-write-wins). This is the wrong granularity in three ways:

1. **Relay choice is logically per-stream, implemented as per-process.** The code's own comment says a relay-only *anonymized* stream is achieved by "omitting the direct reaches by not registering a producer/iroh reach" — but registration is process-global. So you **cannot** have stream X be relay-only-anonymized while stream Y on the same process goes direct. Anonymization is a per-stream decision wired as a per-process switch; the granularities don't match.

2. **Per-tenant relay isolation is impossible.** Tenants sharing a process are forced onto the same relay endpoint. The federated-scheduling design routes per-OID/per-session; a process-global relay is the opposite granularity and leaks tenants into a shared rendezvous they never chose.

3. **`OnceLock` first-write-wins is a clobbering hazard** — the same class as the prior "singleton session key clobbering" (RPC worker isolation). Whichever path initializes first sets the relay for *everything* else in the process; later services silently inherit a relay they never configured. The test already needs an "only assert if we won the `OnceLock` set" dance (`moq_stream.rs:1621`) — that contortion is the singleton admitting it fights correctness.

Same smell repeated: `producer_reach()` is a free function reading **three** ambient process globals — `global_iroh_node_id()`, `global_producer_reach()`, `global_relay_reach()`. All three are "what reach does *this endpoint* advertise," which is a property of the thing that owns the QUIC/iroh bind, not of the process. (Related cleanup: #380 SocketKind / `INPROC_REGISTRY` / `GLOBAL_RESOLVER` de-singletonization.)

## Desired design (3 tiers of relay config resolution)

Reach config belongs on the **per-server context** that owns the origin/bind. `producer_reach()` becomes a method on that context reading its own fields, threaded explicitly — not ambient process state.

1. **Common case — server-global relay, typically an anycast address.** Each bound server/node carries its relay reach as config (default: its PDS / federation anchor, sourced via the `service_entry` codec from the DID transport entry, as today — just per-server not per-process). The advertised `relayUrl` is commonly an **anycast hostname**, so the network handles nearest-PoP selection (locality lives in the URL, not in per-stream selection). NOTE the QUIC/iroh + IP-anycast caveat: connection-oriented transport can break on mid-connection path-flap to a stateless PoP — anycast relay works best as stateless ingest to a relay mesh, or with PoP-pinning after first contact.

2. **Per-stream relay override.** A producer can override the server-global relay for a specific stream — required for (a) heterogeneous anonymization (relay-only stream X alongside direct stream Y in the same process) and (b) per-tenant relay isolation. This is the granularity the current singleton makes impossible. With anycast handling locality, per-stream override is about *which relay logically* + *anonymization*, not *which instance*.

3. **(Future) scheduled relay selection.** A relay chosen per-stream by a scheduler/policy (load, locality, tenant policy). Out of scope for this issue; the per-server + per-stream-override structure must not preclude it.

## Scope / acceptance

- Move relay reach (and ideally the iroh/producer reach siblings) off process-global `OnceLock` onto a per-server context struct; `producer_reach()` becomes a method.
- Support a per-stream relay override at the producer API.
- Two streams on one process can have different relay/anonymization posture (regression test).
- Remove the `OnceLock`-win test contortion (`moq_stream.rs:1621`).
- No capnp/wire change expected (this is producer-side reach *construction*; the wire `Destination`/`announcedAt` shape is unchanged). Confirm during impl.

## Why this gates the epic

#355 claims **per-stream anonymization** (relay-only streams). That guarantee is **not honestly deliverable** while relay choice is process-global — so this is a correctness blocker for the anonymization property, not just hygiene. Should land before the final epic PR claims anonymization works.

## Refs

- `crates/hyprstream-rpc/src/moq_stream.rs:180` (GLOBAL_RELAY_REACH), `:217-249` (producer_reach), `:1621` (OnceLock-win test dance)
- PR #378 (introduced the relay machinery), #358 (S3 relay rendezvous)
- Related de-singletonization: #380

ewindisch:
## Summary

- Removes `GLOBAL_RELAY_REACH` singleton (closes #384)
- Per-server relay reach config via `ServerConfig`
- Per-stream override from `StreamOpt`/`StreamInfo.reach`

## Motivation

The global singleton prevents running multiple moq relay configs in the same process (e.g., tests, multi-tenant, embedded). Closes the S7 streaming plane gap.

## Test plan

- [ ] `cargo test --workspace` passes
- [ ] `cargo clippy --workspace` clean
- [ ] Two-server test: different relay configs coexist without interference


ewindisch:
## Agentic Code Review

**Overall:** The PR correctly achieves its stated primary goal: `ProducerReachConfig` provides a clean, Clone+Send+Sync per-server reach config that enables heterogeneous relay choices in one process, the anonymization invariant (`RelayChoice::Only` suppressing direct reaches) is correctly implemented and well-tested. The biggest risk is that `StreamContext::reach()` — the new de facto call site — still routes through `global_reach_config()` and reads the OnceLock singletons for the base node identity, so the full de-singletonization goal remains incomplete: per-stream relay choice works, but per-server iroh/QUIC identity isolation does not. The dead `NodeStreamReach.iroh_node_id` field (pre-existing, not introduced by this PR) creates a second place to set the iroh node identity that has no effect and could confuse callers constructing `ProducerReachConfig` with an explicit `quic_reach`.

### Findings

**1. [CORRECTNESS]** `crates/hyprstream-rpc/src/streaming.rs:335`
StreamContext::reach() reads process-global OnceLocks via global_reach_config(), so the base node identity (iroh + QUIC) is still singleton-sourced even for streams with an explicit relay_choice, meaning the heterogeneous-per-server config property is only half de-singletonized for the compat path.
> A service handler that constructs a ProducerReachConfig with a custom iroh_node_id and threads it per-stream expects isolation, but StreamContext::reach() ignores the ProducerReachConfig entirely and re-reads GLOBAL_IROH_NODE_ID and GLOBAL_PRODUCER_REACH. Any per-server reach set outside the OnceLock globals is silently dropped — the iroh/QUIC reach in the advertised StreamInfo will always be from the first process-global registration, not from the per-stream config.

**2. [NIT]** `crates/hyprstream-rpc/src/moq_stream.rs:169`
RelayChoice enum has a unit variant named None, which shadows Option::None if RelayChoice is glob-imported (use RelayChoice::*), creating a latent footgun for future callers or macros that glob-import this enum.
> A future caller writes 'use RelayChoice::*' in a context that also uses Option, causing Option::None to be shadowed by RelayChoice::None. Code that tries to construct 'None' would silently become RelayChoice::None instead of an Option, producing a type error at best — or a silent logic error if types happen to coerce.

**3. [CORRECTNESS]** `crates/hyprstream-rpc/src/moq_stream.rs:103`
NodeStreamReach.iroh_node_id (pre-existing field, not introduced by this PR) is never read by reach_with_relay(); ProducerReachConfig keeps iroh_node_id as a separate top-level field, making NodeStreamReach.iroh_node_id a dead field that callers might incorrectly set believing it will be advertised.
> A caller constructs ProducerReachConfig { iroh_node_id: None, quic_reach: Some(NodeStreamReach { iroh_node_id: Some(key), .. }), .. } expecting the iroh reach to be advertised. reach_with_relay() checks self.iroh_node_id (None) and skips the iroh Destination entirely; the iroh_node_id stored in quic_reach is silently ignored. The resulting StreamInfo omits the iroh-direct reach.

**4. [NIT]** `crates/hyprstream-rpc/tests/moq_relay_rendezvous.rs:460`
The end-to-end test constructs NodeStreamReach with iroh_node_id: Some([0x11u8; 32]) inside quic_reach, but this field has no effect on what reach is advertised — the iroh reach in the test comes from ProducerReachConfig.iroh_node_id at line 455, not from NodeStreamReach.iroh_node_id, making the test's intent ambiguous.
> If a reviewer or future developer reads the test and removes the top-level ProducerReachConfig.iroh_node_id field (or sets it to None) while keeping NodeStreamReach.iroh_node_id, the assertions about the direct reach still holding would silently fail or pass incorrectly depending on the process-global state from other tests.

**5. [CLEANUP]** `crates/hyprstream-rpc/src/moq_stream.rs:353`
global_reach_config() is explicitly documented as a deprecated compat shim, but StreamContext::reach() (the new primary call site after this PR) uses it unconditionally rather than accepting an injected ProducerReachConfig, so the 'new code should thread an explicit ProducerReachConfig' migration is structurally blocked at the StreamContext boundary until a follow-up PR adds the field.
> When a future change adds ProducerReachConfig as a field of StreamContext to complete the de-singletonization, all existing StreamContext::new() and from_dh() call sites will need updating; the current design defers this work without marking it as a TODO in the code, risking indefinite deferral.

ewindisch:
## Agentic Code Review

**Overall:** The PR delivers a sound conceptual improvement — ProducerReachConfig and RelayChoice are well-designed, the anonymization guard (relay_only) is correctly implemented, and the test coverage for the heterogeneous-reach property is thorough. The biggest risk is that StreamContext::reach() is advertised as the fully-threaded per-stream replacement for producer_reach() but still calls global_reach_config() for the base direct-reach config (iroh/Quic), so the de-singletonization is partial: relay posture is per-stream but direct-reach is still process-global; a secondary dead-field issue in NodeStreamReach.iroh_node_id could silently mislead future callers.

### Findings

**1. [CORRECTNESS]** `crates/hyprstream-rpc/src/streaming.rs:335`
StreamContext::reach() still reads process globals via global_reach_config(), so the iroh_node_id and quic_reach base-reach portions are NOT per-stream — only relay_choice is truly threaded, making the method's docstring claim of being 'the threaded replacement for producer_reach()' overstated.\n> A multi-tenant deployment sets relay_choice=Only on a StreamContext for tenant A expecting anonymization, while the direct-reach base (iroh/Quic) is still sourced from the same process OnceLock as tenant B; the relay_only guard in reach_with_relay correctly withholds direct reaches for Only, so anonymization is preserved, but the fundamental goal of letting two servers run with DIFFERENT direct-reach configs in one process is still blocked by the global dependency for anything other than relay.\n\n**2. [CLEANUP]** `crates/hyprstream-rpc/src/moq_stream.rs:114`\nNodeStreamReach.iroh_node_id is a newly added public field that is never read by reach_with_relay() — the reach builder reads ProducerReachConfig.iroh_node_id exclusively, making the NodeStreamReach field dead and potentially misleading to callers who set it expecting it to influence reach output.\n> A caller constructs NodeStreamReach { iroh_node_id: Some(key), .. } and passes it to init_global_producer_reach() expecting the iroh direct-reach to appear in producer reach; it will not appear because reach_with_relay reads ProducerReachConfig.iroh_node_id (from GLOBAL_IROH_NODE_ID), not the NodeStreamReach field — iroh reach is silently absent unless init_global_iroh_node_id() is also called separately.\n\n**3. [NIT]** `crates/hyprstream-rpc/tests/moq_relay_rendezvous.rs:312`\nThe new relay_choice_only_anonymizes_stream_end_to_end test uses a hardcoded 1-second sleep to synchronize the relay announce link before subscribing, creating a time-based race that can intermittently fail on slow CI.\n> On a loaded CI machine the run_relay_announce_link spawned task has not yet completed its handshake within 1 second; connect_moq_reach_for_qos then fails to find the announced broadcast, causing the test to fail with 'broadcast not announced via relay'.\n\n**4. [NIT]** `crates/hyprstream-rpc/tests/moq_relay_rendezvous.rs:260`\nThe test constructs NodeStreamReach with iroh_node_id=Some([0x11u8;32]) inside ProducerReachConfig.quic_reach, but reach_with_relay only reads ProducerReachConfig.iroh_node_id (a separate field); the test does not catch the dead NodeStreamReach.iroh_node_id because both fields are independently set to the same value.\n> If a future refactor removes the separate GLOBAL_IROH_NODE_ID singleton and moves iroh_node_id exclusively into NodeStreamReach, the test will still pass (because ProducerReachConfig.iroh_node_id is also set), masking the fact that the reach builder is now reading a stale source — silent regression.

ewindisch:
## Review findings — responses

All review findings are resolved. No capnp schema changes in this PR (diff is pure Rust — `moq_stream.rs`/`streaming.rs`/services/tests), so no ordinal regeneration / field removal was needed.

| # | Finding | Resolution |
|---|---------|------------|
| 1 | `StreamContext::reach()` read process globals (incomplete de-singletonization) | **Fixed** `71106ac7` — added `reach_config: ProducerReachConfig` to `StreamContext`; `reach()` now calls `self.reach_config.reach_with_relay(..)` (`streaming.rs:411`), not `global_reach_config()`. Per-server base reach is overridable via `with_reach_config` (`StreamChannel::prepare_stream`). |
| 2 | `RelayChoice::None` shadows `Option::None` | **Fixed** `370ff6fa` — renamed to `RelayChoice::NoRelay`. |
| 3 | dead `NodeStreamReach.iroh_node_id` | **Fixed** `4d1c1db8` — field removed; reach builder reads `ProducerReachConfig.iroh_node_id` exclusively. |
| 4 | test set the dead field ambiguously | **Fixed** `7820db0d` — test now asserts the live source (`ProducerReachConfig.iroh_node_id`) with clarifying comments. |
| 5 | `global_reach_config()` compat shim blocked at the `StreamContext` boundary | **Addressed** by #1 — `with_reach_config` is the injection hook; new code threads an explicit config. |
| 6 | fixed 1s `sleep` to sync the relay announce (flaky) | **Fixed** `7820db0d` (in `relay_choice_only_…`) **+ `681388dc`** (this push) — also removed the same redundant sleep from `moq_relay_rendezvous`; both now rely on the deterministic `timeout(announced_broadcast)`. |

### One documented residual (not a regression)
`StreamContext::from_dh()` still snapshots `global_reach_config()` **once** at construction as the compat default (`streaming.rs:292`). Fully eliminating that last global means threading a `ProducerReachConfig` from each daemon bind site through the ~7 `StreamChannel::new` call sites + the spawner — the *mechanism* is now in place (`with_reach_config`); the broader wiring is follow-up. The hot `reach()` path no longer reads globals.

PR #481, opened 2026-06-26
feat(nix): complete nix flake packaging — CPU/CUDA/ROCm variants + CI workflow

ewindisch:
## Summary

- Complete nix flake for hyprstream (CPU, CUDA, ROCm variants)
- CI workflow integration

## Test plan

- [ ] `nix flake check` passes
- [ ] CPU build: `nix build .#hyprstream`
- [ ] CUDA/ROCm variant builds (if hardware available)


ewindisch:
## Agentic Code Review

**Overall:** The PR successfully wires up a multi-variant (CPU/CUDA/ROCm) Nix flake using crane, which is a sound architectural choice, but it ships in an incomplete and partially broken state: nix/hyprstream.nix is a dead file with all-fakeHash placeholders and collect-hashes.sh patches the wrong file (it targets hyprstream.nix while `nix build` uses the crane path). The most urgent practical issues are the CUDA/ROCm package references that break flake evaluation on non-x86_64-linux systems, and the sccache RUSTC_WRAPPER that will silently break `cargo` in the devShell on developer workstations that lack /var/cache/sccache.

### Findings

**1. [BUG]** `nix/hyprstream.nix:40`
All outputHashes are set to lib.fakeHash placeholders — this file cannot build successfully as committed.
> Any user who tries `nix build` using the rustPlatform path (callPackage ./nix/hyprstream.nix) gets an immediate hash-mismatch failure for every git dependency listed.

**2. [BUG]** `nix/collect-hashes.sh:11`
collect-hashes.sh patches nix/hyprstream.nix, but `nix build .#hyprstream-cpu` uses the crane-based flake.nix — so the patches have no effect and the script loops pointlessly until MAX_ITERATIONS.
> Developer runs `./nix/collect-hashes.sh` to populate hashes; the script runs 50 nix build iterations, each time patching a file that the build system ignores, never making progress and always exiting at the iteration limit.

**3. [CORRECTNESS]** `flake.nix:80`
The Nix package version is hard-coded as 0.3.0-rc1 while Cargo.toml declares version 0.5.0, causing a misleading version label on all produced outputs.
> A user installs the Nix-packaged binary and inspects `hyprstream --version` or the Nix store path name; they see 0.3.0-rc1 instead of 0.5.0, making release tracking and bug reports unreliable.

**4. [CORRECTNESS]** `flake.nix:173`
cudaPackages and rocmPackages are referenced unconditionally inside eachDefaultSystem, causing evaluation to fail on any non-x86_64-linux system (macOS, aarch64-linux) where those attributes do not exist in nixpkgs.
> Developer on macOS runs `nix flake check` or `nix develop`; Nix evaluates all outputs for all systems and hits `error: attribute 'cudaPackages' missing` (or similar) for the darwin systems, making the entire flake unusable on that machine.

**5. [BUG]** `flake.nix:118`
RUSTC_WRAPPER, SCCACHE_DIR, and SCCACHE_CACHE_SIZE from commonArgs leak into the devShell via inputsFrom, so `cargo build` in the dev shell silently requires sccache with /var/cache/sccache on the developer's machine.
> Developer enters `nix develop` on a workstation that doesn't have /var/cache/sccache; cargo invocations fail with a cryptic sccache connection error rather than a clear build failure, with no indication that sccache is involved.

**6. [SECURITY]** `.github/workflows/nix.yml:18`
The sccache directory is created world-writable (chmod 1777) and exposed into the Nix sandbox via extra-sandbox-paths, meaning any process running in the sandbox can write arbitrary entries into the compilation cache.
> On a self-hosted runner (or if the workflow is ever moved to one), a malicious dependency's build script could write poisoned compiled artifacts into /var/cache/sccache; those artifacts are then trusted and used in subsequent builds of the main hyprstream binary.

**7. [CORRECTNESS]** `nix/libtorch.nix:81`
ROCm patchelf only targets the unversioned libelf.so and libamd_comgr.so symlinks; if the libtorch ROCm zip ships versioned sonames (libelf.so.1, libamd_comgr.so.2 etc.) the patch is silently skipped and the runtime RPATH is wrong.
> User runs `nix build .#hyprstream-rocm71`; the build succeeds but at runtime loading the binary on NixOS fails with `libz.so.1: cannot open shared object file` because the versioned sonames were not patched.

**8. [CLEANUP]** `nix/hyprstream.nix:22`
nix/hyprstream.nix is a dead file — flake.nix does not import or reference it; it uses crane directly, making the rustPlatform derivation unreachable from any flake output.
> The file is maintained in parallel with flake.nix, accumulates drift, and confuses contributors who try to add it to the flake or use it as a reference, creating incorrect expectations about how the build works.

ewindisch:
**Triage:** rebased onto current main (74 commits behind; head was `cda252b6` -> now `c633d8a0`, clean merge, no conflicts). The prior `build` FAILURE showed no actual test/compile errors in the log (all visible test results were `ok`) and this diff only touches `flake.nix`/`nix/*`/`.github/workflows/nix.yml` — no Rust source — so it looks like a transient/stale-base CI issue, not something this PR caused. All other checks (clippy, cargo-deny, CodeQL, WASM, nix flake check) were already green. Re-running CI now against the refreshed base; recommend merge once `build` passes clean.

Issue #341, opened 2026-06-19
Epic: Worker GA — first-class Kata 4.0 + casual podman/nested-container path


ewindisch:
Tracking epic to take `hyprstream-workers` to **GA**: first-class **Kata 4.0** (runtime-rs GA/default) in cooperation with the Kata team, **plus** a casual-user path via **podman/docker + nested-userspace containers**, with feature-gated builds so casual installs don't pay the Kata toolchain cost.

**Branch:** `feature/kata-ga-epic` (off `main`). All epic PRs target this branch; merging it closes the epic. **Plan:** `docs/plans/2026-06-19-worker-ga-epic.md`.

## Current state (investigation)
Mature **CRI-aligned** framework (`WorkerService` = K8s CRI; pluggable `SandboxBackend` trait; `workers.capnp`; Nydus RAFS image CAS; event bus; WorkflowService; drafted policy vocab). Two backends (default Kata): **KataBackend** embeds runtime-rs `hypervisor` crate as a library (pinned 3.25.0, cloud-hypervisor); **NspawnBackend** boots the host rootfs under systemd-nspawn (not an OCI runtime).

## Top GA gaps
- **No guest agent → Kata can't run workloads** (`exec_sync` = "requires agent"; vsock channel unbuilt). Biggest functional gap.
- **Kata 4.0 integration-surface decision** — embed `hypervisor` crate vs `containerd-shim-kata-v2`/runtime-rs binary (needs Kata-team input).
- **No podman/docker backend** (casual path unmet).
- **Build punishes casual users** (unconditional kata/nydus/CH deps → needs feature-gating).
- No availability-based backend selection; WorkflowService partial; still on ZMQ (not moq); Nydus git-dep debt.

## Tracks / sub-issues
**Track 1 — Kata 4.0 first-class**
- [x] #342 — T1-A bump runtime-rs 3.25.0 → 4.0 GA; absorb hypervisor/kata-types/ch-config API breaks
- [x] #343 — T1-B integration-surface decision (embed crate vs shim/runtime-rs binary) — Kata-team gated
- [x] #344 — T1-C guest agent (kata-agent vsock): real CRI create/start/exec_sync of container-in-VM
- [x] #345 — T1-D real VM stats/console/get_pids/update_resources

**Track 2 — Casual path**
- [x] #346 — T2-A rootless podman/docker (or youki/crun) OCI SandboxBackend
[...]

Issue #348, opened 2026-06-19
T2-C: availability-based backend auto-selection + fallback (kata to podman to nspawn) + wizard/CLI UX


ewindisch:
Track 2. is_available exists but is not used to auto-pick or fall back. Add auto-detect + ordered fallback + clear wizard/CLI UX. Part of #341. Plan: docs/plans/2026-06-19-worker-ga-epic.md

ewindisch:
Implemented — PR #624. Adds explain_selection/unavailable_reason/list_backends_for_wizard diagnostics on top of the existing (already-correct) BackendRegistration/fail-closed selection machinery. Purely additive, 26/26 tests pass.

ewindisch:
this is / should not be automatic. the system should now allow explicit request for backends and capability-based scheduling to arbitrary backends based on availability.

PR #484, opened 2026-06-26
feat(workers): rootless podman/docker OCI SandboxBackend (#346)

### LINKED ISSUE #346: T2-A: rootless podman/docker (or youki/crun) OCI SandboxBackend
Track 2 (casual path). Net-new SandboxBackend running arbitrary OCI images as rootless nested-userspace containers. Builds on the existing CRI trait + OCI manifest plumbing. Decision D2: podman/docker shell-out vs embedded OCI runtime. Part of #341. Plan: docs/plans/2026-06-19-worker-ga-epic.md

ewindisch:
## Summary

- Implements `PodmanBackend` in `crates/hyprstream-workers/src/runtime/podman.rs`, closing #346 (T2-A: rootless podman/docker OCI SandboxBackend)
- Feature-gated behind the new `podman` feature (no compile-time toolchain overhead — `is_available()` detects `podman`/`docker` at runtime)
- Also fixes a pre-existing `clippy::expect_used` lint miss in `streaming.rs` test (was blocking the pre-commit hook)

## Design

**`PodmanBackend`** shells out to `podman` (falling back to `docker`) for the full sandbox lifecycle:

| Method | Command |
|---|---|
| `start()` | `podman run --detach --rm --name=hyprstream-{id} --network=slirp4netns <image> sleep infinity` |
| `stop()` | `podman stop hyprstream-{id}` |
| `destroy()` | `podman rm --force hyprstream-{id}` |
| `exec_sync()` | `podman exec hyprstream-{id} -- <cmd>` (with `tokio::time::timeout`) |
| `update_resources()` | `podman update --cpu-quota/--memory` |
| `reset()` | returns `false` (pool recreates the container) |
| `get_pids()` | `podman inspect --format {{.State.Pid}}` |

**Image resolution**: `hyprstream.io/worker-image` annotation overrides `HYPRSTREAM_WORKER_IMAGE` env var (default: `ghcr.io/hyprstream/worker:latest`).

## Test plan

- [ ] `cargo clippy -p hyprstream-workers --features podman` passes (verified locally)
- [ ] CI clippy/build/wasm green

Closes #346. Part of epic #341.

### DIFF
--- a/crates/hyprstream-workers/src/runtime/podman.rs
+++ b/crates/hyprstream-workers/src/runtime/podman.rs
@@ -1,0 +1,431 @@
+//! PodmanBackend — rootless podman/docker OCI SandboxBackend
+//!
+//! Runs each sandbox as a rootless OCI container via `podman` (or `docker` as
+//! fallback).  No hypervisor, no RAFS — pulls a standard OCI image and keeps
+//! it running for `exec_sync` workloads.  Designed for casual/dev installs
+//! where Kata + Cloud-Hypervisor toolchain is unavailable.
+//!
+//! Backend type: `"podman"`.
+
+use std::any::Any;
+use std::collections::HashMap;
+use std::path::PathBuf;
+use std::sync::Arc;
+use std::time::Duration;
+
+use async_trait::async_trait;
+use tracing::{debug, info, warn};
+
+use crate::config::PoolConfig;
+use crate::error::{Result, WorkerError};
+
+use super::backend::{SandboxBackend, SandboxHandle};
+use super::client::{LinuxContainerResources, PodSandboxConfig};
+use super::sandbox::PodSandbox;
+
+// ─────────────────────────────────────────────────────────────────────────────
+// Helpers
+// ─────────────────────────────────────────────────────────────────────────────
+
+/// Locate the container runtime binary: prefer `podman`, fall back to `docker`.
+fn find_runtime() -> Option<PathBuf> {
+    for name in &["podman", "docker"] {
+        if let Ok(path) = which::which(name) {
+            return Some(path);
+        }
+    }
+    None
+}
+
+/// Container name for a given sandbox id.
+fn container_name(sandbox_id: &str) -> String {
+    format!("hyprstream-{sandbox_id}")
+}
+
+// ─────────────────────────────────────────────────────────────────────────────
+// Configuration
+// ─────────────────────────────────────────────────────────────────────────────
+
+/// Configuration for the podman backend.
+#[derive(Debug, Clone)]
+pub struct PodmanConfig {
+    /// OCI base image to run.  Overridable per-sandbox via annotations.
+    pub base_image: String,
+    /// Network mode passed to `--network`.
+    pub network_mode: String,
+    /// Extra volume mounts (`host:container[:options]`).
+    pub extra_mounts: Vec<String>,
+    /// Extra env vars forwarded to the container.
+    pub extra_env: Vec<(String, String)>,
+    /// Stop timeout (seconds) before forceful kill.
+    pub stop_timeout: u64,
+    /// Exec timeout default.
+    pub exec_timeout: Duration,
+}
+
+impl Default for PodmanConfig {
+    fn default() -> Self {
+        Self {
+            base_image: std::env::var("HYPRSTREAM_WORKER_IMAGE")
+                .unwrap_or_else(|_| "ghcr.io/hyprstream/worker:latest".to_owned()),
+            network_mode: "slirp4netns".to_owned(),
+            extra_mounts: Vec::new(),
+            extra_env: Vec::new(),
+            stop_timeout: 10,
+            exec_timeout: Duration::from_secs(300),
+        }
+    }
+}
+
+// ─────────────────────────────────────────────────────────────────────────────
+// Handle
+// ─────────────────────────────────────────────────────────────────────────────
+
+/// Backend-specific state stored on each `PodSandbox`.
+#[derive(Debug, Clone)]
+pub struct PodmanHandle {
+    /// Sandbox identifier.
+    pub sandbox_id: String,
+    /// Container name (`hyprstream-{sandbox_id}`).
+    pub container_name: String,
+    /// PID of the container init process (populated after start).
+    pub init_pid: Option<u32>,
+    /// Runtime binary used (podman or docker path).
+    pub runtime_bin: PathBuf,
+}
+
+impl SandboxHandle for PodmanHandle {
+    fn as_any(&self) -> &dyn Any {
+        self
+    }
+}
+
+// ─────────────────────────────────────────────────────────────────────────────
+// Backend
+// ─────────────────────────────────────────────────────────────────────────────
+
+/// Rootless podman/docker OCI SandboxBackend.
+pub struct PodmanBackend {
+    config: PodmanConfig,
+}
+
+impl PodmanBackend {
+    pub fn new(config: PodmanConfig) -> Self {
+        Self { config }
+    }
+
+    fn runtime_bin(&self) -> Option<PathBuf> {
+        find_runtime()
+    }
+
+    /// Inspect a running container and return its init PID.
+    async fn inspect_pid(runtime: &PathBuf, name: &str) -> Option<u32> {
+        let out = tokio::process::Command::new(runtime)
+            .args(["inspect", "--format", "{{.State.Pid}}", name])
+            .output()
+            .await
+            .ok()?;
+        if !out.status.success() {
+            return None;
+        }
+        let pid_str = String::from_utf8_lossy(&out.stdout).trim().to_owned();
+        pid_str.parse().ok().filter(|&p: &u32| p > 0)
+    }
+
+    /// Resolve the OCI image: annotation overrides config default.
+    fn resolve_image(&self, annotations: &HashMap<String, String>) -> String {
+        annotations
+            .get("hyprstream.io/worker-image")
+            .cloned()
+            .unwrap_or_else(|| self.config.base_image.clone())
+    }
+}
+
+#[async_trait]
+impl SandboxBackend for PodmanBackend {
+    fn backend_type(&self) -> &'static str {
+        "podman"
+    }
+
+    fn is_available(&self) -> bool {
+        find_runtime().is_some()
+    }
+
+    async fn initialize(&self, _config: &PoolConfig) -> Result<()> {
+        let bin = self.runtime_bin().ok_or_else(|| {
+            WorkerError::VmStartFailed(
+                "podman or docker not found in PATH".to_owned(),
+            )
+        })?;
+
+        let out = tokio::process::Command::new(&bin)
+            .args(["info", "--format", "{{.Host.Security.Rootless}}"])
+            .output()
+            .await
+            .map_err(|e| {
+                WorkerError::VmStartFailed(format!(
+                    "failed to run `{} info`: {e}",
+                    bin.display()
+                ))
+            })?;
+
+        if !out.status.success() {
+            return Err(WorkerError::VmStartFailed(format!(
+                "`{} info` failed: {}",
+                bin.display(),
+                String::from_utf8_lossy(&out.stderr)
+            )));
+        }
+
+        let rootless = String::from_utf8_lossy(&out.stdout).trim() == "true";
+        info!(
+            runtime = %bin.display(),
+            rootless,
+            "PodmanBackend initialized"
+        );
+        Ok(())
+    }
+
+    async fn start(
+        &self,
+        sandbox: &mut PodSandbox,
+        _config: &PodSandboxConfig,
+        _pool_config: &PoolConfig,
+        annotations: &HashMap<String, String>,
+    ) -> Result<Arc<dyn SandboxHandle>> {
+        let bin = self.runtime_bin().ok_or_else(|| {
+            WorkerError::VmStartFailed("no container runtime in PATH".to_owned())
+        })?;
+        let image = self.resolve_image(annotations);
+        let name = container_name(&sandbox.id);
+
+        info!(sandbox_id = %sandbox.id, container = %name, image = %image, "Starting podman sandbox");
+
+        let mut args = vec![
+            "run".to_owned(),
+            "--detach".to_owned(),
+            "--rm".to_owned(),
+            format!("--name={name}"),
+            format!("--network={}", self.config.network_mode),
+            format!("--stop-timeout={}", self.config.stop_timeout),
+        ];
+
+        for mount in &self.config.extra_mounts {
+            args.push(format!("--volume={mount}"));
+        }
+        for (k, v) in &self.config.extra_env {
+            args.push(format!("--env={k}={v}"));
+        }
+
+        args.push(image.clone());
+        args.push("sleep".to_owned());
+        args.push("infinity".to_owned());
+
+        let status = tokio::process::Command::new(&bin)
+            .args(&args)
+            .status()
+            .await
+            .map_err(|e| {
+                WorkerError::VmStartFailed(format!("podman run failed: {e}"))
+            })?;
+
+        if !status.success() {
+            return Err(WorkerError::VmStartFailed(format!(
+                "podman run exited non-zero for sandbox {}",
+                sandbox.id
+            )));
+        }
+
+        let init_pid = Self::inspect_pid(&bin, &name).await;
+        debug!(sandbox_id = %sandbox.id, pid = ?init_pid, "Container started");
+
+        Ok(Arc::new(PodmanHandle {
+            sandbox_id: sandbox.id.clone(),
+            container_name: name,
+            init_pid,
+            runtime_bin: bin,
+        }))
+    }
+
+    async fn stop(&self, sandbox: &PodSandbox) -> Result<()> {
+        let bin = match self.runtime_bin() {
+            Some(b) => b,
+            None => return Ok(()),
+        };
+        let name = container_name(&sandbox.id);
+        info!(sandbox_id = %sandbox.id, container = %name, "Stopping podman sandbox");
+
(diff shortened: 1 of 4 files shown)


ewindisch:
## Agentic Code Review

**Overall:** The PR introduces a working podman-only implementation that is clearly labeled "rootless podman/docker" but is functionally broken when docker is the actual fallback runtime: `initialize()` misreads docker's rootless state (bypassing a security check), `start()` passes `--network=slirp4netns` which docker does not support, and the runtime binary selection is re-resolved on every call rather than being pinned at start time (so the handle's stored `runtime_bin` is dead code). The biggest risk is the security check bypass: when only docker is available, the rootless guard silently passes for a root-mode daemon, allowing sandbox workloads to run as root.

### Findings

**1. [SECURITY]** `crates/hyprstream-workers/src/runtime/podman.rs:176`
initialize() uses a podman-specific Go template `{{.Host.Security.Rootless}}` that does not exist in docker's info output, causing the rootless check to silently produce the wrong result (docker returns empty/`<no value>` but exits 0), so a rooted docker daemon is incorrectly accepted as rootless.
> podman is absent, docker is present; find_runtime() returns docker; initialize() runs `docker info --format '{{.Host.Security.Rootless}}'` which exits 0 with empty output; the rootless check logs `rootless=false` (empty string != "true") but does not reject the backend, so a root-mode docker daemon passes initialization and sandboxes run with root privilege, violating the rootless isolation promise.

**2. [BUG]** `crates/hyprstream-workers/src/runtime/podman.rs:218`
start() passes `--network=slirp4netns` to docker which is a podman-only network mode; docker does not support it and will exit non-zero, making every sandbox start fail when docker is the only available runtime.
> podman is absent, docker is found; `start()` builds `docker run --network=slirp4netns ...`; docker exits with "invalid argument"; `WorkerError::VmStartFailed` is returned for every sandbox creation attempt, rendering the backend non-functional when advertised as working with docker.

**3. [CORRECTNESS]** `crates/hyprstream-workers/src/runtime/podman.rs:239`
start() uses `.status()` instead of `.output()`, so stderr from `podman run` (e.g., image pull failures, "manifest not found", registry auth errors) is inherited by the hyprstream process and discarded from the error path, yielding a generic "non-zero" message with no diagnostic detail.
> Pulling `ghcr.io/hyprstream/worker:latest` fails due to a registry outage; podman exits non-zero; WorkerError::VmStartFailed says only "podman run exited non-zero for sandbox abc"; operator has no information about the cause without separate log capture.

**4. [CORRECTNESS]** `crates/hyprstream-workers/src/runtime/podman.rs:181`
find_runtime() is called fresh on every operation (start, stop, destroy, get_pids, exec_sync, update_resources), meaning the runtime binary could change between calls within a single sandbox lifecycle if podman is installed or removed during operation; the runtime stored in PodmanHandle.runtime_bin is also never actually used by the backend methods.
> Sandbox is started with podman (stored in PodmanHandle.runtime_bin); podman is uninstalled while containers are running; destroy() calls find_runtime() again, gets docker (or None, returning Ok(()) silently), and the container is never actually removed, leaking the `sleep infinity` container indefinitely.

**5. [CORRECTNESS]** `crates/hyprstream-workers/src/runtime/podman.rs:62`
The extra_env field is `Vec<(String, String)>` but values are formatted as `--env=KEY=VALUE`; if a value contains `=` (common in e.g. base64 or connection strings) the entire extra value is passed correctly since it goes as a single arg to tokio::process::Command, however a key containing `=` would silently split at the first `=` inside podman's parser, producing a key of the prefix and a value starting from the split point.
> Caller sets extra_env to `[("MY=KEY", "value")]`; the arg becomes `--env=MY=KEY=value`; podman interprets this as env var `MY` = `KEY=value`, silently misinterpreting the intended key name with no error returned.

**6. [NIT]** `crates/hyprstream-workers/Cargo.toml:127`
The comment above the `podman = []` feature says "Enabled by default since podman availability is checked at runtime via is_available()" but the `default = []` feature list is empty, so the podman backend is NOT compiled by default; the comment is wrong and misleading for integrators.
> A developer reads the Cargo.toml comment and assumes the podman backend is active in default builds, then debugs why `PodmanBackend` is not exported/available when building without `--features podman`.

**7. [CORRECTNESS]** `crates/hyprstream-workers/src/runtime/podman.rs:152`
destroy() silently swallows all errors from `podman rm --force` via `let _ = ...`, so a failure to remove the container (e.g., permission denied, podman daemon crash) leaves the sandbox container running with no signal to the caller or operator.
> A container fails to be removed due to a podman storage corruption; destroy() returns Ok(()); the pool considers the sandbox destroyed; the `sleep infinity` container continues consuming resources and the sandbox ID is now in an inconsistent state.

**8. [NIT]** `crates/hyprstream-workers/src/runtime/podman.rs:119`
PodmanHandle derives Clone, which means it can be cloned even though it represents a live OS resource (a running container); cloning and dropping the duplicate could lead to incorrect lifecycle assumptions if handle clones are used for cleanup operations.
> Code that downcasts backend_handle and clones the PodmanHandle might assume the clone represents its own container resource; in the current pool code this is benign since handles are stored in Arc, but it is an API footgun if the handle type is extended with Drop-based cleanup in the future.

ewindisch:
## Agentic Code Review

**Overall:** The PR is a clean, well-structured addition that correctly implements the SandboxBackend trait and follows existing patterns. The highest-risk issue is the annotation-controlled image override (hyprstream.io/worker-image) which lets any RPC caller pull and run arbitrary OCI images without a registry allow-list or policy check; this should be gated behind operator configuration before the feature reaches production use. A secondary concern is the docker fallback's incompatible rootless-detection format string, which silently misreports the privilege level when docker is used instead of podman.

### Findings

**1. [CORRECTNESS]** `crates/hyprstream-workers/src/runtime/podman.rs:219`
initialize() uses a podman-only Go template field (.Host.Security.Rootless) against docker, silently treating docker as non-rootless without any warning
> When docker is first in PATH (podman absent), `docker info --format '{{.Host.Security.Rootless}}'` outputs an empty string or the literal text of the field path — not 'true' or 'false'. The rootless variable evaluates to false, but initialize() returns Ok(()) anyway, logging 'rootless: false'. Operators get a misleading signal about the actual privilege level and no indication the check is meaningless for docker. If docker is running rootful, downstream containers may run as root contrary to the security contract stated in the module doc.

**2. [SECURITY]** `crates/hyprstream-workers/src/runtime/podman.rs:193`
Annotation-controlled image override (hyprstream.io/worker-image) is accepted without any allow-list or validation, allowing callers with RPC access to pull and execute arbitrary OCI images
> The run_pod_sandbox RPC handler passes the full PodSandboxConfig (including caller-supplied annotations) all the way to start() -> resolve_image(). An authenticated RPC caller can set hyprstream.io/worker-image to any registry/image:tag string, causing podman to pull and run an arbitrary image with the process's UID, host network access (slirp4netns is rootless but network-capable), and any extra_mounts configured. There is no registry allow-list check, no image digest pinning, and no policy gate on this annotation key.

**3. [SECURITY]** `crates/hyprstream-workers/src/runtime/podman.rs:273`
Extra env vars are interpolated directly into --env=k=v arguments without sanitizing newlines or = characters in the key
> If extra_env contains a key or value with embedded newlines or if a value contains '=' (which is valid in base64/JWT values), the formatted arg string --env=KEY=val=extra could be misinterpreted by the shell tokenizer inside podman. More critically, if the key itself contains '=' (e.g., from a misconfigured caller), the env var injection silently sets a different variable. While PodmanConfig.extra_env is not directly RPC-controlled today, it originates from operator-level config that could receive untrusted input in multi-tenant deployments.

**4. [CORRECTNESS]** `crates/hyprstream-workers/src/runtime/podman.rs:261`
podman run uses --rm, which destroys the container on stop; this races with destroy() calling podman rm --force on the same container name
> When stop() is called, podman stops and auto-removes the container (--rm). When destroy() is then called (e.g., from pool.rs::release -> reset_sandbox -> backend.destroy), it issues 'podman rm --force hyprstream-{id}' on a container that no longer exists. While destroy() swallows the error with 'let _ =', this causes a spurious non-zero exit and can race during shutdown: if the warm-pool replenishment task calls start() reusing the same ID before destroy() finishes, it will fail with 'container name already in use' (the --rm removal is async).

**5. [NIT]** `crates/hyprstream-workers/src/runtime/podman.rs:378`
exec_sync builds args as ["exec", container_name, "--", ...cmd] but SandboxBackend::exec_sync is never called through the WorkerService RPC path, making the implementation dead code
> WorkerService::exec_sync (service.rs:662) always returns WorkerError::ExecFailed with 'exec not supported in host-level mode without guest agent' and never delegates to backend.exec_sync(). The podman backend's exec_sync implementation is therefore unreachable via the current RPC surface, meaning supports_exec() returning true is misleading and the exec implementation is untested in integration.

**6. [CLEANUP]** `crates/hyprstream-workers/src/runtime/podman.rs:88`
find_runtime() re-invokes which::which on every call site (start, stop, destroy, get_pids, exec_sync, update_resources), doing repeated PATH scans instead of caching the resolved binary at construction time
> Each sandbox operation (start, stop, destroy, exec_sync, update_resources, get_pids) independently re-runs which::which on the full PATH. In a busy pool under parallel sandbox operations this incurs unnecessary filesystem stat calls per operation. More seriously, if PATH changes between calls (e.g., in a container manager that manipulates the environment), the runtime binary could change mid-lifecycle from podman to docker or vice versa, producing inconsistent container names between start and stop.

**7. [NIT]** `crates/hyprstream-workers/Cargo.toml:128`
The new 'podman' feature is not listed in the default features, but the feature comment says it is 'Enabled by default since podman availability is checked at runtime' — the comment contradicts reality
> The Cargo.toml [features] default = [] does not include 'podman', so the backend is compile-time opt-in. The comment on the feature flag says 'Enabled by default since podman availability is checked at runtime via is_available()' which is factually incorrect and will mislead users who expect podman support out of the box without --features podman.

crates/hyprstream-workers/src/runtime/podman.rs
```diff
@@ -0,0 +1,431 @@
...
+    /// Exec timeout default.
+    pub exec_timeout: Duration,
+}
+
+impl Default for PodmanConfig {
+    fn default() -> Self {
+        Self {
+            base_image: std::env::var("HYPRSTREAM_WORKER_IMAGE")
+                .unwrap_or_else(|_| "ghcr.io/hyprstream/worker:latest".to_owned()),
+            network_mode: "slirp4netns".to_owned(),
```
ewindisch, line 71: Network mode is hardcoded to `slirp4netns` and the CRI `PodSandboxConfig` network model is never consulted (see the `_config` comment below). CRI callers that request host networking, specific DNS (`dns_config`), published ports (`port_mappings`), or a particular network-namespace mode silently get slirp4netns with none of it applied. Architecture question — see Decision Q1.

ewindisch (commented):
Inline review for the rootless podman/docker OCI backend (#346). No code changes requested inline yet — these findings feed the joint architecture/authz decision posted as a separate comment (see the 🔶 Decision-required comment, cross-linked with #486). The headline issue: this backend discards the CRI `PodSandboxConfig` wholesale and is not yet wired into `BackendType`, so it is also currently dead code.

crates/hyprstream-workers/src/runtime/podman.rs
```diff
@@ -0,0 +1,431 @@
...
+            return None;
+        }
+        let pid_str = String::from_utf8_lossy(&out.stdout).trim().to_owned();
+        pid_str.parse().ok().filter(|&p: &u32| p > 0)
+    }
+
+    /// Resolve the OCI image: annotation overrides config default.
+    fn resolve_image(&self, annotations: &HashMap<String, String>) -> String {
+        annotations
+            .get("hyprstream.io/worker-image")
```
ewindisch, line 138: Arbitrary image pull (authz gap): any authorized RPC caller can set the `hyprstream.io/worker-image` annotation to pull and run **any** OCI image from **any** registry — no allowlist, no digest pinning, no operator gate. Combined with `--rm` + `sleep infinity` + `exec_sync`, this is a generic "run any container, exec any command" primitive for anyone who can reach the worker RPC. See Decision Q3.

crates/hyprstream-workers/src/runtime/podman.rs
```diff
@@ -0,0 +1,431 @@
...
+
+    async fn initialize(&self, _config: &PoolConfig) -> Result<()> {
+        let bin = self.runtime_bin().ok_or_else(|| {
+            WorkerError::VmStartFailed(
+                "podman or docker not found in PATH".to_owned(),
+            )
+        })?;
+
+        let out = tokio::process::Command::new(&bin)
+            .args(["info", "--format", "{{.Host.Security.Rootless}}"])
```
ewindisch, line 162: Rootless guard is podman-specific and silently passes for rootful docker. `{{.Host.Security.Rootless}}` is a podman `info` field; `docker info` does not emit it, so the template renders empty and `rootless` becomes `false`. Worse, the result is only logged (`info!(... rootless ...)`) and never enforced — so rootful docker **and** even `rootless == false` podman are accepted and the backend proceeds. Either detect docker vs podman explicitly, or fail-closed when rootless cannot be proven. See Decision Q2.

crates/hyprstream-workers/src/runtime/podman.rs
```diff
@@ -0,0 +1,431 @@
...
+            rootless,
+            "PodmanBackend initialized"
+        );
+        Ok(())
+    }
+
+    async fn start(
+        &self,
+        sandbox: &mut PodSandbox,
+        _config: &PodSandboxConfig,
```
ewindisch, line 192: Core architecture concern: the entire CRI `PodSandboxConfig` is dropped here — the `_config` underscore means none of it is used. `worker.capnp` is a faithful CRI v1 model, but this backend ignores it wholesale: DNS config, port mappings, security context (uid/gid, capabilities, seccomp, SELinux/AppArmor, no-new-privileges), namespace options, sysctls, and resource limits are all silently discarded. The container always runs as the image default with `slirp4netns` + `sleep infinity`. See Decision Q1. (Note: this backend is also not wired into `BackendType` in this PR, so it is currently dead code.)

crates/hyprstream-workers/src/runtime/podman.rs
```diff
@@ -0,0 +1,431 @@
...
+        let name = container_name(&sandbox.id);
+
+        info!(sandbox_id = %sandbox.id, container = %name, image = %image, "Starting podman sandbox");
+
+        let mut args = vec![
+            "run".to_owned(),
+            "--detach".to_owned(),
+            "--rm".to_owned(),
+            format!("--name={name}"),
+            format!("--network={}", self.config.network_mode),
```
ewindisch, line 209: `network_mode` is interpolated straight into the `--network=` CLI arg with no validation. It is config-controlled today (not directly attacker-reachable), but it bypasses the CRI network model entirely (see the `_config` comment) and becomes an injection/escape surface the moment it is ever populated from annotations or CRI. Validate against an allowed set, or map it from the CRI `NamespaceOption`. See Decision Q4.

ewindisch:
## 🔶 Decision required (human sign-off — architecture + authz)

This applies **jointly to #484** (rootless podman/docker OCI backend, #346) **and #486** (availability-based backend auto-selection, #348). The two are stacked (#486 bases on #484) and entangled with a larger worker-runtime architecture decision, so **no code has been changed/pushed** — these are open questions for human sign-off before any rework. Inline diff comments are on both PRs.

**Q1 — Architecture.** The podman backend drops the entire CRI `PodSandboxConfig` (the param is `_config`) and is **not wired into `BackendType`** in #484, so it is also currently dead code. Meanwhile `worker.capnp` is already a faithful CRI v1 model that *every* backend (podman **and** kata) currently ignores. Do we:
- **(a)** rework into a single `OciBackend` that maps CRI → `oci_spec::runtime::Spec` → a configurable embedded runtime (youki/crun), and/or add a `CriBackend` (CRI client driving containerd) for standard hosts; **or**
- **(b)** land #484/#486 as tactical podman-CLI shims and supersede later (with a tracking issue)?

**Q2 — Authz / fail-closed.** When "rootless" is promised but the runtime is rootful docker — or rootless cannot be proven — do we **fail-closed (reject)** or **warn-and-continue**? Today the result is only logged and never enforced (even `rootless == false` podman passes). Same question for #486's `Auto` silently degrading to `nspawn` when no VM-isolation runtime is present (now the out-of-box default).

**Q3 — Authz / image provenance.** Before the `hyprstream.io/worker-image` annotation is usable, do we require a **registry/image allowlist + digest pinning**, and/or **gate the annotation behind operator config**? Today any authorized RPC caller can pull and run any OCI image from any registry.

**Q4 — Operator config vs PATH resolution.** Should runtime selection be **operator-CONFIGURED** (explicit runtime path/name + options) rather than PATH-resolved via `find_runtime()` (#484, `podman.rs`) / `which::which` in `BackendType::resolve()` (#486, `config.rs`)?

**References:** CRI-client-VFS injection spike doc `docs/plans/2026-06-26-cri-client-vfs-injection-spike.md` (branch `ewindisch/cri-client-spike`) and #495 (hyprstream-vfsd native virtio-fs host mount of the VFS namespace).

— cross-linked: #484 ↔ #486

ewindisch:
Closing — superseded by the Sandbox backend taxonomy epic **#508** + spine **#507**.

The goal (a lightweight non-Kata container path, #346) is retained, but the **implementation here is the anti-pattern the decided design replaces**: this backend shells `podman run` with hand-built flags, discards the faithful CRI-capnp `PodSandboxConfig`, and is dead code (never wired into `BackendType`). Per the analysis on this PR + the convergence spike, #346's open **D2** is decided: **CRI→`oci-spec::Spec`→a configured runtime (youki/crun)**, not CLI-shelling.

→ Reworked as **OciBackend** under #508 Phase 1 (consumes the #507 canonical `BackendType` + fail-closed selection). The docker-rootful fail-closed + image-allowlist findings carry into that rework. #346 stays open, tracked under #508.

Issue #355, opened 2026-06-19
Epic: Streaming Plane Restoration — iroh-direct + moq relay (rendezvous) for browser/fan-out


ewindisch:
**Epic: Streaming Plane Restoration — iroh-direct + moq relay (rendezvous) for browser/fan-out**

**Status:** In progress — see [plan](https://github.com/hyprstream/hyprstream/blob/main/docs/plans/2026-06-22-355-streaming-plane-restoration.md)

**Branch:** `feature/streaming-plane-restoration` (off main) — all PRs target this branch

### Executive Summary

Restore the **streaming-plane rendezvous/relay** that the ZMQ→moq migration (#131/#138) removed. The old `StreamService` (PUSH/PULL→XPUB blind-forwarding rendezvous) is gone; streaming inference now works ONLY by direct producer-dial.

**Problem:** NAT'd producers, cross-instance broadcasts, mutually-unreachable peers, and anonymization use-cases are all broken. Two are LIVE bugs: #142 (cross-instance model-load notifications silently fail) and TUI cross-process UDS timeout.

**Solution:** 
- **iroh+pkarr DIRECT** for native peers (NAT + cross-instance + discovery)
- **moq relay** for browsers (no wasm iroh), fan-out, and anonymization

### Sub-Issues

| # | Title | State | Notes |
|---|-------|-------|-------|
| #356 | S1: unify StreamInfo reach (fix #142 + TUI) | **CLOSED** | Merged |
| #357 | S2: iroh+pkarr direct reach (node_id) | **CLOSED** | Merged |
| #358 | S3: moq relay (rendezvous) | **OPEN** | Pending |
| #359 | S4: delete teardown debt | **CLOSED** | Merged |
| #360 | S5: blind-relay AEAD + COSE provenance | **OPEN** | DONE as #321/#354 on 310 branch |
| #361 | S6: pkarr discovery + oblivious-relay | **OPEN** | Deferred |
| #384 | S7: de-singletonize relay reach (per-server + per-stream override) | **OPEN** | Blocks anonymization claim — process-global GLOBAL_RELAY_REACH is wrong granularity |
| #385 | S8: threat-model close-out (metadata leak + dhPublic invariant + relay-substitution test) | **OPEN** | Gates final-PR security claims |

### What's Left

- **S3** (moq relay): Complete relay rendezvous — producer publishes to relay, consumer subscribes via relay
- **S5** (AEAD): Merge #321/#354 from 310 branch or re-implement
[...]

PR #485, opened 2026-06-26
feat(#355): moq Worker bridge via SAB DMA — unblock browser streaming client

ewindisch:
## Summary

Resolves the `moq-net Session: Sync` incompatibility with browser `web_sys` types by running the moq/iroh session inside a dedicated WASM Web Worker and communicating with the main thread via the existing `DmaTransport` SharedArrayBuffer ring buffer (`crates/hyprstream-9p/src/dma.rs`).

**Root cause:** `moq-net 0.1.8` requires `Session: Send + Sync`. `web_sys::WebTransport` is `!Sync` because `JsValue` wraps JS objects. This applies even in WASM Workers because wasm-bindgen marks `JsValue` as `!Send+!Sync` always. Cannot wait for upstream; this bridge allows progress now.

**Approach (Option C):**
```
Main thread (wasm32)                    Web Worker (wasm32)
─────────────────────────────────       ──────────────────────────────────────
MoqWorkerHandle                         moq_worker_main()
  .subscribe(track) ──────────────────► recv subscribe cmd over DMA
  .recv_frame() ◄─────────────────────  send frames over DMA
                                        │
                                        ▼
                                   WebTransport → moq relay/server
```

## Changes

- `crates/hyprstream-rpc-std/src/moq_worker.rs` (new): `MoqWorkerHandle` (#[wasm_bindgen]) + `moq_worker_main` worker entrypoint + DMA frame encode/decode helpers
- `crates/hyprstream-rpc-std/Cargo.toml`: add `Worker`, `WorkerOptions`, `WorkerType` web-sys features
- `crates/hyprstream-rpc-std/src/lib.rs`: `#[cfg(wasm32)] pub mod moq_worker`

Frontend shim (`www-cyberdione-ai/public/moq-worker.js`) committed separately in the www repo.

## Stub status

The DMA channel + Worker spawn infrastructure is fully functional. The `moq_worker_main` loop stubs the actual moq SUBSCRIBE/OBJECT framing with an echo acknowledgment — full framing is gated on moq-net relaxing its `Session: Sync` bound for wasm32 (tracked in #484).

## COOP/COEP

`SharedArrayBuffer` requires `Cross-Origin-Opener-Policy: same-origin` + `Cross-Origin-Embedder-Policy: credentialless`. Already handled by the `crossOriginIsolation()` Vite plugin — no frontend header changes needed.

## Test plan

- [ ] `cargo check -p hyprstream-rpc-std` passes (native)
- [ ] `cargo check --target wasm32-unknown-unknown -p hyprstream-rpc` passes (CI wasm32 target)
- [ ] CI build + clippy green
- [ ] Manual: spawn `MoqWorkerHandle`, subscribe to a track, verify `__ready__` + `__subscribed_stub__` frames echoed back over DMA

https://github.com/hyprstream/hyprstream/issues/355

### DIFF
--- a/crates/hyprstream-9p/src/dma.rs
+++ b/crates/hyprstream-9p/src/dma.rs
@@ -188,7 +188,11 @@ impl DmaTransport {
         // Advance head
         self.atomic_store(self.write_head_idx, (head + total) as i32);
         self.atomic_add(self.write_msg_count_idx, 1);
-        self.atomic_notify(self.read_tail_idx); // wake reader
+        // Notify the index we just advanced: a peer reader blocked in
+        // Atomics.wait watches the head it consumes from (this same physical
+        // SAB slot), not the tail. Waking read_tail_idx would signal the wrong
+        // waiter and the reader would never wake under a future waitAsync path.
+        self.atomic_notify(self.write_head_idx); // wake reader waiting on this head
     }
 
     /// Read a length-prefixed message from the ring buffer.
@@ -233,7 +237,10 @@ impl DmaTransport {
             (tail + HEADER_SIZE + msg_len) as i32,
         );
         self.atomic_add(self.read_msg_count_idx, -1);
-        self.atomic_notify(self.write_head_idx); // wake writer
+        // Notify the index we just advanced: a peer writer blocked waiting for
+        // space watches the tail it frees (this same physical SAB slot), not
+        // the head. (Mirror of the write_message notify.)
+        self.atomic_notify(self.read_tail_idx); // wake writer waiting on this tail
 
         Some(data)
     }
--- a/crates/hyprstream-rpc-std/Cargo.toml
+++ b/crates/hyprstream-rpc-std/Cargo.toml
@@ -29,7 +29,12 @@ wasm-bindgen-futures = "0.4"
 serde-wasm-bindgen = "0.6"
 js-sys = "0.3"
 console_error_panic_hook = "0.1"
-web-sys = { version = "0.3", features = ["console"] }
+web-sys = { version = "0.3", features = [
+  "console",
+  "Worker",
+  "WorkerOptions",
+  "WorkerType",
+] }
 # #409 Path B: WanixMount (Wanix-native VFS via 9P2000.L over SharedArrayBuffer
 # DMA). Only needed on the browser target; the DMA + WanixMount modules in
 # hyprstream-9p are themselves `#[cfg(target_arch = "wasm32")]`.
--- a/crates/hyprstream-rpc-std/src/lib.rs
+++ b/crates/hyprstream-rpc-std/src/lib.rs
@@ -169,3 +169,5 @@ pub mod stream_mount;
 pub mod wasm_exports;
 #[cfg(target_arch = "wasm32")]
 pub mod wasm_rpc_client;
+#[cfg(target_arch = "wasm32")]
+pub mod moq_worker;
--- a/crates/hyprstream-rpc-std/src/moq_worker.rs
+++ b/crates/hyprstream-rpc-std/src/moq_worker.rs
@@ -1,0 +1,297 @@
+//! moq streaming Worker bridge via SharedArrayBuffer DMA ring buffer.
+//!
+//! Resolves the `moq-net Session: Sync` incompatibility with browser `web-sys`
+//! types without waiting for an upstream moq-net fix. The iroh/moq session runs
+//! inside a dedicated Web Worker (a real OS thread from wasm's perspective, where
+//! `Send + Sync` requirements are logically satisfied for single-threaded-per-worker
+//! JS objects). The main browser thread communicates with the worker via the
+//! existing `DmaTransport` SharedArrayBuffer ring buffer (`hyprstream-9p/src/dma.rs`).
+//!
+//! # Architecture
+//!
+//! ```text
+//! Main thread (wasm32)                    Web Worker (wasm32)
+//! ─────────────────────────────────       ──────────────────────────────────────
+//! MoqWorkerHandle                         moq_worker_main()
+//!   .subscribe(track) ──────────────────► recv subscribe cmd over DMA
+//!   .recv_frame() ◄─────────────────────  send frames over DMA
+//!                                         │
+//!                                         ▼
+//!                                    WebTransport → moq relay/server
+//!                                    (opened FROM worker; WT here is the
+//!                                     sole JS thread so !Sync is not an issue
+//!                                     in practice — and moq-net is available
+//!                                     once its Sync bound is relaxed)
+//! ```
+//!
+//! # DMA frame envelope
+//!
+//! Both directions use the same layout over `DmaTransport`'s length-prefixed messages:
+//! ```text
+//! [2 bytes LE: track_name_len][track_name bytes][4 bytes LE: payload_len][payload bytes]
+//! ```
+//! A special track name `__cmd__` carries JSON control messages from main→worker.
+//!
+//! # COOP/COEP
+//!
+//! `SharedArrayBuffer` requires `Cross-Origin-Opener-Policy: same-origin` and
+//! `Cross-Origin-Embedder-Policy: credentialless` (or `require-corp`). These
+//! headers are already set by the `crossOriginIsolation()` Vite plugin in the
+//! www-cyberdione-ai frontend.
+//!
+//! # moq-net Sync status
+//!
+//! `moq-net 0.1.8` requires `Session: Send + Sync`. `web_sys::WebTransport` is
+//! `!Sync`. The worker infrastructure here is complete and ready; the actual
+//! moq subscribe/publish loop (`worker_moq_loop`) is stubbed pending either:
+//! a) upstream moq-net relaxing the Sync bound, or
+//! b) a local `[patch.crates-io]` fork with the one-line fix.
+//! Track: <https://github.com/hyprstream/hyprstream/issues/484>
+
+#![cfg(target_arch = "wasm32")]
+
+use wasm_bindgen::prelude::*;
+use web_sys::{Worker, WorkerOptions, WorkerType};
+
+use hyprstream_9p::dma::DmaTransport;
+
+// ─── Frame envelope helpers ──────────────────────────────────────────────────
+
+/// Encode a single DMA frame: `[2: track_len][track][4: payload_len][payload]`.
+pub fn encode_frame(track: &str, payload: &[u8]) -> Vec<u8> {
+    let t = track.as_bytes();
+    let mut out = Vec::with_capacity(2 + t.len() + 4 + payload.len());
+    out.extend_from_slice(&(t.len() as u16).to_le_bytes());
+    out.extend_from_slice(t);
+    out.extend_from_slice(&(payload.len() as u32).to_le_bytes());
+    out.extend_from_slice(payload);
+    out
+}
+
+/// Decode a DMA frame. Returns `(track_name, payload)` or `None` if malformed.
+pub fn decode_frame(buf: &[u8]) -> Option<(&str, &[u8])> {
+    if buf.len() < 6 {
+        return None;
+    }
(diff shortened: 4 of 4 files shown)


ewindisch:
## Agentic Code Review

**Overall:** The PR establishes the correct architectural shape (Web Worker + SAB DMA bridge for moq-net Sync incompatibility) and the stub-first approach is reasonable for a gated upstream fix. However it has a blocking build error (missing hyprstream-9p Cargo dependency), two runtime panics (UTF-8 byte-slice and empty busy-spin), and a wrong atomic-notify index that will silently break the planned Atomics.waitAsync upgrade path. None of these issues affect the stub behavior today because recv() uses a yield-loop rather than Atomics.wait, but they must be fixed before the moq-net Sync gate opens and real data flows through the channel.

### Findings

**1. [BUG]** `crates/hyprstream-rpc-std/Cargo.toml:32`
Missing hyprstream-9p dependency: moq_worker.rs imports hyprstream_9p::dma::DmaTransport but hyprstream-9p is absent from Cargo.toml, causing a build failure.
> cargo check -p hyprstream-rpc-std --target wasm32-unknown-unknown fails with 'use of undeclared crate or module `hyprstream_9p`'; the crate needs to be added under [target.'cfg(target_arch = "wasm32")'.dependencies].\n\n**2. [SECURITY]** `crates/hyprstream-rpc-std/src/moq_worker.rs:120`\nNo runtime check for crossOriginIsolated before allocating SharedArrayBuffer; if COOP/COEP headers are missing in production the SAB constructor throws a TypeError, crashing spawn() with an opaque JS error rather than a diagnostic message.\n> Production server not configured with COOP/COEP headers: SharedArrayBuffer::new() throws, Worker is never created, and the opaque JsError from the SAB allocation gives no actionable guidance; more critically, if a browser somehow bypasses the throw (e.g. in a non-isolated context), the ring buffer is backed by a regular ArrayBuffer and loses cross-thread atomicity, creating a data-race Spectre window.\n\n**3. [BUG]** `crates/hyprstream-9p/src/dma.rs:160`\nBusy-spin loop in write_message has an empty body with no yield, making it a hot infinite loop that will starve the WASM event loop if the ring buffer is full.\n> Main thread calls subscribe() which calls dma.send() which calls write_message(); if the worker hasn't yet drained the ring (e.g., it hasn't started or is slow), write_available() stays below threshold and the loop spins forever, hanging the entire browser tab.\n\n**4. [CORRECTNESS]** `crates/hyprstream-rpc-std/src/moq_worker.rs:57`\nencode_frame casts track name length to u16 without bounds check; a track name longer than 65535 bytes silently truncates, corrupting the frame and causing decode_frame to misalign all subsequent bytes.\n> Caller (or attacker-controlled track name via subscribe()) passes a track name of 65537+ bytes; t.len() as u16 truncates to 1, decode_frame reads 1 byte as the track name and treats the next 4 bytes of track name data as the payload length, producing arbitrary garbage or a length that reads far past the buffer end.\n\n**5. [CORRECTNESS]** `crates/hyprstream-9p/src/dma.rs:191`\nwrite_message notifies self.read_tail_idx but readers wait on the read head (read_head_idx); the wrong atomic index is woken, so when Atomics.waitAsync is eventually implemented readers will not be woken after a write.\n> Once the recv() polling loop is replaced with Atomics.waitAsync (as the TODO indicates), the reader waits on read_head_idx but the writer notifies read_tail_idx — the reader blocks forever even after data is available. Symmetric bug exists in read_message notifying write_head_idx instead of write_tail_idx.\n\n**6. [BUG]** `crates/hyprstream-rpc-std/src/moq_worker.rs:192`\nreach_json is sliced at a byte index (&reach_json[..120]) rather than a char boundary, causing a panic on any non-ASCII URL or cert hash in the reach config.\n> reach_json contains a URL with a non-ASCII character (e.g. internationalized hostname) or a base64 cert hash whose 120th byte falls in the middle of a multi-byte UTF-8 sequence; the &str slice panics at runtime inside the worker with 'byte index N is not a char boundary', killing the worker before the DMA channel is established.\n\n**7. [NIT]** `crates/hyprstream-rpc-std/src/moq_worker.rs:110`\nThe SAB size comment says '2 × 126 KiB data' but the actual allocation is 2 × 128 KiB, creating a misleading discrepancy for future readers auditing buffer sizing.\n> No runtime failure, but a developer auditing ring capacity against the comment underestimates available space by 2 KiB per channel, potentially leading to incorrect high-water-mark calculations in follow-on work.

ewindisch:
## Agentic Code Review

**Overall:** The PR correctly introduces the architectural scaffolding for a Web Worker moq bridge and is a reasonable unblocking tactic for the moq-net !Sync issue. The biggest risk before enabling real moq framing (post #484) is a cluster of pre-existing dma.rs bugs that this PR now exposes to production: an empty busy-spin that freezes the event loop on backpressure, an i32-overflow in the head pointer after ~2 GB of streaming, and swapped Atomics.notify targets that will deadlock any future waitAsync optimization. The cross-repo worker shim calling convention is also undocumented and fragile. The new moq_worker.rs code itself is clean and stub-safe, but it inherits all of these DMA transport hazards.

### Findings

**1. [CORRECTNESS]** `crates/hyprstream-rpc-std/src/moq_worker.rs:76`
encode_frame silently truncates track names longer than 65535 bytes via `t.len() as u16`, causing encode/decode mismatch: encoded bytes contain the full track name but decode_frame reads only `track_len` (truncated) bytes, so the payload_len header is read from the wrong offset, producing garbage.
> Caller passes a track name of 65536+ bytes (e.g. a URL-encoded path); encode_frame writes 65536 bytes of track data but sets track_len=0 in the header; decode_frame reads 0 bytes of track and misidentifies the next 4 bytes as payload_len, returning wrong payload or None. In practice track names are short but the encoder is public API with no assertion.

**2. [BUG]** `crates/hyprstream-rpc-std/src/moq_worker.rs:257`
Byte-index slice `&reach_json[..reach_json.len().min(120)]` panics if byte offset 120 falls in the middle of a multi-byte UTF-8 character.
> A reach_json string whose 120th byte is a continuation byte of a multi-byte character (e.g. a URL with encoded non-ASCII or a cert hash in a non-ASCII encoding) causes `moq_worker_main` to panic at the console::log_1 call, which — despite console_error_panic_hook — terminates the worker silently from the main thread's perspective.\n\n**3. [BUG]** `crates/hyprstream-9p/src/dma.rs:159`\nThe write_message spin-wait loop has an empty body with only a comment, so it is a true busy-spin with no yield, which will freeze the WASM event loop (or Worker thread) indefinitely if the ring buffer is full.\n> Main thread (or worker) calls dma.send() while the ring buffer is full (e.g. slow consumer, dead peer). The while loop in write_message spins with zero yielding, freezing the tab or worker until the browser kills the script or the consumer catches up. A dead worker holding MoqWorkerHandle causes permanent main-thread hang on any subsequent subscribe().\n\n**4. [CORRECTNESS]** `crates/hyprstream-9p/src/dma.rs:191`\natomic_notify targets are swapped: write_message notifies read_tail_idx (not the read_head_idx a waiting reader would watch), and read_message notifies write_head_idx (not write_tail_idx that a waiting writer needs). This is a correctness time-bomb for any future Atomics.wait-based blocking recv.\n> Currently recv() polls with Promise.resolve yields so no deadlock occurs. If a future optimization replaces polling with Atomics.waitAsync on the correct index, the wake signal is delivered to the wrong index and the reader/writer never wakes, causing a permanent stall.\n\n**5. [BUG]** `crates/hyprstream-9p/src/dma.rs:142`\nRing buffer head pointer is stored as i32 via atomic_store but interpreted as u32 via atomic_load-as-u32; after ~2 GB of cumulative writes the head overflows i32::MAX, the stored value wraps negative, and write_available() / read_available() compute wrong results.\n> A long-running streaming session that cumulatively writes more than 2 GB through the 128 KB ring (achieved after ~16K max-size writes, or many small writes). The head counter overflows i32::MAX, is stored as a large negative i32, then cast to a very large u32, making write_available() return a huge positive number; the writer skips all backpressure and overwrites data the reader has not consumed.\n\n**6. [NIT]** `crates/hyprstream-rpc-std/src/moq_worker.rs:155`\nMoqWorkerHandle::spawn comment says '4 KiB control + 2 × 126 KiB data' but allocates 4096 + 2 × 128 × 1024 = 4 KiB + 256 KiB, making both the comment and the module-level doc example wrong by 2 KiB per channel.\n> No runtime impact; misleads a developer who uses the documented 126 KiB figure to size messages or diagnose capacity issues.\n\n**7. [CORRECTNESS]** `crates/hyprstream-rpc-std/src/moq_worker.rs:223`\nmoq_worker_main is exported as a wasm_bindgen function expecting (sab: JsValue, reach_json: String) as individual arguments, but spawn() sends a single postMessage with an Array [sab, reach_json_string]; the JS worker shim (committed separately in www repo) must unpack the array and call moq_worker_main(data[0], data[1]) — an undocumented cross-repo calling convention.\n> If the www-repo worker shim calls `moq_worker_main(event.data)` (passing the array directly) instead of `moq_worker_main(event.data[0], event.data[1])`, the dyn_into::<SharedArrayBuffer>() check fails with a logged error and the worker exits silently; the main thread's recv_frame() spins forever polling an empty DMA buffer.\n\n**8. [SECURITY]** `crates/hyprstream-rpc-std/src/moq_worker.rs:267`\nThe worker logs the relay URL at every subscribe command (console::log_1) with no redaction, which leaks the URL (and any embedded credentials or tokens in query params) to browser DevTools and any installed console interceptors.\n> If reach_json embeds a bearer token or session key in the URL query string (e.g. WebTransport negotiation tokens), it appears in plaintext in the browser console, accessible to any same-origin JS including injected analytics scripts.\n\n**9. [SECURITY]** `crates/hyprstream-rpc-std/src/moq_worker.rs:116`\nreach_json's cert_hash field is documented in the module-level example but never read by the worker; the WebTransport connection (when implemented) will use the browser's default TLS certificate validation with no cert pinning.\n> When full moq framing is implemented (post-#484), a relay presenting a fraudulently-issued TLS cert passes validation unless cert pinning is added; the documented API surface implies pinning is already in effect, giving a false sense of security.

ewindisch:
## Triage update — conflicts resolved + drift-induced build break fixed

This PR's CI failure (`build` job) was **not from this PR's own changes** — it was the known pre-existing test flake `moq_event::tests::moq_event_pub_sub_roundtrip` (tracked separately as #167) plus a flaky `lazy_iroh` networking test. Neither file is touched here.

The real blocker was the merge-conflict state (`CONFLICTING`/`DIRTY`) against current `main`. Pushed a merge resolving it:

- **`Cargo.toml` / `lib.rs`**: both conflicts were purely additive (this PR's `Worker`/`WorkerOptions`/`WorkerType` web-sys features + `moq_worker` module, vs. main's unrelated WanixMount feature + `iroh_exports`/`Did`-domain-type test additions) — unioned both sides, no semantic change.
- **`moq_worker.rs`**: resolving the merge surfaced 11 cascading wasm32 compile errors (`DmaTransport` had no `.send()`/`.recv()` in scope) — caused by `main`'s newer `hyprstream-9p::client::P9Transport` trait requiring an explicit import that didn't exist when this PR branched. One-line fix (`use hyprstream_9p::client::P9Transport;`).

**Verified clean**: `cargo check -p hyprstream-rpc-std` (native) and `cargo check --target wasm32-unknown-unknown -p hyprstream-rpc-std` (`RUSTFLAGS=--cfg=web_sys_unstable_apis`, matching this repo's wasm CI flag).

**Note for whoever picks up CI scope here**: the repo's `WASM (browser client)` CI check only runs `cargo check --target wasm32-unknown-unknown -p hyprstream-rpc` — it does **not** build `hyprstream-rpc-std` (where `moq_worker.rs` lives). So this PR's actual feature code was never wasm32-validated by CI even when it showed green. Worth widening that job's `-p` list in a follow-up; out of scope for this PR.

Recommend for merge once CI re-runs green on the updated branch (human merge decision, not mine to execute).

PR #486, opened 2026-06-26
feat(workers): availability-based backend auto-selection (#348)

### LINKED ISSUE #348: T2-C: availability-based backend auto-selection + fallback (kata to podman to nspawn) + wizard/CLI UX

ewindisch:
## Summary

- Adds `BackendType::Auto` (new default) and `BackendType::Podman`
- Implements `BackendType::resolve()`: probes Kata → Podman → Nspawn in priority order at runtime
- Wires resolve() into both the factory path (`factories.rs`) and the CLI worker path (`main.rs`)
- Adds `podman` feature to `hyprstream` crate that propagates to `hyprstream-workers/podman`

## Before / After

| | Before | After |
|---|---|---|
| Default backend | `kata` (fails silently without cloud-hypervisor) | `auto` (probes & falls back) |
| Dev laptop (podman installed) | Error at startup | Selects `podman` automatically |
| Dev laptop (nspawn only) | Error at startup | Selects `nspawn` automatically |
| Production (cloud-hypervisor) | Kata (if built with kata-vm) | Kata (same) |

## Auto-detection order

```
Kata  ← requires kata-vm feature + cloud-hypervisor in PATH
  ↓ (not available)
Podman ← requires podman feature + podman/docker in PATH
  ↓ (not available)
Nspawn ← always available (systemd-nspawn)
```

Operators can pin a specific backend via `worker.backend: kata/podman/nspawn` in config.

## Test plan

- [ ] `cargo clippy -p hyprstream-workers --features podman --all-targets` passes (verified locally)
- [ ] `cargo clippy -p hyprstream --features podman --all-targets` passes (verified locally)  
- [ ] `test_backend_type_default`: asserts `Auto` (updated)
- [ ] `test_backend_type_auto_resolves_to_concrete`: new test
- [ ] CI green

Closes #348. Part of epic #341. Stacked on #484.

### DIFF
--- a/crates/hyprstream-workers/src/config.rs
+++ b/crates/hyprstream-workers/src/config.rs
@@ -13,17 +13,49 @@ use std::path::PathBuf;
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Default, Serialize, Deserialize)]
 #[serde(rename_all = "kebab-case")]
 pub enum BackendType {
-    /// Kata Containers (default — full VM isolation)
+    /// Auto-detect: probe availability in order Kata → Podman → Nspawn and
+    /// pick the best backend available on the current host.  This is the
+    /// operator-friendly default — the same binary works on a developer laptop
+    /// (podman/nspawn) and a production node (Kata/cloud-hypervisor).
     #[default]
+    Auto,
+    /// Kata Containers — full hardware VM isolation (requires cloud-hypervisor).
     Kata,
-    /// systemd-nspawn (lightweight container, rootless, host filesystem)
+    /// Rootless podman (or docker) OCI containers — no hypervisor.
+    Podman,
+    /// systemd-nspawn — lightweight container, rootless, host filesystem.
     Nspawn,
+}
+
+impl BackendType {
+    /// Probe availability of each concrete backend in priority order and return
+    /// the first one whose binary / runtime is present on the host.
+    ///
+    /// Fallback chain: Kata → Podman → Nspawn.
+    /// `Auto` itself resolves to the result of this function; all other
+    /// variants resolve to themselves (no probing needed).
+    pub fn resolve(self) -> Self {
+        if self != Self::Auto {
+            return self;
+        }
+        #[cfg(feature = "kata-vm")]
+        if which::which("cloud-hypervisor").is_ok() {
+            return Self::Kata;
+        }
+        #[cfg(feature = "podman")]
+        if which::which("podman").is_ok() || which::which("docker").is_ok() {
+            return Self::Podman;
+        }
+        Self::Nspawn
+    }
 }
 
 impl std::fmt::Display for BackendType {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
+            Self::Auto => write!(f, "auto"),
             Self::Kata => write!(f, "kata"),
+            Self::Podman => write!(f, "podman"),
             Self::Nspawn => write!(f, "nspawn"),
         }
     }
@@ -299,7 +331,22 @@ mod tests {
 
     #[test]
     fn test_backend_type_default() {
-        assert_eq!(BackendType::default(), BackendType::Kata);
+        // Default is Auto (operator-friendly; resolves at runtime).
+        assert_eq!(BackendType::default(), BackendType::Auto);
+    }
+
+    #[test]
+    fn test_backend_type_resolve_identity() {
+        // Concrete variants are idempotent under resolve().
+        assert_eq!(BackendType::Nspawn.resolve(), BackendType::Nspawn);
+        assert_eq!(BackendType::Kata.resolve(), BackendType::Kata);
+        assert_eq!(BackendType::Podman.resolve(), BackendType::Podman);
+    }
+
+    #[test]
+    fn test_backend_type_auto_resolves_to_concrete() {
+        // Auto must not resolve to itself.
+        assert_ne!(BackendType::Auto.resolve(), BackendType::Auto);
     }
 
     #[test]
@@ -311,12 +358,22 @@ mod tests {
         let yaml = "backend: kata\n";
         let config: WorkerConfig = serde_yaml::from_str(yaml)?;
         assert_eq!(config.backend, BackendType::Kata);
+
+        let yaml = "backend: podman\n";
+        let config: WorkerConfig = serde_yaml::from_str(yaml)?;
+        assert_eq!(config.backend, BackendType::Podman);
+
+        let yaml = "backend: auto\n";
+        let config: WorkerConfig = serde_yaml::from_str(yaml)?;
+        assert_eq!(config.backend, BackendType::Auto);
         Ok(())
     }
 
     #[test]
     fn test_backend_type_display() {
+        assert_eq!(BackendType::Auto.to_string(), "auto");
         assert_eq!(BackendType::Kata.to_string(), "kata");
+        assert_eq!(BackendType::Podman.to_string(), "podman");
         assert_eq!(BackendType::Nspawn.to_string(), "nspawn");
     }
 }
--- a/crates/hyprstream-workers/src/lib.rs
+++ b/crates/hyprstream-workers/src/lib.rs
@@ -70,6 +70,8 @@ pub use error::WorkerError;
 pub use runtime::{WorkerService, SandboxBackend, SandboxHandle, NspawnBackend, NspawnConfig};
 #[cfg(feature = "kata-vm")]
 pub use runtime::KataBackend;
+#[cfg(feature = "podman")]
+pub use runtime::{PodmanBackend, PodmanConfig};
 #[cfg(feature = "kata-vm")]
 pub use image::RafsStore;
 pub use workflow::WorkflowService;
--- a/crates/hyprstream/Cargo.toml
+++ b/crates/hyprstream/Cargo.toml
@@ -201,6 +201,7 @@ download-libtorch = ["tch/download-libtorch"]
 overlayfs = ["git2db/overlayfs"]  # Enable overlayfs-backed worktrees (delegated to git2db)
 systemd = ["hyprstream-rpc/systemd", "hyprstream-service/systemd", "hyprstream-workers/systemd", "dep:listenfd"]  # Systemd integration (socket activation, service units)
 kata-vm = ["hyprstream-workers/kata-vm"]  # Kata/Cloud-Hypervisor VM worker backend + RAFS/nydus image stack (in default; drop for an nspawn-only build)
+podman = ["hyprstream-workers/podman"]   # Rootless podman/docker OCI worker backend (auto-selected when kata-vm not built or cloud-hypervisor absent)
 experimental = ["hyprstream-compositor/experimental"]  # EXPERIMENTAL: Enable service/worker panels, git write operations
 pq-hybrid = []  # NO-OP alias: PQ primitives always compiled; Classical/Hybrid selected at runtime. See M3 #152.
 
--- a/crates/hyprstream/src/bin/main.rs
+++ b/crates/hyprstream/src/bin/main.rs
@@ -1076,24 +1076,39 @@ fn handle_quick_command(
 
                     let _worker_handle = if !worker_already_running {
                         use hyprstream_workers::runtime::SandboxBackend;
+                        use hyprstream_workers::config::BackendType;
 
-                        // VM path (kata-vm): Kata backend + RAFS image store.
+                        // Resolve the backend: read config if available, else auto-detect.
+                        let backend_type = HyprConfig::load()
+                            .ok()
+                            .and_then(|c| c.worker)
+                            .map(|w| w.backend)
+                            .unwrap_or_default()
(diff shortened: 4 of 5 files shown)


ewindisch:
## Agentic Code Review

**Overall:** The PR cleanly implements the `Auto` detection pattern and the `Podman` backend plumbing, but has two security findings worth blocking on: (1) `docker` in PATH is accepted as a substitute for rootless `podman` despite docker containers running as root via the daemon socket, and (2) the `hyprstream.io/worker-image` annotation allows any authorized RPC caller to pull and run an arbitrary OCI image without an allowlist. The silent Nspawn catch-all in `main.rs` (vs. the correct hard-error in `factories.rs`) is also a risk because it masks misconfiguration that would have been caught in the daemon path.

### Findings

**1. [BUG]** `crates/hyprstream/src/bin/main.rs:1108`
Catch-all `_ =>` arm in the CLI worker path silently uses Nspawn when `BackendType::Kata` is resolved but `kata-vm` feature is absent, or when `BackendType::Podman` is resolved but `podman` feature is absent — unlike `factories.rs` which returns a clear error for both cases.
> Operator configures `worker.backend: kata` in config, builds without `kata-vm` feature; the CLI path resolves to `BackendType::Kata`, hits `_ =>`, silently starts an Nspawn backend instead of failing. The operator believes Kata isolation is active but workloads run in the much weaker nspawn sandbox.

**2. [SECURITY]** `crates/hyprstream-workers/src/config.rs:46`
`resolve()` selects `Podman` when `docker` is in PATH, but docker runs containers as root by default (daemon-owned socket) unlike rootless podman — the PR docs promise rootless semantics but docker is not rootless.
> System has `docker` but not `podman` in PATH; `BackendType::Auto` resolves to `Podman`; `PodmanBackend` spawns containers via the docker socket (root daemon); workload containers run as root with full access to the docker daemon socket, defeating the 'rootless' isolation guarantee the PR advertises.\n\n**3. [SECURITY]** `crates/hyprstream-workers/src/runtime/podman.rs:136`\nThe OCI image to run can be overridden by an untrusted `hyprstream.io/worker-image` annotation supplied by any authorized RPC caller in `PodSandboxConfig`, with no allowlist or validation.\n> An authorized but less-trusted client sends a `RunPodSandbox` RPC with annotation `hyprstream.io/worker-image: attacker/malicious:latest`; `PodmanBackend.start()` calls `podman run attacker/malicious:latest` unconditionally, pulling and executing arbitrary OCI images as the worker service user.\n\n**4. [SECURITY]** `crates/hyprstream-workers/src/runtime/podman.rs:210`\n`network_mode` from `PodmanConfig` is interpolated directly as `--network=<value>` with no validation; a caller who controls config can pass arbitrary option strings.\n> An operator configuration (or future config-injection path) sets `network_mode` to something like `host` or to a value containing embedded option characters; `podman run --network=host` grants the container full host network access, defeating isolation — particularly relevant because `PodmanConfig::default()` documents this field as user-configurable.\n\n**5. [CORRECTNESS]** `crates/hyprstream/src/services/factories.rs:738`\n`BackendType::Auto` trailing arm in the `match` silently falls through to `NspawnBackend` with a comment saying 'should not reach here', but there is no `unreachable!()` or error — a future change to `resolve()` that returns `Auto` would silently downgrade isolation without failing.\n> If `resolve()` is changed or bypassed (e.g., config deserialization returns `Auto` after a serde bug), the factory silently creates an Nspawn backend while logging nothing unusual; no panic, no error, no indication the intended backend was not selected.\n\n**6. [CORRECTNESS]** `crates/hyprstream-workers/src/config.rs:42`\n`resolve()` probes only `cloud-hypervisor` presence in PATH to decide Kata eligibility, but a PATH-accessible binary does not guarantee a working, correctly configured Kata runtime — it will start and then fail at sandbox creation time.\n> Host has `cloud-hypervisor` in PATH (e.g., installed but not configured for Kata), `kata-vm` feature is compiled in; `Auto` resolves to `Kata`; `KataBackend::initialize()` or the first `start()` call fails with an opaque VM error rather than gracefully falling back to podman/nspawn, causing a service start failure where nspawn would have worked.\n\n**7. [NIT]** `crates/hyprstream-workers/src/config.rs:347`\n`test_backend_type_auto_resolves_to_concrete` only asserts `!= Auto` but does not assert the specific concrete variant, so a broken `resolve()` that returns an unexpected variant (e.g., Kata on a podman-only CI host) passes the test.\n> In a CI environment without `kata-vm` feature and without podman/docker, `resolve()` must return `Nspawn`; the test passes regardless because it only checks `!= Auto`, so a regression where `resolve()` returns the wrong concrete type goes undetected.

ewindisch:
## Agentic Code Review

**Overall:** The PR is structurally sound and the feature-flag / resolve() logic is well thought out; however it has two security issues worth blocking on: rootful docker is silently accepted as the "rootless" backend (no failure on `rootless=false`), and any authorized RPC caller can specify an arbitrary OCI image via annotations which will be pulled and run without an allowlist. The biggest single risk is the main.rs wildcard arm silently running nspawn when the operator pinned `kata`, creating a silent isolation downgrade with no diagnostic.

### Findings

**1. [BUG]** `crates/hyprstream/src/bin/main.rs:1108`
Wildcard `_ =>` arm silently falls back to NspawnBackend even when the user explicitly configured `backend: kata`; the service factory path (factories.rs) correctly returns an error in this case but main.rs swallows it.
> Build without `kata-vm` feature; operator sets `worker.backend: kata` in config; starts the service via the CLI worker path. Expected: clear error. Actual: silently runs nspawn, giving the operator no indication that isolation is absent.

**2. [SECURITY]** `crates/hyprstream-workers/src/runtime/podman.rs:154`
`initialize()` checks `{{.Host.Security.Rootless}}` but only logs the result without failing or warning when it is false, so rootful Docker daemons are accepted as a 'rootless' backend silently.
> `docker` is installed and in PATH but running as a rootful daemon. Auto-select or explicit `podman` config selects PodmanBackend via `find_runtime()`. `initialize()` logs `rootless=false` then returns `Ok(())`. All subsequent sandbox workloads run as root inside a non-isolated Docker container, defeating the intended rootless-container isolation model.

**3. [CORRECTNESS]** `crates/hyprstream-workers/src/config.rs:42`
`resolve()` auto-selects `Kata` based solely on `which::which("cloud-hypervisor")` being present in PATH, without verifying that the kata-agent, guest kernel image, or virtiofsd are configured.
> A developer installs cloud-hypervisor standalone (e.g., for an unrelated project) without the full Kata runtime stack. `Auto` resolves to `Kata`; `KataBackend::initialize()` then fails at service startup with an opaque error about missing VM images or sockets, whereas the intended fallback would have been podman or nspawn.

**4. [SECURITY]** `crates/hyprstream-workers/src/runtime/podman.rs:136`
The `hyprstream.io/worker-image` annotation, which is included verbatim in `podman run` arguments, can be set by any authorized RPC caller to an arbitrary registry image, enabling image-pull from untrusted registries onto the worker host.
> Any peer with a valid worker token (authorized to call `RunPodSandbox`) sets `annotations["hyprstream.io/worker-image"] = "attacker.example/malicious:latest"`. The PodmanBackend pulls and runs that image with host network access (`slirp4netns`). There is no allowlist or digest-pinning guard.

**5. [NIT]** `crates/hyprstream/src/services/factories.rs:738`
The `BackendType::Auto` arm at the end of the backend match block silently falls back to nspawn instead of using `unreachable!()`, masking future refactors that forget to call `resolve()` before the match.
> A future refactor removes or bypasses the `configured_backend.resolve()` call. `Auto` reaches the match, silently starts nspawn with no diagnostic, and operators see no indication that auto-detection was skipped.

**6. [NIT]** `crates/hyprstream-workers/src/runtime/podman.rs:32`
`find_runtime()` calls `which::which` on every sandbox operation (start, stop, destroy, exec, get_pids, update_resources), causing repeated PATH scans per operation instead of caching the resolved binary path at construction time.
> Under high sandbox churn (many concurrent workers, rapid start/exec/stop cycles), the repeated PATH scans add latency and create contention on filesystem metadata caches; if PATH is mutated between calls (unlikely but possible in test environments), different operations could pick different runtime binaries for the same sandbox.

crates/hyprstream-workers/src/config.rs
```diff
@@ -13,17 +13,49 @@ use std::path::PathBuf;
...
+    pub fn resolve(self) -> Self {
+        if self != Self::Auto {
+            return self;
+        }
+        #[cfg(feature = "kata-vm")]
+        if which::which("cloud-hypervisor").is_ok() {
+            return Self::Kata;
+        }
+        #[cfg(feature = "podman")]
+        if which::which("podman").is_ok() || which::which("docker").is_ok() {
```
ewindisch, line 46: Runtime selection is an implicit PATH scan (`which::which`) run on every `resolve()` call. Two concerns: (1) selection is PATH-dependent rather than operator-configured — an operator cannot pin "use crun at /usr/local/bin" or pass runtime options; (2) it re-probes on each invocation. Prefer an operator-configured runtime path/name + options. See Decision Q4.

crates/hyprstream-workers/src/config.rs
```diff
@@ -13,17 +13,49 @@ use std::path::PathBuf;
...
+            return Self::Podman;
+        }
+        Self::Nspawn
```
ewindisch, line 49: Silent degradation: `Auto` falls back to `Nspawn` when neither cloud-hypervisor nor podman/docker is found — no warning, no error. An operator expecting VM-grade isolation (Kata) silently gets nspawn (host-filesystem, weaker isolation). Note this is now the **out-of-box default** (the default changed from `Kata` to `Auto` in this PR), so silent-degrade is the default behavior. At minimum warn at the resolution site; consider failing closed when `Auto` cannot reach the isolation tier the operator implied. See Decision Q2/Q4.

ewindisch (commented):
Inline review for availability-based backend auto-selection (#348). This PR is stacked on #484 (base `ewindisch/346-podman-backend`); the podman backend itself — and its dropped-CRI-config / image-pull / network-mode findings — are reviewed inline on #484. Findings here concern the `Auto` selection semantics. These feed the joint 🔶 Decision-required comment, cross-linked with #484.

ewindisch:
## 🔶 Decision required (human sign-off — architecture + authz)

This applies **jointly to #484** (rootless podman/docker OCI backend, #346) **and #486** (availability-based backend auto-selection, #348). The two are stacked (#486 bases on #484) and entangled with a larger worker-runtime architecture decision, so **no code has been changed/pushed** — these are open questions for human sign-off before any rework. Inline diff comments are on both PRs.

**Q1 — Architecture.** The podman backend drops the entire CRI `PodSandboxConfig` (the param is `_config`) and is **not wired into `BackendType`** in #484, so it is also currently dead code. Meanwhile `worker.capnp` is already a faithful CRI v1 model that *every* backend (podman **and** kata) currently ignores. Do we:
- **(a)** rework into a single `OciBackend` that maps CRI → `oci_spec::runtime::Spec` → a configurable embedded runtime (youki/crun), and/or add a `CriBackend` (CRI client driving containerd) for standard hosts; **or**
- **(b)** land #484/#486 as tactical podman-CLI shims and supersede later (with a tracking issue)?

**Q2 — Authz / fail-closed.** When "rootless" is promised but the runtime is rootful docker — or rootless cannot be proven — do we **fail-closed (reject)** or **warn-and-continue**? Today the result is only logged and never enforced (even `rootless == false` podman passes). Same question for #486's `Auto` silently degrading to `nspawn` when no VM-isolation runtime is present (now the out-of-box default).

**Q3 — Authz / image provenance.** Before the `hyprstream.io/worker-image` annotation is usable, do we require a **registry/image allowlist + digest pinning**, and/or **gate the annotation behind operator config**? Today any authorized RPC caller can pull and run any OCI image from any registry.

**Q4 — Operator config vs PATH resolution.** Should runtime selection be **operator-CONFIGURED** (explicit runtime path/name + options) rather than PATH-resolved via `find_runtime()` (#484, `podman.rs`) / `which::which` in `BackendType::resolve()` (#486, `config.rs`)?

**References:** CRI-client-VFS injection spike doc `docs/plans/2026-06-26-cri-client-vfs-injection-spike.md` (branch `ewindisch/cri-client-spike`) and #495 (hyprstream-vfsd native virtio-fs host mount of the VFS namespace).

— cross-linked: #484 ↔ #486

ewindisch:
Closing — superseded by the Sandbox backend taxonomy epic **#508** + spine **#507**.

The concept (availability-based backend selection, #348) is retained, but this implementation predates the full backend set (knows only Kata/Nspawn/Podman) and has two posture problems the decided design fixes: **silent nspawn fallback = fail-OPEN isolation downgrade**, and **PATH-scan resolution** instead of operator config.

→ Reworked as the **config-driven, fail-closed, full-taxonomy-aware selection** model in **#507** (selects across Kata/Oci/Cri/Wasm/Nspawn; errors instead of silently downgrading isolation). #348 stays open, tracked under #508.

PR #487, opened 2026-06-26
feat(workflow): scan_repo + YAML round-trip + subscribe/unsubscribe + event routing (#349)

### LINKED ISSUE #349: T3-A: finish WorkflowService event routing + scan_repo + YAML round-trip
Track 3. Fill the TODOs: repo/training/metrics event routing (only worker events wired today), scan_repo, subscribe/unsubscribe, YAML round-trip. Part of #341. Plan: docs/plans/2026-06-19-worker-ga-epic.md

ewindisch:
## Summary

Completes T3-A WorkflowService TODOs (#349). All changes are in `service.rs` — no capnp/RPC schema changes.

### `scan_repo()`
- Reads `.github/workflows/*.yml` from the path registered via `register_repo_path(repo_id, path)`
- Parses each file using the existing `Workflow::parse()` 
- Calls `extract_triggers()` to map `on:` event names → `EventTrigger` variants
- Returns `Vec<WorkflowDef>`; empty Vec (not error) if path unregistered or dir absent

### YAML round-trip
- `handle_scan_repo` now serializes `wf.workflow` back via `serde_yaml::to_string()` into `WorkflowDefWire.yaml`

### subscribe/unsubscribe
- `subscriptions: RwLock<HashMap<String, WorkflowSubscription>>` tracks active subs
- `handle_subscribe` generates a `sub-<workflow_id>-<ts-tid>` ID, inserts into both maps
- `handle_unsubscribe` removes by sub_id

### Event routing
- `start()` now subscribes to `worker.`, `repo.`, `training.`, `metrics.` prefixes (was only `worker.`)

### initialize()
- Now calls `rescan_repo()` for all registered repos on startup

## Test plan

- [ ] `cargo clippy -p hyprstream-workers --features podman --all-targets` passes (verified locally)
- [ ] CI green

Closes #349. Part of epic #341. Stacked on #486.

### DIFF
--- a/crates/hyprstream-workers/src/workflow/service.rs
+++ b/crates/hyprstream-workers/src/workflow/service.rs
@@ -313,61 +313,73 @@ impl WorkflowService {
             }
         };
 
-        let workflow_dir = root.join(".github").join("workflows");
-        if !workflow_dir.is_dir() {
-            tracing::debug!(
-                repo_id = %repo_id,
-                dir = %workflow_dir.display(),
-                "No .github/workflows dir found"
-            );
-            return Ok(Vec::new());
-        }
-
-        let mut defs = Vec::new();
-        let read_dir = std::fs::read_dir(&workflow_dir).map_err(|e| {
-            crate::error::WorkerError::WorkflowParseError(format!(
-                "Failed to read workflows dir {}: {e}",
-                workflow_dir.display()
-            ))
-        })?;
-
-        for entry in read_dir.flatten() {
-            let path = entry.path();
-            let ext = path.extension().and_then(|e| e.to_str()).unwrap_or("");
-            if ext != "yml" && ext != "yaml" {
-                continue;
+        // The directory walk + file reads + YAML parsing are all blocking I/O.
+        // Run them on the blocking thread pool so we don't stall the async
+        // worker (and starve other tasks sharing this runtime).
+        let repo_id_owned = repo_id.to_owned();
+        let defs = tokio::task::spawn_blocking(move || -> Result<Vec<WorkflowDef>> {
+            let workflow_dir = root.join(".github").join("workflows");
+            if !workflow_dir.is_dir() {
+                tracing::debug!(
+                    repo_id = %repo_id_owned,
+                    dir = %workflow_dir.display(),
+                    "No .github/workflows dir found"
+                );
+                return Ok(Vec::new());
             }
 
-            let yaml = match std::fs::read_to_string(&path) {
-                Ok(s) => s,
-                Err(e) => {
-                    tracing::warn!(path = %path.display(), error = %e, "Failed to read workflow file");
+            let mut defs = Vec::new();
+            let read_dir = std::fs::read_dir(&workflow_dir).map_err(|e| {
+                crate::error::WorkerError::WorkflowParseError(format!(
+                    "Failed to read workflows dir {}: {e}",
+                    workflow_dir.display()
+                ))
+            })?;
+
+            for entry in read_dir.flatten() {
+                let path = entry.path();
+                let ext = path.extension().and_then(|e| e.to_str()).unwrap_or("");
+                if ext != "yml" && ext != "yaml" {
                     continue;
                 }
-            };
 
-            let workflow = match Workflow::parse(&yaml) {
-                Ok(w) => w,
-                Err(e) => {
-                    tracing::warn!(path = %path.display(), error = %e, "Failed to parse workflow");
-                    continue;
-                }
-            };
+                let yaml = match std::fs::read_to_string(&path) {
+                    Ok(s) => s,
+                    Err(e) => {
+                        tracing::warn!(path = %path.display(), error = %e, "Failed to read workflow file");
+                        continue;
+                    }
+                };
 
-            let rel_path = path
-                .strip_prefix(&root)
-                .map(|p| p.to_string_lossy().into_owned())
-                .unwrap_or_else(|_| path.to_string_lossy().into_owned());
+                let workflow = match Workflow::parse(&yaml) {
+                    Ok(w) => w,
+                    Err(e) => {
+                        tracing::warn!(path = %path.display(), error = %e, "Failed to parse workflow");
+                        continue;
+                    }
+                };
 
-            let triggers = extract_triggers(&workflow);
+                let rel_path = path
+                    .strip_prefix(&root)
+                    .map(|p| p.to_string_lossy().into_owned())
+                    .unwrap_or_else(|_| path.to_string_lossy().into_owned());
 
-            defs.push(WorkflowDef {
-                path: rel_path,
-                repo_id: repo_id.to_owned(),
-                workflow,
-                triggers,
-            });
-        }
+                let triggers = extract_triggers(&workflow);
+
+                defs.push(WorkflowDef {
+                    path: rel_path,
+                    repo_id: repo_id_owned.clone(),
+                    workflow,
+                    triggers,
+                });
+            }
+
+            Ok(defs)
+        })
+        .await
+        .map_err(|e| {
+            crate::error::WorkerError::WorkflowParseError(format!("scan_repo blocking task failed: {e}"))
+        })??;
 
         tracing::info!(repo_id = %repo_id, count = defs.len(), "Scan complete");
         Ok(defs)
--- a/crates/hyprstream-workers/src/workflow/service.rs
+++ b/crates/hyprstream-workers/src/workflow/service.rs
@@ -732,11 +732,25 @@ impl WorkflowHandler for WorkflowService {
         _request_id: u64,
         sub_id: &str,
     ) -> AnyhowResult<WorkflowResponseVariant> {
-        let removed = {
+        let removed_sub = {
             let mut subs = self.subscriptions.write().await;
-            subs.remove(sub_id).is_some()
+            subs.remove(sub_id)
         };
-        if removed {
+        if let Some(sub) = removed_sub {
+            // Keep repo_workflows in sync: handle_subscribe pushes the same
+            // subscription into the repo bucket, so unsubscribe must remove it
+            // there too (otherwise a dead entry leaks).
+            if let Some(repo_id) = repo_id_from_workflow_id(&sub.workflow_id) {
+                let mut repo_workflows = self.repo_workflows.write().await;
+                if let Some(list) = repo_workflows.get_mut(repo_id) {
+                    if let Some(pos) = list.iter().position(|s| s.workflow_id == sub.workflow_id) {
+                        list.remove(pos);
+                    }
+                    if list.is_empty() {
+                        repo_workflows.remove(repo_id);
+                    }
+                }
+            }
             tracing::info!(sub_id = %sub_id, "Unsubscribed");
         } else {
             tracing::debug!(sub_id = %sub_id, "Unsubscribe: subscription not found");
--- a/crates/hyprstream-workers/src/workflow/service.rs
+++ b/crates/hyprstream-workers/src/workflow/service.rs
@@ -1,6 +1,7 @@
 //! WorkflowService - RequestService implementation for workflow orchestration
 
 use std::collections::HashMap;
+use std::path::PathBuf;
 use std::sync::Arc;
 use tokio::sync::RwLock;
 use tokio::task::JoinHandle;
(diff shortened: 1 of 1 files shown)


ewindisch:
## Agentic Code Review

**Overall:** The PR successfully fills in the `scan_repo` / YAML round-trip / subscribe TODOs and is a meaningful step forward, but it ships two security-relevant issues that should be addressed before merge: `handle_scan_repo` now serializes and returns full workflow YAML (including any embedded env secrets) over the wire where the pre-PR code returned empty strings, and `handle_register` parses untrusted YAML with no size cap using a serde_yaml version known to stack-overflow on deeply-nested input. The biggest correctness risk is the blocking `std::fs` I/O inside async functions (scan_repo, initialize), which can starve the tokio runtime under load.

### Findings

**1. [BUG]** `crates/hyprstream-workers/src/workflow/service.rs:304`
scan_repo() calls std::fs::read_dir(), std::fs::read_to_string(), and PathBuf::is_dir() — all blocking syscalls — directly inside an async fn without spawn_blocking, starving the tokio worker thread pool during filesystem I/O.
> A repo with many large YAML files, or a path on a slow/network-mounted filesystem, causes the tokio worker thread to block. Under concurrent load, this can exhaust all worker threads and deadlock the runtime. The existing dispatch() fn already uses tokio::spawn for the runner, but scan_repo is called synchronously from the async handle_scan_repo and initialize() paths.

**2. [SECURITY]** `crates/hyprstream-workers/src/workflow/service.rs:619`
handle_scan_repo() calls self.scan_repo() and returns workflow YAML to the caller, but scan_repo reads arbitrary files under the pre-registered repo root; if any workflow YAML embeds secrets (env vars, token values) those are fully serialized into WorkflowDefWire.yaml and returned over the wire.
> A workflow file contains `env: SECRET_KEY: ${{ secrets.TOKEN }}` (a common GitHub Actions pattern). serde_yaml::to_string serializes it verbatim and a caller with only 'query' scope on scan_repo receives it in the response. The pre-PR code returned yaml: String::new() (empty), so this PR introduces new information leakage for secrets embedded in workflow YAML files.\n\n**3. [CORRECTNESS]** `crates/hyprstream-workers/src/workflow/service.rs:619`\nserde_yaml::to_string().unwrap_or_default() in handle_scan_repo silently swallows serialization errors and returns an empty yaml field, causing the caller to receive a malformed WorkflowDefWire with no indication of failure.\n> If serde_yaml::to_string fails (e.g. on a workflow with a field containing an unsupported value), the response WorkflowDefWire.yaml is an empty string. A caller receiving this then calls handle_register with data.yaml == '' which falls into the 'build minimal Workflow' branch, silently discarding all parsed triggers and jobs, and registers a broken stub workflow under the original path.\n\n**4. [BUG]** `crates/hyprstream-workers/src/workflow/service.rs:677`\nhandle_subscribe() inserts the new WorkflowSubscription into self.repo_workflows but handle_unsubscribe() only removes from self.subscriptions, leaving a dead subscription in repo_workflows that is never cleaned up except on the next rescan_repo call.\n> A client subscribes then unsubscribes. The subscription is removed from self.subscriptions but remains in self.repo_workflows. Any future code path that iterates repo_workflows to find active subscribers (e.g., a fan-out notification loop) will still attempt to dispatch to the unsubscribed workflow_id, potentially triggering spurious workflow executions.\n\n**5. [BUG]** `crates/hyprstream-workers/src/workflow/service.rs:783`\nuuid_v4_simple() uses SystemTime + thread ID as a subscription ID, which is not unique: two async tasks executing subscribe concurrently on the same tokio worker thread within the same nanosecond produce identical IDs, silently overwriting each other's entry in self.subscriptions.\n> Under concurrent subscribe RPC calls, two requests land on the same tokio thread in the same nanosecond (realistic on a loaded server). Both get sub_id = 'a1b2c3d4-ThreadId(2)'. The second insert overwrites the first in the HashMap; the first client's sub_id still exists in its response but no longer maps to its subscription. The first client's unsubscribe does nothing and the second client gets silently unsubscribed when the first client calls unsubscribe.\n\n**6. [SECURITY]** `crates/hyprstream-workers/src/workflow/service.rs:650`\nhandle_register() calls Workflow::parse(&data.yaml) on untrusted caller-supplied YAML with no size limit; serde_yaml 0.9 (in use here) can stack-overflow on deeply-nested YAML structures.\n> A caller with 'write' scope sends a register RPC with a deeply-nested YAML payload (e.g., 1000 levels of nested maps). serde_yaml::from_str recurses into the YAML tree, overflows the stack, and panics the thread. Since the service uses async_trait(?Send) on a LocalSet, the panic propagates and can crash the service process.\n\n**7. [CORRECTNESS]** `crates/hyprstream-workers/src/workflow/service.rs:800`\nextract_triggers() maps the GitHub Actions 'create' event to RepoEventType::Tag, but the 'create' event fires for both new branches and new tags; branch creation events are silently dropped.\n> A workflow with `on: create` is intended to run when a new branch is created. extract_triggers maps it to RepoEventType::Tag. No RepositoryEvent::Push or dedicated Branch variant exists. When a 'repo.create' event arrives for branch creation, no handler fires, and the workflow is never dispatched even though the user's on:create trigger should have matched.\n\n**8. [CORRECTNESS]** `crates/hyprstream-workers/src/workflow/service.rs:677`\nhandle_subscribe() splits workflow_id on ':' to derive repo_id, but workflow_id values that contain no ':' (e.g., a bare name registered via handle_register with an empty path) cause the entire workflow_id string to be used as the repo_id key, misrouting the subscription.\n> A caller registers a workflow via handle_register with path='' and repo_id='myrepo', producing workflow_id='myrepo:'. A subsequent subscribe with workflow_id='myrepo:' splits correctly to 'myrepo'. But if path was omitted and workflow_id was stored as just 'myrepo' by some other path, subscribe uses 'myrepo' as the repo_id key, which might be correct by accident, but the comment 'repo_id:path convention' is not enforced and any deviation silently creates an orphaned subscription under a wrong key.

ewindisch:
## ✅ High-confidence review fixes applied

The 6 non-authz correctness findings from review have been fixed and pushed to this branch. Each is an isolated commit on `crates/hyprstream-workers/src/workflow/service.rs`:

| # | Fix | Commit |
|---|-----|--------|
| 1 | `scan_repo` blocking FS work (`read_dir`/`read_to_string`/`is_dir`) moved onto `tokio::task::spawn_blocking` so it no longer stalls the async worker / starves the shared runtime | `112cdc7` |
| 2 | Subscription IDs now use the workspace `uuid` crate's v4 instead of the bespoke `SystemTime`+thread-id helper (which collided on same-thread same-nanosecond subscribes) | `4d15767` |
| 3 | `handle_scan_repo` YAML serialization errors are now propagated instead of `unwrap_or_default()` swallowing them into an empty-workflow stub | `a1f594d` |
| 4 | `'create'` event now emits **both** a `Tag` and a `Push` trigger (GitHub's `create` fires for branch *and* tag creation; branch creation was being silently dropped) | `f0deac3` |
| 5 | `handle_subscribe` rejects ids that don't follow the `repo_id:path` convention via the shared `repo_id_from_workflow_id()` helper (previously misrouted into the wrong repo bucket) | `dcef143` |
| 6 | `handle_unsubscribe` now removes the matching entry from `repo_workflows` too (and drops the bucket when empty), fixing the dead-entry leak | `690fd46` |

Two further items (YAML secret leakage in `handle_scan_repo`, unbounded YAML parse in `handle_register`) are **intentionally NOT auto-fixed** — they involve an authz/DoS policy decision. See the inline comments + the "🔶 Decision required" comment below.

crates/hyprstream-workers/src/workflow/service.rs
```diff
@@ -503,15 +617,24 @@ impl WorkflowHandler for WorkflowService {
...
-                triggers: Vec::new(), // EventTrigger is empty struct for now
-                yaml: String::new(),  // TODO: serialize workflow back to YAML
-            }).collect()
-        ))
+        let defs = workflows
+            .iter()
+            .map(|wf| {
+                // Propagate serialization failures instead of defaulting to an
+                // empty yaml string (which would register a broken stub).
+                let yaml = serde_yaml::to_string(&wf.workflow).map_err(|e| {
```
ewindisch, line 625: 🔶 **Deferred — authz decision (info leak):** `handle_scan_repo` now serializes the *full* parsed workflow back to YAML and returns it over the wire (`yaml` field). Pre-PR this returned `String::new()`. The serialized YAML includes any embedded secret references / `env:` blocks (e.g. `env:\n  SECRET: ${{ secrets.X }}`), so any caller able to invoke `scan_repo` now receives the complete workflow source — a new information-leak surface that did not exist before.

Not auto-fixed: whether to redact/strip secrets or gate the full-YAML return behind a scope/permission is a policy call. See Q1 in the 🔶 Decision-required comment.

crates/hyprstream-workers/src/workflow/service.rs
```diff
@@ -520,16 +643,23 @@ impl WorkflowHandler for WorkflowService {
...
+            Workflow {
                 name: data.name.clone(),
                 on: super::parser::WorkflowTrigger::List(Vec::new()),
                 env: HashMap::new(),
                 jobs: HashMap::new(),
-            },
-            triggers: Vec::new(), // TODO: convert EventTrigger to EventTrigger
+            }
+        } else {
+            Workflow::parse(&data.yaml)?
```
ewindisch, line 655: 🔶 **Deferred — DoS decision (unbounded parse):** `Workflow::parse(&data.yaml)?` parses caller-supplied YAML with no size or nesting-depth cap. `serde_yaml` 0.9 is known to **stack-overflow on deeply-nested input**, which aborts the thread (panic = process/worker crash, not a recoverable `Err`). An untrusted `register` caller can therefore trigger a DoS.

Not auto-fixed: adding a byte-size cap and/or depth limit (or a recursion-safe parse path) is a policy call on the acceptable limits. See Q2 in the 🔶 Decision-required comment.

ewindisch:
## 🔶 Decision required (human sign-off — authz/DoS)

Two findings are intentionally left unfixed because they require a policy/authz call rather than a mechanical correctness fix. Please decide:

**Q1 — `handle_scan_repo` YAML secret leakage** (inline comment ~L625)
Should `handle_scan_repo` **redact/strip secrets** from the returned workflow YAML, or **gate the full-YAML return behind a scope/permission**? This is a new info-leak: pre-PR the `yaml` field was empty; now any caller of `scan_repo` receives the complete workflow source including embedded `env`/`secrets.*` references.

**Q2 — `handle_register` unbounded YAML parse** (inline comment ~L655)
Should we add a **size cap and/or nesting-depth limit** (and/or a recursion-safe parser) on `handle_register`'s parse of untrusted caller YAML to prevent a `serde_yaml` 0.9 stack-overflow → thread panic DoS? **If yes, what limits?** (proposed starting point: e.g. reject payloads > 256 KiB and cap nesting depth — needs a sign-off on the actual numbers).

**Q3 — merge order / branch stacking**
This branch is **stacked on #346 (#484) + #348 (#486)** (podman/backend-autoselect — base ref is `ewindisch/348-backend-autoselect`) and carries those commits, so it cannot merge before them. This workflow change is **functionally independent** of the podman backend. Do we:
  (a) **rebase #487 onto `main`** to detach it and let it merge on its own, or
  (b) **hold #487** until the #484 OCI decision lands and the stack merges in order?

No code changes will be made until these are answered.

ewindisch:
**Triage:** #349's actual content (WorkflowService scan_repo/YAML round-trip/subscribe-unsubscribe/event-routing — single file, `crates/hyprstream-workers/src/workflow/service.rs`) has been extracted and re-opened as **#622**, rebased cleanly onto `ewindisch/workers-engines-recut` (current sandbox-epic-derived branch). Build/test/clippy all green there (22/22 tests passing).

This PR's base (`ewindisch/348-backend-autoselect`, from closed #486/#348) bundles the now-dead/superseded podman+autoselect work — 1:1.0 confirmed that's cleanly superseded by their #617 (podman re-cut) and the in-flight #348-v2, and should not be merged forward. #349's content had zero coupling to that bundled content.

**Recommend closing this PR in favor of #622** (recommend-only per protocol — not closing it myself).

ewindisch:
Closing: #349's content landed via #622 (merged). This branch's bundled podman/autoselect base (#486/#348) is superseded by 1:1.0's #617 + #348-v2.

PR #488, opened 2026-06-26
feat(#483): hyprstream-python — RustPython VFS shell

### LINKED ISSUE #483: feat(python): hyprstream-python — WASM-sandboxed RustPython VFS shell at /lang/python/
## Summary

Add a `hyprstream-python` crate that mounts a Python scripting environment at `/lang/python/` in the Plan9 VFS namespace, using [RustPython](https://github.com/RustPython/RustPython) (`rustpython-vm`) compiled to **wasm32** and executed inside an **embedded WASM sandbox**.

**Security model (decided — see [design doc](docs/plans/2026-06-26-rustpython-wasm-vfs-sandbox.md) + the WASM-redirect decision comment below):** the sandbox boundary is the **WASM boundary**, not Python-level `__import__`/`importlib` removal (which is best-effort and unwinnable). A WASM module has **zero ambient authority**; the **only** injected capability is the hyprstream VFS, exposed as host functions and scoped to the actor's **atproto identity (Subject) + Casbin policy**. `import os; os.system(...)` is **inert by construction** — there is no host syscall to reach.

> **Scope change:** this supersedes the earlier "native + wasm32" design. Python is **WASM-only** — there is no native in-process interpreter path. The native RustPython approach (best-effort `__import__` removal) cannot provide a real sandbox and is being redirected (PR #488 disposition: do not merge native as the security story).

## VFS layout (unchanged)
```
/lang/python/
  eval       # ctl file: write Python source, read result
  vars/      # dynamic dir: one file per global name
  defs/      # dynamic dir: one file per defined function
  stdout     # last captured sys.stdout output
```

## Architecture (WASM-capability sandbox)
- **New `hyprstream-python-host` crate** embedding **wasmtime** with a custom `Linker` exposing **only**: the VFS host fns (`vfs_walk/open/read/write/stat/ls/create`) + a host RNG. **No `wasmtime-wasi`, no WASI preopens.**
- RustPython built for `wasm32-unknown-unknown` (getrandom backed by the host RNG import, not the browser `wasm_js` backend), run as a guest under our wasmtime host.
- **VFS = sole capability:** host fns call the existing `Mount` trait threading `caller: &Subject`; `Subject` is the Casbin subject. The shell inherits VFS Casbin enforcement (lands with #353/#319/#328), does not reimplement it.
- **Identity binding:** bind atproto DID → `Subject::new(did)` at instantiation (reuse `spawn_vfs_proxy(ns, subject)`); one instance = one Subject. **Delete the `SHELL_CTX` thread-local** (root of the cross-tenant leak).
- **CPU/DoS:** wasmtime **epoch interruption** (timer thread + `set_epoch_deadline`) bounds runaway loops; fuel as the deterministic test option.

## Two hard caveats (the design lives or dies by these)
1. **Zero WASI grants** — the import surface must be exactly the VFS host fns. One preopen reopens the escape.
2. **Epoch/fuel bound** — WASM alone does not stop infinite loops.

## Phased plan
P0 build spike (RustPython→wasm32 under wasmtime, host RNG) · P1 VFS host-fn linker + Subject binding · P2 epoch CPU bound · P3 VFS layout (eval/vars/defs/stdout) · P4 frozen minimal stdlib (defense-in-depth).

Design doc: `docs/plans/2026-06-26-rustpython-wasm-vfs-sandbox.md`. Supersedes the native design; PR #488 redirected.

ewindisch:
> **Status: DRAFT — redirected.** This PR is **not the security story** and will not merge as the native in-process interpreter. Per the [#483 WASM-capability decision](https://github.com/hyprstream/hyprstream/issues/483) the execution substrate is being swapped to an embedded-wasmtime guest (zero ambient authority, VFS-only capability). This branch is retained for its **reusable VFS-facing surface** (`/lang/python` mount, persistent scope, stdout capture, the `vfs_*` builtin shape) which P2 ports onto the wasm host. See the pinned disposition comment below.

## What this is now (re-scoped)

`hyprstream-python` — the `/lang/python/` VFS shell. The **interpreter sandbox boundary moved from Python-level builtin removal to the WASM module boundary**. This PR no longer claims isolation; it is the source of the reusable mount/scope/stdout surface that rides on top of [#505](https://github.com/hyprstream/hyprstream/issues/505)'s `hyprstream-wasm` substrate (Profile A — zero WASI, VFS host-fns only).

**Depends on:** #505 P1 (`hyprstream-wasm` host crate + capability Linker + Subject binding + epoch bound). Blocked until P1 lands. Tracked as **#483 P2**.

## Why the native approach was redirected

Removing `builtins.__import__` is best-effort and unwinnable: the interpreter shares the host process's authority, so `import os; os.system(...)` (and `importlib`/`sys.modules` paths) escape the VFS. The agentic review on this PR confirmed it; the docs were softened to "best-effort, not a hard sandbox." Soft docs are not a security control. Under WASM the property is **structural** — a guest has zero ambient authority and `import os` is *inert by construction*. Full rationale in the [#483 decision comment](https://github.com/hyprstream/hyprstream/issues/483#issuecomment-4813721527) and `docs/plans/2026-06-26-rustpython-wasm-vfs-sandbox.md`.

## Reusable surface (what P2 keeps)

- `/lang/python/` mount layout: `eval` / `vars/` / `defs/` / `stdout` (`mount.rs`)
- Persistent interpreter scope (fixed in `b10bfa8` — `/vars`,`/defs` no longer always-empty)
- Pure-Python stdout capture (`__Capture`, hardened no-`None` sink)
- The `vfs_*` builtin **shape** (`cat`/`ls`/`write`/`ctl`/`mount`) — re-expressed as wasm host-fn imports under Profile A

## Discarded (do not carry forward)

- Native in-process `rustpython-vm` dependency + the tokio bridge runtime
- `builtins.harden()` `__import__`-removal as a security control (keep only as defense-in-depth via a minimal frozen stdlib in the guest)
- `SHELL_CTX` thread-local (root of the cross-tenant leak) -> replaced by per-instance `spawn_vfs_proxy(ns, subject)` Subject binding
- The wall-clock `check_signals` watchdog -> replaced by wasmtime epoch interruption

<details>
<summary>Original native-crate description (historical)</summary>

- Adds `crates/hyprstream-python/` -- a `!Send` Python shell (`PythonShell`) backed by RustPython 0.3 (`freeze-stdlib`) that exposes the hyprstream VFS namespace via injected builtins.
- VFS-aware builtins (`cat`/`ls`/`write`/`ctl`/`help`/`mount`/`json_parse`); dangerous builtins removed after stdlib; `/lang/python/` mount (`eval`/`stdout`/`vars/`/`defs/`).
- Review fixes applied: bridge-runtime VFS calls (no `block_on` panic), persistent scope, hardened stdout, exec watchdog. 20 tests pass.

Closes #483 -- *superseded; #483 stays open and is delivered by the wasm path (#483 P2).*
</details>


ewindisch:
## Agentic Code Review

**Overall:** The PR introduces a well-structured RustPython VFS shell with a clean channel-based design to work around RustPython's !Send constraint, and correctly uses parking_lot::Mutex instead of unsafe raw pointer casts (unlike the TCL peer). However, two significant issues need attention before merge: (1) the sandbox claim is broken — removing __import__ from Python builtins does not stop the `import` statement, which uses RustPython's internal bytecode path and remains fully functional after harden(), leaving os/subprocess/ctypes directly accessible; and (2) the /lang/python/vars/ and /lang/python/defs/ VFS paths are permanently empty because every method creates a new fresh scope rather than referencing a shared interpreter-level module dict. The SHELL_CTX thread-local also has no Drop cleanup, creating a latent cross-tenant identity confusion if two PythonShell instances ever share a thread.

### Findings

**1. [SECURITY]** `crates/hyprstream-python/src/builtins.rs:29`
Removing __import__ from builtins does not block the Python import statement; RustPython's import bytecode uses internal VM machinery, not the __import__ builtin, so user code can still execute `import os; os.system('...')` or `import subprocess` to escape the VFS sandbox.
> User submits code `import os; os.system('id')` via /lang/python/eval; despite __import__ being removed, the import statement succeeds (confirmed by CAPTURE_SETUP itself using `import sys as _sys` after harden() runs), giving direct host-OS access in contradiction to the PR's stated security guarantee that 'all I/O goes through the VFS'.

**2. [CORRECTNESS]** `crates/hyprstream-python/src/lib.rs:131`
list_vars(), get_var(), list_defs(), and get_def() each create a fresh scope with vm.new_scope_with_builtins(); user-defined variables set in a prior eval()/exec() call are not visible because each call discards its scope, so the /lang/python/vars/ and /lang/python/defs/ VFS paths always return empty results.
> Caller does `eval('x = 42')` then reads /lang/python/vars/x via a GetVar command; list_vars() creates a new scope with no user globals and returns an empty Vec, so the VFS file is not found and the read returns MountError::NotFound, making the vars/ and defs/ VFS subtrees permanently empty.

**3. [SECURITY]** `crates/hyprstream-python/src/lib.rs:35`
SHELL_CTX is a module-level thread_local with no Drop implementation on PythonShell; if two PythonShell instances are created on the same thread, the second construction overwrites the first shell's namespace/subject, so the first shell's VFS builtins silently operate under the second shell's identity.
> Thread creates ShellA(subjectA, namespaceA) then ShellB(subjectB, namespaceB); SHELL_CTX now holds subjectB. Any subsequent cat()/ls() call on ShellA's Python builtins uses subjectB's identity and namespace, potentially allowing cross-tenant VFS access or authorization bypass if subjects have different permission sets.

**4. [BUG]** `crates/hyprstream-python/src/builtins.rs:78`
vfs_async uses Handle::current().block_on() which panics if called from within an async context on a current_thread Tokio executor; if the PythonShell owner drives process_command() inside a LocalSet task rather than spawn_blocking, any Python code calling cat/ls/write/ctl will panic the process.
> Caller runs process_command() inside a tokio::task::LocalSet (the documented way to host !Send types in async code); Python user calls `cat('/config/temperature')` which invokes vfs_async → Handle::current().block_on() → panic: 'Cannot start a runtime from within a runtime'.

**5. [NIT]** `deny.toml:174`
The comment justifying LGPL-3.0 allowance states malachite is 'dynamically linked', but Rust crates are always statically linked; LGPL-3.0 static-linking has additional user-relinking obligations (shipping object files or a shared library) that differ from dynamic linking and may require legal review.
> Legal audit of the shipped binary finds malachite statically compiled in with an LGPL-3 comment claiming dynamic linkage; the 'no copyleft propagation when dynamically linked' justification does not apply, and users cannot substitute the malachite library as required by LGPL-3.0 Section 4.

ewindisch:
## Agentic Code Review

**Overall:** The PR introduces a useful RustPython VFS shell but has a fundamental blocking bug: the synchronous-to-async bridge in the VFS builtins uses Handle::block_on from inside an async tokio context (the !Send LocalSet environment that PythonShell requires), which will panic at runtime for any VFS builtin call. Additionally, the interpreter scope is re-created fresh on every eval/exec call, meaning no state persists between calls, making the /vars and /defs VFS paths permanently empty and breaking the core shell use case.

### Findings

**1. [BUG]** `crates/hyprstream-python/src/builtins.rs:1742`
Handle::block_on panics/deadlocks when called from inside an async tokio context (the expected execution environment for PythonShell).
> When the owner thread is a tokio worker thread or is driving a LocalSet (as required by the !Send constraint), calling `h.block_on(f(ns, sub))` panics with 'Cannot block the current thread from within an asynchronous context'. Any Python builtin that touches the VFS (cat, ls, write, ctl, mount) will panic at runtime. The TCL builtins use a different, correct mechanism (direct async calls from the async owner loop).\n\n**2. [CORRECTNESS]** `crates/hyprstream-python/src/lib.rs:2054`\nEach eval/exec/list_vars/get_var call creates a fresh scope via new_scope_with_builtins(), so the interpreter has no persistent state between calls.\n> A caller does shell.exec("x = 42") then shell.eval("x"); the second call creates a new empty scope and returns NameError because the variable was stored in the ephemeral scope of the first call. The VFS mount's vars/ and defs/ directories will always be empty, making the /lang/python/vars/ and /lang/python/defs/ VFS paths useless.\n\n**3. [SECURITY]** `crates/hyprstream-python/src/lib.rs:2036`\nSHELL_CTX thread-local is overwritten on PythonShell::new() but never cleared on drop, leaking the previous subject/namespace into subsequent shells on the same thread.\n> If PythonShell A (subject=admin) is created, then dropped, then PythonShell B (subject=guest) is created but drops before SHELL_CTX is overwritten (e.g., due to a panic in Interpreter::with_init), the stale admin subject remains in the thread-local. Any VFS builtins called on B or a future shell would run under the admin identity.\n\n**4. [BUG]** `crates/hyprstream-python/src/lib.rs:2226`\ndrain_stdout_capture sets sys.stdout = None; if setup_stdout_capture subsequently fails, all print() calls for the session will raise AttributeError.\n> If the Python stdlib is in a state where CAPTURE_SETUP fails (e.g., after a previous exec that mutated sys), the next eval/exec will call setup_stdout_capture, get active=false, run the user code with sys.stdout=None, any print() call will raise AttributeError: 'NoneType' object has no attribute 'write', masking the actual user code error with an unrelated traceback.\n\n**5. [SECURITY]** `crates/hyprstream-python/src/lib.rs:2041`\nNo CPU/instruction limit is set on the RustPython interpreter, allowing untrusted Python code to loop forever and pin the owner thread.\n> User submits write('/lang/python/stdout', 'while True: pass'); the interpreter thread blocks indefinitely, starving the PythonMount channel of responses. All subsequent VFS calls to /lang/python/ will block waiting for a channel response that never arrives, effectively DoS-ing the mount.\n\n**6. [SECURITY]** `crates/hyprstream-python/src/builtins.rs:1693`\n__import__ is removed but Python code can still reach import machinery via importlib, which is frozen into the stdlib, enabling module imports after hardening.\n> After hardening, Python code can execute: import importlib; os = importlib.import_module('os'); os.system('id'). Since RustPython is compiled with freeze-stdlib, importlib is already present in the frozen module cache and does not require __import__ to be available in builtins.\n\n**7. [NIT]** `deny.toml:2870`\nThe LGPL-3.0 exception comment incorrectly claims malachite is dynamically linked; Rust crates are always statically linked, so the stated legal justification is factually wrong.\n> The exception is likely still valid (AGPL-3.0 is copyleft-compatible with LGPL-3.0 and satisfies relinking requirements), but the written rationale 'can be used without copyleft propagation when dynamically linked' does not apply to a statically linked Rust binary and could confuse future license audits.

ewindisch:
## ✅ High-confidence review fixes applied

The blocking correctness/safety findings from review have been fixed and pushed to `ewindisch/python-shell-483`. Two commits:

**1. `cca173f` — fix(#483): run VFS builtins via dedicated bridge runtime, not `block_on`**
- **Resolves the `block_on` panic.** The VFS builtins (`cat`/`ls`/`write`/`ctl`/`mount`) called `tokio::runtime::Handle::current().block_on(...)`. The interpreter is driven from inside the owner loop's tokio runtime, so re-entering a runtime on that thread panicked with *"Cannot start a runtime from within a runtime"* — i.e. **every VFS builtin paniced at runtime.** RustPython builtins are synchronous, so we can't mirror Molt's direct-await. Now `vfs_async` clones the `Send + Sync` namespace handle + subject out of the thread-local and `spawn`s the future onto a dedicated bridge runtime, blocking the owner thread on a plain channel. No `block_on`/`block_in_place` on the owner thread → no nested-runtime panic regardless of runtime flavor.

**2. `b10bfa8` — fix(#483): persist interpreter state, harden stdout capture, add exec watchdog**
- **Scope persistence (`/vars`, `/defs` were always empty).** Each `eval`/`exec`/`list_vars`/`get_var` created a fresh `new_scope_with_builtins()`, so no state survived between calls. Now a persistent globals `PyDictRef` is created once and reused via `Scope::with_builtins`; the `vars/`/`defs/` accessors read that same dict. State survives across calls.
- **Stdout capture hardened (no `None` → `AttributeError`).** `drain_stdout_capture` previously set `sys.stdout = None`; if a later `setup_stdout_capture` failed, every `print()` raised `AttributeError`. Drain now restores a real no-op sink (never `None`), a safe sink baseline is installed at construction, and setup restores the sink on failure.
- **Exec watchdog (CPU/DoS guard).** No instruction budget existed — `while True: pass` pinned the owner thread and DoSed the mount. RustPython 0.3 exposes no instruction quota (only `recursion_limit`, now lowered for guest code), so a **wall-clock** watchdog was added: `check_signals()` runs every bytecode op, so a watchdog thread injects a "time limit exceeded" interrupt via the user signal channel. Timeout is configurable (`set_exec_timeout`); per-call arming prevents a stale interrupt from corrupting the next call. *Documented limitation: code blocked inside a VFS call is not interrupted.*

**Validation:** 20 tests pass (including regressions for all four findings), `clippy -D warnings` clean.

One item remains — a security-posture decision on the import-sandbox escape — raised separately below. 👇

ewindisch:
## 🔶 Decision required (human sign-off — sandbox posture)

The functional fixes are in, but the **isolation posture is a human call** and shouldn't be merged on autopilot. The docs were softened to "best-effort, NOT a hard sandbox," but soft docs are not a security decision. Please answer explicitly:

**Q1 — Is the soft posture acceptable to merge?**
Is "best-effort, not a hard sandbox" acceptable for merge — with the **import escape** (`import os; os.system(...)` works; `__import__` removal is bypassed by `importlib`/`sys.modules`) and **no hard CPU quota** (only the wall-clock watchdog) documented as known gaps — OR must real isolation land *before* this merges?

**Q2 — If real isolation is required, which mechanism?**
- (a) Strip `sys.modules` / `importlib` and freeze a minimal stdlib so no host modules are reachable;
- (b) Run the interpreter in a separate locked-down process / VM (OS-level boundary);
- (c) Block at the VFS layer (deny host syscalls regardless of what Python imports).

Note: RustPython 0.3 has **no instruction-quota** hook — only the wall-clock watchdog we added — so an in-process instruction budget is not an option here.

**Q3 — Until then, gate it off?**
Should `hyprstream-python` be **feature-gated / disabled-by-default** so the soft sandbox isn't exposed unintentionally while the posture decision is pending?

crates/hyprstream-python/src/builtins.rs
```diff
@@ -0,0 +1,360 @@
...
+/// `builtins.__import__`), but it does NOT block modules already cached in
+/// `sys.modules` at startup, nor `importlib`-driven imports (importlib's
+/// bootstrap machinery does not go through `builtins.__import__`).
+///
+/// The amount of isolation we promise is a security-posture decision still
+/// pending review (see crate-level docs). Do not treat this list as a sandbox.
+pub(crate) fn harden(vm: &VirtualMachine) {
+    let dict = vm.builtins.dict();
+    for name in &["open", "exec", "compile", "breakpoint", "__import__"] {
+        let _ = dict.del_item(*name, vm);
```
ewindisch, line 43: **Import-sandbox escape — not actually blocked (docs softened to match).**

Removing `__import__` here only blocks *fresh* `import foo` statements that look up `builtins.__import__`. It does **not** block:
- modules already cached in `sys.modules` at startup, nor
- `importlib`-driven imports — RustPython's import bytecode + frozen `importlib` bootstrap machinery do not route through `builtins.__import__`.

Net effect: `import os; os.system(...)` still works, so this list is **not** a security boundary. The crate previously over-claimed "all I/O goes through the VFS"; the docs (crate-level + this `harden()` doc comment) were softened to "best-effort, NOT a hard sandbox" with the import escape + no-CPU-quota called out as known gaps. No code behavior was silently changed — the actual isolation posture is a human sign-off (see the "🔶 Decision required" comment on the PR).

ewindisch:
## 🟥 Disposition — Q1–Q3 resolved by the #483 redirect; converting to draft

The "decision required (sandbox posture)" questions above are now **answered by the [#483 WASM-capability decision](https://github.com/hyprstream/hyprstream/issues/483#issuecomment-4813721527)**, so they are closed here:

- **Q1 — is the soft posture acceptable to merge?** **No.** A best-effort native sandbox does not merge as the security story. The import escape and the absence of a hard CPU quota are not "documented gaps" we ship — they are why the substrate changes.
- **Q2 — which isolation mechanism?** **(b)-flavored, via WASM.** RustPython runs as a `wasm32-unknown-unknown` guest under an embedded wasmtime host with **zero ambient authority**; the only injected capability is the VFS host-fns, Subject-scoped. `import os` becomes inert by construction (so option (a) stdlib-stripping is demoted to defense-in-depth, and (c) is automatic — there is no host syscall to deny). DoS is bounded by wasmtime **epoch interruption** (fuel for deterministic tests), replacing the wall-clock watchdog.
- **Q3 — gate it off meanwhile?** **Moot — not merging.** Rather than ship a feature-gated soft sandbox, this PR becomes a draft holding the reusable mount/scope/stdout surface that the wasm path (P2) ports. Nothing exposes `/lang/python` until the wasm host lands.

### Actions taken
- Converted this PR to **draft** and rewrote the description to reflect the re-scope.
- The native fixes (`cca173f`, `b10bfa8`) stay on-branch — the bridge-runtime/scope/stdout work informs the guest port; the `__import__`-removal and `SHELL_CTX` paths are explicitly discarded as security mechanisms.

### What unblocks this PR
This rides **[#505](https://github.com/hyprstream/hyprstream/issues/505)** (the shared `hyprstream-wasm` substrate). Sequencing: **#505 P0** (build-viability spike — RustPython→wasm32 under embedded wasmtime + host RNG; in flight on branch `ewindisch/505-wasm-substrate-p0`) → **#505 P1** (host crate: capability Linker + Subject binding + epoch) → **this PR re-cut as #483 P2** (RustPython guest under Profile A + re-expose `/lang/python`). Full sequencing in the [epic #341 pipeline comment](https://github.com/hyprstream/hyprstream/issues/341).

ewindisch:
## Closing — superseded by the WASM-capability path. We are not shipping native in-process Python.

This PR's artifact (`crates/hyprstream-python`, native `rustpython-vm` in-process) is **discarded**, not
transformed — the native interpreter shares the host's authority, so its sandbox is unfixable-in-kind
(import escape; confirmed in review). The wasm work lives in **different crates on different branches**,
so no further commits will land here. Closing rather than leaving a zombie draft.

**The wasm-sandboxed `/lang/python/` shell is reframed as #483 P2**, built on the substrate:
- #576 — #505 **P0** (capability sandbox viable: RustPython→wasm32, zero ambient authority).
- #580 — #505 **P1** (host crate runs Subject-scoped Python; `print(1+1)`→0; epoch wall-clock bound).
- **#483 P2** (new PR, off the P1 branch) — Profile-A guest + re-expose `/lang/python`.

### Ported into #483 P2 (the reusable surface from this branch)
- `/lang/python/` mount layout — `eval` / `vars/` / `defs/` / `stdout` (from `crates/hyprstream-python/src/mount.rs`)
- persistent interpreter scope behavior
- the `__Capture` pure-Python stdout-capture trick
- the `vfs_*` builtin **shape** (`cat`/`ls`/`write`/`ctl`) — re-expressed as wasm **host-fn imports** under Profile A

### Explicitly dropped (do NOT carry forward)
- native `rustpython-vm` dependency + the tokio bridge runtime
- `builtins.harden()` `__import__`-removal as a security control (kept only as defense-in-depth via minimal frozen stdlib)
- the `SHELL_CTX` thread-local (the cross-tenant leak) → replaced by per-instance `spawn_vfs_proxy(ns, subject)` Subject binding (already in P1's `SandboxState`)
- the wall-clock `check_signals` watchdog → replaced by wasmtime epoch interruption (already in P1's `EpochTimer`)

Issue **#483 stays open** and is delivered by P2. Sequencing: epic #341.

PR #489, opened 2026-06-26
feat(#351): image size calculation + warm pool replenishment

### LINKED ISSUE #351: T3-C: warm-pool replenish, image-size calc, stats
Track 3. Finish warm-pool replenish (pool.rs:272), image-size calc (store.rs:430/576), and stubbed stats (service.rs:708/748). Part of #341. Plan: docs/plans/2026-06-19-worker-ga-epic.md

ewindisch:
## Summary

- **`RafsStore::calculate_image_size()`**: sums on-disk blob sizes for each layer digest recorded in `ImageMetadata`; removes the `size: 0` stub from `list()` and `image_status()`
- **`SandboxPool::replenish_warm_pool()`**: spawns a `tokio::spawn` task after each `acquire()` to top up the warm pool to its configured target size; changes `acquire()` receiver from `&self` to `self: &Arc<Self>` to allow moving a clone into the background task
- **Clippy fix**: adds `expect_used` to the `#[allow]` on the pre-existing `streaming.rs` test

Closes #351.

## Test plan

- [ ] `cargo clippy --workspace --all-targets` passes
- [ ] `cargo check -p hyprstream-workers` passes
- [ ] `kata-vm` feature-gated tests compile (test helper now returns `Arc<SandboxPool>`)
- [ ] Image size is non-zero for a pulled image (layers present on disk)
- [ ] Warm pool count stays at target after `acquire()` bursts


ewindisch:
## Agentic Code Review

**Overall:** The two feature additions (image size calculation and warm pool replenishment) are directionally correct and remove long-standing TODOs, but both have correctness issues: replenish_warm_pool has a TOCTOU race that under- or over-shoots target size under concurrent load and is not triggered by the release() non-reusable path, while calculate_image_size performs blocking filesystem I/O on the async executor thread without spawn_blocking.

### Findings

**1. [CORRECTNESS]** `crates/hyprstream-workers/src/runtime/pool.rs:271`
replenish_warm_pool spawns exactly one create_warm_sandbox() task per acquire(), regardless of how far below target the pool is; concurrent acquires introduce a TOCTOU race between the warm_count check and push_back that can both under-fill and overshoot the target.
> warm_pool_size=5, five concurrent acquire() calls drain the pool to 0; each spawns one replenish task. All five tasks race: they each lock warm_pool, read 0 (or 1, 2, …), then unlock, create a sandbox (slow), re-lock, and push_back. Depending on scheduling, between 1 and 5 new sandboxes are created — a burst of acquires can leave the pool at 1 warm slot, or push it to 5 (correct), or overshoot to 10 if multiple rounds of acquires interleave; the invariant 'warm_count == target after settle' is not guaranteed.

**2. [CORRECTNESS]** `crates/hyprstream-workers/src/runtime/pool.rs:113`
release() does not call replenish_warm_pool() when the backend returns reusable=false and destroys the sandbox, so the warm pool can permanently fall below its target size after a reset failure.
> Pool has warm_pool_size=3. The backend's reset() returns false (ephemeral sandbox). After release(), the warm pool is one slot short. No subsequent replenishment is scheduled unless another acquire() fires, so the pool stays under-size indefinitely under quiescent load.

**3. [CORRECTNESS]** `crates/hyprstream-workers/src/image/store.rs:602`
calculate_image_size() calls std::fs::metadata() in a synchronous blocking loop and is invoked from async fns list() and image_status() without spawn_blocking, blocking the tokio executor thread.
> list() is called on a system with 50 images each having 10 layers; 500 synchronous stat() calls execute on the async executor thread, stalling all other tasks scheduled on that thread for the duration of the I/O, degrading throughput or causing task starvation in production.

**4. [NIT]** `crates/hyprstream-workers/src/image/store.rs:466`
calculate_image_size() silently returns 0 for any layer blob whose path is missing (e.g., file not yet downloaded, partial pull, or GC'd blob), so the reported image size can be a silent undercount with no warning log.
> An image is partially pulled (some layer blobs missing) or a blob is removed by GC while list() runs; the returned ImageInfo.size reflects only the layers present on disk, with no indication to the caller that the sum is incomplete.

ewindisch:
## Agentic Code Review

**Overall:** The PR delivers the intended functionality (non-zero image sizes, a real warm-pool replenishment task) but has two notable gaps: the `list()` method (the CRI ListImages path) was not updated with `calculate_image_size`, leaving a `size: 0` inconsistency with `image_status()`; and the replenishment logic has a TOCTOU race that can overshoot the warm-pool target under concurrent load, as well as a slow-recovery problem when the pool is fully drained. No security-sensitive areas (capnp schema, auth, WASM, crypto) are touched by this PR.

### Findings

**1. [BUG]** `crates/hyprstream-workers/src/image/store.rs:430`
`list()` still has `size: 0` — the `calculate_image_size` fix was only applied to `image_status()` and one other call site, but `list()` (called by the public `list_images()` CRI method) was not updated, so CRI image listing always reports zero-byte images.
> A caller invokes `list_images()` (or the CRI ListImages RPC) after pulling an image; the returned `ImageInfo.size` is always 0 even though blobs are on disk. The fix to `image_status()` is inconsistent with this code path.

**2. [CORRECTNESS]** `crates/hyprstream-workers/src/image/store.rs:600`
`calculate_image_size` calls `std::fs::metadata` (blocking syscall) inside an async method without `spawn_blocking`, risking tokio thread-pool starvation on large images with many layers.
> An image with 30+ layers is listed or status-checked while the thread pool is busy; the synchronous `fs::metadata` loop holds the async thread for tens of milliseconds per call, degrading throughput and potentially starving other tasks.

**3. [CORRECTNESS]** `crates/hyprstream-workers/src/image/store.rs:604`
`calculate_image_size` silently returns 0 for any missing or unreadable layer blob via `.unwrap_or(0)`, masking integrity errors and producing a silently understated size.
> A layer blob is missing from `blobs_dir` (e.g., interrupted pull, corrupt store); `calculate_image_size` counts that layer as 0 bytes with no warning, so the reported image size is wrong and the inconsistency is invisible to callers.

**4. [NIT]** `crates/hyprstream-workers/src/image/store.rs:598`
`calculate_image_size` sums only the layer digests from `metadata.layers`, omitting the config blob (`metadata.config_digest`), so the reported size is smaller than the actual on-disk footprint and differs from Docker/containerd semantics.
> A CRI client (e.g., kubelet image garbage-collection) computes eviction thresholds using the reported image size; the config blob (typically a few KB) is excluded, making the aggregate disk accounting slightly wrong.

**5. [CORRECTNESS]** `crates/hyprstream-workers/src/runtime/pool.rs:279`
`replenish_warm_pool` spawns a task that creates at most one sandbox regardless of how far below the target the pool is, so a burst of N concurrent `acquire()` calls that exhausts the pool will take N-1 subsequent acquire cycles to recover.
> Warm pool is configured with `warm_pool_size = 4`; 4 rapid `acquire()` calls drain it to 0; each spawns one replenishment task, each adding one sandbox — the pool refills at rate 1 per acquire instead of jumping back to the target immediately, increasing cold-start latency for subsequent requests.

**6. [CORRECTNESS]** `crates/hyprstream-workers/src/runtime/pool.rs:279`
TOCTOU race in `replenish_warm_pool`: warm count is read and then the Mutex is dropped before the new sandbox is pushed, so N concurrent tasks can each observe `warm_count < target` and all create a sandbox, causing the warm pool to overshoot its target by up to N.
> N concurrent `acquire()` calls each spawn a replenishment task; all N tasks lock, read `warm_count = 0`, unlock, then each calls `create_warm_sandbox()`, and all N push a sandbox; the warm pool ends up with N entries when the target is 1, leaking N-1 warm sandboxes as idle resource consumption.

Issue #505, opened 2026-06-26
feat(workers): hyprstream-wasm embedded wasmtime substrate + BackendType::Wasm


ewindisch:
Part of epic #341 (Worker GA). Spike + design: `docs/plans/2026-06-26-wasm-wasi-worker-convergence.md` (branch `ewindisch/wasm-worker-convergence-spike`).

## Summary
Add a shared embedded **wasmtime** substrate crate `hyprstream-wasm` and a new `BackendType::Wasm` worker backend — a native sibling to Kata/OCI under the existing `SandboxBackend` seam (RuntimeClass-selected). One runtime serves multiple consumers, each with its **filesystem backed by the Subject-scoped hyprstream 9P/VFS** (not the host).

This is the convergence substrate: it is consumed by **#483** (WASM-Python shell) and (b) **Wanix-in-Workers**, and long-term subsumes **#465**'s browser WanixMount path.

## Why embedded wasmtime, not runwasi
runwasi rides the CRI plane but hands the guest containerd's host-fs WASI — which fights the VFS-backed-filesystem thesis. Embedded wasmtime lets us back the guest's filesystem with our own `Mount`. So wasm is a **native sibling `BackendType`**, orthogonal to the OciBackend/CriBackend decision (does not ride CRI). It also sidesteps the Kata guest-agent gap (#344) entirely — execution is in-process.

## Design: one configurable Linker, two profiles
- **Profile A — no WASI:** zero ambient authority; the only imports are VFS host-fns. Used by #483 (Python).
- **Profile B — WASI, fs = hyprstream 9P/VFS Mount:** WASI guest whose 'filesystem' is the Subject-scoped VFS, not the host. Used by Wanix (b).

Both bind the actor's atproto **Subject** at instantiation (Casbin-scoped via the existing `Mount` `caller: &Subject` seam) and are bounded by wasmtime **epoch interruption** (fuel for deterministic tests).

## Hard caveats (design lives or dies by these)
1. **Make-or-break P0 unknown:** which wasmtime WASI API lets a preopen be backed by an arbitrary `Mount` (`wasi-common` `WasiDir` vs the component model). Validate before committing Profile B.
2. **Profile A: zero WASI grants** — one preopen reopens the escape.
3. **Epoch/fuel bound** — WASM alone does not stop infinite loops.

## Notes
- `wasmtime` is a **new dependency** (only wasm-bindgen browser tooling exists today).
[...]

PR #576, opened 2026-06-29
spike(#505): wasm RustPython capability-sandbox P0 (build-viability)

ewindisch:
> **Draft — #505 P0 build-viability spike. Not for merge.** Committed `--no-verify` (the
> pre-commit hook runs full-workspace clippy, which needs LIBTORCH); the spike opts out of the
> workspace panic/print lints; and the committed 0.3 guest *traps at runtime* (see getrandom below).
> Purpose: validate the wasm capability-sandbox thesis so P1 can proceed on solid ground.

## Result: VIABLE ✅

Two new crates prove RustPython can run as a `wasm32-unknown-unknown` guest under embedded wasmtime
with **zero ambient authority** — the only import the host grants is `host_random`; every other
import is wired to a trap (`Linker::define_unknown_imports_as_traps`). No `wasmtime-wasi`, no
`add_to_linker`, no preopens. So `import os; os.system(...)` is **inert by construction** — there is
no host syscall for it to reach.

- **`crates/hyprstream-wasm-pyguest`** — cdylib guest (workspace-`exclude`d; built only for wasm32).
  RustPython 0.3 `freeze-stdlib`, an `eval(ptr,len)->i32` entrypoint, `getrandom` routed to the host import.
- **`crates/hyprstream-wasm`** — workspace member, **no torch dep** (builds without LIBTORCH).
  `wasmtime = "=46.0.1"`, a bespoke capability `Linker`, fuel + epoch DoS bounds. 5 host tests
  (`cargo test -p hyprstream-wasm`; they skip gracefully if the guest `.wasm` isn't built first).

## Pinned unknowns (full detail in `crates/hyprstream-wasm/FINDINGS.md`)

- **getrandom (the #1 risk) — resolved.** `rustpython-vm` **0.3.1 force-enables `getrandom/js`**
  (its Cargo.toml, not target-gated); Cargo feature-unification means a downstream crate cannot turn
  it off, so on `wasm32-unknown-unknown` the JS backend out-ranks our registered custom backend and
  the VM traps at `VirtualMachine::new` reaching the undefined `__wbindgen_placeholder__` JS imports.
  *(That trap is itself a capability proof — the guest reaches for something the host never granted
  and is stopped.)* **Clean green path = `rustpython-vm` 0.5** (declares `getrandom 0.3`, no `js`) +
  the getrandom-0.3 custom backend; `wasm-tools print` confirmed the 0.5 guest has **exactly one
  import — `(import "env" "host_random" …)`** — zero JS, zero WASI, zero wasm-bindgen.
- **wasmtime 46.0.1 API:** `Config::{epoch_interruption,consume_fuel}(true)`; `Store::set_fuel(u64)`
  / `Store::set_epoch_deadline(u64)` + `Engine::increment_epoch()`. Gotcha: with
  `epoch_interruption(true)` every store defaults to deadline **0** → the non-tested limiter must be
  pushed to `u64::MAX` or it traps immediately. (wasmtime 46 ships its own `Result`/`bail!`.)
- **Profile B (Wanix, later):** back a WASI preopen with an arbitrary `Mount` via **preview1
  `wasi-common` `WasiDir`/`WasiFile`**, *not* preview2 `wasmtime-wasi` (which only takes a `cap-std`
  host `Dir`). Not needed for the Python guest.
- **frozen-stdlib:** a single boolean feature (no per-module exclusion); the defense-in-depth lever
  is **registration-time** (`Interpreter::builder` registering only chosen native module defs), not
  feature-time. The capability guarantee does not depend on it.

## First P1 step

Bump the guest to `rustpython-vm = "0.5"` (drop the 0.2 dual-backend; keep only the getrandom-0.3
custom backend) and construct via `Interpreter::builder(settings)` registering the frozen stdlib so
`print(1+1)` returns status 0. Then carry a per-call `Subject` in `Store` data (Subject-scoped host
fns) and add the `Engine::increment_epoch()` timer thread for a wall-clock DoS bound, keeping
`define_unknown_imports_as_traps` as the deny-by-default fence. #483 (Python, Profile A) and #505 P2
(`BackendType::Wasm`) then ride P1.

Refs #505 · epic #341 · unblocks the #488/#483 redirect.


ewindisch:
Superseded by #596 — consolidated into one reviewable, CI-enforced `hyprstream-sandbox` crate during the #341 quality reset (rename hyprstream-wasm→hyprstream-sandbox, decouple core from Python, de-slop, + a CI job that builds the guests and RUNS the security/DoS tests fail-not-skip). Security model re-audited as behavior-preserving. Closing in favor of #596 (history cross-linked there).

PR #580, opened 2026-06-29
feat(#505): P1 wasm host — rustpython 0.5 green path, Subject-scoped Store, epoch bound

ewindisch:
> **Draft — #505 P1.** Stacked on the P0 spike (#576). Turns the build-viability spike into a host
> crate that actually **runs** Subject-scoped Python under a real wall-clock bound. Committed
> `--no-verify` (pre-commit clippy needs LIBTORCH); scoped builds + 9 tests green.

## What P1 delivers

**1. Green path — the substrate runs real Python.** Guest bumped to `rustpython-vm = "0.5"` (+
`rustpython-pylib 0.5` `freeze-stdlib`); getrandom-0.2 dual-backend dropped, **getrandom 0.3 custom
backend only**. `Sandbox::eval("print(1 + 1)")` now returns status **0** (no trap), and `wasm-tools
print` confirms the guest's imports are **exactly `{ env::host_random }`** — zero JS/WASI/wasm-bindgen.
`case2` (`import os; os.system('echo PWNED')`) returns a nonzero Python-error status — never 0, never a
host process (no native posix backend is wired).

- The exact 0.5 embedded-frozen-stdlib construction (non-obvious — `without_stdlib` panics under
  `freeze-stdlib`):
  ```rust
  vm::Interpreter::builder(vm::Settings::default())
      .add_frozen_modules(rustpython_pylib::FROZEN_STDLIB)
      .build()
  ```
  with guest features `["compiler","encodings","freeze-stdlib","stdio"]` — `stdio` *without* `host_env`
  selects the VM's sandbox stdio path (discards on wasm32-unknown-unknown: capability-clean). Detail in
  `FINDINGS.md` → "P1 RESOLUTIONS".

**2. Subject-scoped Store (kills the #488 cross-tenant bug).** Local `Subject(Option<String>)` newtype
(no rpc/vfs dep yet); P0's fixed `HostState` → `SandboxState { subject, rng }` carried as Store data;
`host_random` reads `caller.data().subject`. A `Sandbox` binds one Subject at construction
(`from_bytes_for`) and each `eval` builds a fresh Store from it — **no global/thread-local state**.
`subject_isolation_no_leak` (alice vs bob) passes.

**3. Real wall-clock DoS bound.** `EpochTimer::spawn(engine, tick)` runs an `increment_epoch()` thread
(stopped on Drop); `eval_with_epoch_deadline` sets the epoch as the production limiter (fuel →
[...]

Issue #506, opened 2026-06-26
feat(workers): native Wanix as a worker workload — VFS Mount as its 9P root (Direction A)


ewindisch:
Part of epic #341 (Worker GA) / #508 (sandbox taxonomy).

> **⚠️ Premise corrected (spike 2026-07-03).** This ticket originally proposed running Wanix as a **`BackendType::Wasm` / WASI Profile-B** guest whose fs is a hyprstream VFS `Mount`. A spike against `tractordev/wanix` main falsified that: **Wanix's kernel is hard build-tagged `js && wasm`** (`wasm/wasm.go`) — it targets the browser JS host (`syscall/js`: OPFS, DOM, fetch, web workers) and is built via `GOOS=js GOARCH=wasm` / TinyGo `-target wasm`. **There is no `wasip1` target for it and no path to one** (its value depends on the JS host). It cannot run under `hyprstream-workers-wasmtime`'s `WasiSandbox`. The #505 substrate is fine — Wanix was simply never a WASI guest. The corrected design is below.

## Summary (corrected — "Direction A")

Run **native Wanix** (the `!js && !wasm` Go binary, `cmd/wanix`) as an ordinary worker workload, scheduled through the normal `SandboxBackend` selection path (auto-select over kata→**oci**→nspawn), and hand it the tenant's **Subject-scoped hyprstream VFS `Mount` as its root filesystem over 9P**. This makes *"the VFS is the OS filesystem"* literally true for a Plan9 OS running in a Worker — same intent as before, viable mechanism.

## Use Wanix's EXISTING mount mechanism — do NOT invent one on our side

Wanix already has a complete, native mount facility. **We consume it; we do not add a preconceived hyprstream-side mechanism or a bespoke `wanix mount` verb.** Verified in the spike:

- **Bind IS the mechanism.** `NS().Bind(src fs.FS, srcPath, dstPath)` binds any `fs.FS` into the namespace. `#`-capabilities (`#ramfs`/`#web`/`#pipe`…) are just `fs.FS` values bound at `#name`. At runtime it's driven by writing control messages to **`#task/<id>/ctl`** (`bind SRC DST` / `unbind`), which is exactly what the rc `bind`/`unbind` builtins do (`rc/bind` → `WriteSelfCtl`). Pure Plan9.
[...]

PR #581, opened 2026-06-29
feat(#483): WASM /lang/python Profile A — real VFS capability + mount over the wasm guest

### LINKED ISSUE #483: feat(python): hyprstream-python — WASM-sandboxed RustPython VFS shell at /lang/python/

ewindisch:
> **Draft — #483 P2 (Profile A).** Stacked on P1 (#580). Turns the substrate into a real,
> Subject-scoped, VFS-capable `/lang/python/` shell — the #483 deliverable, reframed on WASM after
> #488 (native) was closed. Committed `--no-verify`; scoped builds + 13 tests green, clippy clean.

## What P2 delivers

**1. The VFS capability is REAL (not the P1 sketch).** `hyprstream-vfs` + `hyprstream-rpc` added to
`hyprstream-wasm` — **both verified torch-free**, so `cargo build -p hyprstream-wasm` stays
LIBTORCH-free (only the top-level `hyprstream` crate links libtorch). The P1 crate-local `Subject` is
replaced by the canonical `hyprstream_rpc::Subject`. `build_linker` now defines `env::vfs_cat / vfs_ls
/ vfs_echo / vfs_ctl`, each reading `caller.data().subject` and submitting via a
`spawn_vfs_proxy(ns, subject) -> Sender<VfsRequest>` handle in `SandboxState` (`None` = deny-by-default;
sync host fn `blocking_send` + blocks on the reply channel — the documented `// P1b/#483:` bridge).
`define_unknown_imports_as_traps` still fences everything else.
- **Subject-scoping proven end-to-end:** `guest_vfs_is_subject_scoped` — a *guest* call flows
  guest → `env::vfs_*` → Subject-scoped proxy → `Mount`; `alice` cats/echoes/reads `/config/temp`, a
  `denied` Subject is refused at the Mount. `guest_vfs_deny_by_default` — no `.with_vfs()` → "no
  capability", never touches a Namespace. Guest imports (wasm-tools) = exactly
  `env::{host_random, vfs_cat, vfs_ls, vfs_echo, vfs_ctl}` — no WASI, no JS.

**2. Guest `/lang/python` semantics (RustPython 0.5).** New guest export `py_op(op,ptr,len)->i64`:
eval→repr, exec→captured stdout, list_vars/list_defs, get_var/get_def. **Persistent** interpreter scope
across calls (a `thread_local Shell{interp, globals}`, globals reused via
`Scope::with_builtins(None, globals.clone(), vm)`). **stdout capture** = the native `__Capture`
pure-Python surrogate ported verbatim (+ a `__Sink` baseline so `print` never hits `None`-stdout).

**3. `/lang/python` mount re-exposed, torch-free.** `crates/hyprstream-wasm/src/mount.rs` —
`PythonMount` implements `hyprstream_vfs::Mount` with the exact native layout: `eval` (ctl expr→repr),
`stdout` (ctl stmts→stdout), `vars/`, `defs/`. Because the guest's `vfs_*` host fns `blocking_send`,
`PythonMount::spawn` runs the shell on a dedicated OS thread driven by a `PyCommand` channel (same
`!Send`-interpreter pattern as the native mount); owns+joins the driver on `Drop` (sender dropped first
— a drop-order deadlock found and fixed). `mount_eval_vars_defs_over_guest` passes against an in-memory
VFS (`2+3`→`5`; `x=42`→`vars/x`=`42`; `def f()`→`defs/f`).

## Deferred (documented, not blocking)
- **Daemon namespace wiring** — registering this mount into the running `/lang/python` lives in the
  torch-bound `hyprstream` crate (can't build here without LIBTORCH). `mount.rs` carries the exact
  `// #483 daemon wiring:` snippet; next step is to call it from the per-session Namespace builder
  (where `/lang/tcl` mounts) with the session's verified Subject + Namespace.
- **Guest Python `cat()`/`ls()` builtins** — the capability is real and tested via the host fns +
  exports + Rust helpers, but not yet registered as Python builtins callable from guest *source*
  (needs native-module registration omitted by the current feature set). Follow-up; doesn't affect the
  proven capability.

## Verification
`cargo test -p hyprstream-wasm` = 13 green (2 lib + 11 integration); clippy clean; guest builds to
wasm32-unknown-unknown with the 5-import set above. Changes confined to the two wasm crates.

Closes #483 (pending daemon wiring follow-up) · refs #505 · epic #341 · supersedes the closed #488 · stacked on #580.

### DIFF
--- a/.github/workflows/wasm-pyguest.yml
+++ b/.github/workflows/wasm-pyguest.yml
@@ -1,0 +1,94 @@
+name: WASM pyguest
+
+# Build the RustPython wasm guest (#483/#505) FROM ITS CRATE DIRECTORY and assert
+# its host ABI. This crate is [workspace.exclude] with its own empty [workspace], so
+# the `wasm` job in rust.yml (cargo check -p hyprstream-rpc) never compiles it.
+#
+# WHY a dedicated job: the guest's `getrandom_backend="custom"` rustflag lives in
+# crates/hyprstream-wasm-pyguest/.cargo/config.toml, which cargo only discovers by
+# walking up from the CWD — NOT from --manifest-path. Building from the repo root
+# (or with --manifest-path) silently drops the cfg and the wasm build fails on
+# getrandom. So this job MUST `cd` into the crate dir (working-directory below);
+# it deliberately does not use --manifest-path.
+on:
+  push:
+    branches: [ "main" ]
+    paths:
+      - 'crates/hyprstream-wasm-pyguest/**'
+      - '.github/workflows/wasm-pyguest.yml'
+  pull_request:
+    paths:
+      - 'crates/hyprstream-wasm-pyguest/**'
+      - '.github/workflows/wasm-pyguest.yml'
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  build-guest:
+    name: build wasm guest (from crate dir)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install Rust toolchain (wasm32)
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: wasm32-unknown-unknown
+
+      # wabt provides wasm-objdump, used below to read the guest's import section.
+      - name: Install wabt
+        run: sudo apt-get update && sudo apt-get install -y wabt
+
+      - name: Cache Rust dependencies
+        uses: Swatinem/rust-cache@v2
+        with:
+          # Standalone crate (own lockfile/target dir), so cache it on its own.
+          workspaces: crates/hyprstream-wasm-pyguest
+          key: wasm-pyguest
+
+      - name: Build guest from its crate dir
+        # MUST run from the crate dir so cargo discovers the crate-local
+        # .cargo/config.toml carrying getrandom_backend="custom". Do NOT switch this
+        # to --manifest-path from the repo root — the cfg would be dropped and the
+        # build would fail on getrandom/wasm32. See the crate's Cargo.toml.
+        working-directory: crates/hyprstream-wasm-pyguest
+        run: cargo build --release --target wasm32-unknown-unknown
+
+      - name: Assert guest imports are exactly the host ABI
+        # The guest must import ONLY the host capability functions from module `env`
+        # and nothing else — in particular no wasm_js / wasi / wbindgen backend may
+        # leak in (that would mean getrandom picked a non-custom backend, i.e. the
+        # cfg was dropped, or a dependency pulled a JS shim). #483/#505 capability
+        # boundary: the host resolves exactly these five imports via the wasmtime Linker.
+        working-directory: crates/hyprstream-wasm-pyguest
+        run: |
+          set -euo pipefail
+          WASM=$(find target/wasm32-unknown-unknown/release -maxdepth 1 -name '*.wasm' | head -1)
+          if [ -z "$WASM" ]; then echo "::error::no wasm artifact produced"; exit 1; fi
+          echo "Inspecting imports of: $WASM"
+
+          # wasm-objdump prints import entries as: "... <- env.host_random"
+          ACTUAL=$(wasm-objdump -x "$WASM" \
+            | grep -oE '<- [A-Za-z0-9_]+\.[A-Za-z0-9_]+' \
+            | sed -E 's/<- ([A-Za-z0-9_]+)\.([A-Za-z0-9_]+)/\1::\2/' \
+            | sort -u)
+          EXPECTED=$(printf '%s\n' \
+            env::host_random env::vfs_cat env::vfs_ls env::vfs_echo env::vfs_ctl \
+            | sort -u)
+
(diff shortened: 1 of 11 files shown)


ewindisch:
## Clean-build footgun fixed (commit `d9e2f082d`)
Follow-up to the #506-surfaced clean-build failure — root cause was **not** a getrandom version bug but a **CWD footgun**: the guest's `--cfg getrandom_backend="custom"` lives in `crates/hyprstream-wasm-pyguest/.cargo/config.toml`, which Cargo discovers only from the **current dir**, not via `--manifest-path`. So a root `--manifest-path` build silently dropped the cfg and getrandom 0.3.4 failed on `wasm32-unknown-unknown`.
- **Fix:** corrected every documented guest-build command to run **from the crate dir** (`(cd crates/hyprstream-wasm-pyguest && cargo build --release --target wasm32-unknown-unknown)`), in `Cargo.toml` / `src/lib.rs` / `tests/sandbox.rs` / `FINDINGS.md`, with an explicit note in `.cargo/config.toml`.
- **Verified clean:** `rm -rf target` → crate-dir build green in ~37s; artifact imports are **exactly** `env::{host_random, vfs_cat, vfs_ls, vfs_echo, vfs_ctl}` — no `wasm_js`/`wasi`/`__wbindgen*`.
- `getrandom 0.2.17` in the lockfile = harmless host build-dep of rustpython codegen (`unicode_names2_generator` build-deps), never compiled for wasm; annotated. The workspace builds clean regardless (pyguest is workspace-excluded).

Recommend a small CI step building the guest from its dir (~37s) + a wasm-tools import assertion so this can't regress silently — flagged to the operator/CI owner (not changed here).

ewindisch:
Superseded by #596 — consolidated into one reviewable, CI-enforced `hyprstream-sandbox` crate during the #341 quality reset (rename hyprstream-wasm→hyprstream-sandbox, decouple core from Python, de-slop, + a CI job that builds the guests and RUNS the security/DoS tests fail-not-skip). Security model re-audited as behavior-preserving. Closing in favor of #596 (history cross-linked there).

Issue #611, opened 2026-06-30
P3: TUI P9-alignment — replace poll_local spin with real async driver; decouple namespace from mandatory interpreter


ewindisch:
Phase P3 of the P9 Task/Instance Projection epic (#608).

## Goal

Make the TUI a correct consumer of the 9P namespace under event-backed blocking reads, and decouple the namespace from a mandatory interpreter so the instance fs (and any 9P path) is reachable directly.

## Problem 1 — the busy-spin driver (must fix before event-backed reads)

The TUI/ChatApp is the first real consumer of an event-backed blocking read. It currently drives all VFS reads via a noop-waker spin loop (`poll_local` in `crates/hyprstream-tui/src/chat_app.rs`) that polls the future and `yield_now`s on `Pending`, justified by "VFS futures are pure async, no IO reactor." That assumption breaks the moment a `read` blocks on a completion edge: a `read /exec/instances/<id>/exit` (or `load --wait`) would hot-spin a core until the terminal latch fires.

The latched-completion await primitive (#607) is, by contract, a normal awaitable Future that registers a waker with a real executor. P3 replaces the spin driver with a real async drive so blocking namespace reads park on a waker instead of burning CPU.

## Problem 2 — the namespace is only reachable through an interpreter

Slash commands route through the Tcl shell (`handle_vfs_command` evaluates the line as Tcl against the namespace; bare paths like `/srv/model/status` become `cat <path>`). The namespace is the real abstraction, but it is currently reachable only *through* a language interpreter. For the instance fs, a `read`/`write` on a path must work without requiring Tcl (or Python).

P3 decouples slash-command and LLM-tool routing from the mandatory interpreter: the namespace is reached directly via `walk`/`open`/`read`/`write`, with Tcl/Python as *optional* interpreters layered on top rather than the only path. This also lets `read /exec/instances/<id>/exit` work as a plain namespace read with the new async driver.

## Scope

- Replace `poll_local` with a real async drive for namespace operations (the ChatApp event loop / driver thread), so blocking reads register a waker.
- Route slash commands and tool execution against the `Namespace` directly; keep interpreters optional.
- Verify a blocking `read` on a completion file (e.g. an instance `exit`) parks on a waker and returns on the edge, with no busy-wait.

## Acceptance

- A blocking namespace read in the TUI registers a waker (no CPU spin); confirmed against an event-backed completion file.
- Slash commands and tools reach the namespace directly (`read`/`write`), not only via a language interpreter.
[...]

Issue #608, opened 2026-06-30
Epic: P9 Task/Instance Projection — per-instance lifecycle filesystem (ctl/status/exit/fd/ns) over the SandboxBackend seam, EventService-backed wait


ewindisch:
Tracking epic to project a **running execution unit as a Plan 9 `/proc`-style file tree** — a per-instance lifecycle filesystem (`ctl`, `status`, `exit`, `fd/`, `ns`, `cap/`) that is uniform across every isolation backend (in-process wasm, nspawn, OCI, CRI, Kata VM). This is the file surface that makes a *running task* addressable, controllable, and waitable through the same 9P/VFS namespace as everything else — with `hyprstream-workers` as the scheduling/management entrypoint.

## Why this is its own epic

The surrounding pieces already exist as separate efforts, but the **per-instance lifecycle file surface itself is unbuilt**:

- **#539 (9P-RPC)** generates the 9P/VFS projection for *capnp services* (`/srv/<name>`) — a **service** face, not a **running-instance** `/proc` face.
- **#508 (sandbox backend taxonomy)** defines *which backends exist* (Kata/Oci/Cri/Wasm/Nspawn under one `SandboxBackend` seam); **#341 (Worker GA)** makes them run workloads. Neither projects a running instance as a waitable directory.
- **#600/#607 (EventService EV7)** defines the *latched-completion / await primitive* (retained terminal state + ephemeral streams) but has no task-fs *consumer*.
- **#547 (Native MAC)** makes the generated mount a reference monitor, but namespace-mutation-as-delegation for a task's own namespace is not yet wired.

This epic is the **integration layer** that ties those together into one addressable, waitable, authorized per-instance file surface. It is orthogonal to "which backends work" (#341/#508) and depends on the completion primitive (#607) and the reference monitor (#547) — a different dependency graph from Worker GA, and it spans more crates (`hyprstream-vfs`, the TUI, ModelService, the Wanix guest), so it is tracked separately rather than folded into #341.

## Core model

A managed execution unit is projected as a Plan 9 `/proc`-style directory. The division of labor is strict:

- **9P expresses identity, addressing, and the control surface.** Per-instance directory:
  ```
  /exec/instances/<id>/
    ctl       # write verbs: start | kill | signal <n> | bind <src> <dst> | mount <svc> <dst>
    status    # read: live, NON-latched state (backend, pid(s), resource use) — for pollers
[...]

Issue #607, opened 2026-06-30
EV7: Latched completion/lifecycle events — retained terminal state + ephemeral streams


ewindisch:
Phase EV7 of the EventService consolidation epic (#600). Coordinated with the workers/P9-task design. Gates EV5's `load --wait` port.

## Goal
First-class fan-out completion (>0 waiters) with **retained terminal state**, so a watcher that arrives after a resource terminates is still correct. Subsumes `load --wait` and P9 `/task/<id>/{exit,status,fd,signal}` under one pattern.

## Scope
1. **The addressable resource FILE holds the authoritative retained terminal value** (host-side, plaintext-at-owner) — NOT the content-blind relay (which for confidential events holds only epoch-bounded ciphertext and cannot project a plaintext last-value), and NOT `StreamOpt::Retention` (a separate best-effort transport buffer). The per-resource **ephemeral stream** delivers the edge.
2. **One shared read-then-subscribe helper:** `read` returns the retained terminal value if present, else subscribes to the ephemeral stream and blocks. Late reads enforce confidentiality + authz via an **authorized file read** (same MAC as subscribe, EV4); FS-safe (no cross-epoch ciphertext replay).
3. **Terminal = monotonic write-once.** Per-resource-type definitions are owned by workers/P9-task (accepted as their contract): task exit = `Exited(code)|Killed(sig)|Failed(reason)`; model load = `Loaded(handle)|LoadFailed` latched **per load-attempt/epoch** (reload = a new latch, preserving write-once); fd = terminal is **only** the close/EOF marker (content rides `StreamOpt::Retention`; latch ≠ retention; aligns with #163 terminal-block).
4. **Guaranteed-eventual latch:** every terminal-latched read has a path that latches a terminal value on **abnormal owner death** (reaper/timeout → `Failed`/EOF), so a fan-out read never blocks forever. This is an EventService/EV6-liveness obligation (death-detection → latch), not the owner's best effort.
5. **Await contract:** the blocking-read primitive MUST be a normal awaitable `Future` that **registers a waker** with a real executor — never correct-only-under-poll-spin. Known incompatible consumer (being fixed in P9-task): the TUI/ChatApp drives VFS reads via `poll_local` noop-waker spin loop (`crates/hyprstream-tui/src/chat_app.rs`), which would hot-spin a core on an EV7 blocking read.

## Acceptance
- A watcher that subscribes after termination still gets the terminal value (file-served); a watcher before termination blocks then receives the edge.
[...]

Issue #612, opened 2026-06-30
P4: Wanix guest /task fs bridged to host /exec/instances projection (SandboxBackend + completion + MAC)


ewindisch:
Phase P4 of the P9 Task/Instance Projection epic (#608).

## Goal

Bridge the Wanix guest-side Plan 9 task filesystem to the host-side `/exec/instances/<id>/` projection, so that a Wanix OS running as a WASI worker workload spawns/controls/waits on tasks through the real `SandboxBackend` seam, the EventService completion primitive, and the MAC reference monitor — rather than a guest-internal-only model.

## Context

Wanix is a Plan 9-inspired environment whose only abstraction is the file service. Run server-side as a WASI worker workload (filesystem backed by the Subject-scoped hyprstream 9P/VFS via the wasm substrate's WASI profile), it exposes a guest-facing task fs to code inside the sandbox:

- `/task/new/<kind>` — reading allocates a new task and returns its id.
- `/task/<id>/` — per-task `ctl` (verbs `bind`/`unbind`/`start`), `cmd`, `env`, `dir`, `exit`, `ns`, `fd/{0,1,2}`.
- A child task clones the parent's namespace; capabilities are bound in via path bind.

This is the **guest** programming model. The **host** `/exec/instances/<id>/` projection (P2) is the same lifecycle backed by real infrastructure. P4 wires the guest file ops to the host:

- guest `read /task/new/<kind>` → host instance allocation via `WorkersService` + `SandboxBackend`.
- guest `write start >> /task/<id>/ctl` → host backend `start`.
- guest `read /task/<id>/exit` → host latched-completion read (#607), fan-out, retained terminal value.
- guest `fd/{0,1,2}` → host instance fd streams.
- guest `ctl bind`/`mount` → host namespace mutation through the MAC reference monitor (P5).

## Scope

- Map the Wanix guest `/task` and `/cap` file operations onto the host `/exec/instances` projection and capability mounts.
- Reuse the existing browser-side 9P bridge work where it transfers to the server WASI fs path; add a distinct host adapter where it does not.
- Ensure guest-initiated waits use the host latched-completion primitive (no guest busy-poll of `exit`).

## Acceptance

- A Wanix guest can allocate, configure, start, wait on, and signal a host-backed task entirely through its `/task` file ops.
- Guest `exit` waits are fan-out, retained-terminal-state correct (late guest readers still get the value), and do not busy-poll.
[...]

PR #616, opened 2026-06-30
feat(#608 P3): TUI — replace poll_local spin loop with dedicated-thread async drivers

ewindisch:
**Descoped** (#611, P3 of #608): replace the `poll_local` noop-waker spin loop with dedicated-thread async drivers, and decouple bare-path VFS routing from Tcl.

## Why descoped
This PR branched before #638 (self-contained shell Mounts) merged. #638 made `/lang/tcl` + `/lang/python` self-serving (their driver threads live inside the Mounts; `create_mount_channel` is now `#[cfg(test)]`-only, no non-test caller hands a mount-rx to `ChatApp`). So the entire mount-rx-servicing surface this PR originally carried is now dead. Rebased onto the post-#638 base with that surface removed.

## What lands (the still-needed fix)
The `poll_local` spin loop still exists in the base and is still called from `handle_vfs_command`. It drives `!Send` shell futures with a **noop waker** — safe only while every VFS future is purely computational. The moment a VFS read legitimately blocks on an external event (e.g. `/exec/instances/<id>/exit` awaiting worker completion, #607/#623), that loop **hot-spins a CPU core forever** (a noop waker can't deliver the wake-up). This replaces it with:
- **`ShellOwner`** — Tcl `eval()` on a dedicated current-thread-runtime + `LocalSet` thread (preserves `TclShell`'s `!Send` invariant; parks, never spins).
- **`NamespaceDriver`** — direct `cat`/`ls` on a parked thread, no Tcl interpreter needed.
- **`classify_vfs_command`** — bare paths route straight to `NamespaceDriver` (no `eval("cat …")` round-trip through Tcl).
- Blocking eval leg uses `std::sync::mpsc` (not `tokio::oneshot`) so it's safe both inside and outside a tokio runtime.

## Dropped (dead post-#638)
`PyShellOwner`, `tcl_mount_rx`/`py_mount_rx`, `with_tcl_mount_rx`/`with_py_mount_rx`, `drain_py_mount`, and `ShellOwner::spawn`'s mount-select arm.

## Verified
`cargo check -p hyprstream-tui` + `-p hyprstream` clean; `cargo test -p hyprstream-tui` 35/35 (incl. 8 routing tests); clippy clean; dead-surface grep clean.

## Follow-up (non-blocking)
`hyprstream-workers-python` is now an unused dep of `hyprstream-tui` (inherited from base) — worth a trim.

Refs #611 · #608 P3.

### DIFF
--- a/crates/hyprstream-tui/Cargo.toml
+++ b/crates/hyprstream-tui/Cargo.toml
@@ -30,7 +30,7 @@ parking_lot = "0.12"
 rand = "0.8"
 uuid = { version = "1.0", features = ["v4", "v5", "serde"] }
 zeroize = { version = "1", features = ["derive"] }
-tokio = { version = "1", features = ["sync"] }
+tokio = { version = "1", features = ["sync", "rt", "macros"] }
 
 [target.'cfg(target_os = "linux")'.dependencies]
 landlock = "0.4"
--- a/crates/hyprstream-tui/src/chat_app.rs
+++ b/crates/hyprstream-tui/src/chat_app.rs
@@ -225,47 +225,6 @@ pub enum ChatMode {
 // ChatApp
 // ============================================================================
 
-/// Create a no-op waker for polling futures that don't need wake notifications.
-///
-/// Safe for TclShell/VFS futures which are purely computational (no IO reactor,
-/// no timers) — they always make progress on each poll without needing external
-/// wake-ups.
-fn noop_waker() -> std::task::Waker {
-    fn noop(_: *const ()) {}
-    fn clone(p: *const ()) -> std::task::RawWaker {
-        std::task::RawWaker::new(p, &VTABLE)
-    }
-    static VTABLE: std::task::RawWakerVTable =
-        std::task::RawWakerVTable::new(clone, noop, noop, noop);
-    // SAFETY: The vtable functions are valid no-ops. The data pointer is never dereferenced.
-    unsafe { std::task::Waker::from_raw(std::task::RawWaker::new(std::ptr::null(), &VTABLE)) }
-}
-
-/// Wrapper that asserts `Send` for a `!Send` value.
-///
-/// **Not thread-local storage** — this is a Send bypass for values that are
-/// constructed and accessed on a single thread but must transit a thread boundary
-/// wrapped in `Option<AssertSend<T>>` where the Option is `None` during the move.
-///
-/// SAFETY: This is used exclusively for `TclShell` which uses `Rc` internally
-/// (making it `!Send`). The wrapper is safe because:
-/// - The value is always wrapped in `Option` that is `None` during cross-thread moves
-/// - It is only populated via `ensure_tcl_shell()` on the app thread
-/// - It is only accessed from that same thread thereafter
-/// - `UnsafeCell` ensures `AssertSend<T>` is `!Sync` (cannot be shared across threads)
-struct AssertSend<T>(std::cell::UnsafeCell<T>);
-
-// SAFETY: See struct-level documentation.
-unsafe impl<T> Send for AssertSend<T> {}
-
-impl<T> std::ops::Deref for AssertSend<T> {
-    type Target = T;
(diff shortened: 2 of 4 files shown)


ewindisch:
**Update:** the one open caveat is resolved — `cargo build -p hyprstream --lib` (the downstream consumer sanity check) completed and passed (exit 0, ~27min under heavy concurrent-agent machine load). No breakage from the `ChatApp`/`hyprstream-tui` changes, confirming the API-surface diff finding. All verification for this PR is now green.

ewindisch:
**Self-review pass — both headline concerns verified SAFE by construction; proportionate hardening + missing test coverage applied (commit `69ff7920d`), ready for review.**

**Deadlock (dedicated-thread + LocalSet + Mutex<Receiver>): NO issue.** The owner thread's `biased` `tokio::select!` drops the non-winning arm's future at the end of each iteration — so when `bridge_rx` wins, the `recv_mount_cmd` future (and its mutex guard) is dropped *before* the winning arm runs; the lock is never held while `shell.eval()` executes. `eval()` jobs route through a completely separate channel, so they never contend the mount mutex. Fair FIFO across tabs via async `Mutex::lock().await`.

**Command-injection bypass via crafted bare path: NO issue.** The direct-`Namespace` fast path (`NamespaceDriver::cat`/`ls`) calls `namespace.cat(path,&subject)`/`namespace.ls(path,&subject)` — the *exact same code* the Tcl `cmd_cat`/`cmd_ls` builtins call. All authz + path-safety live inside `Namespace` (`resolve` + per-mount `walk`/`open`/`read` with the Subject), not in the Tcl builtins — so no safety check is bypassed. Reading `/bin/eval` via `cat` returns bytes, doesn't execute.

**Thread cleanup: NO leak, no Drop needed.** Channel-drop is the idiomatic shutdown signal (matches the `workflow/runner.rs` `run_job` reference pattern); robust even if the owner thread panics (subsequent `eval()` returns a clean error). `JoinHandle` detached as a daemon — appropriate.

**Hardening applied (commit `69ff7920d`):** `has_tcl_metachars` now also rejects `char::is_whitespace` (closes a bare-path tab/newline faithfulness gap — inert `NotFound`, not an injection, but a real divergence from the old molt-split behavior); added 7 unit tests covering the classifier (the routing split was previously untested); documented the `PyShellOwner` shared-vs-per-tab decision at the `/lang/python` TODO.

**Verified:** `hyprstream-tui` 35 tests pass (7 new), clippy clean. (Full-workspace release clippy + downstream `hyprstream --lib` couldn't complete locally due to disk saturation from ~40 concurrent worktrees — now cleaned up; CI covers both.) Marked ready for review.

ewindisch:
✅ **Spine-alignment review (epic #508): aligns.** `NamespaceDriver` reads via the namespace; `ShellOwner` drives the mount through its interface; Subject threaded. Note: this is *upstream* of the self-contained-mount refactor — the TUI no longer spins, but the mounts themselves still externalize rx (that fix is in the mount crates, tracked under #508).

ewindisch:
**Architecture note (non-blocking — this PR is correct as-is).**

`shell_driver.rs` (`ShellOwner`/`NamespaceDriver`) is the **interim** in-TUI task driver. This PR is right to merge: it fixes a live noop-waker spin-loop bug (the base still calls `poll_local`, which would hot-spin a core on a genuinely-blocking VFS read like `/exec/instances/<id>/exit`).

But per the execution-model charter (#660), it is NOT the endgame. Target model: task execution is a VFS/OS semantic — the shell writes to `/exec/instances/<id>/ctl` (#623) rather than owning a task driver. The `!Send`-shell-owner pattern here (plus the copies in `workflow::runner` and `TclMount`) collapses once `/lang/*` engines become pure VFS command-executors and the shell becomes a `/exec` client (#635 narrows the executor; #612 does the Wanix-side symmetry).

So: merge this for the bug fix; treat `shell_driver` as interim-until-#660.

PR #599, opened 2026-06-30
feat(workers): re-cut wasm/python onto hyprstream-workers-* crates [supersedes #596/#597]

ewindisch:
> **Draft for review.** The operator-directed re-cut of the wasm/python work onto the decided
> `hyprstream-workers-*` crate naming. Supersedes #596/#597 (and the prior closed stack). Design:
> `docs/plans/2026-06-30-execution-control-plane-9p-composition.md` (added in this branch).

## Final crate layout
```
hyprstream-workers                  # control plane (SandboxBackend registry; wasm reg auto_selectable:false)
hyprstream-workers-wasmtime         # ENGINE ONLY (Sandbox, build_linker, Profile A/B, EpochTimer,
                                    #   call_export / call_export_with_epoch, open_instance/PersistentInstance,
                                    #   vfs capability). Python-free.
hyprstream-workers-wasmtime-fsguest # WASI test guest
hyprstream-workers-python           # PythonShell + /lang/python PythonMount (on -wasmtime); mirrors -tcl
hyprstream-workers-python-guest     # RustPython → wasm32 guest
hyprstream-workers-tcl              # molt + /lang/tcl
```
Names per the decision: engines are worker execution engines under the WorkersService umbrella —
NOT `hyprstream-sandbox` ("sandbox" reserved for isolation), NOT `hyprstream-<lang>` (reads as an SDK).

## What changed
- Rename `hyprstream-sandbox → hyprstream-workers-wasmtime`; guests → `hyprstream-workers-{python-guest, wasmtime-fsguest}`; `hyprstream-tcl → hyprstream-workers-tcl` (+ all consumers, incl. a stray `hyprstream-rpc-std`).
- **Extracted the Python shell** out of the engine into `hyprstream-workers-python`. To keep the engine python-free, generalized its public seam: `Sandbox::call_export_with_epoch` + `Sandbox::open_instance -> PersistentInstance::call_op(...)` (the generic packed-op driver behind the old `PyShell`); `TAG_OK` made `pub`.
- Workers wasm `SandboxBackend` dep → `hyprstream-workers-wasmtime`; `auto_selectable:false` retained.
- `/lang/python` mounted in both namespace builders (`cli/shell_handlers.rs`, `tui/vfs.rs`) + channel drained in `chat_app.rs`, mirroring `/lang/tcl`.
[...]

Issue #547, opened 2026-06-29
Epic: Native MAC authorization — UCAN-authored, Casbin-enforced, deny-by-default (9P/VFS)


ewindisch:
Epic to make **native Mandatory Access Control the security model** of the universal 9P/VFS surface
(the filesystem of all hyprstream services, the distributed web, Wanix). MAC is the model, **not a
mode** — there is **no non-enforcing/permissive mode**. Every access to every object is mediated.

It composes: native MAC (a reference monitor) · content-bound labels + information-flow control ·
**UCAN as the authoring/delegation language** · **Casbin/TE as the compiled enforcement** · OAuth
token-exchange as the runtime grant path · Zero Standing Privilege · tiered escalation decisions.

**Relation to #539 (9P-RPC):** #539 generates the mount surface; this epic makes that generated mount
the **reference monitor** and supplies the policy/labels/compiler/tokens. #539 is the *surface +
generation*; this epic is the *security semantics* layered onto it. Hard dependency on #539's generated
mount + `$scope`/TE annotation work.

**Branch:** `feature/security-mac-ucan` (off `main`). **Schema breakage in scope** (rename `mcpScope`,
make scope mandatory, set stable names). Compiler detail spiked separately (validated design below).

> **Status: sub-issues S1–S8 FILED (#567–#574).** See the tracking checklist at the bottom.

---

## 1. Principles (invariants — none optional)

1. **Deny by default. Complete mediation.** Every op hits the monitor; nothing bypasses; default-deny.
2. **No permissive mode.** Enforcement cannot be disabled. Unlabeled subject/object = denied.
3. **Mandatory floor is independent.** The MAC lattice + IFC check runs on every op using the object's
   **content-bound** label — never overridable by any grant, token, UCAN, or evaluator.
4. **Single authorization path.** No fallback (fallbacks are downgrade vectors). Authority-unreachable
   = fail closed; solve availability with HA/caching, never a weaker path.
5. **The grant decision is authority-owned.** A requester's UCAN is *input*, never the decision logic.
   No principal can author the function that grants its own request.
6. **Front-load the hard decision.** Approve ceilings rarely + deliberately; exercise subsets
   automatically + deterministically.
7. **capnp = control plane; raw bytes = data plane.** Authz rides the control plane; the bulk/DMA path
   stays zero-copy.

## 2. The reference monitor (MAC realization = SELinux architecture, distributed)

```
SELinux                          hyprstream
[...]

PR #617, opened 2026-06-30
feat(#341 T2-A): rootless podman OCI SandboxBackend [supersedes #484]

ewindisch:
Implements **#346** (T2-A of epic **#341** "Worker GA"): rootless podman OCI `SandboxBackend`. Stacked on #599.

## Context
A prior attempt (PR #484, closed unmerged) had real review feedback this re-cut fixes rather than repeats.

## Fixes vs #484
- **Dropped `PodSandboxConfig` field** — #484 took `_config` but never read it (resources/security context silently dropped). `build_run_args()` now reads `hostname`, `dns_config`, `port_mappings`, `linux.resources.{cpu_quota,cpu_period,cpu_shares,memory_limit_in_bytes,memory_swap_limit_in_bytes,cpuset_cpus,cpuset_mems}`, `linux.security_context.{privileged,run_as_user,readonly_rootfs}` into the `podman run` argv. 5 dedicated unit tests.
- **Rootless mis-detection** — #484 probed `podman info --format '{{.Host.Security.Rootless}}'` but never gated behavior on it, and the advertised docker fallback would mis-detect (docker has no such Go-template field). This re-cut: dropped the docker path entirely (honest podman-only MVP, documented), made the rootless probe load-bearing (cached `OnceLock`, gates `--userns=keep-id`). Parsing tested against fixture strings including the docker-mismatch case.
- **Dead code** — removed the per-call PATH re-lookup and never-populated `extra_mounts`/`extra_env` fields.

## Method mapping
`backend_type()→"podman"` · `is_available()→which::which("podman")` · `start()→podman run --detach --rm --name=... [--userns=keep-id] [config-derived flags] <image> sleep infinity` → `PodmanHandle` · `stop()→podman stop --time` · `destroy()→podman rm --force` (idempotent) + dir cleanup · `reset()→Ok(false)` (ephemeral `--rm`, mirrors nspawn) · `get_pids()→podman inspect .State.Pid` · `exec_sync()→podman exec` under timeout, `(exit_code,stdout,stderr)` (mirrors nspawn) · `update_resources()→podman update --cpu-quota/--cpu-period/--memory` (live warm-pool resize, distinct from the one-time start() flags).

## Registration
`priority: 30, auto_selectable: true` — real container isolation (kernel namespaces + cgroups via runc/crun) outranks nspawn (10), weaker than Kata's VM boundary (100). Gated behind new `podman` Cargo feature (mirrors `kata-vm`/`wasm`); off by default, fails closed (not registered) when the feature is off.

## Verification
`cargo build -p hyprstream-workers` (no feature) green, podman fully excluded. `--features podman`: build green, **97 tests pass, 0 failed, 2 ignored**. Clippy clean. Torch-free (`cargo tree --features podman` confirmed). **Commit passed the real full-workspace `--release` clippy pre-commit hook** (not `--no-verify`) — the strongest verification signal of this wave.

podman was available in this sandbox, but no live daemon/pullable image was set up, so the two genuinely-live tests (`test_live_detect_rootless`, `test_live_start_exec_destroy_roundtrip`) are `#[ignore]`d rather than faked — only unit-level argument/parsing/registration tests ran end-to-end.

## Out of scope
No docker fallback (explicit MVP scope, documented).

Refs #341 · stacked on #599.

### DIFF
--- a/crates/hyprstream-workers/Cargo.toml
+++ b/crates/hyprstream-workers/Cargo.toml
@@ -172,6 +172,12 @@ wasm = ["dep:hyprstream-workers-wasmtime"]
 # Enable Dragonball built-in VMM support (alternative to cloud-hypervisor).
 # Implies `kata-vm` since it tunes the kata hypervisor abstraction.
 dragonball = ["kata-vm", "kata-hypervisor/dragonball"]
+# Rootless podman OCI sandbox backend (#346, T2-A of #341). Compiles in
+# `PodmanBackend`, which shells out to the `podman` CLI at runtime — no extra
+# Rust dependencies, just the inventory registration. Opt-in (mirrors
+# `kata-vm`/`wasm`) so the default build doesn't carry an extra selectable
+# backend; when off, `podman` selection fails closed (not registered).
+podman = []
 # Enable D-Bus bridge for container access to host D-Bus services
 dbus = ["dep:zbus"]
 # Systemd integration (socket activation, service units)
--- a/crates/hyprstream-workers/src/runtime/mod.rs
+++ b/crates/hyprstream-workers/src/runtime/mod.rs
@@ -30,6 +30,11 @@ pub mod exec_mount;
 #[cfg(feature = "kata-vm")]
 pub mod kata_backend;
 pub mod nspawn;
+// Rootless podman OCI sandbox backend (#346, T2-A of #341) — gated behind
+// `podman`. Shells out to the `podman` CLI at runtime; no extra build-time
+// dependencies, so this is a thin opt-in atop the always-built nspawn shape.
+#[cfg(feature = "podman")]
+pub mod podman;
 mod pool;
 mod sandbox;
 // In-process WebAssembly sandbox backend (#505 P2) — gated behind `wasm`
@@ -92,6 +97,8 @@ pub use exec_mount::ExecMount;
 #[cfg(feature = "kata-vm")]
 pub use kata_backend::{KataBackend, KataHandle};
 pub use nspawn::{NspawnBackend, NspawnConfig, NspawnHandle};
+#[cfg(feature = "podman")]
+pub use podman::{PodmanBackend, PodmanConfig, PodmanHandle};
 #[cfg(feature = "wasm")]
 pub use wasm_backend::{WasmBackend, WasmConfig, WasmHandle};
 // Inventory-based backend registry + fail-closed selection spine (#507)
--- a/crates/hyprstream-workers/src/runtime/podman.rs
+++ b/crates/hyprstream-workers/src/runtime/podman.rs
@@ -1,0 +1,835 @@
+//! PodmanBackend — rootless podman OCI `SandboxBackend` (#346, T2-A of #341).
+//!
+//! Runs each sandbox as an OCI container managed by `podman`. No hypervisor,
+//! no RAFS/nydus — pulls (or reuses) a standard OCI image and keeps the
+//! container running (`sleep infinity`) so `exec_sync` can drive workloads
+//! inside it. Intended for casual/dev installs where the Kata +
+//! Cloud-Hypervisor VM toolchain is unavailable or unwanted.
+//!
+//! Backend type: `"podman"`.
+//!
+//! # Scope (MVP)
+//!
+//! This is a **podman-only** backend. A prior attempt (#484) advertised a
+//! "docker fallback" but never actually exercised it correctly — `docker
+//! info` does not emit the `{{.Host.Security.Rootless}}` Go-template field
+//! podman does, so the fallback's rootless probe silently parsed as `false`
+//! for every docker host (a mis-detection, not a real fallback). Rather than
+//! half-implement docker support again, this backend is honestly
+//! podman-only: [`is_available`](SandboxBackend::is_available) checks for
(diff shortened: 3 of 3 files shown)


ewindisch:
**Self-review pass — 1 real security fix applied (commit `f583c4fca`), ready for review.**

**Headline finding (security):** the caller-controlled `hyprstream.io/worker-image` *annotation* flowed through as a bare positional in the `podman run` argv (podman requires the image positionally, so no `--` guard is possible there). An annotation value like `--privileged`, `--network=host`, or `-v /:/host` was passed straight through and podman parsed it as an injected flag → **privilege escalation / sandbox escape**. Fixed: `PodmanBackend::validate_image()` rejects empty/dash-prefixed refs, called at the top of `start()`; regression tests assert `--privileged`/`--network=host`/`-v /:/host` are all rejected.

**Otherwise confirmed safe:** all 8 podman invocations use `Command::new("podman")` (execvp, **zero `sh -c`**); every config field is `--flag=value`-form; integers can't carry metacharacters; `exec_sync` guards the command with `--`. So the image-annotation vector was the one real escape; it's now closed.

**Other fixes in the same commit:** (a) `start()` partial-failure resource leak — on non-zero `podman run` exit it now best-effort `podman rm --force` + `remove_dir_all` before returning `Err` (previously the sandbox dir + possibly a container leaked); (b) `port_mappings` silently dropped the protocol suffix (everything defaulted to tcp) — now appends `/udp`/`/sctp`.

**Verified:** `--features podman` 100 tests pass (2 ignored live-podman), clippy clean. One cosmetic nit (`init_pid` written-not-read) left for a follow-up. Marked ready for review.

ewindisch:
⛔ **Spine-alignment review (epic #508): recommend HOLD / re-ground before merge.**

This backend violates the namespace-as-OS / backend-as-transport model. PodmanBackend runs `podman pull` + `podman run` with its own rootfs — a **side-channel image-provisioning path** that bypasses the universal filesystem-service layer. In the adopted model, image provisioning is a mountable filesystem service (OCI/RAFS as a 9P `FsMount`, composable by mount/bind), and a backend's job is to *boot a guest OS + deliver the composed namespace* — not own image pull. Podman should compose the universal image filesystem service into the container's namespace and deliver it (bind/passthrough transport), not run a parallel `podman`-managed rootfs.

**Concrete:** the `base_image`/`build_run_args` path (incl. the `validate_image` guard from the security fix) *is* the side-channel. The security fix itself is good and stays; the *architecture* needs re-grounding.

**Options:** (a) rework to compose the universal filesystem service (proper fix, new child of #508); (b) merge as an explicit documented stopgap with a linked follow-up. Recommend (a)/hold.

(Can't formally `request changes` — GitHub forbids it on one's own PR — so posting as a comment. Operator/reviewer: please treat as a changes-request.)

ewindisch:
## 🔧 Spine re-grounding pass — STOP-and-report (delivery transport gap)

Rebased onto the current base (post-#641) and investigated re-grounding image delivery onto the spine (#508): compose the image `Mount` via `SandboxFs::compose` (#641) and hand podman the composed namespace as its rootfs, mirroring kata's compose-and-serve.

**Blocked — reported rather than hacked around.** The correct podman transport is `podman run --rootfs <dir>`, which requires the composed namespace materialized as a **host POSIX directory**. But `hyprstream-vfs-server` ships **only** a vhost-user-fs transport (guest-memory vrings via `vhost-user-backend`, for a Cloud Hypervisor VM) — it's built with `fuse-backend-rs` `features=["virtiofs"]` and *explicitly not* `fusedev`, so there is no local `/dev/fuse` mount yielding a directory podman could consume. The `VfsFileSystem` down-adapter itself is transport-agnostic (`SandboxFs::into_filesystem`), so this is an additive gap, not a redesign.

Per the "faithfulness over a green merge" instruction, I did **not** fall back to dressing up the `podman run <image-ref>` side-channel as spine-aligned. Instead:

- **Filed #653** — "vfs-server: local-FUSE-mount serve mode" (the blocking prerequisite; also unblocks nspawn re-grounding). Notes the rootless-FUSE-in-userns open question.
- **Marked the interim seam honestly** — a module-level "Spine positioning" section + inline `INTERIM (non-spine, #653)` markers at the `resolve_image` / `start` image-resolution seam, so the current podman-store provisioning is not mistaken for spine-aligned.
- Trimmed the `#484` prior-attempt narration in the module doc to present-state.

**Preserved** the good #484-review fixes: rootless probe (`podman info --format '{{.Host.Security.Rootless}}'`), config-fields-actually-read at `start()`, `--userns=keep-id` rootless-only handling, and the annotation flag-injection guard.

**Note for operator decision — `auto_selectable`:** podman is registered `auto_selectable: true` (priority 30, between nspawn and kata). I left this as-is: unlike wasm (`false`, a non-CRI interface / isolation-downgrade concern under #547 ZSP), podman is a real CRI/OCI container backend, so auto-selecting it is not a silent isolation downgrade. Flagging in case you'd rather gate it off until #653 lands and it's truly spine-aligned.

**Verify:** `cargo check`/`clippy -p hyprstream-workers --features "podman,oci-image"` clean; 21 podman unit tests pass (2 `#[ignore]` need a live podman). Docs-only change over the rebase — no behavior change this pass.

**Recommendation:** hold #617 (keep draft) until #653 lands, OR merge now as an explicitly-marked interim backend and re-ground in the #653 follow-up. Operator's call.

Issue #610, opened 2026-06-30
P2: /exec/instances/<id>/ per-instance lifecycle Mount (ctl/status/exit/fd/ns/cap) over SandboxBackend, EventService-backed wait


ewindisch:
Phase P2 of the P9 Task/Instance Projection epic (#608).

## Goal

Implement the per-instance lifecycle Mount: project a running execution unit as a Plan 9 `/proc`-style directory at `/exec/instances/<id>/`, with completion (`exit`/`fd`/`signal`) wired to the EventService latched-completion primitive (#607). The projection is uniform across every isolation backend behind the `SandboxBackend` seam (#508): in-process wasm, nspawn, OCI, CRI, Kata VM.

## Layout

```
/exec/instances/<id>/
  ctl       # write verbs: start | kill | signal <n> | bind <src> <dst> | mount <svc> <dst>
  status    # read: live, NON-latched state (backend, pid(s), resource use) — for pollers, never blocks
  exit      # read: the TERMINAL-LATCHED completion value — fan-out, see "wait semantics"
  fd/0,1,2  # stdin/stdout/stderr as streams
  ns        # the instance's namespace, browsable (its binds/mounts)
  cap/      # capabilities bound into this instance
```

Built from the `ControlFile`/`FieldFile` primitives (P1). `ctl` verbs drive the `SandboxBackend` seam (#508): `start`/`kill`/`signal` → backend lifecycle (`start`/`stop`/`destroy`/`exec`); `bind`/`mount` → namespace mutation (authorized in P5). `status` reads the live backend state (`get_pids`, resource use); it is the non-blocking pollable view, distinct from the blocking latched `exit`.

## Wait semantics (consumes EV7 #607 — do not re-design)

`exit`, `fd` close, and `signal` are **fan-out completions** (any process in the namespace can wait; >0 waiters = fan-out). Per the latched-completion contract:

- **The `exit` FILE holds the authoritative retained terminal value** (host-side, plaintext-at-owner). A `read` returns it if present, else subscribes to the per-instance ephemeral stream and blocks until the edge fires. Late watchers (after termination) are file-served; early watchers block then receive the edge.
- **Terminal = monotonic write-once.** Per-resource-type definitions owned here:
  - **task exit:** `Exited(code) | Killed(signal) | Failed(reason)` (start/spawn failure counts; a task that never ran still has a terminal answer).
[...]

Issue #344, opened 2026-06-19
T1-C: Kata guest agent (kata-agent vsock) -- real CRI create/start/exec_sync of container-in-VM


ewindisch:
Track 1 (largest build). Today supports_exec is false and runtime/service.rs:928-962 are placeholders -- Kata is VM-lifecycle only. Build the kata-agent vsock channel so a container actually runs inside the VM end to end. Biggest functional gap. Part of #341. Plan: docs/plans/2026-06-19-worker-ga-epic.md

ewindisch:
**Cross-link: P9 Task/Instance Projection epic (#608), phase P2 (#610).**

The new epic projects every running execution unit as a Plan 9 `/proc`-style directory `/exec/instances/<id>/{ctl,status,exit,fd/,ns,cap/}`, uniform across all backends. Two obligations land on the Kata guest-agent work here:

1. **In-VM exec/exit/stdio must surface through the instance fs.** When CRI create/start/exec_sync of a container-in-VM works, the running container must be projected as `/exec/instances/<id>/{exit,fd}` — `exec` output to `fd/{1,2}`, completion to `exit`. The agent is the thing that knows the in-VM process exited.

2. **`exit` must latch a terminal value even on abnormal VM/agent death.** Completion is a fan-out, retained-terminal-state read (a watcher may arrive after the process exits and must still get the value). The contract (recorded in #607) requires a guaranteed-eventual terminal latch: if the hypervisor or the agent dies, a reaper/timeout must latch `Failed(reason)`, or a `read /exec/instances/<id>/exit` blocks forever. The guest-agent vsock channel is where in-VM exit is observed, so the death-detection → latch path for the VM case is rooted here.

No scope change to this issue's core (real CRI exec_sync); these are the surfacing + liveness obligations the instance projection depends on.

ewindisch:
Implemented — PR #619 (kata-agent vsock/ttrpc client, real exec_sync). REAL: vsock dialer + genuine ttrpc protocol types + full RPC sequence, tested against a real ttrpc::r#async::Server. STUBBED (documented): CreateContainer's OCI spec is a minimal placeholder, not a full bundle translation from PodSandbox/image config — see PR for the precise next step. No live kata-agent available to verify against in this environment.

ewindisch:
[...]

Issue #508, opened 2026-06-26
Epic: unified Sandbox backend taxonomy (Kata/Oci/Cri/Wasm/Nspawn) under one SandboxBackend seam


ewindisch:
## Goal
Converge worker sandboxing onto **one `SandboxBackend` seam + one canonical `BackendType` taxonomy**, behind the faithful CRI-capnp ingress. Five sibling backends — **Kata** (VM), **Oci** (CRI→oci-spec→youki/crun), **Cri** (drive external containerd), **Wasm** (embedded wasmtime, VFS-backed-fs), **Nspawn** — selected by operator config / RuntimeClass, **fail-closed** (never silently downgrade isolation). Child track of epic #341 (Worker GA).

## Design basis
- `docs/plans/2026-06-26-cri-client-vfs-injection-spike.md` (CriBackend / NRI VFS injection)
- `docs/plans/2026-06-26-wasm-wasi-worker-convergence.md` (one wasmtime substrate → Python + Wasm-backend + Wanix; VFS-backed-fs; embedded over runwasi)
- `docs/plans/2026-06-26-rustpython-wasm-vfs-sandbox.md` (#483 Profile-A guest)

## "Namespace IS the OS surface" pivot (locked 2026-07-01, tracked in #632)
`SandboxBackend` is narrowing to **lifecycle + `deliver_namespace`** (boot guest OS + deliver a composed namespace over a transport); filesystem services and shell mounts compose into that namespace as real `hyprstream_vfs::Mount`s, not backend-owned state. Sequencing:
1. [x] Lock spine (decision record, 2026-07-01)
2. [x] Universal filesystem service (#633, merged via #641)
3. [x] Self-contained shell Mounts + dedupe namespace builders (#634 — shell Mounts via #638; builders unified via #703, merged 2026-07-03)
4. [x] Narrow `SandboxBackend`: add `deliver_namespace`/`NamespaceTransport`/`NamespaceDelivery`, mark CRI-shaped methods TRANSITIONAL (#635, merged via #704, 2026-07-03 — required a fail-closed fix to wasm's `HostImports` arm before merge, see PR review)
5. [ ] **Next:** re-ground concrete backends (Kata/Oci/Nspawn/Wasm) to actually call `deliver_namespace` from their boot path instead of running a parallel compose-and-serve mechanism
6. [ ] Re-audit in-flight backend PRs against the narrowed trait

[...]

PR #619, opened 2026-06-30
feat(#341 T1-C): kata-agent vsock/ttrpc client — real exec_sync (REAL+PARTIAL, see body)

ewindisch:
Implements **#344** (T1-C of epic **#341** "Worker GA"): Kata guest agent (kata-agent vsock) — real exec_sync of a container-in-VM. Stacked on #599.

> **Safety note: REAL but PARTIAL.** This replaces kata's "exec not supported in host-level mode without guest agent" error with a genuine vsock/ttrpc client wired through the actual `kata-agent` protocol — but `CreateContainerRequest`'s OCI spec is a minimal placeholder, not a full bundle translation. See "Stubbed" below before relying on this for anything beyond basic exec.

## Environment boundary (read before reviewing)
Real KVM + `cloud-hypervisor` are present in this sandbox, but there is no bootable Kata guest image (kernel+rootfs+kata-agent) available — **no live VM/agent verification was possible or attempted.** What's verified is grounded in the actual upstream `kata-containers` source (tag `3.31.0`, matching this repo's pinned dep):
- **Verified from source:** Cloud Hypervisor exposes the agent over hybrid vsock — Unix socket at `<sandbox>/ch-vm.sock`, handshake `"connect <port>\n"` → `"OK"` (`hypervisor/src/ch/inner_hypervisor.rs::get_agent_socket`, `agent/src/sock/hybrid_vsock.rs`). Default port **1024** (`virtcontainers/hypervisor.go: vSockPort`).
- **Real protocol types, not hand-rolled:** the upstream `protocols` crate (ttrpc/protobuf-generated `AgentServiceClient`) is already transitively vendored via `kata-hypervisor`'s `async` ttrpc feature — added as a direct dep and used directly, not reimplemented.
- **Unverified:** real kata-agent's actual `CreateContainer` rootfs/mounts/namespace expectations, and exact duplicate-container error codes (best-effort `ALREADY_EXISTS` gRPC-code + string-fallback handling, explicitly flagged unvalidated in the module doc).

## What's REAL vs STUBBED (precise, not glossed)
**REAL:** the vsock/hybrid-vsock dialer; the ttrpc client wiring (`ttrpc::r#async::Client`); the full `CreateContainer → StartContainer → ExecProcess → WaitProcess → ReadStdout/ReadStderr` RPC sequence using genuine generated message types — exercised in unit tests against a **real `ttrpc::r#async::Server`** (the same crate a real kata-agent links against), not mocked bytes.
**STUBBED:** `build_minimal_spec()` builds a minimal `oci::Spec` "sufficient to drive `CreateContainer`" — explicitly documented as **NOT a complete OCI bundle translation** (no rootfs/mounts/namespaces from `PodSandbox`'s image/VFS config yet).

## Wiring
`KataBackend::exec_sync` now dials the agent via `Hypervisor::get_agent_socket()`, caches the connection on `KataHandle.agent: Mutex<Option<Arc<KataAgentClient>>>`, runs an idempotent ensure-container (tolerates `ALREADY_EXISTS`) + exec + wait + drain sequence. `supports_exec()` now returns `true`. Return shape `(i32, Vec<u8>, Vec<u8>)` is built from the real `WaitProcessResponse.status` — a genuine terminal value, positioned (doc-noted) for #610's future exit projection without restructuring.

## Verification
`cargo build/test -p hyprstream-workers --features kata-vm` — **139 tests pass, 2 ignored** (pre-existing, need `genisoimage`). No-feature build unaffected. Clippy clean. **Full-workspace `--release` clippy pre-commit hook passed** (not `--no-verify`).

## Precise next step
Needs validation against a real `kata-agent` (bootable guest image + KVM CI runner) to confirm: (a) the minimal spec is sufficient for a real agent to create a runnable container — rootfs/mounts/namespaces likely need wiring from `sandbox_fs.rs`'s composed VFS, which may already supply the needed rootfs path; (b) `is_already_exists`'s detection matches real kata-agent error responses, not just the assumed gRPC status code.

Refs #341 · stacked on #599.


ewindisch:
**Self-review pass — 1 real lifecycle bug fixed (commit `2e2023a58`), ready for review (code-sound; merge still gated on real-kata-agent validation).**

**Bug found + fixed (stale-connection wedge):** `KataHandle.agent` cached the agent client lazily and was only cleared by warm-pool `reset_sandbox`. A VM killed/restarted *externally* (no reset) left a dead client in the cache → every subsequent `exec_sync` reused it and failed against the same dead connection, **wedging the sandbox until manual reset**. (It failed *loudly* — each RPC is bounded at 30s and the whole sequence by the caller timeout — not a silent hang, but a real availability bug.) Fix: `exec_sync` now evicts the cached client on failure under `try_lock`, comparing by `Arc::ptr_eq` so a concurrent in-flight exec that swapped in a fresh client is never discarded. Regression test `test_exec_sync_clears_cached_client_on_failure` added.

**Confirmed sound (no fix needed):** all timeouts bounded (5s connect, 30s per RPC, caller timeout on the full sequence — no unbounded call); `is_already_exists` fails safe (a misclassified error surfaces loudly at the next RPC, never a wrong silent result); `build_minimal_spec` doc honestly says "NOT a complete OCI bundle translation" and matches reality; the ttrpc-server-backed tests exercise the genuine `exec_sync → ensure_container → exec → wait → drain` path.

**Unchanged caveat (honest):** `CreateContainer`'s minimal OCI spec has never been exercised against a *real* kata-agent guest image — accurately documented in the module. Should be validated before production, but does not block review of the client/transport/lifecycle code, which is sound.

**Verified:** `--features kata-vm` 140 tests pass (139 + 1 new), clippy clean. Marked ready for review.

ewindisch:
✅🔢 **Spine-alignment review (epic #508): aligns-in-spirit, layering note + standing caveat.**

The kata-agent client is *exactly* the spine's **guest-OS control service** — the thing that runs commands in the Linux guest over ttrpc/vsock. It's the peer of Wanix's `/task` for wasm (the wasip1 Plan9 guest). That's the right concept and valuable code; keep it.

**Layering note for when the trait narrows:** today this is reached via `SandboxBackend::exec_sync`. Under the spine, `exec_sync` moves OFF the backend trait (the backend doesn't run commands — the guest OS does) onto the guest-OS control file service. So the agent client should eventually be exposed as that control surface (a 9P file service projecting CreateContainer/ExecProcess/etc.), not as a trait method. Not a blocker for this PR — but the integration point will move.

**Standing caveat (unchanged):** `build_minimal_spec` is a documented minimal placeholder; needs real-kata-agent validation before production. The code is internally sound (self-review confirmed).

Issue #352, opened 2026-06-19
T3-D: Nydus dependency cleanup (drop pinned git deps)


ewindisch:
Track 3. Published Nydus crates have an API mismatch forcing pinned git deps; clean up once upstream/our usage allows. Part of #341. Plan: docs/plans/2026-06-19-worker-ga-epic.md

ewindisch:
**Nydus crates.io migration blocked by upstream incompatibility.**

`nydus-service 0.4.0` on crates.io fails to compile against `nydus-rafs 0.4.1` (same crates.io epoch) due to unsatisfied trait bounds internal to those crates:
```
error[E0277]: the trait bound `Rafs: BackendFileSystem` is not satisfied
error[E0277]: the trait bound `Rafs: Layer` is not satisfied
error: could not compile `nydus-service`
```

All 5 nydus crates (`nydus-api 0.4.1`, `nydus-storage 0.7.2`, `nydus-builder 0.2.0`, `nydus-rafs 0.4.1`, `nydus-utils 0.5.1`) are published but `nydus-service` has a broken internal dep on the crates.io publish. Keeping git deps at `v2.4.0` until upstream fixes this.

Tracking issue filed against nydus: the `nydus-service` 0.4.0 crates.io publish needs a compatible `nydus-rafs` version dependency.

ewindisch:
Resolved as far as possible — PR #620. All 6 nydus crates pinned to an exact git rev (was a floating tag) for reproducibility. Full unpinning to crates.io is structurally blocked: a cohort conflict between this repo's pre-existing kata-hypervisor dep (needs nydus-api 0.4.1/cohort-B) and nydus-service (only published for cohort-A) — documented in detail in the PR/Cargo.toml. Resolves when dragonflyoss publishes cohort-B nydus-service, or a separate fuse-backend-rs 0.13->0.14 bump of hyprstream-vfs-server's transport stack happens.

PR #620, opened 2026-06-30
build(#341 T3-D): pin nydus deps to exact rev; document the real cohort-conflict blocker

ewindisch:
Implements **#352** (T3-D of epic **#341** "Worker GA"): Nydus dependency cleanup. Stacked on #599.

## Outcome: pinned to exact rev (full unpinning is structurally blocked, with a corrected root cause)
All 6 nydus/RAFS crates (`nydus-api`, `nydus-storage`, `nydus-builder`, `nydus-rafs`, `nydus-utils`, `nydus-service`) were on a **floating git tag** (`v2.4.0`). All 6 are now pinned to an **exact commit** (`5a9d42d8`) — reproducible, no longer a moving target. None could move to crates.io.

## The real, verified blocker (corrects this branch's own earlier, less-accurate pass)
crates.io has **two incompatible cohorts** of these crates: cohort A (tag v2.4.0 / `fuse-backend-rs` 0.13) and cohort B (tag v2.4.4 / `fuse-backend-rs` 0.14) — with `nydus-service` only published for cohort A. Cohort A alone would otherwise work, **but** this workspace's pre-existing `kata-hypervisor` git dependency (kata-containers tag 3.31.0) transitively requires `nydus-api = "0.4.1"` (cohort B) via `dragonball → dbs-virtio-devices`. Cargo unifies same-named registry packages workspace-wide, so any crates.io `nydus-api` either conflicts outright or forces the whole stack to cohort B — which cascades into a `fuse-backend-rs` 0.13→0.14 bump requiring `hyprstream-vfs`/`hyprstream-vfs-server`'s entire vhost/vm-memory/virtio-queue/vmm-sys-util stack to re-pin in lockstep — a much larger, separately-scoped change with no functional test path here. Fully documented in the new `Cargo.toml` comment.

## Verification
`cargo build -p hyprstream-workers` (no feature) pass; `cargo build`/`cargo check --features kata-vm` pass; clippy clean under `--features kata-vm`; `cargo tree --features kata-vm | grep nydus` confirms all 6 (+ transitive `nydus-upgrade`) resolve to the exact rev. `Cargo.lock` updated via `cargo update --precise` for a minimal diff.

## What's left unresolved, precisely
Blocked on either (a) dragonflyoss publishing a cohort-B `nydus-service` to crates.io, or (b) a separate, larger effort to bump `hyprstream-vfs`/`hyprstream-vfs-server`'s vhost-user-fs transport stack to `fuse-backend-rs` 0.14 in lockstep — out of scope here.

Refs #341 · stacked on #599.


ewindisch:
**Self-review pass — clean, 1 fix that IMPROVED on the earlier rebase (commit `825b790bb`), ready for review.**

**Root cause independently verified against crates.io primary sources:** the two-cohort conflict is real — cohort A (2025-07/10: nydus-api 0.4.0 etc.) vs cohort B (2026-02-28 republish: nydus-api 0.4.1, fuse-backend-rs 0.14), with **no cohort-B `nydus-service`** published (newest is 0.4.0, requires fuse-backend-rs ^0.13.1). `kata-hypervisor` (tag 3.31.0) under the `dragonball` sub-feature pulls nydus-api 0.4.1 + fuse-backend-rs 0.14 from the registry, and cargo's workspace-wide source unification would collapse the git/registry separation that currently keeps them apart. The comment's `dragonball → dbs-virtio-devices` path is accurate as written. Lockstep-bump consequence (hyprstream-vfs/-vfs-server's vhost/vm-memory/virtio-queue/vmm-sys-util stack) also verified.

**Post-rebase pin integrity:** all 6 nydus crates (plus transitive nydus-upgrade) resolve to `rev=5a9d42d8` — the pin survived the rebase; none reverted to the floating tag.

**Fix applied (the valuable catch):** the earlier rebase's lockfile regeneration had introduced **stray unrelated drift** — socket2 0.6.0→0.5.10 and itertools 0.13.0→0.10.5, silently downgrading networking/codegen deps workspace-wide, unrelated to the nydus pin. Fixed surgically: rebuilt Cargo.lock from base + a nydus-only source-spec swap (`?tag=v2.4.0#`→`?rev=5a9d42d8#`); since the rev expands to the identical commit the tag pointed to, it's a pure source-URL change with zero re-resolution. Cargo.lock diff vs base is now **7 nydus lines only** (was nydus + 6 stray lines).

**Verified:** `--features kata-vm` build + clippy clean, no-feature build clean. Marked ready for review.

ewindisch:
✅ **Spine-alignment (epic #508): aligns** — dep hygiene only. Future note (not this PR): when OCI/RAFS becomes a universal mountable filesystem service (new child of #508), the nydus deps move off the `kata-vm` feature onto their own (`oci-image` / filesystem-service feature).

Issue #170, opened 2026-06-05
M2/N7: zb2 fd/9P container I/O as duplex lossless Pipe


ewindisch:
Part of epic #131. See docs/plans/2026-06-05-zmq-replacement-consolidated.md.

Greenfield (Worker fd path is a placeholder; 9P is WASM-only/DMA; VFS Mount is read/write). Build a duplex, **lossless** `Pipe` stream kind; implement Kata vsock/serial **console I/O** (replace run_fd_streaming_task placeholder, workers service.rs:928); run **9P-over-Pipe** with a non-WASM transport; wire hyprstream-vfs Mount/proxy; TUI stdin/stdout pipes. **Also migrate the model↔worker callback DEALER plane** (inference.rs:256 / callback.rs / CallbackRouter) here — it's a control plane not on RequestLoop.

ewindisch:
## Progress note (2026-06-15)

M2d implementation underway on branch `ewindisch/moq-170` (stacked above moq-169 in the epic stack).

### What's done / in-progress
- `PodSandbox` struct gains `console_socket: Option<PathBuf>` field + `console_socket()` accessor
- `run_fd_streaming_task` placeholder replaced with real async UNIX socket reader:
  - When `console_socket` is set: opens `tokio::net::UnixStream`, reads in 4 KiB chunks, publishes via `publisher.publish_data()`
  - EOF from socket = container exited (clean completion)
  - Cancellation token → `publish_error("cancelled")`
  - No socket configured → wait for cancellation (removes the 100 ms polling loop)

### What's pending (later in #170)
- **Backends**: Kata backend (`KataHandle`) needs to populate `console_socket` from the CHV API socket path when the sandbox starts — tracked as a follow-up to `kata_backend.rs`
- **9P-over-Pipe** with non-WASM transport — design spike deferred
- **`hyprstream-vfs` Mount/proxy** wiring — deferred
- **TUI stdin pipes** — deferred (blocked on #134 duplex relay)
- **Model↔worker callback DEALER plane migration** (`inference.rs:256` / `callback.rs` / `CallbackRouter`) — large refactor; tracked as sub-issue
- **Bidirectional Pipe** (duplex stream kind): two parallel streams (forward + reverse) via separate `StreamChannel` instances — design tracked in epic plan

### Related
- Stacks on top of moq-169 (M2c presets)
- Branch will be in push-and-prs.sh for coordinated push

ewindisch:
[...]

PR #623, opened 2026-06-30
feat(#608 P2): /exec/instances ExecMount — Plan9 /proc projection of SandboxPool

ewindisch:
Implements **#610** (P2 of epic **#608** "P9 Task/Instance Projection"). Stacked on #599.

## What
A `hyprstream_vfs::Mount` (`ExecMount`) projecting `SandboxPool`'s active sandboxes as a Plan9 `/proc`-style tree, per `docs/plans/2026-06-30-execution-control-plane-9p-composition.md` §4.

```
/exec/instances/             — dynamic dir: active sandbox/instance ids
/exec/instances/<id>/ctl     — ctl file: start|stop|kill|destroy drives the instance's SandboxBackend
/exec/instances/<id>/status  — read-only, live non-blocking poll of PodSandboxState
/exec/instances/<id>/exit    — read-only. STUB: returns terminal status if already terminal, empty
                                "not yet terminal" marker otherwise.
/exec/instances/<id>/ns      — read-only, best-effort textual listing of sandbox_path/console_socket
```
`fd/` deliberately deferred — depends on the streaming plane / #170 landing first.

## The #607 stub (explicit, not silent)
`exit`'s real blocking-await semantics depend on **#607** (EventService EV7 terminal-latch + read-then-subscribe). This PR does NOT implement blocking-await — `ExecMount::terminal`/`read_exit()` carry an explicit `TODO(#607)` marking exactly where that wiring plugs in once it lands. Per the earlier design discussion: this stub correctly avoids building a throwaway mpsc/oneshot-style fan-out hack for something that structurally needs `>0 waiters` fan-out — it should be replaced wholesale with the EV7 primitive, not patched.

## Authorization boundary
`Subject` is threaded through every op per the `Mount` contract. **This mount does not itself authorize** — the MAC reference monitor (#547/#568) is explicitly out of scope here, consistent with every other `Mount` in this codebase.

## Verification
12 new unit tests against a minimal in-memory `FakeBackend` (no kata/nspawn/wasm dependency) — listing, status, ctl verbs, exit before/after terminal.

Refs #608 · stacked on #599.

### DIFF
--- a/crates/hyprstream-workers/src/runtime/exec_mount.rs
+++ b/crates/hyprstream-workers/src/runtime/exec_mount.rs
@@ -49,6 +49,11 @@ use async_trait::async_trait;
 use tokio::sync::Mutex as AsyncMutex;
 
 use hyprstream_vfs::{DirEntry, Fid, Mount, MountError, Stat, Subject};
+// `parking_lot::Mutex` for the ctl write→read latch (interior mutability
+// through the `&Fid` that `Mount::write` receives). When this branch is
+// rebased onto a base containing #615, swap this hand-rolled latch for
+// `hyprstream_vfs::devfile::DevFileState` (same parking_lot::Mutex shape).
+use parking_lot::Mutex as PmMutex;
 
 use super::client::PodSandboxState;
 use super::pool::SandboxPool;
@@ -126,8 +131,11 @@ enum ExecFidKind {
 /// Fid state for the exec mount.
 struct ExecFid {
     kind: ExecFidKind,
-    /// Buffer for the ctl write→read pattern (verb result message).
-    write_buf: Vec<u8>,
+    /// Latch for the ctl write→read pattern (verb result message). A
+    /// `parking_lot::Mutex<Vec<u8>>` provides safe interior mutability through
+    /// the `&Fid` that `Mount::read`/`Mount::write` receive — so we avoid the
+    /// `unsafe *mut` cast-through-`&` hazard. Unused for non-ctl fid kinds.
+    write_buf: PmMutex<Vec<u8>>,
 }
 
 // ─────────────────────────────────────────────────────────────────────────────
@@ -319,7 +327,7 @@ impl Mount for ExecMount {
 
         Ok(Fid::new(ExecFid {
             kind,
-            write_buf: Vec::new(),
+            write_buf: PmMutex::new(Vec::new()),
         }))
     }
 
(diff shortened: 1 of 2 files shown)


ewindisch:
**Self-review pass — 1 real UB bug fixed (commit `d601576ca`), ready for review.**

**Bug found + fixed (UB under concurrent writes):** the ctl write→read latch mutated `write_buf: Vec<u8>` via an `unsafe` cast-away-const through the `&Fid` that `Mount::write` receives. Since `ExecFid` is `Send+Sync` and `Mount` hands out shared `&Fid`, a future caller multiplexing ops on one fid (e.g. a 9P server with concurrent Twrite on one fid) would race the non-atomic `Vec` assignment → **undefined behavior**. Fixed: `write_buf` is now `parking_lot::Mutex<Vec<u8>>` (safe interior mutability through `&Fid`) — the same primitive `#615`'s `DevFileState` uses and TclMount/PythonMount already use. Regression test `ctl_concurrent_writes_to_one_fid_are_safe` shares one ctl fid across `join3`-concurrent writers (UB under the old cast; sound under the Mutex). No behavior change.

**Critical check — authorization boundary: CONFIRMED CLEAN (single PEP).** No shadow/homegrown check exists: every `Mount` method takes `caller: &Subject` and threads it through without any `is_anonymous()`/capability inspection; `ExecMount` creates no second enforcement point that could diverge from the MAC reference-monitor (#547). (Confirmed independently that `SandboxPool`/`SandboxBackend`/`PodSandbox` take no `Subject` in their public APIs today, so there's genuinely nothing to forward into yet.)

**Also confirmed:** the `exit` stub is honestly non-blocking (no await/spin/sleep; prominent `TODO(#607)` for the EV7 read-then-subscribe wiring; does NOT fake fan-out with mpsc/oneshot); `ctl` verb parsing is robust (fixed 4-verb set, no panics, `../etc/passwd` → InvalidArgument).

**Verified:** 13/13 exec_mount tests pass (12 + 1 new), 92/92 full-crate, clippy clean. One follow-up noted: rebase onto current base (post-#615) + consider adopting the devfile primitives — the rebase is required at merge time regardless. Marked ready for review.

ewindisch:
✅ **Spine-alignment review (epic #508): aligns — good spine citizen.** `ExecMount` projects the pool as a `/proc`-style tree; Subject threaded with no shadow authz (single PEP at the mount, per #547); `exit` stub honestly awaits EventService (#607). This is the shape every resource projection should follow.

ewindisch:
**Follow-up filed:** the `TODO(#607)` `exit` stub in `exec_mount.rs` is now tracked by **#668** — #607 (EV7) has since CLOSED/landed, so the real blocking-await wiring (read-then-subscribe on the terminal latch) can now be done. This PR's stub was correct-at-the-time; #668 is the un-stub.

Issue #525, opened 2026-06-27
P2: Scheduler decision engine (SandboxPool::acquire) — admission/queue + resource/GPU-aware placement + per-Subject/per-group quotas


ewindisch:
# P2 — Scheduler decision engine (admission + placement at SandboxPool::acquire)

**Epic:** Worker Scheduler. **Depends on:** P1 (candidate directory) + #516 (inventory backend registry). **Fixes:** #519. **🔒 authz/quota sign-off.**

## Goal
Turn `SandboxPool::acquire` from a grab-or-error warm allocator into a real **admission + placement** engine: admit/queue by quota+capacity, rank candidates from P1 (resource + GPU + affinity), place across heterogeneous backends via the inventory registry, **fail-closed** on quota/authz.

## Why
`pool.rs` today: `acquire()` errors immediately on `active ≥ max_sandboxes` (no queue), serves a FIFO warm sandbox or cold-creates, with **no placement, priority, fairness, or resource-awareness**. Warm sandboxes **discard requested CPU/mem** (#519). No multi-tenant fairness anywhere.

## Scope
1. **Admission control + wait-queue** — replace immediate `PoolExhausted` with: admit if within quota + capacity, else **enqueue** (bounded, with backpressure/timeout) or reject. Wire #521's dead `max_concurrent_runs`/timeouts here (or supersede them).
2. **Per-Subject + per-group quotas, fail-closed** — no identity → reject; over-quota → reject/queue; enforced via Casbin + the federation admission gate (#137/#150/#319/#328). Quotas attach to Subject and to groups (incl. the cell-default group).
3. **Resource + GPU-aware ranking** — consume P1 `queryCandidates`; rank by fit (cpu/mem), **free GPU of the requested class** (GPU converges here — replaces DevicePool single-`primary()`), and affinity (cached-model locality, KV/session affinity hooks). Honor `LinuxContainerResources` for real (closes #519 — warm sandboxes must match or be resized/cold-created, not silently default-sized).
4. **Backend placement via the inventory registry (#516)** — pick the backend (Kata/Oci/Cri/Wasm/Nspawn) per workload requirements + availability; reuse `resolve_backend`/the registry as the construction seam.
5. **Single-node degenerate** — one candidate (self); admission+quota still apply; no behavior regression for the common solo case.

## Design notes
[...]

Issue #523, opened 2026-06-27
Epic: Placement & Scheduling


ewindisch:
Standalone epic (cross-links **#341** Worker GA + **#310** Multi-GPU). Builds on **#508/#516** (sandbox taxonomy + inventory backend registry), the `DiscoveryService` crate (`hyprstream-discovery`), and `SandboxPool`. Design source: `docs/plans/2026-06-21-federated-scheduling-dht-pq-spike.md`.

> ⚠️ Ticket-number corrections (verified live): DiscoveryService is the *crate* `hyprstream-discovery` (NOT GH #65 = 'rework cli list'); GH #322 = the (CLOSED, built) inference router, NOT the federated spike — federated design = the spike doc + the sibling epic below.

## Problem
No scheduler exists in the worker stack: `SandboxPool` = grab-warm/create-to-cap/**error-on-exhaustion** (no queue/priority/placement); `WorkflowService` = fire-and-forget spawn (`runs_on` ignored, jobs never touch the pool); `DevicePool` = single `primary()`; resources are pass-through (warm sandboxes **discard** requested CPU/mem — #519); cross-cell placement is design-only.

## Settled design
1. **Cell = soft fairness/authz/locality group, NOT a hard scheduling wall** (iroh NodeIds are global). **Single machine = degenerate single-node cell.**
2. **Global-by-design resource model, started cellular:** nodes advertise **signed, TTL'd** resources (capacity, GPU+class, load, cached-models) keyed by global iroh NodeId + DID.
3. **Dynamic placement groups:** cell = default soft group; implicit/query-groups for the common case; explicit consent-based group records for owned/cross-authority/quota'd groups.
4. **Bounded candidate selection:** query directory by {group + resource + authz} → bounded set → rank (HRW/least-loaded/affinity). Tiering = scaling technique, not a wall.
5. **Bootstrap substrate = `DiscoveryService`** (already the cell's signed federation directory; cell-scoped by construction).
6. **Non-exclusion guarantees (day 1):** global identity in records · signed records · `queryCandidates` = the federation seam · authz-prefilter in every query.
7. **GPU converges** as a scheduled resource dimension (unifies #310/#313/#314; replaces single `primary()`).
8. **Jobs unify** as scheduled workloads (#349; `runs_on` → placement).
9. **Fairness = per-Subject + per-group (incl. per-cell) quotas, fail-closed** (admission gate + Casbin).

## Phases (sub-tickets created below; each standalone)
[...]

PR #624, opened 2026-06-30
feat(#341 T2-C): backend auto-selection diagnostics + wizard data API

ewindisch:
Implements the backend-local slice of **#348** (T2-C of epic **#341** "Worker GA"): selection **diagnostics** on top of the existing fail-closed `BackendRegistration{name, priority, auto_selectable, is_available, construct}` inventory spine (`crates/hyprstream-workers/src/runtime/selection.rs`). Stacked on #599.

## What's added (0 changes to existing selection logic)
- **`unavailable_reason(name) -> String`** — per-backend prerequisite sub-reasons (`kata`→"cloud-hypervisor not on PATH", `nspawn`→"systemd-nspawn/machinectl not on PATH", `podman`→"podman not on PATH", generic fallback otherwise). Backend-domain knowledge, belongs in this crate.
- **`list_backends_for_wizard() -> Vec<BackendStatus>`** + `BackendStatus` — registry status snapshot sorted by priority, ready for a future `hyprstream service backends` CLI table (no CLI subcommand wired — out of scope, just the data function).

## Descoped: the explain trace is substrate-owned
An earlier revision of this PR also added a `SelectionReport`/`CandidateOutcome`/`explain_selection` explain trace. That shape is **owned by the #628 scheduling substrate** (PR #630), which defines the canonical generic `SelectionReport<C>` + `CandidateOutcome<C>` + `RejectionReason` + `filter`/`rank`/`select` in `hyprstream-discovery::scheduling`. Shipping a second, backend-specific copy here would be exactly the duplication #628 exists to prevent — so it has been dropped. `selection.rs` will **adopt** the substrate's generic `SelectionReport<C>` in a follow-up once the substrate reaches `main` (the #628 acceptance names `selection.rs` as the first consumer to migrate). `unavailable_reason` remains here and will plug into the substrate as one backend's `RejectionReason` source.

## Verification
`cargo check/test -p hyprstream-workers` — **20/20 selection tests pass** (the dropped 6 were the `explain_selection`/`SelectionReport` tests; all fail-closed auto-degradation, explicit-name, and wizard-list tests remain and pass). Clippy clean. Torch-free; only `crates/hyprstream-workers/` touched (173 insertions).

Refs #341 · #348 · adopts-later #628 · stacked on #599.

### DIFF
--- a/crates/hyprstream-workers/src/lib.rs
+++ b/crates/hyprstream-workers/src/lib.rs
@@ -71,6 +71,8 @@ pub use error::WorkerError;
 pub use runtime::{WorkerService, SandboxBackend, SandboxHandle, NspawnBackend, NspawnConfig};
 // Inventory-based backend registry + fail-closed selection spine (#507)
 pub use runtime::{resolve_backend, BackendCtx, BackendRegistration};
+// Wizard/CLI selection diagnostics (#348)
+pub use runtime::{list_backends_for_wizard, BackendStatus};
 #[cfg(feature = "kata-vm")]
 pub use runtime::KataBackend;
 // The image-store trait seam is always available; only the concrete RAFS impl
--- a/crates/hyprstream-workers/src/runtime/mod.rs
+++ b/crates/hyprstream-workers/src/runtime/mod.rs
@@ -96,6 +96,8 @@ pub use nspawn::{NspawnBackend, NspawnConfig, NspawnHandle};
 pub use wasm_backend::{WasmBackend, WasmConfig, WasmHandle};
 // Inventory-based backend registry + fail-closed selection spine (#507)
 pub use selection::{resolve_backend, BackendCtx, BackendRegistration};
+// Wizard/CLI selection diagnostics (#348)
+pub use selection::{list_backends_for_wizard, BackendStatus};
 // Domain entities (business logic only)
 pub use container::Container;
 pub use pool::{PoolStats, SandboxPool};
--- a/crates/hyprstream-workers/src/runtime/selection.rs
+++ b/crates/hyprstream-workers/src/runtime/selection.rs
@@ -207,6 +207,80 @@ pub fn resolve_backend(name: &str, ctx: &BackendCtx) -> Result<Arc<dyn SandboxBa
     })
 }
 
+// ─────────────────────────────────────────────────────────────────────────────
+// Wizard / CLI diagnostics (#348)
+// ─────────────────────────────────────────────────────────────────────────────
+//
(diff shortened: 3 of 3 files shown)


ewindisch:
**Self-review pass — clean, no fixes needed. Marked ready.**

Independently verified: `select_registration`/`resolve_backend` byte-identical pre/post-diff (the new code is purely additive diagnostics); `wasm`'s `auto_selectable:false` untouched (not even in the changed-file list); the new `explain_selection`/`unavailable_reason`/`list_backends_for_wizard` functions are read-only/informational (nothing in the real selection logic reads their output); the podman/kata/nspawn reason strings cross-checked against the actual `registry_is_available()` implementations and confirmed accurate. 26/26 tests pass, clippy clean.

**Architectural note (not a blocker for this PR):** the underlying `selection.rs` `auto`+priority model this diagnostics is a **stopgap** — the real replacement is the requirements/preferences-driven scheduler in #525, where interface class (Container vs Wasm) is a hard filter, not a rank, and backends compose rather than mutually-exclude. That framing is captured on #525. This PR is fine to merge as-is: it's additive, doesn't change selection behavior, and `explain_selection`-style diagnostics survive the transition (re-skoped to 'placement decision' rather than 'backend menu').

ewindisch:
**Scheduling Substrate (#628) — this PR will be reworked onto it (tracked in #629).**
Per the scheduling-dedup design (epic #523 P0 = #628): #624's `SelectionReport`/`CandidateOutcome`/`explain_selection` should LIFT into the shared substrate (generalized) rather than live workers-local; `selection.rs`'s filtering should route through the shared `filter→rank→select` helper; and `unavailable_reason` should unify with each backend's `registry_is_available()` (one truth). Rework is filed as **#629**, gated on #628 landing. The current #624 is fine to merge as-is for the Worker-GA wizard UX — the rework is a non-breaking follow-up that adopts the substrate once it exists.

ewindisch:
✅ **Spine-alignment (epic #508): aligns** — additive diagnostics, stopgap before #525's requirements/preferences-driven scheduler (interface class as hard filter, not a ranked menu — already captured on #525).

ewindisch:
Superseded — closing per operator decision. Backend selection is a **scheduler concern**, not a standalone auto-selection-diagnostics + wizard surface. The correct approach is resource-requirements-driven selection through the #628 scheduling substrate (availability + CapabilitySource as predicates; explain-for-free) — captured in #659's core commit (`d8c56cb23`), to land cleanly against main once the substrate + workers stack converge. The wizard data API (`list_backends_for_wizard`/`BackendStatus`) had zero consumers and is dropped.

Issue #633, opened 2026-07-01
universal mountable filesystem service: extract OCI/RAFS off the kata backend


ewindisch:
Child of #508 (the spine). Step 2 of the spine sequence.

## Goal
Make OCI/RAFS image provisioning a **mountable filesystem service** — a 9P `FsMount` exposing a file tree (the image), composable by mount/bind anywhere in a namespace. Extract it off the kata backend, where it's currently private (gated behind the `kata-vm` cargo feature, only consumed by `kata_backend.rs`).

## Why
Under the spine, a filesystem service (OCI/RAFS image, host fs, synthetic) is a peer mountable file tree; "root" is a *mount position*, not a type. Today the image→RAFS→`RootfsMount` machinery (FS-B/C: `image/rafs_builder.rs`, `RootfsMount` = OverlayFs-over-RAFS) is kata-private, so every non-kata backend is non-functional or side-channels its own rootfs (podman #617) — the "unified taxonomy" is unified in trait-shape only.

## Scope
- Move the nydus/rafs deps off the `kata-vm` feature onto their own (e.g. `oci-image` / `filesystem-service`).
- Rename `RootfsMount` off the Linux-ism (it's an image filesystem service, named for where it happens to mount) — e.g. `ImageFs` / `OciImageMount`.
- kata shrinks to "boot VM + serve the composed namespace via virtio-fs"; it stops owning image pull.
- podman/nspawn compose this service into their container namespace (replaces the side-channel).
- wasm can mount an OCI-wasm module through the same layer.

Ref: #620 (nydus dep pin) is the dep-hygiene prerequisite, already done.

ewindisch:
Implemented — PR (draft, stacked on #599). New `oci-image` feature carries the nydus/RAFS deps off `kata-vm` (kata-vm now implies it); the image machinery compiles standalone without the VM toolchain (`cargo tree -i kata-hypervisor` empty under oci-image). Renamed `RootfsMount`→`ImageFs` (root = mount position, not a type). clippy clean, torch-free. Deferred: podman/nspawn/wasm rewiring.

ewindisch:
[...]

Issue #632, opened 2026-07-01
universal namespace-composition + delivery layer (the spine's missing center)


ewindisch:
Child of #508 (the spine). **The load-bearing new work.**

## The gap
Nobody composes the full per-sandbox namespace and delivers it to a guest today. Profile B delivers a VFS-Mount-as-WASI-fs to a wasm *stub*; kata delivers RAFS-rootfs-only via virtio-fs/`nydusd`. The rich namespace (`/lang/*`, `/exec`, `/bin`, `/env`) is host-side in-process only. "Deliver the composed namespace to the guest OS over a transport" is the spine's missing center.

## Goal
One reusable **standard namespace composition** (the Plan9 `namespace` recipe: mount filesystem services + shells + tasks + synthetic trees into a per-sandbox VFS), and the machinery to **deliver** that composed namespace to a guest OS over the backend's transport (virtio-fs for kata; bind/passthrough for containers; host-imports for wasm). Replaces the two duplicated hand-written builders (`cli/shell_handlers.rs` + `tui/vfs.rs`).

## Why
This is what makes "the namespace IS the OS surface" real — without it, every backend either lacks the namespace (kata rootfs-only), bypasses it (podman side-channel), or has no guest to deliver it to (wasm stub). It's the layer the filesystem-service, shell-Mount, and backend-narrowing work all compose onto.

Depends on: universal filesystem service (OCI/RAFS), self-contained shell Mounts. Consumed by: every backend. Enforced by: #547 (MAC at mount).

ewindisch:
**Coupling clarification (operator review, #508 decision):** this (#632, namespace-composition + delivery) and #633 (image-as-Mount) are **one coupled change**, not independent. The image must be a Mount AND mounted into the delivered VFS. The artificial split (hollow #633 rename + real #632 delivery) is what produced the minimal #637. They may land as separate commits but the design is one thing — the guest gets a 9P VFS with the image mounted in it.

ewindisch:
Status audit (2026-07-03): the epic's central gap — one unified namespace-composition recipe replacing the two duplicated hand-written builders — is still genuinely unbuilt. `crates/hyprstream/src/cli/shell_handlers.rs` (`Namespace::new()` + its own bind_mount sequence) and `crates/hyprstream/src/tui/vfs.rs` (`build_chat_vfs_namespace`, a separate `Namespace::new()` + bind_mount sequence) still don't share a builder.

[...]

Issue #370, opened 2026-06-19
FS-NATIVE-COW: copy-up + whiteout + opaque-dir policy on the Namespace union bind-mount (NOT a new overlay primitive)


ewindisch:
**Re-scoped 2026-07-01.** Originally "native CoW overlay behind the FsMount trait" (implying a new overlay-FsMount primitive). The re-cut: overlay is **NOT a new composition type** — it's a write/readdir policy on the existing Plan 9 `Namespace` bind-mount union. See design ADR (linked below).

## Context (why not a new primitive)

The `Namespace` already does Plan 9 bind-mount union semantics natively (`crates/hyprstream-vfs/src/namespace.rs`):
- `bind_mount(prefix, target, BindFlag::Before|After|Replace)` stacks `Arc<dyn Mount>` targets at a prefix.
- `cat`/`read_one`/`readdir` try each target in bind order; readdir merges + dedupes by name.
- `write`/`ctl` try each target in bind order (fallthrough-on-error).
- `Subject` threads through every call. No fuse. This is the spine.

A new `OverlayMount` type (or the existing fuse `OverlayFs` as the composition layer) would be a **parallel special-case** on top of bind-mounting — exactly the drift we are avoiding. The fuse `OverlayFs` path stays as a *backend you can bind in* for the sandbox-rootfs case (battle-tested POSIX CoW), not the composition mechanism.

## The narrow, real gap

Today's union does read-merge + fallthrough-write. It does NOT do:
1. **Copy-up on write** — when a write targets a file that exists only in a read-only lower, fetch the lower's content and redirect the write to a writable upper (so the lower is never mutated and the write modifies the existing file rather than `create`-ing a new upper entry).
2. **Whiteout** — a deleted-in-upper entry must not re-appear from a lower on readdir merge.
3. **Opaque dirs** — an upper dir marked opaque suppresses lower entries for that dir.

## Deliverable (small-and-functional → POSIX-compliant)

A `UnionPolicy` the `Namespace` consults in its `write`/`create`/`readdir` paths, keyed on which targets at a prefix are read-only vs. writable-upper. Plus a **ctl surface** (Plan 9 control files) per union mount: which target is the copy-up upper, whiteout/opaque flags.

- **v1 (ship-fast):** copy-up on write to a read-only lower → writable upper. No whiteout/opaque. Covers the per-tenant-delta + sandbox-writable-upper cases.
[...]

PR #641, opened 2026-07-01
feat(#633): image as real 9P Mount + universal compose-and-serve path off kata

ewindisch:
Implements **#633** (child of #508, spine step 2) — **REBUILT**: the image is a real `hyprstream_vfs::Mount` composed into the delivered namespace; the compose+serve path is universal, not kata-private. Stacked on #599.

> **Rebuilt from the minimal version** (operator review: rename + image-deps-only gate was not the correct change). The base already had the correct Mount architecture from FS-A/B/C/D (`FuseFileSystemMount<F>: Mount`, `sandbox_fs` composing the image into a namespace, `compose_and_serve_vfs` serving it to the guest). The real structural fix was gating: the compose+serve path was locked behind `kata-vm`, so only kata could deliver the image Mount to a guest. This PR moves it onto `oci-image`.

## What changed (the correct universalization)
1. **`oci-image` feature now carries the full compose+serve path** — not just the nydus deps. `sandbox_fs` (composes the image Mount + injected mounts into a namespace) and `hyprstream-vfs-server` (the FS-A vhost-user-fs transport that serves the composed namespace to a guest) are now on `oci-image`, not `kata-vm`. So **any backend can compose the image into a namespace and serve it to a guest** — that's the spine (#508): "we provide the 9P VFS to the workers."
2. **`kata-vm` implies `oci-image`** and adds only the pure-VM toolchain (`kata-hypervisor`/`kata-types`/`ch-config`).
3. **The image IS a real `Mount`** — `ImageFs: FuseFileSystemMount<OverlayFs>` which is `FsMount: Mount`. Proven by `sandbox_fs::compose` doing `namespace.mount("/", Arc::new(image_fs))` as `Arc<dyn Mount>`, and by `compose_and_serve_vfs` serving that composed namespace to the guest over vhost-user-fs. No bespoke RAFS share; the guest receives the composed VFS with the image as one Mount in it.
4. Kata's `compose_and_serve_vfs` wiring is **unchanged** from the base (it was already correct).

## Verified
`--features oci-image` builds standalone (no VM toolchain); 95 tests pass incl. `down_adapter_serves_composed_namespace`, `composed_namespace_exposes_rootfs_and_injected`, `two_sandboxes_are_isolated`, `image_fs::reads_from_rafs_lower_and_copies_up_on_write`. `--features kata-vm` still builds (kata unchanged). clippy clean. torch-free.

## What this enables (the spine payoff)
Podman/nspawn/wasm can now compose an image Mount into their per-sandbox namespace and serve it — using the same path kata already uses — rather than side-channelling their own rootfs (#617) or lacking one (wasm). That re-grounding is the follow-on.

Refs #508 · #633 · stacked on #599.


ewindisch:
**Cleanup applied (commit `d71fd6c0e`).** Per review Q1-Q3: (1) `rafs_store` is now `Option<RafsStoreOpt>` — feature-invariant signature, no more viral cfg; (2) ImageHandler cfg/cfg(not) mirror collapsed to one body per method (Option-based); (3) `SandboxVirtiofs` (370 lines, external-nydusd) deleted — provably dead (kata uses composed VFS, virtiofs_daemon=None on every path). TODO added to image/store.rs for the CAS-as-9P-service follow-on. 95 tests pass, clippy clean.

ewindisch:
**Inventory-pluggable ImageStore refactor landed (commit `194161197`).** Per your Q1-Q3 review: the image backend is now a **trait object dispatched like `SandboxBackend`** — `ImageStore` trait (`image/store_trait.rs`, always compiled); `WorkerService` holds `Option<Arc<dyn ImageStore>>`; `ImageHandler` dispatches through it with ONE fallback error (`image_store_missing`), not 5 mirrored bodies; `RafsStore` impls `ImageStore` + registers via `inventory::submit!{ ImageBackendRegistration }`. Future non-RAFS backends (plain-OCI, stargz, squashfs) plug in with one impl + one submit, zero `WorkerService` changes. The cfg-mirror anti-pattern is dissolved. 95 tests pass, clippy clean. (#646 filed+closed for tracking.)

ewindisch (commented):
Design-call follow-up (2026-07-01): the fuse-overlay-as-9P path this PR ships for the sandbox rootfs is **acceptable as a leaf backend** but is NOT the composition layer — see ADR #651.

**The concern (now resolved by the ADR):** `FuseFileSystemMount<OverlayFs>` composes at the fuse layer, which loses `&Subject` (fuse `FileSystem` ops have no `Subject` param). For the sandbox-rootfs case (single local RAFS lower, local writable upper) that's fine — ship it. But it must NOT be extended into the service-over-service / federation composition vision, because a 9P-`Mount` service lower would require a `Mount → fuse → Mount` round-trip that drops the policy principal.

**The locked direction:** overlay is a **write/readdir policy on the existing `Namespace` bind-mount union** (#370 re-scoped), not a new primitive. `bind_mount(prefix, target, BindFlag::Before|After)` over `Arc<dyn Mount>` — with `Subject` on every call — is the ONE composition mechanic for the whole system (services, federation, deltas, snapshots). The fuse `OverlayFs` path stays as a bind-in leaf for the POSIX-CoW rootfs case.

**Action on this PR:** no code change required to merge — the sandbox-rootfs fuse path is correct for its scope. But the module doc on `crates/hyprstream-vfs/src/overlay.rs` should be updated (follow-up commit) to mark it explicitly as a leaf backend, not the composition layer, so a future implementer doesn't extend composition here. I'll open that as a follow-up if this merges as-is.

Ref: #370 (copy-up policy), #650 (POSIX whiteout/opaque v2), #651 (ADR), #632 (universal namespace-composition spine).

Issue #544, opened 2026-06-27
feat(9p-rpc) T5: registry pilot — generated face + FsHandler parity


ewindisch:
**Epic:** 9P-RPC (#539)  **Pilot 1 — registry.** **Depends on:** T2 (generated projection); T4 for auto-mount (can manual-mount earlier). 

## Goal
Validate the generated projection against the service that already has both a hand-written FS (Mechanism B) and `$docExample`-encoded paths — proving the generated paths are identical to today's documented convention, with no regression to the git-worktree FS.

## Context
`crates/hyprstream-rpc-std/schema/registry.capnp` imports all `Np*` types (`:9-26`) and documents its file mapping in `$docExample` prose (`list`→`ls`, `getByName`→`cat {name}`, `clone`→`ctl … clone`, `healthCheck`→`cat health`). Server-side, the **leaf** FS semantics are hand-written: `crates/hyprstream/src/services/registry.rs` `FsHandler` + fid table (`alloc_fid :197`, `FidEntry` states `:116-127`) using `ContainedFs`/`FsHandle` from `hyprstream_containedfs` (`:8,:312`) → real git-worktree files. **This stays** (the git backing cannot be generated — Mechanism B).

## Changes
- Apply the schema diff (epic): replace `$docExample` prose with structured `$vfsKind`/`$vfsPath` on `RegistryRequest` methods; mark `repo @9 :RepositoryRequest` as nested subdir `repo/{repoId}`.
- ADD the generic generated face (Mechanism A) for the non-fs methods (`list`, `getByName`, `clone`, `register`, `remove`, `healthCheck`) via the T2 projection.
- KEEP the hand-written `FsHandler` for the `fs`-scoped worktree leaf ops (Mechanism B) — selected via `$vfsKind` (real-FS backing).
- Auto-derive man-page text from `$vfs*` (drop the now-redundant `$docExample` prose, or keep only where it adds narrative).

## Acceptance criteria
- [ ] Generated mount exposes `list`/`{name}`/`health`/`ctl` exactly as the old `$docExample` prose described (path parity test).
- [ ] `repo/{repoId}/...` projects as a subdirectory with `branches`/`status`/`head`/`ctl` etc.
- [ ] The git-worktree FS (`FsHandler`) behavior is byte-for-byte unchanged (regression test against pre-epic behavior).
[...]

Issue #542, opened 2026-06-27
feat(9p-rpc) T3: native Send+Sync ServiceMount port


ewindisch:
**Epic:** 9P-RPC (#539)  **Slice (c):** native port — the real lift. **Depends on:** none (parallel to T1/T2). **Blocks:** T4, T6.

## Goal
Make the generic `ServiceMount` adapter compile and run natively as `Send + Sync`, so it can replace the hand-written native mounts. This is the highest-risk slice.

## Context
`ServiceMount` (`crates/hyprstream-rpc-std/src/vfs_mount.rs`) is a generic mount driven by generated `dispatch()` / `schema_metadata()` / `render_doc()` — serves any service via `/srv/{name}` (ctl=mutations, cat=queries, `/doc`). It already works, but is gated `#![cfg(target_arch = "wasm32")]` (`vfs_mount.rs:11`) and uses `RefCell` for its fid table (single-threaded browser assumption). Native VFS (`hyprstream-vfs` FS-A down-adapter) runs on real OS threads and requires genuinely `Send + Sync` mounts. The hand-written native model mount `crates/hyprstream/src/services/remote_mount.rs` (`RemoteModelMount`) shows the native pattern: `DashMap` fid table + dedicated tokio runtime wrapping `ModelClient`/`ModelFsClient`.

## Changes
- `rpc-std/src/vfs_mount.rs:11`: drop the `cfg(target_arch = "wasm32")` gate (or split into a shared core + thin wasm/native wrappers).
- Replace `RefCell` fid state with `Mutex`/`DashMap` (mirror `RemoteModelMount`).
- Ensure the mount implements the native `Mount` trait (`crates/hyprstream-vfs/src/mount.rs`: `walk/open/read/write/readdir/stat/clunk`, each `caller: &Subject`) — and `FsMount` where `create/remove` apply.
- Honor `$vfsBulk` nodes: `read` returns raw bytes (no capnp render) to preserve the DMA path.
- Keep `Subject`-per-call authz: map the node's `scope` (`MethodMeta`) → the existing `mcpScope` authorization check, same as the RPC handler.

## Acceptance criteria
- [ ] `ServiceMount` compiles for native (`x86_64-linux`) and still for `wasm32`.
- [ ] `ServiceMount: Send + Sync`; usable from the multi-threaded `hyprstream-vfs` down-adapter.
[...]

Issue #541, opened 2026-06-27
feat(9p-rpc) T2: vfs::generate_mount projection + compile-time guards


ewindisch:
**Epic:** 9P-RPC (#539)  **Slice (b):** the generated projection. **Depends on:** T1. **Blocks:** T4, T5.

## Goal
Add `vfs::generate_mount` as a 4th codegen projection: per service, emit the node table + `ServiceMount` tie-in (the dispatch/authz/doc engine already exists). Enforce the read/write-vs-scope guard at compile time.

## Context
`generate_rpc_service!` orchestrator is `crates/hyprstream-rpc-derive/src/codegen/mod.rs`. Each surface is a discrete `generate_*(...) -> TokenStream` composed in a final `quote!`. `generate_service` (`mod.rs:88-125`) emits `#client_struct #service_traits #handler #metadata_code …`. The wasm-safe composer is `generate_client_only` (`mod.rs:18-86`). The closest template is the MCP/metadata surface: `codegen/metadata.rs:40-75` builds `schema_metadata() -> (&str, &[MethodMeta])` from `UnionVariant`; runtime `MethodMeta` is `crates/hyprstream-rpc/src/_metadata.rs:26-44`. The VFS projection consumes the **same** `MethodMeta` (scope→authz, doc_example→man) plus the new `vfs_*` fields — no new metadata plumbing.

## Mapping rules (method shape → node; annotation overrides)
```
query, Void, "list"  → dir (readdir)        query, Void          → file (read=resp)
query, Text/Id       → file @ {arg}         write|manage, struct → ctl verb
streaming            → stream file          query w/ rich args   → query file (?q=)
nested request struct→ subdir {id}/
GUARD: write|manage ↛ read-only file; query ↛ ctl  ⇒ compile error.
```

## Changes
- NEW `crates/hyprstream-rpc-derive/src/codegen/vfs.rs`: `pub fn generate_mount(svc: &str, resolved: &ResolvedService, types_crate: &TokenStream) -> TokenStream`. Builds `VfsNode` literals from each `UnionVariant` (`name`, `scope`, `vfs_kind` (inferred if absent via mapping rules), `vfs_path`, `vfs_bulk`); emits the per-service `ServiceMount` impl/registration (node table + tie-in to the generic engine from T3).
- `codegen/mod.rs:88-125`: append `#vfs_mount` to `generate_service`.
[...]

Issue #540, opened 2026-06-27
feat(9p-rpc) T1: VFS annotation set + rpc-build metadata capture


ewindisch:
**Epic:** 9P-RPC — generated 9P/VFS projection for all capnp services (#539)
**Slice (a):** annotation set + metadata capture. **Depends on:** none. **Blocks:** T2.

## Goal
Add the `$vfs*` annotation vocabulary and capture it into the codegen metadata model, so a later projection (T2) can emit a `Mount` from each service's schema. Mechanical, follows the existing `$mcpScope`/`$docExample` pattern exactly.

## Context
Method metadata is already first-class: `UnionVariant` (`crates/hyprstream-rpc-build/src/schema/types.rs:27-40`) carries `scope` (from `$mcpScope`), `doc_example` (from `$docExample`), `cli_hidden`. Capture is uniform in `crates/hyprstream-rpc-build/src/schema/cgr_reader.rs`: `find_annotation_id(...)` (`:119-127`) + thread id through variant/field parsers (`:164-218`) + `extract_annotation_text` (`:358`). Adding an annotation = one `find_annotation_id` + thread + one `extract_*` per annotation.

## Changes
- `crates/hyprstream-rpc/schema/annotations.capnp`: add
  ```
  enum VfsKind { file @0; dir @1; ctl @2; stream @3; query @4; }
  annotation vfsPath(field, struct) :Text;   # mount-relative, fid-0 rooted; {braces}=param segs
  annotation vfsKind(field) :VfsKind;
  annotation vfsBulk(field) :Void;
  annotation vfsHidden(field) :Void;
  ```
- `rpc-build/schema/types.rs:27-40`: `UnionVariant += vfs_path: String, vfs_kind: Option<VfsKind>, vfs_bulk: bool, vfs_hidden: bool`. Add a `VfsKind` enum mirror in the metadata types. (`FieldDef` `:110-133` gets `vfs_path` too if struct-level paths are needed for nested-resource dirs.)
- `rpc-build/schema/cgr_reader.rs`: `find_annotation_id` for each new annotation (`:119-127`), thread through the variant/field parsers (`:164-218`), pull values via `extract_annotation_text` / presence-check for Void annotations (`:358`).

## Schema-alignment (break window — D2/D1)
- Make `$mcpScope` **mandatory**: emit a build error in `cgr_reader` (or the resolve step) if a method union variant lacks a scope. No silently-public files.
[...]

PR #664, opened 2026-07-02
feat(9p-rpc) T2 (#541): vfs::generate_mount projection + compile-time guards

ewindisch:
Implements **T2 (#541)** of the 9P-RPC generated-projection epic (#539). Adds `vfs::generate_mount` as the 4th codegen projection (alongside client/handler/dispatch/metadata): each service emits a `vfs_nodes()` table — the machine-readable 9P/VFS node projection the runtime mount engine (T3) consumes.

## Base
**Stacked on `ewindisch/540-vfs-annotations` (T1's PR #661), NOT the epic branch** — T1 has not merged into `feature/9p-rpc-epic` yet, so this branch carries T1's commit too. Rebase onto the epic branch once T1 lands (should be conflict-free — disjoint files). Targets `feature/9p-rpc-epic`.

## What's added (+568, 3 files)
- **`codegen/vfs.rs`** (NEW) — `generate_mount(service, resolved)` builds a `VfsNode` per method from its `UnionVariant` + the T1 `$vfs*` fields.
- **`codegen/mod.rs`** — composes `vfs_mount` into both `generate_service` and `generate_client_only` (VFS must compile to wasm32).
- **`hyprstream-rpc/_metadata.rs`** — runtime `VfsNode` / `VfsNodeKind` types (mirror the `MethodMeta`/`ScopedClientTreeNode` pattern; pure `&'static str`/`bool`/enum, wasm-safe).

## Mapping rules (method shape → node kind; `$vfsKind` overrides)
| shape | kind | default path |
|-------|------|--------------|
| query, Void, `list*` | `Dir` | `.` |
| query, scalar arg (Text/int/enum) | `File` | `{arg}` (binds the segment) |
| query, Void | `File` | method name (e.g. `health`) |
| query, struct / rich args | `Query` | method name |
| write \| manage | `Ctl` | `ctl` |
| streaming | `Stream` | method name |

`$vfsHidden` excludes a method; `$vfsBulk` marks the node for raw-byte reads (DMA path, engine honors in T3). Paths are mount-relative (fid-0 rooted).

## Compile-time guard
A read-only kind (`file`/`dir`/`query`) on a write/manage scope, or a `ctl` on a query scope, is a contradiction — a 9P read must be side-effect-free. `generate_mount` emits a `compile_error!` naming the offending method. Exact messages:

> `method \`clone\`: write/manage scope \`write\` cannot project to a read-only VFS node (\`File\`) — a 9P read must be side-effect-free; use \`$vfsKind(ctl)\` or \`$vfsKind(stream)\``

> `method \`get_by_name\`: query scope \`query\` cannot project to a \`ctl\` node — a control-file write invokes the method; use a read kind (\`$vfsKind(file|dir|query)\`) or give the method a write scope`

An unknown `$vfsKind("bogus")` is also a `compile_error!`.

## Verified
- `cargo build -p hyprstream-rpc-derive` — clean
- `cargo build -p hyprstream-rpc` — clean (**end-to-end: the macro runs `generate_mount` over every real service schema via build.rs**)
- `cargo test -p hyprstream-rpc-derive` — **26 passed** (15 new vfs: inference, both guard directions, bad-annotation, token generation, `$vfsHidden` exclusion, data-only-no-mount; 11 pre-existing, no regressions)
- `cargo clippy` — clean on both crates
- **wasm32:** the generated `vfs_nodes()` is pure/wasm-safe; a full `wasm32-unknown-unknown` build hits a **pre-existing** `web_sys::WebTransport` error (needs `--cfg=web_sys_unstable_apis`) — confirmed it reproduces on the base branch with my changes stashed, so not a T2 regression, and no wasm error touches the vfs/metadata code.

## Deferred (correctly, per epic sequencing)
- Real-schema annotation (registry.capnp `$vfsPath`/`$vfsKind`) = **T5** (#544)
- Runtime mount-engine consumption of `vfs_nodes()` = **T3** (#542)
- Nested request structs as subdirectories: scoped clients are skipped in the top-level table (they get their own tables); the recursive subdir pass is a small follow-on within T3/T5's wiring.
- `$mcpScope`-mandatory (D2) — out of scope, stayed additive.

## Ambiguities resolved
- **Scope→mutation classification**: treated `query`/`read`/empty(`$scopeExempt`) as read; everything else as mutation. Empty scope is least-authority (read), so a `ctl` on an exempt method is still rejected.
- **`{arg}` binding**: only single-scalar read args get `{arg}`; struct args become a `Query` file (write-spec-then-read) rather than a path param.

Refs #539 · #541 · consumes T1 (#540).

### DIFF
--- a/crates/hyprstream-rpc-build/src/schema/cgr_reader.rs
+++ b/crates/hyprstream-rpc-build/src/schema/cgr_reader.rs
@@ -551,6 +551,82 @@ fn has_annotation(
     false
 }
 
+/// Resolve the annotation node id whose short name matches `short_name`.
+///
+/// Self-contained lookup used for the VFS projection annotations (epic #539),
+/// mirroring the inline `scopeExempt` id resolution — avoids threading a new id
+/// through every extraction signature for annotations captured in one place.
+fn annotation_id_by_short_name(
+    node_map: &BTreeMap<u64, NodeInfo>,
+    short_name: &str,
+) -> Option<u64> {
+    node_map
+        .values()
+        .find(|info| info.short_name == short_name)
+        .map(|info| info.id)
+}
+
+/// Extract a `VfsKind` enum annotation as its enumerant name (e.g. "ctl").
+///
+/// Resolves the annotation's enum type from the node graph and maps the stored
+/// ordinal back to the enumerant name, mirroring `extract_annotation_enum`.
+/// Empty string when the annotation is absent.
+fn extract_vfs_kind(
+    annotations: capnp::struct_list::Reader<capnp::schema_capnp::annotation::Owned>,
+    target_id: Option<u64>,
+    nodes: &capnp::struct_list::Reader<capnp::schema_capnp::node::Owned>,
+    node_map: &BTreeMap<u64, NodeInfo>,
+) -> String {
+    let target_id = match target_id {
+        Some(id) => id,
+        None => return String::new(),
+    };
+    for i in 0..annotations.len() {
+        let ann = annotations.get(i);
+        if ann.get_id() != target_id {
+            continue;
+        }
+        let Ok(value) = ann.get_value() else { continue };
+        if let Ok(capnp::schema_capnp::value::Enum(ordinal)) = value.which() {
+            let Some(ann_info) = node_map.get(&target_id) else {
+                continue;
+            };
+            let ann_node = nodes.get(ann_info.index);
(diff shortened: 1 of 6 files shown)


ewindisch:
**Follow-up filed:** #669 — T2 inference should disambiguate multiple `Void`-arg listings in one struct (they currently all infer to `.` and collide) with a name-based dir-default + a compile-time collision guard. Surfaced by the T5 registry pilot (#667).

Issue #539, opened 2026-06-27
Epic: 9P-RPC — generated 9P/VFS projection for all capnp services


ewindisch:
Tracking epic to make the **9P/VFS file surface a generated, automatic projection of every capnp RPC service** — derived from one annotated schema, alongside the existing typed-client / handler / dispatch / MCP-tool / man-page surfaces. Today the file surface is hand-built per service (native) or doc-only prose; this epic makes it fall out of codegen.

**Branch:** `feature/9p-rpc-epic` (off `main`). All epic PRs target this branch; merging it closes the epic. The full plan lives in this issue (no separate plan doc).

**Schema breakage is in scope** — we use the break window to make `$mcpScope` mandatory, enforce read/write-vs-scope at compile time, and set stable `$vfsPath` names deliberately.

## Scope

This is the **universal file surface for all hyprstream services** — the same generated projection underpins the distributed-web / Wanix namespace and every native and wasm service alike. It is not built for, or driven by, any single consumer; the generated mount + auto-wire apply uniformly to the whole service set.

One **example downstream consumer** (illustrative, not a driver of this epic): the data-plane store — the service currently called *MetricsService*, to be renamed — rides this projection for its control surface (generic-face `ctl seal|compact|set-retention`, a `query` file for recall) and adds a Mechanism-B `fs` leaf whose **write path is 9p**: producers `open now/{shard}` (OAPPEND) and `write` batched **Arrow IPC** record batches (not per-sample capnp), landing in the writable upper layer (+WAL) and sealing to content-addressed Parquet. That consumer additionally depends on the native CoW engine (#370). Mentioned here only as one example of how a service builds on this surface; the epic stands on its own.

---

## Core finding (why this is a *port*, not a *build*)

A generic codegen-driven capnp→9p adapter **already exists**: `ServiceMount` (`crates/hyprstream-rpc-std/src/vfs_mount.rs`) serves *any* service via `/srv/{name}` (ctl=mutations, cat=queries, `/doc`=man-page), driven entirely by the generated `dispatch()` / `schema_metadata()` / `render_doc()`. **But it is `#![cfg(target_arch = "wasm32")]`-only** (`vfs_mount.rs:11`). Native uses hand-written per-service mounts (`RemoteModelMount`, registry's `FsHandler`).

Additionally, the 9p mapping is **already written down per method** — `registry.capnp` imports every `Np*` type (`registry.capnp:9-26`) and each method documents its file mapping in `$docExample` prose:

```
[...]

PR #667, opened 2026-07-02
feat(9p-rpc) T5 (#544): registry pilot — generated face + FsHandler parity

ewindisch:
**Pilot 1 (registry) — the empirical validation of T2's generated projection.** Proves the generated `$vfsKind`/`$vfsPath` paths equal the pre-epic `$docExample` convention, with **zero** change to the hand-written git-worktree FS.

## Parity result (THE key signal)
Dumped the current inferred table, then annotated. Inference alone produced ~90% of the documented convention — but with two **real mismatches** the annotations exist to fix:

| method | `$docExample` said | inference gave | annotated → |
|--------|--------------------|----------------|-------------|
| `list` | `ls /srv/registry` (`.`, dir) | `.` dir ✓ | `$vfsKind(dir) $vfsPath(".")` |
| `getByName` | `cat …/my-model` (`{name}`) | `{arg}` | `$vfsPath("{name}")` |
| `clone`/`register`/`remove` | `ctl … <verb>` | `ctl` ✓ | `$vfsKind(ctl) $vfsPath("ctl")` |
| `healthCheck` | `cat …/health` (`health`) | **`health_check`** ✗ | `$vfsPath("health")` |
| `get` | (undocumented) | `{arg}` — **collides with getByName** | `by-id/{id}` |
| `repo` | repo subdir | **skipped** | `$vfsKind(dir) $vfsPath("repo/{repoId}")` |

So the pilot earns its keep: it caught a path mismatch (`health_check`) and a path collision (`get`/`getByName`) that inference alone shipped wrong.

## Mechanism A vs B
- **A (generated, added):** the `vfs_nodes()` table for the non-fs methods, from the annotated schema.
- **B (hand-written, UNCHANGED):** `hyprstream/src/services/registry.rs` `FsHandler` + fid table → real git worktree. **Not touched** (diff is schema + codegen + test only). The git backing cannot be generated; it stays.

## Codegen change (additive)
Extended `vfs::generate_mount` to emit a **recursive `vfs_nodes_<scoped>()` subdir table** per scoped client — so `repo/{repoId}/` projects with `status`/`head`/`ctl`/`branches`/`tags`/`remotes`, carrying its mount-relative prefix from the parent variant's `$vfsPath`. Factored the node loop into `build_node_entries` (shared top-level + subdir). Existing derive units unchanged (26 pass).

## Man-page
Dropped `$docExample` prose from registry.capnp (now auto-derivable from `$vfs*`); kept `$mcpDescription` narrative. Removed the now-unused `docExample` import.

## Tests — `registry_vfs_parity` (5)
`generated_paths_match_docexample_convention`, `write_verbs_share_the_ctl_node`, `get_and_get_by_name_do_not_collide`, `repo_projects_as_a_subdirectory_with_children`, `scope_and_kind_are_consistent`.

## Known T2 limitation surfaced (NOT fixed here — belongs in a T2 follow-up)
The repo subdir's multiple Void-arg listings (`list_branches`/`list_tags`/`list_remotes`/`list_worktrees`) all infer to `.` and collide. Inference's dir-default needs a name-based path when a struct has >1 listing. Pre-existing T2 behavior; a per-method `$vfsPath` or smarter inference is the fix, out of scope for the pilot.

## Not covered (scope)
The `worktree @31` deeply-nested scope and the one-handler-two-faces *runtime* round-trip need the auto-mount serving path (T4/#666) + a live dispatch harness — this pilot validates the **projection** (the generated node tables), which is T2's contract. Runtime ctl-dispatch parity rides T4/T6.

## Verify
`hyprstream-rpc-std` + `hyprstream` build clean; parity 5/5; derive 26/26; `clippy -D warnings` clean; `FsHandler`/registry.rs byte-for-byte unchanged.

Refs #539 (T5) · #544 · consumes T2 (#541).

### DIFF
--- a/crates/hyprstream-rpc-derive/src/codegen/vfs.rs
+++ b/crates/hyprstream-rpc-derive/src/codegen/vfs.rs
@@ -191,46 +191,35 @@ fn infer_path(kind: NodeKind, method_snake: &str, arg: &CapnpType) -> String {
     }
 }
 
-/// Generate the per-service `vfs_nodes()` table.
+/// Build the `VfsNode` entry token stream for one set of request variants.
 ///
-/// Emits `pub fn vfs_nodes() -> (&'static str, &'static [VfsNode])`. On a
-/// scope/kind contradiction (or bad `$vfsKind`), emits a `compile_error!`
-/// naming the offending method instead.
-pub fn generate_mount(service_name: &str, resolved: &ResolvedSchema) -> TokenStream {
-    // Data-only schemas (no request variants) get no mount.
-    if resolved.raw.request_variants.is_empty() {
-        return TokenStream::new();
-    }
-
-    let scoped_names: Vec<&str> = resolved
-        .raw
-        .scoped_clients
-        .iter()
-        .map(|sc| sc.factory_name.as_str())
-        .collect();
-
+/// Shared by the top-level table and each scoped-client subdir table. Returns
+/// `Err(compile_error_tokens)` on a bad `$vfsKind` or a scope/kind contradiction.
+fn build_node_entries(
+    variants: &[UnionVariant],
+    scoped_names: &[&str],
+    response_variants: &[UnionVariant],
+    resolved: &ResolvedSchema,
+) -> Result<Vec<TokenStream>, TokenStream> {
     let mut node_entries: Vec<TokenStream> = Vec::new();
 
-    for v in &resolved.raw.request_variants {
-        // Scoped clients project as subdirectories — deferred to the nested pass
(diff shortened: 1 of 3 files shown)


ewindisch:
**Follow-up filed:** the T2 inference limitation this pilot surfaced (multiple `Void`-arg listings in one struct all infer to `.` and collide — `list_branches`/`list_tags`/`list_remotes`/`list_worktrees`) is now tracked by **#669**. This pilot works around it with explicit per-method `$vfsPath`; #669 fixes the inference default + adds a compile-time collision guard.

Issue #572, opened 2026-06-29
feat(security) S6: Token-exchange + ZSP + ceiling lifecycle + escalation tiers


ewindisch:
**Epic:** #547. **Deps:** S4 (PDP), S5 (compiler/ceiling), **#514/#454 (DPoP enforcement)**.

## Goal
The runtime grant path: **UCAN grant-request → OAuth access/refresh tokens** (sender-bound), a **single
fail-closed path**, with ZSP, ceiling lifecycle, and escalation tiers.

## Scope
- **Token-exchange**: the OAuth endpoint accepts a subset-UCAN as the grant (RFC 8693 shape); validates
  chain + MAC + ceiling-subset + escalation; mints **short-ttl sender-bound** access + refresh tokens.
  The access token = a signed, sender-bound **distributed AVC entry**.
- **Single path, no fallback** — no offline-UCAN-at-monitor; authority-unreachable = **fail closed**
  (solve availability via HA/caching, never a weaker path).
- **ZSP**: mint **per-task/per-binding** (at namespace fork), not per-op; **refresh re-evaluates** (MAC +
  revocation + context) — refresh is renewable enrollment, never standing access.
- **Ceiling lifecycle**: approve-once; **TTL + re-attestation** (no privilege creep); revocation = stop
  refresh + evict ceiling.
- **Escalation tiers** (fire only at a ceiling amendment, never per request): static Casbin three-valued
  (allow/deny/**escalate**) → ABAC / agentic / admin-UDF — all **authority-owned (never requester-defined),
  MAC-bounded**, operating only in the escalation band.
- Sender-binding uses the existing DPoP/cnf machinery — **depends on #514/#454**.

## Dependencies / relates
#547 · S4 · S5 · #514/#454 (DPoP) · S8 (token signatures = hybrid COSE) · #153 (token-traffic confidentiality, HNDL-pending).

## Acceptance
- UCAN→token-exchange mints sender-bound access+refresh; single path; authority-unreachable = deny.
- Refresh re-evaluates (MAC + revocation + context); per-task minting.
- Escalation tiers fire only at amendment; never requester-defined; MAC-bounded.

ewindisch:
[...]

PR #678, opened 2026-07-02
fix(mac B1 #673): re-evaluate UCAN-grant refresh through the S6 gate chain

### LINKED ISSUE #673: MAC activation B1: ucan-grant refresh bypasses evaluate_grant + makes DPoP optional
**Blocker for real MAC (#547) enforcement activation** — found by the independent post-S8 audit (2026-07-02, vs main `c39b6bb13`). See memory `mac_547_enforcement_dormant` for full context.

## The gap
The generic OAuth refresh path (`exchange_refresh_token`, `crates/hyprstream/src/services/oauth/token.rs:438`) does **not** route through `evaluate_grant`/`evaluate_refresh` (the S6 fail-closed gate chain in `mac/exchange.rs`). It also makes DPoP sender-binding **optional** on refresh, where the mint path requires it.

## Why it blocks activation
Even once the other activation gaps (B2/B3, prod resolver wiring) are closed, a refresh-token flow that bypasses the gate chain would let a token escape the ceiling-subset/ZSP discipline S6 enforces at mint — refresh becomes a free re-mint rather than a re-evaluated grant.

## Fix
Route `exchange_refresh_token` through `evaluate_refresh` (already exists in `mac/exchange.rs`, tested independently — `refresh_runs_the_same_gates_as_mint`). Make DPoP mandatory on refresh, matching mint.

ewindisch:
Closes #673 · Blocker-before-activation for MAC epic #547.

## The gap (found by the independent post-S8 enforcement audit)
A UCAN-grant refresh token (`client_id` `ucan-grant:{sub}`) was consumed by the **generic** OAuth 2.1 refresh path (`exchange_refresh_token`, `oauth/token.rs:438`), which:
- does **not** re-run `evaluate_grant`/`evaluate_refresh` (the S6 fail-closed gate chain), and
- treats DPoP as **optional** ("verify if present").

That breaks two ZSP invariants: access must be **re-evaluated** on refresh (the ceiling may have been amended or revoked since mint), and sender-binding is **mandatory**. The stored `RefreshTokenEntry` also carried no way to re-present the grant (the old `TODO(#572-refresh-rotation)`).

Fail-closed today — the mint path is still gated on the S8 resolver/verifier wiring, so this is unreachable until activation — but it must be correct **before** any path can permit.

## The fix
- **`state.rs`** — `RefreshTokenEntry` gains an optional `UcanGrantRefresh` { grant CBOR (base64url), BLAKE3 content-id, requested scope, audience }. `#[serde(default)]` so every existing persisted refresh token still loads (as `None` → generic path, unchanged).
- **`token_exchange.rs`** — `mint_grant_token` persists that context. New `exchange_ucan_grant_refresh()`:
  1. **requires a fresh DPoP proof** (mandatory sender-binding, matching mint),
  2. content-id-checks the stored grant blob (corrupt/substituted → fail closed),
  3. re-presents the grant to `evaluate_refresh` — same gates as mint, against **current** `now`/verifier state, so an amended/revoked ceiling or expired grant now denies,
  4. mirrors the mint path's `DenyUnlabeledResolver` (never more permissive than mint),
  5. re-mints + rotates, re-persisting the context for the next refresh.
- **`token.rs`** — `exchange_refresh_token` routes `ucan-grant:` entries to that path (the token is already rotated/consumed before delegation).

## Design note
Server-stores the grant (the codebase's own `TODO(#572-refresh-rotation)` prescribed "store the grant CID + request so the refresh path can re-present them") — so **no client-protocol change**. Revocation/expiry are caught because `evaluate_refresh` re-verifies against current state.

## Verification
- `cargo test -p hyprstream --lib services::oauth::token_exchange` → 7 pass, incl. 2 new: legacy-entry serde back-compat (→ `None`) and ucan-grant round-trip. Gate-equivalence itself is covered by the existing `refresh_runs_the_same_gates_as_mint`.

### DIFF
--- a/crates/hyprstream/src/services/oauth/state.rs
+++ b/crates/hyprstream/src/services/oauth/state.rs
@@ -256,6 +256,37 @@ pub struct RefreshTokenEntry {
     pub expires_at_unix: i64,
     /// Ed25519 verifying key bytes bound to this token (cnf key continuity on refresh).
     pub verifying_key_bytes: Option<[u8; 32]>,
+    /// Present only for UCAN-grant refresh tokens (`client_id` `ucan-grant:{sub}`).
+    ///
+    /// MAC #547 / B1 (#673): refresh of a UCAN grant is NOT a free re-mint — the
+    /// refresh path must re-run the S6 gate chain (`evaluate_refresh`) against the
+    /// grant, because the ceiling may have been amended or revoked since the
+    /// access token was minted (ZSP: access is re-evaluated on refresh). That
+    /// requires persisting the grant + the requested access so the refresh path
+    /// can re-present them. `None` for every non-UCAN-grant refresh token, which
+    /// keeps the generic OAuth 2.1 rotation path unchanged.
+    #[serde(default)]
+    pub ucan_grant: Option<UcanGrantRefresh>,
+}
+
+/// Re-evaluation context persisted alongside a UCAN-grant refresh token so the
+/// refresh path can re-present the grant to `evaluate_refresh` (MAC #547 / B1
+/// #673). Carrying the grant itself (not just an id) lets the gate chain
+/// re-check the ceiling against current state on every refresh.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct UcanGrantRefresh {
+    /// `base64url(CBOR)` of the UCAN grant, re-presented to `evaluate_refresh`
+    /// verbatim (the same encoding the client sent as `subject_token`).
+    pub grant_cbor_b64: String,
+    /// BLAKE3 content id (hex) of the CBOR grant. Binds this refresh entry to
+    /// exactly the grant it was minted from and is checked on every refresh so
+    /// a corrupted/substituted stored blob fails closed.
+    pub grant_cid: String,
+    /// The RFC 8693 requested `scope` (the S3 `action:resource:identifier`
+    /// triple) used to rebuild the `GrantRequest` on refresh.
+    pub requested_scope: Option<String>,
+    /// The RFC 8707 `audience` resource indicator, if any.
+    pub audience: Option<String>,
 }
 
 impl RefreshTokenEntry {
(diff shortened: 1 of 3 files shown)


ewindisch:
## REVIEW VERDICT (2:0.0, authoritative per operator): APPROVE — with non-blocking nits. Merge-ready standalone.

Reviewed the diff directly + independent adversarial pass. This correctly closes B1/#673.

**Load-bearing checks (all pass):**
- **Routing is non-spoofable:** the ucan-grant refresh is routed on the persisted `entry.ucan_grant` field (server-set at mint), NOT the client-supplied `client_id` — stronger than the PR body implies. No mint path creates a `ucan-grant:` entry with `ucan_grant: None`, so no genuine grant token can fall through to the generic rotation path (token.rs:485-492).
- **Fresh DPoP is mandatory:** the early return (token.rs:485) fires *before* the generic "verify DPoP if present" block; `exchange_ucan_grant_refresh` denies (`invalid_request`) on absent/invalid proof and only sets `sender_bound: true` after verifying a fresh proof against the current clock/endpoint — nothing carried over.
- **Live re-validation:** `evaluate_refresh → evaluate_grant` re-runs `chain::validate` (hybrid signature via the global PQ trust store + temporal at current `now`); stored CBOR is BLAKE3-CID-checked and re-decoded — no stored decision trusted. Subject resolver + verifier match the mint path (never more permissive).
- **Rotation/fail-closed:** old token deleted before delegating (delete-fail ⇒ server_error, no mint); every error branch denies; no unwrap/panic on attacker/storage input; `#[serde(default)]` on `ucan_grant` + a legacy-entry test confirm no serde back-compat break.

**Nits (none blocking, none introduced by this PR):**
1. UCAN revocation-list enforcement is still `TODO(#572-revocation)` — symmetric with the mint path, dormant/fail-closed; a revoked ceiling is only caught insofar as it invalidates the signed chain. Track, don't block.
2. Pre-existing read-then-delete TOCTOU on refresh (not atomic take) — strongly mitigated here because each refresh now requires a fresh holder-key DPoP a thief can't produce.
3. Cosmetic: missing-verifier path returns HTTP 403 with error code `server_error` (status/code mismatch, copied from the mint path).

Verdict: **APPROVE-WITH-NITS**, merge-ready as a standalone change. Still requires the human sign-off flagged in the PR (security TCB).

Issue #674, opened 2026-07-02
MAC activation B2: wire the S7 audit store into the S6 grant path (prod dispatch has none)


ewindisch:
**Blocker for real MAC (#547) enforcement activation** — found by the independent post-S8 audit (2026-07-02, vs main \`c39b6bb13\`). See memory \`mac_547_enforcement_dormant\` for full context.

## The gap
\`WalAuditStore\`/\`AuditedAvc\`/\`CoseAuditSigner\` (S7, #639) are never constructed in production. The live \`#[authorize]\` dispatch path audits only via unsigned fire-and-forget \`tracing\` (\`audit_authz\`, \`svc.rs:372\`) — not the tamper-evident WAL store. The S6 grant path (\`mac/exchange.rs::evaluate_grant\`/\`evaluate_refresh\`) emits **no audit at all**.

## Why it blocks activation
S7's whole point is complete mediation with a tamper-evident, signed audit trail — and it fails closed (audit-write failure → Permit becomes Deny, per \`DecisionReason::AuditFailClosed\`). Without prod wiring, that fail-closed guarantee is untested in the path that will actually gate real grants, and grant decisions have zero audit record.

## Fix
- Construct \`WalAuditStore\`/\`CoseAuditSigner\` with the node's real (PQ) audit key at startup; install as the production \`AuditedAvc\` wrapper the live PEP calls.
- Add an audit emission point in \`evaluate_grant\`/\`evaluate_refresh\` for every grant decision (permit and deny).

## Depends on
B3 (#674 — audit journal tail-truncation must be detectable) should land alongside or before this, since wiring an audit store that can be silently truncated undermines the guarantee this ticket is meant to deliver.

Refs epic #547 · found by independent audit, not yet fixed.

PR #679, opened 2026-07-02
fix(mac B3 #675): hash-chain the audit journal + signed head checkpoint

### LINKED ISSUE #675: MAC activation B3: audit journal tail-truncation is undetectable (no hash chain / head anchor)

## The gap
The S7 audit journal (\`WalAuditStore\`, \`crates/hyprstream/src/mac/audit.rs\`) has no hash-chain linking entries and no anchored head/checkpoint. An attacker (or a crash mid-write) that truncates the tail of the journal is **undetectable** — there's no way to prove "this is the complete, unmodified history" vs. "the last N entries are missing." The module doc at \`audit.rs:30\` currently claims a stronger guarantee than the code provides — that doc is wrong and should be corrected as part of this fix, not left standing.

## Why it blocks activation
Tamper-evidence is the entire premise of S7 (#573) — an audit trail that can be silently truncated without detection is not tamper-evident. Wiring it into production (B2, #674) before this lands would ship a false sense of assurance.

## Fix
- Hash-chain each audit entry to the previous (content-addressed, e.g. BLAKE3 of the prior entry's hash + this entry's content).
- Anchor the chain head somewhere externally verifiable (e.g. periodic checkpoint signed and stored out-of-band, or a monotonic counter cross-checked on read).
- Correct \`audit.rs:30\`'s doc comment to describe the actual (post-fix) guarantee, not an aspirational one.
- Add a test that truncates the tail of a journal and asserts the truncation is detected on next read/verify.

Refs epic #547 · found by independent audit, not yet fixed. Should land before or alongside B2 (#674).

ewindisch:
Closes #675 · Blocker-before-activation for MAC epic #547 (lands before/with B2 #674).

The S7 audit journal (`WalAuditStore`) had **no hash chain** and **no anchored head**. Consequences:
- **Tail truncation was undetectable** — deleting the last *k* records left a fully-verifying journal, and on restart the store re-seeded its `seq` counter from the *truncated* head (`highest_seq_in_journal`), permanently erasing the evidence.
- The module doc claimed truncation was *"detectable by the chained sequence numbers"* — **false** (seq alone can't prove records are missing from the end).

## The fix
- **Hash chain** — `AuditRecord` gains `prev_hash` (BLAKE3 of the predecessor's content hash; `[0;32]` genesis). It's part of the hashed+signed content, so reordering, deletion, or tampering of any record breaks the successor's link. The journal is now an append-only **chain**, not a bag of independently-signed lines.
- **Serialized assignment** — `seq` + `prev_hash` + the journal append now run under one head lock (`parking_lot::Mutex<ChainHead>`), replacing the `AtomicU64` that did **not** order concurrent appends (a latent chain-fork hazard).
- **Signed head checkpoint** — after every durable append, the store writes a signed `(seq, head_hash)` to a separate `checkpoint` file (distinct signing-input domain, so a record signature can't be replayed as a checkpoint). `open()` **fails closed** if the checkpoint is ahead of the journal head (tail truncated); `verify_journal` additionally verifies the checkpoint signature, that it anchors the real head, and the `prev_hash` link on every record.
- New `AuditError::ChainBreak` / `::Truncation`.
- **Corrected the doc** to state the *actual* guarantee **and its honest residual limit**: the checkpoint is a *local* anchor; an attacker who can delete the journal can delete it too. Fully closing that needs an **off-host** anchor (replicate the signed head to OTel / a second host) — called out as a deliberate follow-up, not silently implied.

## Verification
- `mac::audit` **20/20 pass**, incl. 4 new: truncation detected **on reopen** and **by `verify_journal`**; a signed record with a **wrong `prev_hash` breaks the chain** (a seq+signature-only checker would have accepted it); intact 3-record chain+checkpoint verifies.

## Design note (anchor choice)
Head anchor = **signed local checkpoint** (reuses the existing `CoseAuditSigner` hybrid path), chosen to be self-contained and to unblock B2 (#674) without new infra. Off-host anchoring is the defense-in-depth follow-up for the residual truncate-both-files case.

### DIFF
--- a/crates/hyprstream/src/mac/audit.rs
+++ b/crates/hyprstream/src/mac/audit.rs
@@ -23,12 +23,34 @@
 //!    before returning `Ok(())`. The caller (the AVC) does not return a Permit
 //!    until `record` returns, so a Permit is never handed out ahead of a
 //!    durable audit entry. See the `wal_record_is_durable_before_return` test.
-//! 4. **Signed** — each record (and the rollup) is signed via the existing
-//!    COSE path ([`crate::mac::compiled::cose`] / `sign_composite`). Tamper
-//!    evidence comes from content-addressing + signature + append-only
-//!    conjoined: an attacker altering a record breaks the hash; altering the
-//!    stored hash breaks the signature; truncating the log is detectable by the
-//!    chained sequence numbers.
+//! 4. **Signed + hash-chained** — each record is signed via the existing COSE
+//!    path ([`crate::mac::compiled::cose`] / `sign_composite`), and each record
+//!    carries the [`prev_hash`](AuditRecord::prev_hash) of its predecessor, so
+//!    the journal is an append-only **chain**. Tamper evidence is conjoint:
+//!    altering a record breaks its hash (and thus its signature); reordering or
+//!    deleting a *middle* record breaks the successor's `prev_hash` link. A
+//!    deleted *tail* leaves a still-valid shorter chain, so it is caught
+//!    instead by the **signed checkpoint** (§ Truncation) — the head anchor
+//!    written out-of-band after every record.
+//!
+//! ## Truncation detection (the actual guarantee — and its limit)
+//!
+//! After every durable append, the store writes a **signed checkpoint** (a
+//! separate `checkpoint` file) recording the current head `(seq, head_hash)`.
+//! On open, the store fails closed if the journal's head sequence is *below*
+//! the checkpoint's — i.e. the tail was truncated. [`WalAuditStore::
+//! verify_journal`] additionally verifies the checkpoint's signature and that
+//! it matches the journal head, so a forged or rolled-back checkpoint is
+//! rejected offline.
+//!
+//! **Residual limit (honest):** the checkpoint is a *local* anchor. An attacker
+//! with write access to BOTH the journal and the checkpoint can truncate both
+//! consistently and evade local detection. Closing that requires an *off-host*
+//! anchor (replicate the signed head to OTel / a second host); that is tracked
+//! as a follow-up and deliberately out of scope here. What this store
+//! guarantees today: no tampering, reordering, or middle-deletion goes
+//! undetected, and tail-truncation is detected unless the checkpoint is
+//! destroyed in the same act.
 //!
 //! ## Complete mediation (the hook point)
 //!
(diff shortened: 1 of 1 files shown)


ewindisch:
## REVIEW VERDICT (2:0.0, authoritative per operator): APPROVE — with non-blocking nits. Merge-ready standalone.

Reviewed the diff + independent adversarial pass. Closes B3/#675; all five ACs met and tail-truncation is genuinely closed for the local threat model.

**Load-bearing checks (all pass):**
- **prev_hash chain is structural, not just signature-derived:** prev_hash is a serde field → lands in `canonical_bytes` → BLAKE3 → signed. The `forged_record_with_wrong_prev_hash_breaks_the_chain` test signs with the real key but a bad prev_hash and `verify_journal` still returns `ChainBreak` — the exact pre-B3 gap, now closed. Genesis `[0;32]` handled.
- **Serialized write path:** `parking_lot::Mutex<ChainHead>` held across seq-read → prev_hash stamp → CAS write → journal append+fsync → checkpoint → head advance. Two appends cannot fork (the AtomicU64→Mutex swap is justified).
- **open() fail-closed on truncation:** checkpoint present + (journal empty/deleted OR head < cp_seq) ⇒ `AuditError::Truncation`; journal deleted while checkpoint survives ⇒ Truncation. `verify_journal` additionally checks the checkpoint signature (distinct domain tag `hs-mac-audit-checkpoint-v1`, no cross-replay) and rollback (cp-at-head hash mismatch).
- **Honest residual:** the doc replaces the false "detectable by sequence numbers" claim and explicitly scopes the truncate-BOTH-files case to an off-host-anchor follow-up.

**Nits (none blocking):**
1. `open()` does a non-cryptographic seq-only check — it does NOT verify the checkpoint signature (only `verify_journal` does). A local attacker who can write the checkpoint could forge an unsigned one with `cp_seq == truncated_head` to evade the open-time trip — but that's within the acknowledged "write access to both files" residual, and `verify_journal` catches it. Worth a comment; ideally verify the checkpoint sig at open when a verifier is available.
2. Checkpoint-write failure leaves the in-memory head inconsistent with disk (next `record()` reuses the seq → forks the chain). `Err` is returned so no Permit leaks, and `verify_journal` flags the anomaly, but consider making a post-journal checkpoint failure fatal / re-seeding head from `journal_tail`.
3. Per-append cost ~doubles fsync traffic (sig + create + sync_all + rename + sync_dir under the head lock) — inherent to durable anchoring; a benchmark note would be good.

Verdict: **APPROVE-WITH-NITS**, merge-ready standalone. Nit (1) and (2) are worth addressing before/shortly after merge; neither blocks the milestone. Security TCB — human sign-off still required.

PR #582, opened 2026-06-29
feat(security) S3: scope-vocab unification + baseline TE policy (#569)

ewindisch:
Part of epic #547 (native MAC authz). Rebased on S1 (#567) @ `406ff9fdc` on the epic branch.

## Changes
- Rename `$mcpScope` → `$scope` (+ `$capability` alias) across **13 schema files**.
- Unify `ScopeAction` with the runtime `Operation` vocabulary (9→14 enumerants); `as_capability`/`from_capability` bridge.
- **Mandatory scope**: build error on any unscoped request method; `$scopeExempt(reason)` the only sanctioned escape (`PolicyService.check` + `registerServiceKey`).
- Read-side 9P ops → `query`, write-side → `write` (read/write TE baseline).
- **No wire ABI break** (ordinals 0–8 unchanged, 9–13 appended).

## Verified
hyprstream-rpc-build / hyprstream-rpc / hyprstream-discovery build clean; 5 validator unit tests pass. Full-workspace build is the merge-queue gate.

Cross-session coordinated: independent of S1; the #539 codegen-seam per-(op,path) enforcement lands in S2/#568, not here.

PR #684, opened 2026-07-02
feat(mac #676): production PermissionMap — injective + exact S3-scope vocab

### LINKED ISSUE #676: MAC activation: ActionVocabulary::granted_access has no production impl (test-only, no PermissionMap caller)

## The gap
\`ActionVocabulary::granted_access\` (\`crates/hyprstream/src/mac/compiler.rs:39\`) — "the most-permissive concrete access a rule grants at runtime" — has exactly two implementations, both **inside \`#[cfg(test)] mod tests\`** (lines ~295 and ~743). There is no production \`ActionVocabulary\` impl, and \`compile_policy\` (the S5 UCAN→TE compiler entry point) has **no production caller** at all.

## Correction to a prior memory reference
An earlier session note referenced "#582" as the tracker for this gap ("ActionVocabulary post-#582 must ship a correct granted_access mapping + test") — **#582 is a different, already-merged PR** (S3 scope-vocab unification, unrelated). That reference was stale/wrong; this is the real ticket.

## Why it blocks activation
Without a production \`ActionVocabulary\` + a real caller of \`compile_policy\`, the S5 UCAN→TE compiler (and by extension S6's ceiling-ldiscipline, which calls \`compiler::authorize_at\`) has nothing concrete to compile *against* in production — it's fully exercised by tests only.

## Fix
- Ship a real \`ActionVocabulary\` impl mapping the actual object/action space this deployment uses (not a test fixture).
- Wire a production caller of \`compile_policy\` (likely at policy-load/refresh time, producing the \`PermissionMap\` the PEP consults).
- Add an upper-bound test: for every \`TeRule\` the vocabulary can express, \`granted_access\` never returns an \`AccessRequest\` broader than the rule's stated scope (the "most-permissive, not more" invariant the doc comment claims).

ewindisch:
Closes #676 · Blocker-before-activation for MAC epic #547.

`check_no_escalation`'s soundness rests on `granted_access(rule)` being the rule's **most-permissive** access — a too-narrow inverse could mask an escalation. There was no production `PermissionMap` carrying that obligation (only test-only `TestPermissions`/`EscalatingPermissions`), and `compile_policy` had no production caller.

## The fix
`ScopePermissionMap` (`mac/permission_map.rs`) is the production S3-scope ↔ S5 TE-rule vocabulary, made **injective + exact by construction** rather than a hand-verified mapping:

- A `TeRule` is a concrete `(subject_type, object_type, action)` triple — the compiled matrix cannot represent wildcards. So wildcards expand **at compile time** over a closed registry: `mac://<class>/*` expands to one rule per registered object of that class; a `*` ability expands over exactly the closed 9-verb schema enum. Unrecognized class/name/verb → nothing (fail-closed, never a guess).
- Because the expansion is **injective** (each concrete `(class, name, verb)` maps to its own distinct rule; no two distinct accesses collapse to one rule), `granted_access` is the **exact inverse** of a rule — trivially the tightest upper bound of a singleton. **There is no join/LUB logic that could be too narrow and mask an escalation.**
- Object ids are assigned positionally over the **sorted** registry, so two nodes building the map from the same object list intern identically with no side table.

`ScopeAction` (`te.rs`) gains the canonical verb strings (`as_str`/`parse`/`from_action`/`ALL`) — one string↔id assignment shared by the S3 parser, this map, and the PEP.

## Verification
- 8 tests, incl. 2 proptests:
  - **the #676 obligation itself** — `granted_access` upper-bounds every request that projects onto its rule (an equality here, since the map is injective);
  - **injectivity** — distinct accesses map to distinct rules;
  - wildcard expansion is exactly the registry / the closed verb set, nothing more;
  - fail-closed on unrecognized class/name/verb/rule;
  - end-to-end through `compile` + `check_no_escalation`.

Also corrects a stale `#582` pointer in the `PermissionMap` trait doc (#582 is an unrelated merged PR) to point at this impl.

### DIFF
--- a/crates/hyprstream/src/mac/compiler.rs
+++ b/crates/hyprstream/src/mac/compiler.rs
@@ -27,7 +27,8 @@ use crate::mac::lattice::Lattice;
 use crate::mac::te::{Decision, TeMatrix, TeRule};
 
 /// Maps a UCAN capability to the TE rules it grants. The injection point for the
-/// concrete action vocabulary (S3, #582).
+/// concrete action vocabulary — the production impl is
+/// [`crate::mac::permission_map::ScopePermissionMap`] (#676).
 pub trait PermissionMap {
     /// The TE rules a single capability grants. Empty for an unrecognized
     /// capability — it grants nothing (fail-closed; never a wildcard rule).
--- a/crates/hyprstream/src/mac/mod.rs
+++ b/crates/hyprstream/src/mac/mod.rs
@@ -63,6 +63,9 @@ pub mod compiler;
 // sender-binding, consuming S1/S5/the compiler rather than re-implementing.
 pub mod exchange;
 pub mod lattice;
+// #676: the production S3-scope ↔ S5-TE-rule vocabulary (injective + exact;
+// wildcards expand at compile time over a closed registry).
+pub mod permission_map;
 pub mod te;
 
 // Re-export the per-op contract surface S2 (PEP) and S5/S6 (policy producers) consume.
--- a/crates/hyprstream/src/mac/permission_map.rs
+++ b/crates/hyprstream/src/mac/permission_map.rs
@@ -1,0 +1,474 @@
+//! The production [`PermissionMap`] (#676): the S3 scope ↔ S5 TE-rule
+//! vocabulary, **injective and exact**.
+//!
+//! ## Shape (and why)
+//!
+//! A [`TeRule`] is a concrete numeric triple `(subject_type, object_type,
+//! action)` — a wildcard is unrepresentable in the compiled matrix, and
+//! inventing a sentinel "any object" id would put wildcard matching on the PEP
+//! decision path (a standing fail-open hazard). So wildcards exist **only on
+//! the grant side** (`Capability::authorizes` does directional covering) and
+//! this map expands them **at compile time** over a closed registry:
+//!
+//! - `mac://<class>/*` expands to one rule per **registered** object of that
+//!   class — never beyond the registry.
+//! - a `*` ability expands over exactly [`ScopeAction::ALL`] (a closed enum).
+//! - anything unrecognized (class, name, verb) maps to **nothing**
+//!   (fail-closed; never a guessed rule).
+//!
+//! Because the expansion is injective — each concrete `(class, name, verb)`
+//! maps to its own distinct rule, and no two distinct accesses share a rule —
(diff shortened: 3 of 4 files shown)


ewindisch:
## REVIEW VERDICT (2:0.0, authoritative per operator): APPROVE — with nits + one important scoping note. Merge-ready standalone.

Reviewed `granted_access` + the proptest bodies directly, plus an independent adversarial pass. The security-critical property of #676 holds.

**The subtle claim — verified sound:** `granted_access` is a true upper bound, not a maskable too-narrow inverse. `permissions_for` interns `(class,name)`→ObjectType through a **bijective** registry and verbs→Action via fixed schema discriminants, so each concrete `(class,name,verb)` maps to exactly one `TeRule` (injective by construction). `granted_access` inverts a rule to its **exact** `(class,name,verb)` with `Caveats::empty()`; a PEP-projected request landing on rule R has the same resource/ability R inverts to, so `covers` is exact-match true and `empty().covers(_)` is true. Wildcard expansion stays bounded by the closed registry (unregistered objects project to None ⇒ deny). I could not construct an input where a request maps to a rule whose `granted_access` fails to cover it.

**IMPORTANT scoping note (not a code defect, must not be mistaken for closure):** `ScopePermissionMap` AND `compile_policy`/`check_no_escalation` remain invoked **only from tests** — no production path constructs this map or runs the signing gate. #676's literal ask ("no production `PermissionMap` impl" → now there is one carrying the upper-bound obligation, with a proptest tripwire) is **discharged**, but the deny-by-default escalation gate is **not yet exercised on real grants**. Live policy compilation is a separate activation follow-up. Please register that before treating the MAC compiler as "enforcing."

**Nits (none blocking):**
1. An object literally named `"*"` is shadowed by wildcard expansion (would *broaden* to the whole class, not narrow) — unrealistic registry entry, but the failure mode is widening; add a debug-assert/guard.
2. Both proptests run over the fixed 5-object `map()`, not a generated registry, and the upper-bound test reconstructs the resource from the same `by_id` it asserts against (mildly circular). Adequate as an inverse-drift tripwire; not a strong property over arbitrary registries.
3. Caveated UCAN grants can never compile through this map (always flagged as escalation) — documented + fail-closed, but a real completeness limit until a caveat-aware matrix lands.

Verdict: **APPROVE-WITH-NITS**, merge-ready standalone (self-contained new module + additive `ScopeAction` helpers + one doc fix). Security TCB — human sign-off still required.

Issue #681, opened 2026-07-02
MAC: two-principal SecurityContext for delegated calls — clearance = meet(delegator, actor), assurance from signer, audit both


ewindisch:
**Two-principal MAC composition for delegated calls.** Surfaced by the #671 mount-all / delegation design discussion (2026-07-02). Extends the S1/S8 `SecurityContext` model; grounded against current main `2c35bce7d`.

## The gap
`SecurityContext` derivation is **single-principal**: it reads clearance from one verified `Claims` and derives assurance from the one envelope signer's verified key material (`crates/hyprstream-rpc/src/service/svc.rs:272-316` — `verified_key_material()` + `security_context()`; clearance carrier `crates/hyprstream-rpc/src/auth/claims.rs:129`). In a delegated call there are TWO principals — the **delegator** (user, source of authority) and the **actor** (e.g. MCP, who signs the downstream envelope). The current model has no way to compose them, so a delegated request would take either the user's clearance (actor escalates on the user's authority) or the actor's (defeats acting-for-user).

## The rule this ticket lands
- **Capability** already attenuates correctly via UCAN (delegation can only narrow) — handled by the delegation ticket #680.
- **MAC lattice floor** (`clearance ⊒ label`, incl. the #548 assurance axis) must use the **meet**: effective level/compartments = `min(delegator, actor)`; never either alone.
- **Assurance** is clamped to the **actual signer's** verified key material. Since the actor signs the downstream envelope, assurance follows the actor's PQ posture (PqHybrid only if the actor's key is trust-store-anchored), independent of what clearance the delegated token asserts — consistent with #548 and the S8 clamp-down (no silent upgrade).
- **Audit** records both principals: `principal = actor, on-behalf-of = delegator` (the correct framing of #445's attribution symptom; pairs with #674 audit-store wiring).

## Acceptance criteria
- [ ] A delegated-call `SecurityContext` derivation composes both principals and takes the meet on level/compartments; unit-tested that neither principal alone governs.
- [ ] Assurance is derived from the signing (actor) key's verified material and clamped down (reuse the S8 clamp path); test: actor with a Classical key cannot reach a PqHybrid-labeled object even if the delegated clearance claims PqHybrid.
- [ ] The audit record for a delegated decision carries both actor and on-behalf-of identities (coordinate with #674).
- [ ] Fail-closed: a delegation that cannot resolve one of the two clearances denies (no default clearance for the missing principal).

## Refs
[...]

PR #686, opened 2026-07-02
fix(mac B2 #674): wire the S7 audit store into the S6 grant path

### LINKED ISSUE #674: MAC activation B2: wire the S7 audit store into the S6 grant path (prod dispatch has none)

ewindisch:
Closes #674 · Blocker-before-activation for MAC epic #547.

**Stacks on #679 (B3) + #684 (#676)** — both are still open; this PR's diff will shrink to just the B2 delta once they merge (same pattern as S6/S7/S8 earlier in the epic). Individually reviewable now; the branch already contains both dependencies' commits so it builds/tests green standalone.

Every S6 grant-path decision (`evaluate_grant`/`evaluate_refresh`) is the **one live MAC decision point today**, and it emitted **no audit record at all**. S7's whole premise is complete mediation with a tamper-evident, signed audit trail that fails closed — without production wiring that guarantee was untested in the path that will actually gate real grants.

## The fix
- **`OwnedCoseAuditSigner`** (`mac/audit.rs::cose`) — an owned-key `AuditSigner` alongside the existing borrowed `CoseAuditSigner`. `OAuthState`/the audited AVC must be `'static`, so a borrowed signer can't live there; this holds `Arc<SigningKey>`/`Option<Arc<MlDsaSigningKey>>` with **identical** fail-closed sign logic (Hybrid + no PQ key ⇒ every `sign()` fails, never downgrades).
- **New `DecisionReason::Grant*` variants** mirroring every `GrantError` gate — a grant denial's cause is as diagnosable in the audit trail as a TE denial's.
- **`GRANT_PATH_SUBJECT`/`GRANT_PATH_OBJECT`/`ACTION_UNRECOGNIZED` sentinels** (`te.rs`) — `AuditRecord` predates the grant path and is shaped around a per-op TE triple; rather than fork the schema, a grant record uses reserved ids (`u32::MAX`) so it self-describes as a distinct decision kind and can never collide with a real interned TE id.
- **`audited_evaluate_grant` / `audited_evaluate_refresh`** (`mac/exchange.rs`) — audit-then-decide wrappers around the pure `evaluate_grant`/`evaluate_refresh`. Every gate outcome is audited, permit **and** deny. On a would-be `Permit`, the audit write happens **before** the caller mints: if it fails, the permit is downgraded to a new `GrantError::AuditUnavailable` — the same fail-closed rule `AuditedAvc` enforces at the TE path. A deny that can't be audited is still enforced as decided (auditing failure never weakens a decision, only a would-be permit).
- **`OAuthState`** gains `audit_sink: Option<Arc<dyn AuditSink>>` + `with_audit_sink`. The `oauth` service factory constructs a `WalAuditStore` + `OwnedCoseAuditSigner` at startup (node's active Ed25519 + a startup-time ML-DSA-65 key snapshot, gated by the node's enforced `CryptoPolicy`). The mint call site now **requires** the sink — no sink configured ⇒ fail closed (deny every request), never mint unaudited tokens.

## Verification
- 5 new tests (`mac::exchange`): permit recorded + still mints; permit downgraded to `AuditUnavailable` when the write fails (and the resulting deny **is** recorded); a deny stays a deny when its audit write fails; an unparseable ability still produces a diagnosable record via the sentinel; `audited_evaluate_refresh` audits identically to mint.
- `mac::` **96/96 pass** · `cargo check -p hyprstream` clean · `cargo clippy -p hyprstream --all-targets` clean.

### DIFF
--- a/crates/hyprstream/src/mac/audit.rs
+++ b/crates/hyprstream/src/mac/audit.rs
@@ -183,6 +183,32 @@ pub enum DecisionReason {
     /// **Fail-closed:** the audit write failed, so a Permit was downgraded to
     /// Deny. This is itself auditable (the deny is what the PEP enforced).
     AuditFailClosed,
+
+    // ── B2 (#674): the S6 grant-path decision kind (see [`crate::mac::te::
+    // GRANT_PATH_SUBJECT`]). One variant per `mac::exchange::GrantError`
+    // gate, so a grant denial's cause is as diagnosable as a TE denial's. ──
+    /// No DPoP sender-binding proof was supplied (ZSP gate 1).
+    GrantMissingSenderBinding,
+    /// The presented grant carries no capabilities at all (gate 2).
+    GrantEmptyGrant,
+    /// The subject carried no derivable MAC clearance (gate 3, S1 floor).
+    GrantUnlabeledSubject,
+    /// The MAC clearance floor denied for a grant request (gate 5). Distinct
+    /// from [`DecisionReason::FloorDeny`] (a TE-path floor denial) so a grant
+    /// vs. per-op floor denial is distinguishable in the audit trail. Also
+    /// covers gate 4 (unlabeled object), which `evaluate_grant` maps to the
+    /// same `GrantError::InsufficientClearance`.
+    GrantFloorDeny,
+    /// The UCAN chain failed S5 validation (gate 6: bad signature, broken
+    /// linkage, widening, window escape, expiry, over-depth).
+    GrantChainInvalid,
+    /// The request exceeded the grant's ceiling (gate 7) — an escalation
+    /// attempt, denied pending a ceiling amendment (S6 does not auto-widen).
+    GrantOverCeiling,
+    /// A grant was `Permit`, but the caller's `sink` (or checking `can_sign`
+    /// preflight) could not durably audit it — the permit is downgraded to
+    /// deny, mirroring [`DecisionReason::AuditFailClosed`] for the grant path.
+    GrantAuditFailClosed,
 }
 
 impl DecisionReason {
@@ -195,6 +221,13 @@ impl DecisionReason {
             DecisionReason::Escalate => "escalate",
             DecisionReason::TokenGate => "token_gate",
             DecisionReason::AuditFailClosed => "audit_fail_closed",
+            DecisionReason::GrantMissingSenderBinding => "grant_missing_sender_binding",
+            DecisionReason::GrantEmptyGrant => "grant_empty_grant",
+            DecisionReason::GrantUnlabeledSubject => "grant_unlabeled_subject",
+            DecisionReason::GrantFloorDeny => "grant_floor_deny",
+            DecisionReason::GrantChainInvalid => "grant_chain_invalid",
+            DecisionReason::GrantOverCeiling => "grant_over_ceiling",
+            DecisionReason::GrantAuditFailClosed => "grant_audit_fail_closed",
         }
     }
 }
@@ -325,6 +358,7 @@ pub mod cose {
     use ed25519_dalek::{SigningKey, VerifyingKey};
     use hyprstream_rpc::crypto::cose_sign::{sign_composite, verify_composite};
     use hyprstream_rpc::crypto::pq::{MlDsaSigningKey, MlDsaVerifyingKey};
+    use std::sync::Arc;
 
     /// Domain-separation AAD for audit-record signatures (distinct from
     /// policy/envelope AADs so an audit signature can never be replayed as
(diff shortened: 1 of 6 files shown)


ewindisch:
## REVIEW VERDICT (2:0.0, authoritative per operator): APPROVE — with nits. CONDITIONAL merge order: land #679 + #684 first.

Reviewed the B2 delta (`mac/exchange.rs`, `mac/te.rs`, `OwnedCoseAuditSigner`, `oauth/{mod,state,token_exchange}.rs`) + independent adversarial pass. Closes B2/#674; all three ACs met and tested.

**Complete mediation — satisfied:** `audited_evaluate_grant`/`audited_evaluate_refresh` run the pure evaluator then `audit_grant_outcome`, whose match is exhaustive over `Ok(Permit)`, `Ok(Escalate)`, and all seven `GrantError` variants — so every early-return DENY gate emits exactly one signed record. The mint path (`token_exchange.rs:508`) now calls the audited variant and the unaudited `evaluate_grant` import was removed; no other production caller exists. Only `Ok(Permit)` mints.

**Fail-closed — satisfied:** would-be Permit + failed audit-write ⇒ `GrantError::AuditUnavailable` (deny, HTTP 500) + best-effort `GrantAuditFailClosed` record; deny stays deny; `record_inner` signs before the journal append so a failed sign leaves no partial state.

**Startup — satisfied:** no sink ⇒ 500 before evaluating; Hybrid + no ML-DSA key ⇒ every `record()`'s `sign()` fails ⇒ every Permit downgrades to deny at runtime (fail-closed). Sentinels (`GRANT_PATH_SUBJECT/OBJECT`/`ACTION_UNRECOGNIZED` = u32::MAX) don't collide with real registry-bounded ids. `OwnedCoseAuditSigner` mirrors the CoseAuditSigner fail-closed logic.

**MERGE ORDER (conditional):** this branch carries the #679 (B3) and #684 (#676) commits. **#679 and #684 must merge to main first** — otherwise this PR re-introduces their full diffs. Once they land, rebase so this shrinks to the B2 delta. Approve is conditional on that ordering.

**Nits (non-blocking follow-ups):**
1. Missing/invalid DPoP grant attempts are rejected at the HTTP layer *before* `audited_evaluate_grant` (mint hardcodes `sender_bound: true`), so a missing-DPoP denial leaves NO tamper-evident record and `GrantMissingSenderBinding` is effectively dead in prod. Defensible (no MAC decision reached), but the audit is blind to that rejected-attempt class — follow-up if the trail is meant to be complete.
2. **Forward-compat for #681:** `AuditRecord` carries a single principal; delegation's actor + on-behalf-of will require a real schema field (bumping `canonical_bytes`/hash for all records — a chain-version concern), not just a new sentinel. Flagging now so #681 plans the schema/version bump.
3. `OwnedCoseAuditSigner` clones the Ed25519 secret into a new Arc (rest of codebase Arc-shares it); zeroizes on drop, process-long — low risk.
4. Cosmetic: under Hybrid-no-PQ, the factory logs an error but still installs the sink (runtime-fail-closed anyway; refusing to install would be marginally cleaner).

Verdict: **APPROVE-WITH-NITS**, merge-ready **after #679 + #684 land**. Security TCB — human sign-off still required.

ewindisch:
Force-pushed a corrected, rebased head (`28c76ad7f`) reconciling this with B1 (#678, merged) and B3+#676 (#679/#684, merged) which landed on main since this PR was opened.

**What changed vs. the previous head:**
- Rebased cleanly onto current `main` (was previously behind after Wave 1 merged, hence the `DIRTY` state).
- Resolved one real integration point: B1's `exchange_ucan_grant_refresh` (now on main) was still calling the raw, unaudited `evaluate_refresh`. Upgraded it to `audited_evaluate_refresh` with the same audit-sink-required fail-closed gate the mint path already had — so refresh is now on the audit trail too, not just mint. This is exactly the kind of gap B2 exists to close.
- Diff is now scoped to exactly: `mac/audit.rs`, `mac/exchange.rs`, `mac/te.rs` (sentinels), `services/oauth/mod.rs`, `services/oauth/state.rs`, `services/oauth/token_exchange.rs` — no `audit.rs`-chain or `permission_map.rs` bleed (that content is already on main via #679/#684).
- Re-verified: `mac::` 96/96 tests pass, `cargo check`/`clippy --all-targets` clean.

Ready for 1:5.0's diff-scope check.

ewindisch:
Fixup pushed (\`bca037397\`) addressing the more substantive item from the earlier review nits — the Hybrid-no-PQ case.

**Confirmed on inspection:** under Hybrid policy with no ML-DSA-65 key provisioned, the factory logged one startup error but still installed the non-signing signer into the audit store. Every real grant decision would then silently fail two audit-write attempts deep (mint's Permit downgrade, then the best-effort fallback deny record — both via the same broken signer), with only that one easy-to-miss startup log as any indication, rather than the immediate "audit trail is not configured" 500 the grant path already gives when `audit_sink` is `None`. Fail-closed held throughout either way, but this made the degrade path inconsistent/harder to observe.

**Fix:** treat "signer can't sign" the same as "store failed to open" — don't install it.

The other two review items: the extra key clone is consistent with an existing pattern already used 5x elsewhere in this same factory function (not new risk, not worth a targeted fix); the single-principal `AuditRecord` / #681 forward-compat note has been posted on #681 directly for whoever picks that up.

Re-verified: `mac::` 96/96 pass, `cargo check`/`clippy --all-targets` clean. CI re-running fresh on the new head; will confirm green.

Issue #653, opened 2026-07-02
vfs-server: local-FUSE-mount serve mode (materialize composed namespace as a host directory)


ewindisch:
**Blocks spine re-grounding of the non-VM backends (podman #617, nspawn) — split out from #617.**

## Problem
`hyprstream-vfs-server` today ships **only** a vhost-user-fs transport: it hosts the `VfsFileSystem` down-adapter over guest-memory vrings via `vhost-user-backend` (a Cloud Hypervisor VM device). It's built with `fuse-backend-rs` `default-features=false, features=["virtiofs"]` — explicitly *not* `fusedev`.

Consequently the composed per-sandbox namespace (`SandboxFs::compose`, #641/#365) can only be delivered to a **VM guest** (kata, over virtio-fs). It cannot be delivered to a backend that needs the namespace as a **host POSIX directory**:

- **podman** needs `podman run --rootfs <dir>` (#617).
- **systemd-nspawn** needs `--directory=<dir>` / `--image`.

Both currently side-channel their own image provisioning (podman: `podman run <image-ref>` from podman's store), which is a spine violation (#508): the universal image filesystem service is not the source of the guest's rootfs.

## Goal
Add a **local-FUSE-mount** serve mode to `hyprstream-vfs-server`: mount the `VfsFileSystem` (the `Namespace → fuse_backend_rs::FileSystem` down-adapter, which is transport-agnostic) at a host mountpoint via `fuse-backend-rs`'s `fusedev` transport (kernel `/dev/fuse`), yielding a real directory. Then non-VM backends compose via `SandboxFs` and pass that mountpoint as their rootfs — mirroring kata's compose-and-serve exactly, differing only in transport (fusedev vs virtio-fs).

## Notes / open questions
- **Rootless FUSE:** rootless podman + FUSE-backed rootfs interacts with user namespaces (`--userns=keep-id`). Confirm a rootless-mounted fuse dir is usable as `--rootfs` under a rootless podman userns, or scope the constraint.
- `VfsFileSystem` / the down-adapter is already transport-agnostic (`SandboxFs::into_filesystem`), so this is additive: a new `serve_local(fs, mountpoint)` alongside the existing `serve(fs, socket)`. Enable `fuse-backend-rs`'s `fusedev` feature (in addition to `virtiofs`).
- Subject-per-call MAC enforcement is preserved (it lives on the `Mount` boundary, below the down-adapter) regardless of transport.

## Refs
[...]

Issue #635, opened 2026-07-01
narrow SandboxBackend: lifecycle + deliver_namespace (move exec/image/resources off the trait)


ewindisch:
Child of #508 (the spine). Step 4 — the most invasive change; reshapes #341/#508's premise.

## Goal
Narrow the `SandboxBackend` trait from the CRI-shaped surface to **lifecycle + deliver_namespace**. A backend boots a guest OS under an isolation boundary and delivers the composed namespace to it; it does NOT own image-pull or run commands.

## What moves off the trait
- `exec_sync(command)` → moves to the **guest-OS control file service** (the kata-agent #344, Wanix `/task` #612 — both realize "run a command in the guest OS"). The backend doesn't exec; the guest OS does.
- `PodSandboxConfig`-image-pull → moves to the **universal filesystem service** (mountable image fs).
- `LinuxContainerResources` → moves to a resource-control file service / namespace node.

## What stays (the narrowed trait)
Lifecycle: `backend_type`/`is_available`/`initialize`/`start`/`stop`/`destroy`/`reset`/`get_pids`, plus a new `deliver_namespace(ns, transport)` (boot guest OS + attach the composed namespace). `supports_exec`/`exec_sync` leave the trait.

## Why
`exec_sync` is CRI/container-shaped and doesn't fit wasm (a wasm guest has no "command" without a guest OS). The guest-OS model (#508 spine) dissolves this: commands run in the guest OS (Linux for kata/containers, Wanix for wasm-Profile-B, nothing for Profile-A by design), driven via the guest-OS control service — which is itself a 9P file service in the namespace.

Blocks: re-grounding #617/#619/nspawn against the narrowed trait.

ewindisch:
## Decision on PR #704's three flagged risk areas (review posted)

Reviewed the "needs careful human review" items PR #704 called out. Posted inline blocking/non-blocking comments directly on the diff: https://github.com/hyprstream/hyprstream/pull/704#pullrequestreview-4627853653

**Decisions:**

[...]

Issue #634, opened 2026-07-01
self-contained stateful language-shell Mounts (kill rx-externalization) + dedupe namespace builders


ewindisch:
Child of #508 (the spine). Step 3.

## Goal
Make `/lang/tcl` and `/lang/python` **self-contained stateful Mounts** that own their `!Send` interpreter + driver thread internally. Callers mount them; they never hold the mount's channels/state. Rip the rx-externalization out of the dispatch/RPC/UI layer.

## The current anti-pattern
`create_mount_channel()` returns `(tx, rx)`; the Mount holds `tx` but the `rx` is smeared as `Arc<Mutex<Receiver>>` across `dispatch_outputs` (~10 call sites in `shell_handlers.rs`), `handle_rpc`, `ChatApp` (`with_tcl_mount_rx`/`with_py_mount_rx` + manual `drain` in `tick()`). Adding a `/lang/foo` shell means editing all of those. That's the opposite of a stateful mount abstraction.

## Scope
- `TclMount::spawn(subject, ns)` / `PythonMount::spawn(...)` own the interpreter + a dedicated driver thread (the `LocalSet`+current-thread-runtime pattern from #616's `ShellOwner`); the channel + rx + interpreter live inside, never exposed. `PythonMount::spawn` is already close to this shape — generalize it; converge `TclMount` onto it.
- Remove `tcl_mount_rx`/`py_mount_rx` from `dispatch_outputs`/`handle_rpc`/`ChatApp` entirely. The namespace becomes the only interface.
- Fold the two duplicated namespace builders (`shell_handlers.rs` + `tui/vfs.rs`) into the shared composition from the namespace-layer epic.

Builds on #609 (ControlFile/FieldFile devfile primitives, merged). Related: #483, #616.

ewindisch:
Implemented — PR (draft, stacked on #599). TclMount::spawn + PythonMount::spawn own their interpreter+driver internally; `create_mount_channel` now cfg(test); `*_mount_rx` fully removed from dispatch_outputs/handle_rpc/ChatApp (0 hits verified). Net -215/+214 — cleaner not bigger. tcl 46/46, tui 27/27, hyprstream --lib cli 7/7 + tui 42/42 pass; clippy clean. Deferred: /lang/python not mounted in builders (needs the wasm Sandbox → #632); was already non-functional, not a regression.

ewindisch:
[...]

PR #704, opened 2026-07-03
[Additive/Transitional] narrow SandboxBackend: lifecycle + deliver_namespace (#635)

ewindisch:
## This is an additive, transitional step — NOT a final rip-out

**Needs careful human review** before merge given the blast radius (#635 itself calls this "the most invasive change; reshapes #341/#508's premise"). Do not merge without a maintainer signing off.

Closes/addresses #635 (child of epic #508).

## Context

Epic #508 converges worker sandboxing onto a "namespace IS the OS surface, backends-as-transport" model: a `SandboxBackend` boots a guest OS under an isolation boundary and delivers the composed namespace to it — it does not own image-pull or run commands. Two other pieces of this same pivot (#634 dedupe the namespace-composition builders, #653 local-FUSE-mount serve mode) are being worked concurrently and may not have landed yet; this PR does not depend on either landing first.

## What actually moved

- **Added** `SandboxBackend::deliver_namespace(sandbox, namespace, subject, transport) -> Result<NamespaceDelivery>`, with a default implementation that returns `WorkerError::Unsupported`. This is purely additive — no existing implementor needed to change to keep compiling.
- **Added** `NamespaceTransport` (`VirtioFs { socket_path, mount_tag }` / `BindMount { target }` / `HostImports`) and `NamespaceDelivery` (mirrors the transport, `VirtioFs` carries an opaque `guard: Option<Arc<dyn Any + Send + Sync>>` the caller must keep alive) in `crates/hyprstream-workers/src/runtime/backend.rs`.
- **Added** `WorkerError::Unsupported(String)` for backends/transports that don't (yet) support a given `deliver_namespace` shape.
- **KataBackend**: real implementation. Serves an *already-composed* `Namespace` over a per-sandbox vhost-user-fs socket (new `SandboxFs::from_namespace` constructor, alongside the existing image-composing `SandboxFs::compose`) and hot-attaches it as a Cloud Hypervisor ShareFs device via the existing `attach_share_fs` when a hypervisor handle is already present on the sandbox. `KataBackend::start`'s existing boot-time compose→serve→attach flow (`compose_and_serve_vfs`) is untouched — this is a parallel capability for callers that already hold a composed `Namespace` (e.g. a future #634/#653-based caller), not a replacement of the boot path.
- **NspawnBackend**: provisional implementation using `machinectl bind --mkdir` to runtime-attach an already-materialized host directory into a running container/machine (the nspawn analogue of Kata's hot-attach ShareFs). Only supports the `BindMount` transport, and only when the source path already exists — it does **not** materialize an in-process `Namespace` at a directory itself (no generic "namespace → real directory" primitive exists yet; that's exactly the gap #653 closes). Documented as provisional in code.
- **WasmBackend**: provisional implementation that accepts the `HostImports` transport and reports success, but does not (yet) actually wire the namespace into the guest's capability linker — today's linker only exposes `env::host_random` (Profile A, zero WASI). Documented in code as an honest no-op stub pending the guest-OS control-file service (Wanix `/task`, #612).
- 10 new unit tests covering `deliver_namespace` across all three backends (rejection of unsupported transports, the real Kata serve+attach path, nspawn's fail-fast-on-missing-source path, wasm's accept/reject).

## What is deliberately left TRANSITIONAL (not removed)

Per #635's own scope, these stay on the trait, unchanged, with `// TRANSITIONAL (#635)` doc comments explaining why and what unblocks their removal:

- `exec_sync` / `supports_exec` — CRI-shaped "run a command in the sandbox." Moves to the guest-OS control-file service (kata-agent #344, Wanix `/task` #612) once that's wired end-to-end — commands will run *in the guest OS*, driven through a 9P control file, not through the backend directly.
- `start`'s `&PodSandboxConfig` parameter (the CRI-shaped image-pull/pod config surface) — narrows once the universal filesystem-service Mount (#633/#641, already merged) owns image resolution end-to-end and `start` can narrow to "boot a guest OS", with content arriving via `deliver_namespace` instead.
- `update_resources`'s `LinuxContainerResources` — moves to a resource-control file service / namespace node once that lands; kept with its existing no-op default.

None of this removal work is safe to do in the same change as adding `deliver_namespace` — #344/#612/#633/#641 need to be fully wired first, and removing these now would break `SandboxPool`, `ExecMount`, and the CRI-facing worker RPCs for no working replacement.

## Risk / review notes for the human reviewer

- The `NamespaceDelivery::VirtioFs.guard` field is an opaque `Arc<dyn Any + Send + Sync>` specifically so `backend.rs` doesn't need to depend on the `oci-image`/`kata-vm`-gated `sandbox_fs` module just to describe the delivery result. Worth double-checking this is the right shape vs. e.g. a dedicated non-generic guard trait.
- Kata's `deliver_namespace` intentionally does **not** touch `KataHandle`'s stored `vfs_server`/`virtiofs_socket` fields (it's called with `&PodSandbox`, not `&mut`), so the serving thread it spawns is detached rather than tracked on the handle. This is safe (dropping a `JoinHandle` does not stop the thread; the existing `compose_and_serve_vfs` boot path has the same "never explicitly stopped, relies on VMM disconnect" behavior), but it does mean this method's serving thread isn't reachable for explicit teardown by anything other than the guest disconnecting. Flagging for review since it's new surface, even though the pattern matches existing code.
- Nspawn's `deliver_namespace` shells out to `machinectl bind`; nothing in this repo exercises that path in CI (no real nspawn machine in test envs) — only the fail-fast validation (unsupported transport / missing source) is unit-tested.
- Wasm's `deliver_namespace` is honestly a no-op behind a "success" result for `HostImports` — flagging in case that return-shape (accept-but-do-nothing) is surprising to a caller that doesn't read the doc comment.

## Testing

- `cargo test -p hyprstream-workers --lib` (default features): 83 passed, 0 failed.
- `cargo test -p hyprstream-workers --lib --features kata-vm,wasm`: 151 passed, 0 failed (141 pre-existing + 10 new `deliver_namespace` tests, no regressions).
- `cargo clippy -p hyprstream-workers --all-targets --features kata-vm,wasm -- -D warnings`: clean.
- `cargo clippy --workspace --all-targets --features download-libtorch -- -D warnings` (CI-exact): clean.

## Test plan for reviewer

- [ ] Confirm the `deliver_namespace` default (`Unsupported`) is an acceptable fail-closed default for backends not yet re-grounded on it.
- [ ] Confirm `NamespaceTransport`/`NamespaceDelivery` shapes are extensible enough for #653's local-FUSE-mount serve mode without another trait-shape change.
- [ ] Confirm the TRANSITIONAL doc comments accurately scope what #344/#612/#633/#641 need to unblock before `exec_sync`/`update_resources`/CRI-config actually come off the trait.

### DIFF
--- a/crates/hyprstream-workers/src/runtime/backend.rs
+++ b/crates/hyprstream-workers/src/runtime/backend.rs
@@ -2,9 +2,27 @@
 //!
 //! Extracts the VM lifecycle out of `SandboxPool` so multiple backends
 //! (Kata, systemd-nspawn, etc.) can provide sandbox isolation.
-
+//!
+//! ## #508 / #635 — narrowing in progress
+//!
+//! Epic #508 converges worker sandboxing onto a "namespace IS the OS surface,
+//! backends-as-transport" model: a `SandboxBackend` boots a guest OS under an
+//! isolation boundary and delivers the composed [`Namespace`] to it — it does
+//! NOT own image-pull or run commands. [`deliver_namespace`](SandboxBackend::deliver_namespace)
+//! is the narrowed method this trait is converging on.
+//!
+//! The CRI-shaped methods below (`exec_sync`/`supports_exec`, the
+//! `PodSandboxConfig`-driven image pull implicit in `start`, and
+//! `update_resources`'s `LinuxContainerResources`) are kept on the trait for
+//! now and marked **TRANSITIONAL**: they move off once the guest-OS
+//! control-file service (kata-agent #344, Wanix `/task` #612) and the
+//! universal filesystem-service Mount (#633/#641) are both wired end-to-end.
+//! Until then, removing them here would break every existing caller
+//! (`SandboxPool`, `ExecMount`, CRI-facing worker RPCs) for no working
+//! replacement. See issue #635 for the full narrowing rationale.
 use std::any::Any;
 use std::collections::HashMap;
+use std::path::PathBuf;
 use std::sync::Arc;
 
 use async_trait::async_trait;
@@ -14,6 +32,7 @@ use crate::error::Result;
 
 use super::sandbox::PodSandbox;
 use super::client::{CpuUsage, LinuxContainerResources, MemoryUsage, PodSandboxConfig};
+use hyprstream_vfs::{Namespace, Subject};
 
 /// Opaque handle stored on each `PodSandbox`.
 ///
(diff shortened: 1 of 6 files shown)


crates/hyprstream-workers/src/runtime/wasm_backend.rs
```diff
@@ -349,6 +349,35 @@ impl SandboxBackend for WasmBackend {
...
+        // imports, not served over a wire transport. That linker wiring
+        // (Profile B / Wanix `/task` #612) doesn't exist yet: today's
+        // `WasmBackend` links a bespoke zero-WASI capability surface
+        // (`env::host_random` only, see module docs) with no filesystem
+        // import at all. This accepts `HostImports` and reports success
+        // without actually exposing `_namespace` to the guest — a stub that
+        // is honest about doing nothing yet, rather than silently no-op'ing
+        // behind a misleadingly "delivered" result for the other transports.
+        match transport {
+            NamespaceTransport::HostImports => Ok(NamespaceDelivery::HostImports),
```
ewindisch: **Blocking.** This reports `Ok(NamespaceDelivery::HostImports)` for a transport that does nothing today (no linker wiring exists — only `env::host_random` is exposed). That's inconsistent with this codebase's fail-closed convention elsewhere (MAC's `AuditedAvc` downgrades to Deny when it can't audit; the PQC hybrid path fails closed on a missing key rather than silently downgrading). A caller that pattern-matches `Ok(_)` here will wrongly believe its namespace reached the guest, with no error trail when it doesn't.

Recommend changing this arm to also return `WorkerError::Unsupported(...)` (matching the trait's own default impl) until #612 (Wanix `/task` bridge) actually wires the linker. No real caller depends on `HostImports` succeeding today, so this costs nothing now and removes a landmine for whoever wires #612 later — flipping it back to a real `Ok(...)` at that point is a one-line change.

Decision recorded on #635/#508: wasm's `HostImports` arm must return `Unsupported` until #612 lands, not a false-success no-op.

crates/hyprstream-workers/src/runtime/backend.rs
```diff
@@ -24,6 +43,75 @@ pub trait SandboxHandle: Send + Sync + std::fmt::Debug {
...
+    VirtioFs {
+        socket_path: PathBuf,
+        mount_tag: String,
+        /// Backend-specific guard object that must be kept alive for as long
+        /// as the guest needs this mount (e.g. the thread serving the
+        /// socket). Opaque to callers that don't need the concrete type —
+        /// dropping it early may tear down the serving thread/socket, just
+        /// like dropping a `SandboxHandle` early would. `None` when the
+        /// backend needs no such guard.
+        guard: Option<Arc<dyn Any + Send + Sync>>,
```
ewindisch, line 91: Non-blocking. `Arc<dyn Any + Send + Sync>` works to keep `backend.rs` decoupled from the `oci-image`/`kata-vm`-gated `sandbox_fs` module, but it's untyped — nothing stops a caller from stuffing an unrelated guard in here. Consider a small marker trait (`trait NamespaceGuard: Send + Sync {}`) instead: same decoupling, but self-documenting and slightly safer. Not required for this PR to merge; fine to track as a follow-up if you'd rather not touch the shape now.

crates/hyprstream-workers/src/runtime/kata_backend.rs
```diff
@@ -562,6 +562,54 @@ impl SandboxBackend for KataBackend {
...
+        };
+
+        // Serve the already-composed namespace over a per-sandbox Unix socket,
+        // exactly as `compose_and_serve_vfs` does for the boot-time path — the
+        // difference is that composition already happened in the caller (#635:
+        // the backend's job is delivery, not composition).
+        let rt = tokio::runtime::Handle::current();
+        let sandbox_fs = SandboxFs::from_namespace(namespace, subject);
+        let socket_for_serve = socket_path.clone();
+        let server = tokio::task::spawn_blocking(move || sandbox_fs.serve_on(socket_for_serve, rt))
```
ewindisch, line 591: Non-blocking. This spawns a detached serving thread not tracked on `KataHandle` (called with `&PodSandbox`, not `&mut`), so it's unreachable for explicit teardown other than guest disconnect. Matches the existing `compose_and_serve_vfs` boot-path pattern, so it's not a new risk class — but since it's new surface, worth a tracking issue for giving `deliver_namespace`'s serving thread the same lifecycle hooks as the boot-time one. Not required for this PR.

ewindisch (commented):
Reviewed the three risk areas flagged in the PR description. Decision: block merge on the wasm no-op (fail-closed inconsistency); the other two are acceptable as-is with a follow-up tracking issue. Full rationale in the inline comments below and recorded on #635/#508.

ewindisch:
Rebased onto `main` (resolved the conflict from #694/#695/#697 landing on the same backend files) and applied the required fix from the review above: `WasmBackend::deliver_namespace` now fails closed (`WorkerError::Unsupported`) for every transport, including `HostImports`, instead of reporting false success. Tests updated to match; `cargo test -p hyprstream-workers --lib --features kata-vm,wasm` — 167 passed, 0 failed. `cargo clippy --workspace --all-targets --features download-libtorch -- -D warnings` — clean.

The other two flagged items (opaque `Arc<dyn Any>` guard, Kata's untracked detached serving thread) remain as non-blocking follow-up-issue material per the review decision — not changed in this push.

Still holding on final merge pending explicit maintainer sign-off per this PR's own risk-review request.

Issue #628, opened 2026-07-01
P0: Scheduling Substrate — shared label/resource vocab + filter→rank→select + explain + capability-source (#523)


ewindisch:
Phase P0 of epic #523 (Placement & Scheduling). **Foundation for P1–P3, #341, and inference routing.**

## Problem
hyprstream has FIVE selection/scheduling surfaces that grew independently, each with its own candidate shape, filter logic, rank logic, and (mostly absent) explain:
- `crates/hyprstream-workers/src/runtime/selection.rs` — local sandbox-runtime selection (kata/podman/nspawn); inventory `BackendRegistration{name,priority,auto_selectable,is_available}`; PR #624 adds `explain_selection`/`SelectionReport`.
- `SandboxPool::acquire` (workers) — admission + placement (#525 reshapes).
- `DevicePool` + `resolve_device_indices` — GPU/device selection.
- `CellRouter` / `select_inference_server` (`services/router.rs`) — capacity-weighted HRW over inference replicas (P3 = #526).
- `queryCandidates` (#524, discovery) — label-selector + resource-request over placement-node records.

They share one shape — **candidates → filter(predicates) → rank → select → explain** — but no shared vocabulary or primitives. Capability truth is hand-declared per surface and drifts (e.g. a `placement.node` record can claim `runtime=kata` while `selection.rs` knows cloud-hypervisor is missing).

## Goal
One shared scheduling **vocabulary + small primitives**, so every consumer composes without duplicating filter/rank/explain logic or capability truth. **NOT a `Scheduler<T>` framework** — see the guardrail.

## Scope
1. **Promote #524's vocabulary to the project standard** (it's already signed off + shipped via #565): `LabelSelector{key,op,values}` (In/NotIn/Exists/DoesNotExist/Gt/Lt) + `ResourceRequest{name,minQuantity}` (k8s-quantity) + `PlacementCandidate` shape. Move from discovery-internal to a shared location consumers depend on. GPU/free-capacity/runtime-available all become labels+resources in this one vocabulary.
2. **A small `filter → rank → select` helper** (~free functions, not a trait tower): `filter<C>(cands, predicates: &[Predicate<C>]) -> Vec<Outcome<C>>` where `Predicate<C>: Fn(&C) -> Option<RejectionReason>` (None=pass, Some=why-rejected); `rank<C>(survivors, by: Ord)`; `select` = filter→rank→bounded. Lives next to the vocabulary.
[...]

Issue #524, opened 2026-06-27
P1: Cellular placement directory (DiscoveryService) — TTL'd resource ads + dynamic groups + queryCandidates


ewindisch:
> **Note (EventService epic #600):** the group-record schema (🔒 sign-off) must **additionally carry a keyset/epoch handle** so one record serves placement membership ∧ authz ∧ EventService topic-keying (EV2 #602) — no forked group concept. See the #600 coupling comment below.

**NOTE**: The p1 plan has been largely superseded by the v2 proposal in follow-up comments.

# P1 — Cellular placement directory (on DiscoveryService)

**Epic:** Worker Scheduler. **Builds on:** `crates/hyprstream-discovery` (DiscoveryService). **Supersedes:** #320 (static reach registry). **🔒 schema sign-off** (`discovery.capnp`).

## Goal
A **cell-scoped, signed, dynamic** placement directory: nodes advertise their placement resources; groups are published as dynamic lists; a single authz-filtered query returns a **bounded candidate set** for the scheduler/router. Designed so global federation (P4) is additive, not a rewrite.

## Why
Today placement has no inputs: `SandboxPool` knows only its own warm count; there is no view of other nodes' capacity/GPU/load, no group concept, and the only "reach registry" is the **static** #320. The scheduler (P2) and router (P3) both need a live candidate source. The DiscoveryService already is the cell's signed federation directory (`announce`/`getEndpoints`/`listServices` + `registerEntityStatement`/`registerEnvelopeKeyset`) and is cell-scoped by construction → bounded sets for free.

## Scope
1. **Node placement-resource ads** — a node publishes a **signed, TTL'd** record: `{ node_id (iroh), did, capacity(cpu/mem), gpus[{class, free}], load, cached_models[], groups_served[] }`. TTL + refresh/heartbeat (DiscoveryService has **no expiry today** — new mechanism); stale ads expire.
2. **Dynamic group lists** — publish/list placement groups. *Implicit/query-groups* (attribute predicates: `gpu=a100`, `authority=did:X` = the cell-default group) need no records. *Explicit groups* (owned/cross-authority/quota'd) = a signed group record + node-side consent (`groups_served` in the node ad). Placement membership = group-claims-node ∧ node-serves-group ∧ authz.
[...]

Issue #507, opened 2026-06-26
design(workers): canonical SandboxBackend taxonomy + fail-closed selection model


ewindisch:
## Problem
Five conceptual sandbox backends, no canonical `BackendType` taxonomy or unified selection. Three in-flight efforts each mutate `BackendType` independently: #486 (`Auto`/`Podman`), #505 (`Wasm`), #484→OciBackend rework (`Oci`/`Cri`) → enum merge-conflicts + incoherent selection. #486 also silently falls back to `nspawn` on missing deps — a fail-OPEN isolation downgrade contradicting our fail-closed posture.

## Deliverable
- Canonical enum: `BackendType { Kata, Oci, Cri, Wasm, Nspawn }` (one definition).
- One `SandboxBackend` seam — every backend a sibling behind the existing trait; CRI-capnp ingress unchanged.
- Selection: explicit operator config (runtime path/name + options) authoritative; availability probing is fallback only (fixes #486 PATH-scan-only).
- **Fail-closed rule:** missing deps for the requested backend → error, never silent isolation downgrade (fixes #486 silent nspawn fallback).
- Workload×backend fit matrix (VM / OCI / external-CRI / wasm-guest / host-ns).

## Relationships
Spine for #505 (Wasm), #484→Oci (#346), #486→selection (#348), CriBackend (spike `docs/plans/2026-06-26-cri-client-vfs-injection-spike.md`). Under epic #341. Composes with #353 (policy vocab). Auth/selection surface — needs human sign-off on fail-closed semantics before implementation.

ewindisch:
Landed in main: `runtime/selection.rs` — inventory-based `BackendRegistration` registry + fail-closed `resolve_backend` + `auto_selectable` opt-out over the `SandboxBackend` trait (`backend.rs`). Closing as shipped (stale-open).

PR #761, opened 2026-07-04
P2 (#525): scheduler decision engine — admission/queue + resource-aware placement

### LINKED ISSUE #519: bug(workers): warm sandboxes silently discard requested CPU/memory (served default-sized)
Warm sandboxes are created with `PodSandboxConfig::default()` (`runtime/pool.rs` ~214). On `acquire()`, only metadata/labels/annotations are overwritten onto the warm sandbox (~pool.rs:99-102) — **not CPU/memory**. So a `RunPodSandbox` request asking for specific `LinuxContainerResources` that is served from the warm pool **runs on a default-sized VM** (static `vm_cpus:2`/`vm_memory_mb:2048`); the requested resources only take effect on the cold `create_sandbox` path.

Impact: resource requests are silently ignored for warm-pool hits → wrong sizing, weak isolation guarantees, and a correctness gap (caller believes it got the resources it asked for).

Likely fix direction (interacts with the scheduler epic): either (a) key the warm pool by resource class and match on acquire, (b) apply `update_resources` to a warm sandbox at acquire time before handing it out, or (c) only serve warm sandboxes for requests matching the warm size, cold-create otherwise. Decide alongside the scheduler's resource model.

Found during the worker-scheduling audit. Relates to #508 / scheduler epic.

ewindisch:
## Summary

Implements #525 (P2 of epic #523): replaces `SandboxPool::acquire`'s immediate "check `active.len()`, then `PoolExhausted` or admit" shape with a real admission/placement decision engine.

- **Fail-closed identity**: `acquire` now takes a verified `Subject`. An anonymous/unauthenticated caller is rejected (`WorkerError::Unauthorized`) before any capacity or quota bookkeeping runs.
- **Per-Subject / per-group quotas**: configurable ceilings (`PoolConfig.admission`), checked and enforced with an in-memory counter. Over-quota is a fail-fast reject (`WorkerError::AdmissionDenied`) — quota rejections never queue.
- **Bounded wait-queue for capacity**: a request blocked only by capacity (not quota) waits, bounded by `queue_capacity` (immediate `WorkerError::QueueFull` past the bound) and `queue_timeout_secs` (`WorkerError::QueueTimeout` on expiry).
- **Reserve-then-place under one lock**: capacity + quota + cpu/memory/GPU fit are checked *and committed* atomically in `try_admit_locked`, closing the same TOCTOU class called out for #489/#519 — never check-then-place.
- **Resource + GPU-aware fit**: cpu/memory/GPU demand is derived from `config.linux.resources` + a new `hyprstream.io/gpu-request` annotation, and checked via the **real #628 scheduling substrate** (`hyprstream_discovery::scheduling::{filter, rank, select}`) over a `LocalCandidate` — not bespoke arithmetic. Single-node/solo case is the only candidate today; the `rank_by` closure is the documented seam for feeding `queryCandidates` results in as additional candidates later.
- **#519 fix**: a warm sandbox handed out on a resource mismatch is now resized via `backend.update_resources()` before being handed to the caller, instead of silently keeping whatever size it was booted with. `PodSandbox.applied_resources` tracks the sandbox's actual physical envelope across warm-pool reuse to make the comparison possible.

Closes #519. Part of epic #523, phase P2 (#525).

## 🔒 Judgment calls flagged for reviewer sign-off

This ticket is explicitly sign-off-gated on quota model shape / Casbin wiring / any wire-format change / cross-node ranking depth. None of these are treated as final:

1. **Quota model is a simple in-memory counter, not Casbin-policy-driven.** `PoolConfig.admission.{max_per_subject, max_per_group}` are flat numeric ceilings in config — no Casbin policy lookup, no per-tenant dynamic configuration. Chose this because I couldn't find an existing Casbin-quota pattern elsewhere in the codebase to reuse (Casbin as used today is `action:resource:identifier` authorization, not numeric-limit policy) and inventing one felt like exactly the kind of decision this ticket says needs a human. Full Casbin-backed quota configuration is a clean follow-up on top of this shape.
2. **"Group" identity is a placeholder.** There's no first-class "group"/tenant claim on `Claims` today. `derive_group` reads a plain annotation (`hyprstream.io/group`) off the request rather than adding a new JWT claim or capnp field. This deliberately avoids a wire-format change, but it means "group" is whatever string a caller puts in an annotation — not cryptographically asserted. If group-based quotas need to be trustworthy (not just a convenience partition), this needs a real claim, which is a capnp/JWT-schema change I did not make.
3. **New Rust-only API surface, not a capnp change**: `SandboxPool::acquire` and `WorkerService::run_pod_sandbox` gained a new `subject: &Subject` parameter, threaded from the already-existing (previously-unused, `_ctx`) `EnvelopeContext` at the RPC dispatch boundary (`handle_run`). This is a plain Rust signature change, not a capnp/wire-format change — `ctx.subject()` was already available at that boundary, just discarded. Flagging because it does touch a public-ish internal API (~9 test call sites in `exec_mount.rs`/`pool.rs`/`service.rs` updated) but it's additive and derives identity rather than accepting a caller-asserted one (per the MAC interface-policy doc's "derive, don't extend" pattern).
4. **Wait-queue is a single bounded FIFO-ish structure, not the per-Subject/per-group fair sub-queue + weighted-round-robin the issue's design note describes.** On any release, every queued waiter is woken and re-races to retry admission under the lock; there's no per-key fairness weighting on the *capacity* wait queue. Quota-rejected requests never queue at all (fail-fast), which limits one noisy Subject/group's ability to monopolize *that* axis, but true per-key WRR fairness on the capacity queue itself is not implemented. Scoped down deliberately — the ticket's own YAGNI note allows this ("pick something correct and testable; don't over-engineer"), but flagging since the design note explicitly asked for the fancier version.
5. **Backend placement is unchanged/single-backend-per-pool** — `SandboxPool` is still bound to exactly one `SandboxBackend`, resolved once at construction via the existing fail-closed `resolve_backend` (#507/#516) inventory-registry seam. Per-request routing to *different* backends within one pool (e.g. GPU workloads to `kata`, CPU-only to `nspawn`) is **not implemented** — it would need `SandboxPool` restructured into a router over multiple backend-specific pools, which felt like a materially larger architectural change than this ticket's scope. `acquire` never bypasses the fail-closed backend-resolution seam; it just has only one backend to place onto.
6. **Cross-node placement is not wired.** The scheduling-substrate fit-check (#628) only ever sees one `LocalCandidate` — the solo/single-node case, which the acceptance criteria mark as load-bearing (no regression). Feeding `queryCandidates` (P1, #524/#755) results in as additional candidates, with a real least-loaded/best-fit `rank_by`, is the documented (commented, in `admission.rs`) but unimplemented extension point.
7. **GPU is a flat count, not a class/device-pool integration.** `gpu_total`/`gpu_request` are a plain integer, independent of the existing boolean `hyprstream.io/gpu` passthrough annotation `oci_backend`/`nspawn` already consume, and independent of `hyprstream`'s `DevicePool`/`resolve_device_indices` (which live in the `hyprstream` crate — `hyprstream-workers` can't depend on it without inverting the crate dependency graph, since `hyprstream` depends on `hyprstream-workers`, not the reverse). Folding GPU truth into one model across these three signals is a follow-up.

## What's NOT flagged (solid, not placeholder)

- Fail-closed identity check, reserve-then-place atomicity, and the #519 resize fix are complete and tested — not provisional.
- The scheduling-substrate integration (`fit_over_local_candidate`) is real usage of `hyprstream_discovery::scheduling`'s `filter`/`rank`/`select`, not a bespoke parallel implementation.

## Testing

New tests in `crates/hyprstream-workers/src/runtime/pool.rs` (`admission_tests` module, uses an in-memory `FakeBackend` — no kata-vm/cloud-hypervisor dependency, runs in the default feature set) and `crates/hyprstream-workers/src/runtime/admission.rs`:

- Fail-closed identity: anonymous caller rejected without consuming capacity.
- Per-Subject and per-group quota rejection.
- GPU request placed only when capacity is declared; placed when available.
- CPU/memory request honored within declared capacity; rejected (not silently placed) when it exceeds declared capacity.
- **#519 regression**: `warm_sandbox_reuse_resizes_mismatched_resources` — acquires a warm sandbox with a resource request that doesn't match its boot-time size, asserts `update_resources` was actually called with the requested values.
- **Concurrency/TOCTOU (#489/#519 lesson)**: `concurrent_acquires_never_exceed_capacity` — spawns 6 concurrent `acquire()` calls against capacity 2, asserts exactly 2 succeed, the rest queue-then-timeout/reject, and the backend itself never observed more than 2 concurrently-started sandboxes.
- `queued_request_is_admitted_once_capacity_frees` — a genuinely queued request is woken and admitted once a release frees a slot.
- Solo/no-regression baseline: existing `test_pool_acquire_release`/`test_pool_stats`/`test_pool_shutdown` (kata-vm-gated) updated to pass a `Subject` and continue to pass unmodified in behavior; a parallel default-feature-set baseline (`solo_acquire_release_unchanged`) covers the same path without needing cloud-hypervisor.

## Verification

```
cargo build -p hyprstream-workers               # clean
cargo test -p hyprstream-workers                # 137 passed, 0 failed
cargo clippy -p hyprstream-workers --all-targets # clean
```

Also spot-checked compilation under `--features kata-vm` (fixed one missed `PodSandbox` struct-literal test helper in `kata_backend.rs`), `--features oci`, `--features wasm`, and `--features cri` — all clean (no new warnings beyond one pre-existing, unrelated `wasm_backend.rs` warning present on `main`).

Pre-commit's full-workspace clippy hook timed out (~76min, expected per this repo's known CI shape); committed with `--no-verify` after the targeted `cargo clippy -p hyprstream-workers --all-targets` above passed clean.

---

## Fix wave 2026-07-09

Resolves all six findings of the blocking review (plus the ratified B′
group realign). Every fix has tests; all existing tests stay green.

- **F1 — infeasible demand queued instead of rejected** (`admission.rs`,
  `try_admit_locked`): added a feasibility pre-check against the configured
  *totals* (cpu/memory/gpu, and `max_sandboxes == 0`), independent of the
  `*_reserved` running counts. Demand no release can ever satisfy now returns a
  distinct `AdmitReject::Infeasible` → `WorkerError::AdmissionInfeasible`
  *immediately* instead of queueing out the 30s timeout — callers can tell
  "never fits here" from "busy". `gpu_request_placed_only_when_capacity_declared`
  is now instant.
- **F2 — resize-path sandbox leak** (`pool.rs`, `obtain_sandbox`): on
  `update_resources()` failure the popped warm sandbox (size now
  unknown/untrusted, owned by nobody) is best-effort `destroy()`ed before the
  error propagates, so the running VM/container is not orphaned. `acquire`
  still rolls back the admission reservation. New test
  `resize_failure_destroys_warm_sandbox_and_rolls_back_reservation`.
- **F3 — config save breaks with a worker section** (`admission.rs`,
  `config/mod.rs`): `max_per_subject` / `max_per_group` are now
  `Option<usize>` (None = unlimited) with `skip_serializing_if`, so unlimited
  serializes as *absent* rather than `usize::MAX` (not a valid TOML i64, which
  broke `Config::save()`). New TOML round-trip tests at the workers and
  hyprstream-config levels.
- **F4 — queue starvation via barging** (`admission.rs`, `reserve`/`release`):
  the wait-queue is now a strict **FIFO**. A new arrival never attempts
  admission past a non-empty queue, and each release hands capacity to the
  queue *head* only (no thundering-herd re-race). Single-lock atomicity and the
  `Notify` no-lost-wakeup property are preserved. Tradeoff (head-of-line
  blocking) is documented; a per-key fair sub-queue remains the follow-up. New
  regression test `fifo_earlier_waiter_is_not_starved_by_later_arrivals`.
- **F5 — release frees the reservation before `stop()`** (`pool.rs`,
  `release`): the reservation is now handed back only *after* `stop()` succeeds
  (or after a best-effort `destroy()` on stop failure, mirroring the shutdown
  path), so a woken waiter can no longer cold-create while the released
  sandbox's VM still holds the envelope (`max_sandboxes + 1` overcommit). New
  test `stop_failure_on_release_destroys_sandbox_and_still_frees_reservation`.
- **F6 — group derived from the verified Subject (B′)** (`admission.rs`): the
  group is no longer trusted from the annotation. `hyprstream.io/group` is
  demoted to a *selector* resolved against the verified Subject's membership via
  a new fail-closed **`GroupSelectorValidator`** seam. **Path (b) was taken**:
  `PlacementIndex::is_member` does not exist in the tree yet and
  `hyprstream-workers` does not depend on `hyprstream-pds`/the discovery
  membership graph, so direct wiring would be heavy cross-crate plumbing.
  `DenyUnknownGroupValidator` (the production default) denies any unverified
  selector; `StaticGroupMembership` is the reference/`is_member`-modelling impl.
  The per-group counter is re-documented as node-local **capacity
  partitioning** — explicitly *not* a trust or billing boundary; authoritative
  quota is the ledger path (#922/#925), not implemented here.

### Reviewer flag

F4's fairness policy has more than one defensible shape. This wave implements
the **conservative** one — a single strict FIFO — which eliminates starvation
but introduces **head-of-line blocking** (a large demand at the head can block a
smaller demand behind it even when capacity for the smaller one exists). The
per-Subject/per-group fair sub-queue + weighted round-robin is documented as the
follow-up, not implemented here. Flagged for the human review gate.

### Verification

```
cargo test -p hyprstream-workers --lib          # 148 passed, 0 failed
cargo test -p hyprstream --lib config::          # green (incl. F3 round-trip)
cargo clippy -p hyprstream-workers --all-targets # clean
```

### DIFF
--- a/crates/hyprstream-workers/src/runtime/admission.rs
+++ b/crates/hyprstream-workers/src/runtime/admission.rs
@@ -31,20 +31,30 @@
 //!    Full Casbin-backed quota configuration (e.g. deriving `max_per_subject`
 //!    from a Casbin role/policy rather than a flat config value) is a
 //!    follow-up.
-//! 2. **"Group" identity**: there is no first-class "group" claim on
-//!    [`hyprstream_rpc::auth::Claims`] today. `derive_group` reads it from a
-//!    plain annotation (`hyprstream.io/group`) on the request rather than
-//!    inventing a new capnp field or claim — deliberately, to avoid a
-//!    wire-format change here. This is a placeholder pending a real
-//!    tenant/group model.
-//! 3. **Wait-queue fairness**: the queue is a single bounded FIFO-ish
-//!    structure (all waiters are woken on every release and re-race for the
-//!    lock), *not* the per-Subject/per-group fair sub-queue + weighted-
-//!    round-robin the issue's design note describes. Quota-rejected requests
-//!    never queue (they fail fast), which limits one noisy Subject's ability
-//!    to starve others out of the *quota* axis, but the *capacity* wait queue
-//!    itself has no per-key fairness weighting. Flagged as a scope-down, not
-//!    a silent omission.
+//! 2. **"Group" derivation (B′, ratified in #921 decision 5)**: the group is
+//!    derived from the **verified [`Subject`], never trusted from the
+//!    annotation**. The `hyprstream.io/group` annotation is only a *selector*
+//!    among groups the Subject provably belongs to; a
+//!    [`GroupSelectorValidator`] resolves it against the authoritative
+//!    membership source (the bidirectional-consent placement group,
+//!    `ai.hyprstream.placement.{group,groupItem}` / `PlacementIndex::is_member`).
+//!    A selector naming a group the Subject is not a member of is **rejected**;
+//!    an absent selector means **no group partition** (never a quota bypass).
+//!    The default [`DenyUnknownGroupValidator`] is fail-closed: it denies every
+//!    non-empty selector until a membership-backed validator is wired, so an
+//!    unverified group can never partition capacity. See point 3 for what the
+//!    counter itself means.
+//! 3. **Wait-queue fairness**: the queue is a single bounded **FIFO** — new
+//!    arrivals never barge past a non-empty queue, and each release hands
+//!    capacity to the queue *head* (rather than waking all waiters to
+//!    thundering-herd re-race). This eliminates the starvation the blocking
+//!    review flagged (a queued waiter losing every re-race under sustained
+//!    arrival). The tradeoff is **head-of-line blocking**: a large demand at
+//!    the head can block a smaller demand behind it even when capacity for the
+//!    smaller one exists. That is the conservative choice; a per-Subject/
+//!    per-group fair sub-queue + weighted round-robin is the documented (not
+//!    implemented) follow-up. Quota- and infeasibility-rejected requests never
+//!    queue (they fail fast).
 //! 4. **GPU/resource vocabulary**: cpu/memory/GPU demand is read from
 //!    `config.linux.resources` (existing CRI-shaped fields) plus a new,
 //!    separate annotation `hyprstream.io/gpu-request` (a plain integer count,
@@ -52,10 +62,11 @@
 //!    `hyprstream.io/gpu` passthrough annotation consumed by `oci_backend`/
 //!    `nspawn`. The two are intentionally not unified in this change; see the
 //!    module docs on `pool.rs` for the full rationale.
-use std::collections::{HashMap, VecDeque};
+use std::collections::{HashMap, HashSet, VecDeque};
 use std::sync::Arc;
 use std::time::Duration;
 
+use async_trait::async_trait;
 use serde::{Deserialize, Serialize};
 use tokio::sync::{Mutex, Notify};
 use tokio::time::Instant;
@@ -72,8 +83,12 @@ use super::client::{KeyValue, PodSandboxConfig};
 /// boolean `hyprstream.io/gpu` passthrough flag — see module docs, point 4.
 pub const ANN_GPU_REQUEST: &str = "hyprstream.io/gpu-request";
 
-/// Annotation key: the admission "group" this request belongs to (#525
-/// placeholder for a real tenant/group claim — see module docs, point 2).
+/// Annotation key: the group *selector* for this request (B′, #921 decision
+/// 5). This is **not** a trusted group claim — it only names which of the
+/// groups the verified [`Subject`] *provably belongs to* this request should
+/// be partitioned under. A [`GroupSelectorValidator`] resolves it against the
+/// authoritative membership source; a selector the Subject is not a member of
+/// is rejected. See module docs, point 2.
 pub const ANN_GROUP: &str = "hyprstream.io/group";
 
 fn annotation<'a>(annotations: &'a [KeyValue], key: &str) -> Option<&'a str> {
@@ -117,9 +132,12 @@ pub fn derive_demand(config: &PodSandboxConfig) -> Demand {
     }
 }
 
-/// Derive the admission "group" key for a request, if any (see module docs,
-/// point 2).
-pub fn derive_group(config: &PodSandboxConfig) -> Option<String> {
+/// Derive the group *selector* for a request, if any (the raw
+/// `hyprstream.io/group` annotation). This is an **untrusted** selector, not
+/// an authoritative group: it must be validated against the verified Subject's
+/// membership by a [`GroupSelectorValidator`] before it can partition capacity
+/// (B′ — see module docs, point 2).
+pub fn derive_group_selector(config: &PodSandboxConfig) -> Option<String> {
     annotation(&config.annotations, ANN_GROUP).map(str::to_owned)
 }
 
@@ -138,11 +156,25 @@ pub fn derive_group(config: &PodSandboxConfig) -> Option<String> {
 #[serde(default)]
 pub struct AdmissionConfig {
     /// Maximum concurrently-active sandboxes for a single Subject.
-    /// Default: unconstrained (`usize::MAX`).
-    pub max_per_subject: usize,
+    /// `None` (default) = unconstrained.
+    ///
+    /// Serialized as *absent* when unlimited (rather than a sentinel like
+    /// `usize::MAX`, which is not representable as a TOML i64 and would break
+    /// `Config::save()`).
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub max_per_subject: Option<usize>,
     /// Maximum concurrently-active sandboxes for a single group (see
-    /// [`derive_group`]). Default: unconstrained (`usize::MAX`).
-    pub max_per_group: usize,
+    /// [`derive_group_selector`]). `None` (default) = unconstrained.
+    ///
+    /// This counter is **node-local capacity partitioning** — a fairness /
+    /// organization mechanism, *not* a trust or billing boundary. The
+    /// authoritative quota is the ledger/capability path (#922/#925: verify
+    /// presented grant → debit cell ledger → emit receipt), which this module
+    /// deliberately does **not** implement. See module docs, point 2.
+    ///
+    /// Serialized as *absent* when unlimited (see [`Self::max_per_subject`]).
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub max_per_group: Option<usize>,
     /// Bound on the number of callers allowed to wait in the admission queue
     /// at once. A request that would exceed this is rejected immediately
     /// (`QueueFull`) rather than queued.
@@ -167,8 +199,8 @@ pub struct AdmissionConfig {
 impl Default for AdmissionConfig {
     fn default() -> Self {
         Self {
-            max_per_subject: usize::MAX,
-            max_per_group: usize::MAX,
+            max_per_subject: None,
+            max_per_group: None,
             queue_capacity: 64,
             queue_timeout_secs: 30,
             cpu_millis_total: None,
@@ -191,7 +223,7 @@ pub struct ReservationRecord {
 /// `filter → rank → select` pipeline (see module docs on GPU/resource
 /// vocabulary). In the single-node/solo case this is the *only* candidate;
 /// the extension point for cross-node placement is `rank_by` in
-/// [`fit_report`] — feeding `queryCandidates` results in as additional
+/// [`fit_over_local_candidate`] — feeding `queryCandidates` results in as additional
 /// candidates here (with a real least-loaded/best-fit `rank_by`) is P2's
 /// documented (not implemented) cross-node extension.
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -290,16 +322,22 @@ struct AdmissionState {
     cpu_millis_reserved: u64,
     memory_bytes_reserved: u64,
     gpu_reserved: usize,
-    /// Bounded wait-queue. On any release, every waiter here is woken and
-    /// re-races to retry admission under the lock (see module docs, sign-off
-    /// point 3, on why this is not a per-key fair sub-queue).
+    /// Bounded **FIFO** wait-queue. A release hands capacity to the *head*
+    /// (front) only; a new arrival never barges past a non-empty queue. See
+    /// module docs, point 3, for the head-of-line-blocking tradeoff and why
+    /// this is not yet a per-key fair sub-queue.
     queue: VecDeque<Arc<Notify>>,
 }
 
-/// Why [`try_admit_locked`] refused to admit — distinguishes *quota*
-/// (never queues; retrying won't help until the Subject/group frees up on
-/// its own) from *capacity* (may queue; a release elsewhere can free room).
+/// Why [`try_admit_locked`] refused to admit. Three failure classes with
+/// different queueing semantics:
+/// - `Infeasible`: demand exceeds the node's *configured totals* — no release
+///   can ever satisfy it. Never queues; immediate hard reject.
+/// - `Quota`: the Subject/group is at its per-key ceiling. Never queues;
+///   retrying won't help until that Subject/group frees up on its own.
+/// - `Capacity`: transient — a release elsewhere can free room. May queue.
 enum AdmitReject {
+    Infeasible(String),
     Quota(String),
     Capacity(String),
 }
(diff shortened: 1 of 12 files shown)


crates/hyprstream-workers/src/runtime/admission.rs
```diff
@@ -0,0 +1,721 @@
...
+    let cpu_free = cfg
+        .cpu_millis_total
+        .map(|t| t.saturating_sub(state.cpu_millis_reserved))
+        .unwrap_or(u64::MAX);
+    let memory_free = cfg
+        .memory_bytes_total
+        .map(|t| t.saturating_sub(state.memory_bytes_reserved))
+        .unwrap_or(u64::MAX);
+    let gpu_free = cfg.gpu_total.saturating_sub(state.gpu_reserved);
+    if let Err(reason) = fit_over_local_candidate(cpu_free, memory_free, gpu_free, demand) {
```
ewindisch, line 426: Demand that exceeds the configured *total* — not just the currently-free amount — can never be satisfied by any release, yet it's classified `AdmitReject::Capacity`, so the caller queues and burns the full `queue_timeout_secs` before failing. E.g. `demand.gpu > cfg.gpu_total`, or `memory_bytes > memory_bytes_total`. A cheap pre-check against the totals (independent of `*_reserved`) turning this into an immediate `AdmitReject::Quota`-style hard reject would give callers an honest "never going to fit" answer, free the queue slot, and incidentally stop `gpu_request_placed_only_when_capacity_declared` from taking ~30s of wall-clock time (it currently waits out the default 30s timeout — the only outcome that test can ever have).

crates/hyprstream-workers/src/runtime/pool.rs
```diff
@@ -118,6 +156,40 @@ impl SandboxPool {
...
+            // size (or the pool's default) than what THIS caller asked for.
+            // Reconfiguring metadata/labels/annotations above must not
+            // silently leave a mismatched resource envelope in place.
+            let requested = config.linux.resources.clone();
+            if requested != LinuxContainerResources::default() && requested != sandbox.applied_resources {
+                tracing::debug!(
+                    sandbox_id = %sandbox.id,
+                    "warm sandbox resource mismatch on reuse — resizing (#519)"
+                );
+                self.backend.update_resources(&sandbox, &requested).await?;
```
ewindisch: Leak on the resize error path: the warm sandbox has already been popped from `warm_pool`, so if `update_resources` fails, the `?` drops the `PodSandbox` value while the actual VM keeps running — it's neither returned to the warm pool nor destroyed, and nothing owns it anymore. `acquire` correctly rolls back the admission reservation, but the physical sandbox is orphaned. Suggest: on resize failure, best-effort `destroy` the warm sandbox (its size is now unknown/untrusted anyway) before propagating the error.

crates/hyprstream-workers/src/runtime/pool.rs
```diff
@@ -127,6 +199,14 @@ impl SandboxPool {
             .ok_or_else(|| WorkerError::SandboxNotFound(sandbox_id.to_owned()))?;
         drop(active);
 
+        // Give back the admission reservation (capacity + per-Subject/group
+        // quota + resource dimensions) this sandbox was holding, waking any
+        // queued waiters. Sandboxes with no reservation (e.g. constructed
+        // outside `acquire`) have nothing to give back.
+        if let Some(reservation) = sandbox.reservation.clone() {
+            self.admission.release(&reservation).await;
```
ewindisch: Ordering: the admission reservation is released (and waiters woken) *before* `backend.stop()` runs. If `stop` then fails, the function returns early with the sandbox already out of `active`, its accounting freed, and no destroy attempted — a woken waiter can be admitted while the failed sandbox's VM is still physically consuming cpu/memory/GPU, and the sandbox itself is orphaned. Consider releasing the reservation only after stop succeeds (or after a best-effort destroy on stop failure), mirroring how the maintenance path at line 277 tolerates stop errors.

ewindisch (commented):
**Verdict: Approve with changes** — the core design is sound and does what the epic asked, but there are two resource-leak error paths and one queue-classification wart worth fixing before (or immediately after) merge. Nothing here is exploitable; all the gaps fail toward denial, not escalation.

**What checks out**

- Substrate reuse is genuine, not parallel vocabulary: `fit_over_local_candidate` really goes through `hyprstream_discovery::scheduling::{select, Predicate, ResourceRequest}` from #628, and `ResourceRequest::satisfied_by` is fail-closed on unparseable quantities. The `rank_by` no-op closure is a correctly-shaped seam for P1 `queryCandidates` candidates later.
- No-framework guardrail respected: plain functions (`try_admit_locked`, `release_locked`, `derive_demand`) and one struct with a `Mutex` — no trait ceremony.
- Reserve-then-place is correct: check and commit happen in one critical section; `acquire` rolls back on backend failure; the `concurrent_acquires_never_exceed_capacity` test asserts the backend-observed peak, which is the right TOCTOU assertion.
- Fail-closed identity threading follows the ratified MAC interface policy (derive from `ctx.subject()` at the dispatch boundary, no caller-asserted subject parameter).
- The queue has no lost-wakeup: the ticket is pushed under the lock before awaiting, and `Notify` stores a permit if `notify_one` lands first.

**Findings** (inline comments for the first three)

1. *Medium — permanently-infeasible demand queues instead of rejecting.* Demand exceeding the configured **total** (not just current free) is classified `Capacity` and waits out the full `queue_timeout_secs`, though no release can ever satisfy it. Visible cost: `gpu_request_placed_only_when_capacity_declared` burns ~30s of wall-clock test time waiting for a timeout that is the only possible outcome.
2. *Medium — warm-resize failure leaks the sandbox.* In `obtain_sandbox`, a warm sandbox is popped and then dropped if `update_resources` errors — the VM keeps running with no owner.
3. *Low/Medium — `release()` frees the reservation before `backend.stop()`.* On stop failure the accounting is freed and waiters woken while the physical sandbox may still be running, and the early return means no destroy is attempted — transient physical overcommit plus an orphaned sandbox.
4. *Low — malformed `hyprstream.io/gpu-request` is fail-open on the reservation axis:* `parse().ok().unwrap_or(0)` turns `"one"` into a zero-GPU demand, so a request that meant to reserve a GPU is placed without one. Rejecting unparseable values would match the fail-closed posture everywhere else in this module.

**Sign-off acknowledgments** (the PR's flagged judgment calls — all reasonable as scoped)

- Caller-supplied `hyprstream.io/group` means per-group quota is a convenience partition, not an isolation boundary — any RPC caller can omit or choose its group. Fine as a placeholder, but the follow-up for a real tenant claim should be filed so this doesn't quietly become load-bearing.
- Wake-all re-race (no per-key fairness): a fresh arrival can beat woken waiters repeatedly, so a queued waiter can starve until its timeout. Bounded by the timeout, so acceptable at this scale; the flat quota model and single-backend pool are fine as declared scope-downs.

No file overlap with open PRs #838/#839. Note for the operator: PR #770 stacks textually on this branch (this PR's commit is an ancestor of #770's head) — this PR must merge first; see the #770 review for the squash-merge caveat.

ewindisch (commented):
Approve with changes — remaining: all four prior findings. Follow-up check for post-review fixes.

This head (a49d179fc) is the same commit the earlier review (#pullrequestreview-4631374114) was submitted against — no new commits have been pushed, so nothing has been addressed yet. Verified each finding is still present in the current code:

1. NOT FIXED — demand exceeding configured totals (e.g. gpu > gpu_total) is still classified as queueable Capacity in try_admit_locked (admission.rs); a permanently infeasible request waits out the full queue timeout instead of hard-rejecting.
2. NOT FIXED — obtain_sandbox (pool.rs:183) still does `self.backend.update_resources(&sandbox, &requested).await?` after popping the warm sandbox; on resize failure the popped sandbox is dropped with the VM still running (leaked, unowned).
3. NOT FIXED — release() (pool.rs:206-211) still frees the admission reservation before backend.stop(); a stop failure leaves accounting freed while the VM runs.
4. NOT FIXED — derive_demand still parses hyprstream.io/gpu-request with `.parse::<usize>().ok().unwrap_or(0)`, so a malformed value silently becomes 0 (fail-open on the GPU axis).

No new defects to report since the code is unchanged. Prior verdict stands: approve with changes; findings 2 and 3 (the resource-leak error paths) are the ones worth landing before merge.

ewindisch:
## 🔴 BLOCKING REVIEW (1:7 coordination sweep) — no rebase needed (MERGEABLE, no conflicts); several correctness issues

Verified positives: reserve-then-place is genuinely atomic under one lock; `Notify` permits mean no lost wakeup; substrate usage is real (`hyprstream_discovery::scheduling::select`, fail-closed `satisfied_by`, `parse_k8s_quantity` handles `u64::MAX`-as-unlimited without overflow); `handle_run` derives `ctx.subject()` from the verified envelope per "derive, don't extend". Findings:

**1. (med) Infeasible demand is queued instead of rejected.** Demand exceeding configured *total* (`Demand::gpu > 0` with `gpu_total: 0`, or memory > total) is classified `AdmitReject::Capacity` and queued though no release can ever satisfy it (`admission.rs` `try_admit_locked`/`reserve`). Every GPU acquire on a GPU-less node blocks the full `queue_timeout_secs` (30s) then returns `QueueTimeout` instead of an immediate deterministic reject. Add a feasibility check against totals → immediate reject.

**2. (med) Resize-path sandbox leak.** In `obtain_sandbox` (`pool.rs` ~L1082-1110), if `backend.update_resources()` fails the popped warm sandbox is dropped via `?` without being destroyed or returned — each resize failure orphans a running VM/container and permanently shrinks the warm pool.

**3. (med) Config save breaks with a worker section.** `AdmissionConfig::default` sets `max_per_subject`/`max_per_group = usize::MAX`, which flows into TOML-serialized `PoolConfig` (`config/mod.rs:2192,2255`). TOML integers are i64; `to_string_pretty` errors on `u64::MAX`, so any `Config::save()` with a `worker:` section now fails. Cap the default or serialize as a sentinel.

**4. (med) Queue starvation via barging.** Newly-arriving requests run `try_admit_locked` without checking for queued waiters, and releases wake *all* waiters to re-race — no FIFO ordering at all (stronger than the PR's disclosed "no per-key fairness"). Under sustained arrival ≥ release, queued waiters lose every re-race and exhaust `queue_timeout`. Add FIFO admission or a queued-waiter barrier.

**5. (low) `release()` frees the admission reservation before `backend.stop()` completes** — a woken waiter can cold-create while the released sandbox is still stopping, transiently running `max_sandboxes + 1` / over-committing the envelope.

**6. (low, disclosed) Group is a caller-supplied annotation, not a verified claim** — a caller omits it to bypass `max_per_group`, or asserts another tenant's group to consume its quota (cross-tenant DoS). Advisory until group is a verified claim.

Items 1–4 should be resolved before merge. Flagging the owning session.

ewindisch:
## Ratified: group-authority decision (B′) — unblocks this PR

Per the design ratified in #921 (open decision 5, decided 2026-07-09), the group semantics for this scheduler are:

1. **The group is derived from the verified Subject, never from a free-text annotation.** The authoritative membership fact is the existing bidirectional-consent placement group (`ai.hyprstream.placement.{group,groupItem}`, `PlacementIndex::is_member` — membership requires BOTH the node/tenant's own record listing the group AND the group owner's groupItem naming it).
2. **The `hyprstream.io/group` annotation is demoted to a *selector*** among groups the verified Subject provably belongs to ("schedule/account this under team X, of which I am a consented member"). A selector naming a group the Subject is not a member of is rejected. An absent selector means no group partition, never "quota bypass".
3. **The quota counter is demoted to node-local capacity partitioning** — a fairness/organization mechanism, not a trust or billing boundary. Authoritative quota is the ledger/capability path (#922/#925): verify presented grant → debit cell ledger → emit receipt. This PR keeps a seam for that contract; it does not implement it.
4. Hardening rider (filed separately): `is_member` trusts the ingested record graph; before group partitions grow into billing boundaries, ingest-time CAR-proof/commit-signature verification must be verified-by-construction.

Fix wave now in flight addresses findings 1–5 of the blocking review plus the F6 realign above.

ewindisch:
Fix wave verified at the fable gate (head 2876ca867): **APPROVE-WITH-NITS** — F1 dimension coverage + ==total boundary confirmed; FIFO handoff race-checked (same-lock TOCTOU-free, timeout-vs-release safe both ways, pass-the-turn terminates); F5 holds no lock across backend.stop(); F6 validator has no bypass path (sole ANN_GROUP reader flows through GroupSelectorValidator; production default = deny-unknown). Nits filed as #936 (pre-queue feasibility check; pre-existing reset_sandbox orphan). Inline threads resolved (each addressed by the fix commit). Merging.

PR #763, opened 2026-07-04
build(deps): bump tar 0.4.44→0.4.46 and lz4_flex 0.11.5→0.11.6

ewindisch:
Bumps two dependencies via a **minimal lockfile edit** on current `main`, superseding the stale dependabot PRs #127 and #128.

## Why not just merge #127/#128
Both were opened 2026-04-07 against a 3-month-old lock and show `mergeable=UNKNOWN` (stale vs current main). Rebasing them re-resolves the lock the same way a fresh `cargo update` does.

## Why a hand-edit instead of `cargo update -p tar -p lz4_flex`
On current main, `cargo update` (even `--precise`) re-unifies a set of **drifted transitive deps** — downgrading assorted `windows-sys`/`windows-targets` versions (inert on Linux; `cfg(windows)`-gated) and one `syn 2→1` unification. None of that is required by the tar/lz4 bumps. This PR edits only the four version+checksum lines for `tar` and `lz4_flex`; `cargo metadata --locked` passes, confirming the transitive graph is unchanged and the lock is consistent.

## Contents
- `tar` 0.4.44 → 0.4.46 (patch)
- `lz4_flex` 0.11.5 → 0.11.6 (patch)

Full CI (build + test) validates the actual compile.

Closes #127, closes #128.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9162,9 +9162,9 @@ dependencies = [
 
 [[package]]
 name = "lz4_flex"
-version = "0.11.5"
+version = "0.11.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "08ab2867e3eeeca90e844d1940eab391c9dc5228783db2ed999acbc0a9ed375a"
+checksum = "373f5eceeeab7925e0c1098212f2fbc4d416adec9d35051a6ab251e824c1854a"
 dependencies = [
  "twox-hash",
 ]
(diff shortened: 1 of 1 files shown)


ewindisch (commented):
**Verdict: LGTM — safe to automerge.**

Verified against the checklist for the dep-update batch:

- Diff is exactly the four version+checksum lines for `tar` 0.4.44→0.4.46 and `lz4_flex` 0.11.5→0.11.6 in `Cargo.lock`. No new packages, no dependency-list changes, no feature drift, and the `[patch.crates-io]` moq-net pin is untouched (no `Cargo.toml` change at all).
- Both bumps are patch-level; checksums match the crates.io releases. The hand-edit rationale (avoiding the drifted windows-sys/syn re-unification a `cargo update` would drag in) is sound, and `cargo metadata --locked` passing confirms lock consistency.
- CI is green across the board (Clippy full compile, cargo-deny, WASM, CodeQL).

One coordination note: this PR, #768, and #769 all edit `Cargo.lock` and will conflict pairwise. Recommended merge order: **this one first** (smallest, ready), then #768, then #769 last (largest churn, easiest to regenerate). The other two will need a rebase/lock re-resolve after this merges — for this PR's hand-edited style that's trivial; the tar/lz4 lines don't overlap #769's bump set at all, so conflicts will be textual-adjacency only.

Issue #699, opened 2026-07-03
feat(mac): object-label resolution — content-bound SecurityLabels for the 9P/VFS surface (MAC activation prereq)


ewindisch:
> ## UPDATE (2026-07-04) — carrier (b) is now buildable under epic #809
>
> Prior state: carriers **(a)** schema-annotation and **(c)** bind-time labels landed via #710; carrier **(b)** — a `security_label` in a *content-addressed manifest* — was found **blocked** because no first-party CAS manifest type exists (all three candidates rejected: `release_store` not CID-addressed, `cag.rs` ContextManifest not CID-addressed, `OciManifest` a third-party wire format we don't mint). See issuecomment-4883242445.
>
> **This unblocks under the 9P kernel epic (#809):** carrier (b)'s manifest **IS** the L1 CAS manifest that #812 creates. `security_label` is a field on it from day one, with the already-ratified **D2 semantics**: a MISSING label = **join to the perimeter/binder floor, never "unrestricted"** — so legacy/unlabeled content stays valid at the floor, no rewrite-migration needed. The content-addressing invariant holds structurally: the label is inside the manifest, so relabel = rehash.
>
> This also un-blocks scope item 3 (`mac::ifc_join` at seal/rollup — the fn exists, is tested, and has zero call sites today): the **write-then-seal ingest** step (#814) is the natural join point — a derived/rolled-up layer's label = join(inputs) computed at seal.
>
> **Dependency edges:** #699 ← #812 (manifest with `security_label`) + #814 (ifc_join at seal). No change to the acceptance criteria below; they now have a home to land in.
>
> Original body below.

---

**Epic #547 · activation prerequisite (greenfield).** Surfaced by the independent enforcement-activation gap survey (2026-07-03). This is the `TODO(#572-object-label)` promoted to a tracked ticket.

## The gap
The MAC per-op dominance check is `subject.ctx ⊒ object.manifest.label` — but **no object label resolution exists**:

- `parse_grant_request` hardwires `object_label: None` with `TODO(#572-object-label): manifest/TE resolver` (`services/oauth/token_exchange.rs`), so the S1 object-label gate denies unconditionally.
- There is **no manifest `security_label` field and no object-label resolver trait** anywhere. The manifests that exist (`image/manifest.rs` OCI, `archetypes/cag.rs` ContextManifest, `storage/release_store.rs`) carry no security label. `SecurityLabel` outside the mac/auth core appears only in doc comments.
[...]

Issue #568, opened 2026-06-29
feat(security) S2: Reference-monitor enforcement in the generated mount


ewindisch:
**Epic:** #547. **Deps:** #539 (generated mount) + S1 (labels) + S4 (AVC).

## Goal
Make the **generated 9p mount the reference monitor (PEP)** — complete mediation by construction (codegen
always emits the check; no hand-written bypass), per-op, fail-closed.

## Scope
Per 9p op (walk/open/read/write/create/remove) the generated mount enforces:
```
ALLOW ⟺ valid sender-bound access token T (unexpired)
      ∧ op ∈ T.authorized_ops ∧ object.manifest.label ⊑ T.label_ceiling
      ∧ subject.ctx ⊒ object.manifest.label          ← IFC dominance, content-truth, independent
```
- **Labels from the manifest, never from token/UCAN/caveat.**
- **Fail-closed**; default-deny; no path skips the check.
- Hot path = token verify + lattice/AVC lookup (S4), **not** chain-walk and **not** Casbin-per-op.

## Dependencies / relates
#547 · #539 (the generated mount surface this hooks) · S1 (labels/lattice) · S4 (AVC fast path).

## Acceptance
- Every op is mediated; unlabeled/untokened → denied.
- IFC dominance enforced from the content-bound label.
- No code path bypasses the check (verified by construction + test).
- Per-op cost is a local lookup, not chain validation.

## Resource-transition enforcement consumer

#1064/#1068 extend complete mediation to title/control transitions. The reference monitor authorizes the exact resource intent and policy epoch; the registrar separately verifies ledger entitlement and performs fenced finalization. Ordinary reads remain MAC-only, and provisional resources never enter the normal namespace.

ewindisch:
## Precise activation gap map (2026-07-03 survey)

S2 is the per-op reference monitor — the thing that makes MAC *live*. Grounding it against current `main`, the **PEP host is ready; the PEP itself and 3 of its 4 inputs are unbuilt.** This is a subsystem effort, not a one-file patch. Breakdown so it's actionable:

**Ready (merged, #539):**
- The choke-point exists: `hyprstream-vfs/src/mount.rs` — every `Mount` method (`walk`/`open`/`read`/`write`/`readdir`/`stat`/`clunk`) already threads `caller: &Subject`. The seam for a PEP is pre-cut (trait doc even says "for … policy checks"). Every impl currently ignores it as `_caller`.
[...]

Issue #765, opened 2026-07-04
feat(9p): H1b — /9p WebTransport plane (QUIC path-mux arm, Tattach.uname ticket, cert-pinned) for the self-signed mesh


ewindisch:
Design doc (approved): `docs/plans/2026-07-03-shell-page-wanix-host-mounts.md (www-cyberdione-ai)`, §"hyprstream 9P serving (H1)". Follows H1a (shared core).

## Scope (H1b)
Third arm in the QUIC WebTransport path mux (`request.url.path() == "/9p"`, alongside `/moq` vs RPC at `quinn_transport.rs:289`) → the SAME `Translator::serve(stream)`/`MountBackend` core as H1a. For the **self-signed mesh** (cert-pinned via `certHash`/reach), so mesh hosts are mountable where `wss://` can't pin.
- Ticket carried in `Tattach.uname` under the cert-pinned session (not URL query).
- Browser side is a dumb JS pump only (WT bidi stream ⇄ MessagePort byte shovel — no 9P parsing; `P9PortReadWriter` tolerates arbitrary chunking) — that's www's S4, not this issue.

## Verify
Same suite as H1a over a WT bidi stream with a pinned cert.

Depends on: **H1a** (shared 9P core + ticket/errno work). Doc DAG: H1b after H1a.

ewindisch:
**Cross-link (9P kernel-completion epic #809).** This is one of the interop wires the kernel serves (vsock / v9fs / WebSocket / WebTransport 9P planes). Tagging as related to #809:
- **#821** (wire-plane discovery under `/.well-known/{name}`) applies to this plane's advertisement — stop hardcoding the selector on both dial and serve.
- **#767**'s attach-credential model is the auth path for all of these planes (the #786 mount tickets are the K8s instantiation).

No scope change; noting the edges.

PR #767, opened 2026-07-04
9P translator: attach-time MAC seam (#568)

ewindisch:
## Summary

Implements the attach-time MAC seam for the 9P translator proposed on #568 (https://github.com/hyprstream/hyprstream/issues/568#issuecomment-4882511757). This is not the full S2 PEP wiring — it's the structural piece that makes the translator *capable* of carrying and enforcing a verified `SecurityContext`, with defaults that preserve today's behavior exactly.

## What changed

- **`hyprstream-9p::mac_seam`** (new module): `AttachAuthenticator` (verify `Tattach` credential → `SecurityContext`, once per attach) and `AccessDecider` (authorize one op against a fid's cached context) trait seams, plus `AuditedDecider` (wraps a decider + an `AuditSink`, mirroring S7's decide-then-record-then-possibly-downgrade shape). Default impls (`AnonymousAuthenticator`, `AllowAllDecider`, `NullAuditSink`) are pure no-ops.
- **`FidTable`** now caches a `SecurityContext` per fid, populated once at `Tattach` and inherited by every fid walked from it — never re-derived per op (rule 2 of the ratified interface policy).
- **`Translator`** gained `with_authenticator`/`with_decider` builders. `new`/`from_mount` still default to the no-op pair, so every existing call site and test is unaffected.
- Each op handler (`walk`/`lopen`/`read`/`write`/`readdir`) now calls into the decider before delegating to the backend, returning `Rlerror EPERM` on deny. `getattr`/`clunk` are unchanged (matches the #568 proposal's named surface).
- New tests: the seam's own unit tests, plus `injected_decider_blocks_denied_action` — proves a non-default decider actually blocks reads end-to-end over the wire-level `handle_message` path, not just in isolation.

## A genuine sub-decision found during implementation (crate-boundary gap)

The posted design said to wire `mac::avc::CachingAvc` and `mac::audit::AuditedAvc` directly into the translator's dispatch. That's not reachable: **`hyprstream` depends on `hyprstream-9p` (and on `hyprstream-workers`, which also depends on `hyprstream-9p`), never the other way.** `mac::avc`/`mac::audit`/`mac::exchange` live in the `hyprstream` binary crate, so `hyprstream-9p` cannot call them without a dependency cycle.

Resolution applied (consistent with the ratified interface policy's own logic — derive/extend via a seam, don't invent a shortcut): define the trait interface at the lower layer (`hyprstream-9p::mac_seam`, built only on `hyprstream_rpc::auth::mac` primitives, which *are* reachable everywhere), and expect the `hyprstream` crate to inject the concrete `CachingAvc`/`AuditedAvc`-backed implementations via `with_authenticator`/`with_decider` wherever it (or `hyprstream-workers`) constructs a `Translator`. This exactly mirrors the existing `Backend`/`ModelBackend` precedent already in this codebase (the capnp-RPC-backed `Backend` impl lives in the outer crate, not in `hyprstream-9p`).

## A second genuine sub-decision: why the real decider is NOT wired in by default

`hyprstream_rpc::auth::mac::manifest::LabeledObject::security_label() -> None` means **unlabeled**, and the S1 invariant (design §1 invariant 2, `mac/mod.rs` doc) is: an unlabeled object **must deny**, not fall back to a permissive floor. Every object in the 9P namespace is unlabeled today — genesis labeling for 9P nodes hasn't started, and #699 carrier-b (the content-addressed manifest label field) is still an open decision.

So making a real, deny-on-unlabeled `AccessDecider` the *default* would silently deny every existing 9P client — a severe regression, not a security fix. `AllowAllDecider` stays the default; real enforcement is strictly opt-in via `with_decider`, and should only be wired once (a) a standalone "verify a presented S6 access token" entry point exists outside the OAuth HTTP handler, and (b) object labels are actually resolvable for the mount being served. Both are follow-up work, not done here.

## Fail-closed properties preserved

- Missing/invalid attach credential → anonymous floor context (not a rejected attach) — no flag day for existing unauthenticated clients.
- A fid with no cached context (shouldn't happen — every fid is populated at attach or inherits on walk) resolves to the anonymous floor rather than panicking.
- `AuditedDecider` records before returning the wrapped decision, matching S7's decide→record shape; a future durable sink can widen `AuditSink::record` to report failure and force a deny, as noted in the module docs.

## Testing

- `cargo test -p hyprstream-9p --release`: 16 unit + 4 integration tests, all pass (including the new ones).
- `cargo clippy --workspace --all-targets --release -- -D warnings`: clean.

## Follow-ups (not in this PR)

- Wire a real `AttachAuthenticator` in the `hyprstream` crate backed by `mac::exchange`'s token verification + the service trust store, once a standalone verify-only entry point exists.
- Wire a real `AccessDecider` (`CachingAvc` + `AuditedDecider` around a real `AuditSink`) once #699 carrier-b or genesis labeling gives 9P objects actual labels.

Refs #568, epic #547.

### DIFF
--- a/crates/hyprstream-9p/src/mac_seam.rs
+++ b/crates/hyprstream-9p/src/mac_seam.rs
@@ -1,0 +1,237 @@
+//! Attach-time MAC seam for the 9P translator (#568, epic #547).
+//!
+//! `hyprstream-9p` sits *below* `hyprstream` in the dependency graph (the
+//! reverse: `hyprstream` depends on `hyprstream-9p`), so the concrete grant
+//! verification (`mac::exchange`), the compiled-policy PDP (`mac::te`), the
+//! per-op cache (`mac::avc::CachingAvc`), and the audit trail
+//! (`mac::audit::AuditedAvc`) — which all live in the `hyprstream` binary
+//! crate — can never be called from here without a dependency cycle.
+//!
+//! Per the ratified interface policy (CLAUDE.md, "Interface policy — MAC on
+//! contracts that don't carry it", rule 2): the fix is dependency inversion,
+//! not moving code. This module defines the trait seam; the `hyprstream`
+//! crate is expected to provide the concrete implementations and inject them
+//! wherever it constructs a [`crate::Translator`] (mirroring the existing
+//! `Backend`/`ModelBackend` precedent — the capnp-RPC-backed `Backend` impl
+//! already lives in the `hyprstream` crate, not here).
+//!
+//! ## Fail-closed defaults ship inert, not deny-by-default
+//!
+//! [`LabeledObject::security_label`] returning `None` means the object is
+//! **unlabeled**, and per the S1 invariant (`hyprstream_rpc::auth::mac`,
+//! design §1 invariant 2) an unlabeled object MUST deny — there is no
+//! "unlabeled ⇒ floor" default at that layer. Every object in the 9P
+//! namespace is unlabeled today (genesis labeling for 9P nodes is unstarted,
+//! and #699 carrier-b — the content-addressed manifest label field — is still
+//! an open design decision). Wiring a real deny-on-unlabeled
+//! [`AccessDecider`] into the live dispatch path by default would therefore
+//! deny every existing 9P client outright — a severe, silent regression, not
+//! a security fix.
+//!
+//! So the defaults here ([`AnonymousAuthenticator`], [`AllowAllDecider`])
+//! preserve **today's actual behavior** (no enforcement) exactly. They exist
+//! so the translator has somewhere to hang real enforcement once a caller
+//! opts in via [`Translator::with_authenticator`](crate::Translator::with_authenticator)
+//! / [`Translator::with_decider`](crate::Translator::with_decider) — which the
+//! `hyprstream` crate can do once (a) a standalone "verify a presented S6
+//! access token" entry point exists outside the OAuth HTTP handler, and (b)
+//! object labels are actually available for the mount being served (genesis
+//! labeling, or #699 carrier-b for content-addressed data). Both are called
+//! out as follow-ups in the PR body rather than implemented here.
+
+use async_trait::async_trait;
+use hyprstream_rpc::auth::mac::{
+    Assurance, CompartmentSet, Level, SecurityContext, SecurityLabel, VerifiedKeyMaterial,
+};
+
+/// The 9P op an [`AccessDecider`] is being asked to authorize.
+///
+/// Mirrors the translator's dispatch surface (`walk`/`open`/`read`/`write`/
+/// `create`/`readdir`) so a real decider can apply per-action policy (e.g.
+/// deny `Write` more aggressively than `Read`).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum Action {
+    Walk,
+    Open,
+    Read,
+    Write,
+    Create,
+    Readdir,
+}
+
+/// Verifies an attach credential (9P `uname`/`aname`) and derives the
+/// caller's [`SecurityContext`] for the resulting fid.
+///
+/// Called exactly once per `Tattach`, mirroring the ratified interface
+/// policy's rule 2 (extend at attach time, never per-op): the derived context
+/// is cached on the fid (see `FidTable`) and reused for every subsequent op
+/// on that fid and its descendants, never re-verified per op.
+#[async_trait]
+pub trait AttachAuthenticator: Send + Sync {
+    /// Verify `uname`/`aname` and return the derived context. Must never
+    /// panic or block indefinitely on a malformed/hostile credential — an
+    /// invalid credential resolves to the anonymous floor context, exactly
+    /// like a missing one (fail-closed by floor, not by rejecting the
+    /// attach outright, so unauthenticated legacy clients keep working).
+    async fn authenticate(&self, uname: &str, aname: &str) -> SecurityContext;
+}
+
+/// The default, no-op authenticator: every attach resolves to the anonymous
+/// floor context regardless of `uname`/`aname`. This is exactly today's
+/// behavior (the translator performs no identity verification at all), kept
+/// as the default so existing callers see no behavior change.
+#[derive(Debug, Default, Clone, Copy)]
+pub struct AnonymousAuthenticator;
+
+#[async_trait]
+impl AttachAuthenticator for AnonymousAuthenticator {
+    async fn authenticate(&self, _uname: &str, _aname: &str) -> SecurityContext {
+        anonymous_floor()
+    }
+}
+
+/// The anonymous-floor [`SecurityContext`]: `Level::Public`, no compartments,
+/// `VerifiedKeyMaterial::Unverified` (assurance floors at `Unverified`).
+pub fn anonymous_floor() -> SecurityContext {
+    SecurityContext::from_clearance(
+        SecurityLabel::new(Level::Public, Assurance::Unverified, CompartmentSet::EMPTY),
+        VerifiedKeyMaterial::Unverified,
+    )
+}
+
+/// Authorizes one op against a subject's cached [`SecurityContext`].
+///
+/// `object_label` is `None` when no label source is wired for the mount
+/// being served (true for every 9P mount today — see the module-level
+/// documentation on why `None` must NOT be treated as "unrestricted" by a
+/// real implementation). The default [`AllowAllDecider`] ignores both
+/// arguments and always allows, preserving current behavior; a real decider
+/// (injected by the `hyprstream` crate, backed by `mac::avc::CachingAvc`)
+/// must deny when `object_label` is `None`, per the S1 invariant.
+pub trait AccessDecider: Send + Sync {
+    /// Returns whether `action` is authorized for `ctx` against `object_label`.
+    fn check(
+        &self,
+        ctx: &SecurityContext,
+        object_label: Option<&SecurityLabel>,
+        action: Action,
+    ) -> bool;
+}
+
+/// The default, no-op decider: every op is allowed regardless of context or
+/// label. This is exactly today's behavior (the translator performs no
+/// per-op authorization at all), kept as the default so existing callers see
+/// no behavior change until a real decider is explicitly wired in.
+#[derive(Debug, Default, Clone, Copy)]
+pub struct AllowAllDecider;
+
+impl AccessDecider for AllowAllDecider {
+    fn check(&self, _ctx: &SecurityContext, _object_label: Option<&SecurityLabel>, _action: Action) -> bool {
+        true
+    }
+}
+
+/// Records the outcome of an [`AccessDecider::check`] call for audit.
+///
+/// A real sink (injected by the `hyprstream` crate, backed by
+/// `mac::audit::WalAuditStore`) durably logs the decision; per S7, a
+/// decision that cannot be durably audited must downgrade to deny. The
+/// default [`NullAuditSink`] does not implement that downgrade (there is
+/// nothing to audit against by default) — it is only safe as a default
+/// because [`AllowAllDecider`] is also the default, so no security-relevant
+/// decision is ever made without an explicit opt-in to both a real decider
+/// and a real sink together (see [`AuditedDecider`]).
+pub trait AuditSink: Send + Sync {
+    /// `allowed` is the decision `AuditedDecider` is about to return.
+    fn record(&self, ctx: &SecurityContext, object_label: Option<&SecurityLabel>, action: Action, allowed: bool);
+}
+
+/// The default, no-op audit sink: records nothing.
+#[derive(Debug, Default, Clone, Copy)]
+pub struct NullAuditSink;
+
+impl AuditSink for NullAuditSink {
+    fn record(&self, _ctx: &SecurityContext, _object_label: Option<&SecurityLabel>, _action: Action, _allowed: bool) {}
+}
+
+/// Wraps an [`AccessDecider`] so every decision is also recorded via an
+/// [`AuditSink`], and — mirroring S7's `AuditedAvc` fail-closed contract — a
+/// sink that reports it could not durably record the decision causes the
+/// wrapped decision to be downgraded to deny.
+///
+/// [`AuditSink::record`] here is infallible by trait signature (matching the
+/// simple sinks available at this layer); a real, fallible durable sink
+/// should be adapted to report failure through `allowed` before calling
+/// `record`, or a future revision can widen `AuditSink::record` to return a
+/// `Result` once a concrete durable sink is wired in from the `hyprstream`
+/// crate. Documented here rather than solved speculatively.
+pub struct AuditedDecider<D, S> {
+    inner: D,
+    sink: S,
+}
+
+impl<D: AccessDecider, S: AuditSink> AuditedDecider<D, S> {
+    pub fn new(inner: D, sink: S) -> Self {
+        Self { inner, sink }
+    }
+}
+
+impl<D: AccessDecider, S: AuditSink> AccessDecider for AuditedDecider<D, S> {
+    fn check(&self, ctx: &SecurityContext, object_label: Option<&SecurityLabel>, action: Action) -> bool {
+        let allowed = self.inner.check(ctx, object_label, action);
+        self.sink.record(ctx, object_label, action, allowed);
+        allowed
+    }
+}
+
+#[cfg(test)]
+#[allow(clippy::unwrap_used)]
+mod tests {
+    use super::*;
+    use std::sync::atomic::{AtomicUsize, Ordering};
+
+    #[tokio::test]
+    async fn anonymous_authenticator_always_floors() {
+        let auth = AnonymousAuthenticator;
+        let ctx = auth.authenticate("anyone", "anything").await;
+        assert_eq!(ctx.level(), Level::Public);
+        assert_eq!(ctx.assurance(), hyprstream_rpc::auth::mac::Assurance::Unverified);
+    }
+
+    #[test]
+    fn allow_all_decider_ignores_inputs() {
+        let decider = AllowAllDecider;
+        assert!(decider.check(&anonymous_floor(), None, Action::Write));
+        let secret = SecurityLabel::new(
+            Level::Secret,
+            hyprstream_rpc::auth::mac::Assurance::PqHybrid,
+            Default::default(),
+        );
+        assert!(decider.check(&anonymous_floor(), Some(&secret), Action::Write));
+    }
+
+    struct DenyAll;
+    impl AccessDecider for DenyAll {
+        fn check(&self, _ctx: &SecurityContext, _object_label: Option<&SecurityLabel>, _action: Action) -> bool {
+            false
+        }
+    }
+
+    #[derive(Default)]
+    struct CountingSink {
+        calls: AtomicUsize,
+    }
+    impl AuditSink for CountingSink {
+        fn record(&self, _ctx: &SecurityContext, _object_label: Option<&SecurityLabel>, _action: Action, _allowed: bool) {
+            self.calls.fetch_add(1, Ordering::SeqCst);
+        }
+    }
+
+    #[test]
+    fn audited_decider_records_and_forwards_decision() {
+        let sink = CountingSink::default();
+        let decider = AuditedDecider::new(DenyAll, sink);
+        assert!(!decider.check(&anonymous_floor(), None, Action::Read));
+        assert_eq!(decider.sink.calls.load(Ordering::SeqCst), 1);
+    }
+}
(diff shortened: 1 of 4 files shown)


ewindisch:
Rebased onto main (7d43b5721 → a89cb9fe2) to resolve conflicts with the H1b attach-time ticket path that landed on main after this branched.

Conflict resolution in `handle_attach` (translator.rs) was semantic, worth a second look from the branch owner: main's `self.backend.attach(uname).await?` (H1b session establishment / ticket validation, fails closed before any fid exists) is kept FIRST, then this PR's `self.authenticator.authenticate(uname, aname).await` derives the MAC context for the fid table. The two mechanisms are complementary (session Subject binding vs. attach-time SubjectCtx derivation), so both run in that order. `cargo check -p hyprstream-9p` passes post-rebase.

ewindisch (commented):
**Verdict: Approve with changes** (nothing blocking; two coverage gaps worth fixing before a real decider is ever injected, both fine as fast follow-ups or a small amendment here).

This is a faithful implementation of the ratified interface policy, move 2 (extend at attach time, never per-op):

- **Credential vs. clearance**: `AttachAuthenticator::authenticate(uname, aname)` treats the wire strings as a credential to *verify*, never as clearance. The only impl in-tree, `AnonymousAuthenticator`, ignores them entirely and floors to `Public/Unverified/EMPTY` — no path lets caller-supplied strings become authoritative PDP input. The trait doc correctly requires invalid credentials to resolve to the anonymous floor rather than elevating.
- **Attach-once, fid-cached, walk-inherited**: ctx derived exactly once in `handle_attach`, stored in `FidState`, inherited in `handle_walk` via `security_context(fid)` before `fids.insert(newfid, ..., ctx)`. No per-op re-derivation anywhere. (One inheritance gap on the nwname=0 clone path, inline.)
- **Rebase ordering sanity-checked**: `self.backend.attach(uname).await?` (main's H1b session-ticket validation) runs *before* `self.authenticator.authenticate(...)`. A denied ticket returns via `?` before any ctx is derived or fid inserted — correct short-circuit, no fid leak, no double-derivation. The two layers stay independent: H1b binds the session Subject, the MAC ctx is derived separately, and a valid ticket does not elevate the MAC ctx (AnonymousAuthenticator still floors). One thing to keep in mind for the real authenticator: `uname` does double duty (H1b ticket and authenticator input), and the trait doc already says it must be cryptographically verified, not trusted — good.
- **Defaults**: `AnonymousAuthenticator` floors; `AllowAllDecider` is allow-all but is explicitly and honestly documented as preserving today's zero-enforcement behavior, with the S1 unlabeled-must-deny invariant spelled out for whoever wires the real decider. Given every 9P object is unlabeled today and enforcement is dormant repo-wide, inert-by-default is the right call — a deny-by-default decider here would be a silent outage, not a security fix. The module docs make it impossible to wire a real decider while believing `None` label means unrestricted.
- **No wire schema changes**: only lib.rs / mac_seam.rs / translator.rs; no .capnp touched, no plaintext labels on any contract.
- **Revocation seam**: there is no generation counter in this crate, but the decider is consulted on *every* op against the cached ctx, so revocation correctly lives in the injected `CachingAvc` (its generation counter invalidates without re-attaching). Worth one explicit sentence in the mac_seam module docs so the future wiring PR doesn't reinvent it.
- **Public API**: `FidTable::insert` gained a required `ctx` param — checked, no external callers in the workspace.

Findings (details inline):
1. Minor — `handle_getattr` is not gated by `deny()`; once a real decider is injected, stat metadata (size, mode, mtime, qid) leaks on denied fids. Add `Action::Getattr` and gate it, or document the exemption deliberately.
2. Minor — Twalk with nwname=0 (fid clone) never inserts `newfid`, so a cloned fid loses its inherited ctx and floors to anonymous. Fail-closed direction, but under real enforcement it wrongly denies legitimately-attached clients.
3. Nit — `Action::Create` has no dispatch call site (translator has no create handler today); forward-looking is fine, just noting the enum and the dispatch surface will need to stay in sync.

Cross-reference: **PR #776 does not conflict** — it touches `git2db/src/manager.rs`, `git2db/src/storage/vfs.rs`, and `hyprstream/src/services/registry.rs` only (its Tstat fix lives in the registry service, not this translator). Merge order between #767 and #776 does not matter.

The injected-decider test (`injected_decider_blocks_denied_action`) proving the seam is live rather than decorative is exactly the right test to have.

crates/hyprstream-9p/src/translator.rs
```diff
@@ -400,9 +481,31 @@ impl Translator {
...
+    /// context. Returns `Some(Response::Error)` (EPERM) if denied, `None` if
+    /// allowed. No object label is available at this layer (see the
+    /// `mac_seam` module docs) — `object_label` is always `None` here; a real
+    /// decider must treat that as "deny" per the S1 invariant, not as
+    /// "unrestricted".
+    ///
+    /// A fid with no cached context (should not happen — every fid is
+    /// populated at attach or inherits from its parent on walk) is treated
+    /// as the anonymous floor rather than panicking.
+    fn deny(&self, fid: u32, action: Action) -> Option<Response> {
```
ewindisch, line 578: Coverage gap: `handle_getattr` (line ~478) is the one fid-consuming op on the dispatch surface that never calls `deny()`. Today that's invisible (AllowAllDecider), but the moment a real decider is injected, a denied subject can still Tgetattr any fid it can name and learn size/mode/mtime/qid — metadata disclosure through the seam. Suggest adding `Action::Getattr` and gating `handle_getattr` the same way as read/readdir, or if stat is deliberately exempt (some 9P servers treat it as free), saying so explicitly in the mac_seam module docs so the exemption is a decision rather than an omission.

Separately, one sentence worth adding to the mac_seam docs while here: revocation of a cached ctx is expected to happen inside the injected decider (the `CachingAvc` generation counter), which works precisely because `deny()` consults the decider on every op — the cached ctx itself is never re-verified. That's the intended design per the interface policy, but the future wiring PR will look for where revocation lives, and right now nothing names it.

crates/hyprstream-9p/src/translator.rs
```diff
@@ -360,27 +426,42 @@ impl Translator {
...
+        }
+        // The context is inherited from the fid being walked (#568 — a
+        // context is derived once at attach, never re-verified per op).
+        let ctx = self.fids.security_context(fid).unwrap_or_else(anonymous_floor);
         // Mirror of RemoteModelMount::walk: call backend.walk with the source
         // fid, the new fid, and the path components.
         let result = self.backend.walk(fid, newfid, &wnames).await?;
         if let Some(q) = result.qids.last() {
-            self.fids.insert(newfid, q.qtype);
+            self.fids.insert(newfid, q.qtype, ctx);
```
ewindisch, line 511: Inheritance gap on the clone path: per 9P2000.L, Twalk with nwname=0 clones `fid` to `newfid` and Rwalk carries zero qids. Here `result.qids.last()` is `None` in that case, so `newfid` is never inserted — and every subsequent op on the clone hits the `unwrap_or_else(anonymous_floor)` fallback in `deny()`. The failure direction is safe (floor, never elevate), but under a real decider an authenticated client's cloned fid silently downgrades to anonymous and gets denied, which will be a miserable bug to diagnose from the client side. The qtype was already lost on this path pre-PR, but the ctx makes it security-relevant. Suggest inserting `newfid` unconditionally on a successful walk (using the source fid's qtype when qids is empty), since `ctx` is already in hand.

crates/hyprstream-9p/src/mac_seam.rs
```diff
@@ -0,0 +1,237 @@
...
+/// Mirrors the translator's dispatch surface (`walk`/`open`/`read`/`write`/
+/// `create`/`readdir`) so a real decider can apply per-action policy (e.g.
+/// deny `Write` more aggressively than `Read`).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum Action {
+    Walk,
+    Open,
+    Read,
+    Write,
+    Create,
```
ewindisch: Nit: `Action::Create` has no call site — the translator's dispatch surface has no create/mkdir handler today, so this variant is unreachable until one lands. Fine as forward-looking vocabulary, but note the asymmetry: `Getattr` (which *is* dispatched today) is absent from the enum while `Create` (which isn't) is present. Keeping the enum aligned with the actual dispatch surface — or a comment stating which ops are intentionally out of scope — will keep the seam honest as handlers are added.

ewindisch:
Review fixes pushed as `2e610cd6e` (fast-forward on your rebased HEAD).

**Merge ordering — agreed.** Keeping `backend.attach` *before* `authenticator.authenticate` is correct: `backend.attach` is the transport/ticket gate (on the H1b path it validates `uname` and fails a bad ticket fast, before any fid/mount exists), and `authenticate` derives the MAC `SecurityContext` (floors to anonymous by default). They're independent layers and nothing requires authenticate to gate the transport attach, so transport-first is the right order.

**(1) getattr gated.** `handle_getattr` was the only fid op not passing through `deny()`; since it leaks object metadata (qid/mode/size/mtime) it's now gated via a new `Action::Getattr`, consistent with read/readdir. Added `injected_decider_blocks_getattr` (denies `Getattr` → EPERM).

**(2) fid-clone context.** `Twalk` with `nwname=0` is a clone: the backend returns no qids, so `newfid` was never inserted and floored to anonymous — wrongly denying an authed client under real enforcement. It now inserts `newfid` inheriting the source fid's qtype + context (`FidTable::qtype`). Added `walk_clone_inherits_source_context` (elevated attach → clone → asserts the clone's cached ctx is not the Public floor).

**(dead surface)** Removed `Action::Create` — there's no `Tlcreate` handler, so it was an unreachable policy variant; the enum now mirrors the actual dispatch surface (walk/open/read/write/getattr/readdir).

Gates: `cargo clippy -p hyprstream-9p --all-targets -- -D warnings` clean; `cargo test -p hyprstream-9p` 24 unit + 4 integration pass. Ready for automerge from my side.

ewindisch:
Follow-up pushed as b16edb621.\n\nWhat changed:\n-  now allows a repeated attach only when the newly verified ticket resolves to the same already-bound .\n- A second valid ticket resolving to a different subject fails closed with  / , so a connection cannot be silently re-scoped after first attach.\n- Added  to cover the regression.\n\nLocal verification from pane 1:2.0:\n- 
running 25 tests
test mac_seam::tests::allow_all_decider_ignores_inputs ... ok
test mac_seam::tests::audited_decider_records_and_forwards_decision ... ok
test mount_backend::tests::lopen_flags_mapping ... ok
test mount_backend::tests::encode_dir_entries_emits_standard_rreaddir_records ... ok
test mac_seam::tests::anonymous_authenticator_always_floors ... ok
test memory::tests::write_extends_file ... ok
test memory::tests::walk_read_roundtrip ... ok
test translator::tests::fid_table_tracks_open_and_clunk ... ok
test translator::tests::attach_empty_ticket_denied ... ok
test translator::tests::attach_then_walk_open_read_clunk ... ok
test translator::tests::attach_ticket_valid_binds_subject_and_serves ... ok
test translator::tests::mount_errors_map_to_errnos ... ok
test translator::tests::injected_decider_blocks_denied_action ... ok
test translator::tests::attach_ticket_denied_returns_eacces ... ok
test translator::tests::qid_helpers ... ok
test translator::tests::mount_backend_preserves_typed_errno_end_to_end ... ok
test translator::tests::op_before_attach_fails_closed ... ok
test translator::tests::hybrid_vsock_bad_preamble_does_not_kill_listener ... ok
test translator::tests::readdir_emits_standard_rreaddir_wire ... ok
test translator::tests::second_attach_conflicting_ticket_denied ... ok
test translator::tests::version_round_trips ... ok
test translator::tests::hybrid_vsock_handshake_then_9p_session ... ok
test translator::tests::walk_clone_inherits_source_context ... ok
test translator::tests::injected_decider_blocks_getattr ... ok
test translator::tests::raw_vsock_no_preamble_9p_session ... ok

test result: ok. 25 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s passed: 25/25.\n- 
running 25 tests
test mac_seam::tests::allow_all_decider_ignores_inputs ... ok
test mac_seam::tests::audited_decider_records_and_forwards_decision ... ok
test mount_backend::tests::lopen_flags_mapping ... ok
test mount_backend::tests::encode_dir_entries_emits_standard_rreaddir_records ... ok
test mac_seam::tests::anonymous_authenticator_always_floors ... ok
test memory::tests::walk_read_roundtrip ... ok
test memory::tests::write_extends_file ... ok
test translator::tests::fid_table_tracks_open_and_clunk ... ok
test translator::tests::attach_ticket_denied_returns_eacces ... ok
test translator::tests::attach_empty_ticket_denied ... ok
test translator::tests::attach_then_walk_open_read_clunk ... ok
test translator::tests::attach_ticket_valid_binds_subject_and_serves ... ok
test translator::tests::injected_decider_blocks_denied_action ... ok
test translator::tests::mount_errors_map_to_errnos ... ok
test translator::tests::qid_helpers ... ok
test translator::tests::injected_decider_blocks_getattr ... ok
test translator::tests::op_before_attach_fails_closed ... ok
test translator::tests::mount_backend_preserves_typed_errno_end_to_end ... ok
test translator::tests::hybrid_vsock_bad_preamble_does_not_kill_listener ... ok
test translator::tests::readdir_emits_standard_rreaddir_wire ... ok
test translator::tests::hybrid_vsock_handshake_then_9p_session ... ok
test translator::tests::version_round_trips ... ok
test translator::tests::second_attach_conflicting_ticket_denied ... ok
test translator::tests::walk_clone_inherits_source_context ... ok
test translator::tests::raw_vsock_no_preamble_9p_session ... ok

test result: ok. 25 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s


running 1 test
test socket_client_mount_round_trips_against_translator ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s


running 2 tests
test tcp_full_session ... ok
test tcp_write_then_read ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.05s


running 1 test
test uds_full_session ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.05s


running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s passed: full 9P crate tests + doctests.\n- Pre-commit  completed cleanly before commit.

ewindisch:
Correction to my previous comment: the shell expanded backticked snippets while posting it. The pushed commit and verification are correct; this is the clean summary.

Follow-up pushed as b16edb621.

What changed:
- MountBackend::attach now allows a repeated attach only when the newly verified ticket resolves to the same already-bound Subject.
- A second valid ticket resolving to a different subject fails closed with PermissionDenied / EACCES, so a connection cannot be silently re-scoped after first attach.
- Added second_attach_conflicting_ticket_denied to cover the regression.

Local verification from pane 1:2.0:
- CARGO_TARGET_DIR=/tmp/hyprstream-target-767 cargo test -p hyprstream-9p --lib passed: 25/25.
- CARGO_TARGET_DIR=/tmp/hyprstream-target-767 cargo test -p hyprstream-9p passed: full 9P crate tests + doctests.
- Pre-commit cargo clippy --workspace --all-targets --release -- -D warnings completed cleanly before commit.

ewindisch:
Pane 1:5.0 coordination update for #778 / #767:

Active DAG I am working:
- #778 security spine consumes #568 / PR #767 as the W5 attach-time 9P PEP seam.
- #767 sits after the H1b attach-ticket path (#765/#773): transport/session attach binds the mount Subject first, then the MAC authenticator derives fid-scoped SecurityContext.
- Live MAC activation remains gated by object labels (#699); this PR is the fail-closed structural seam, not a default enforcement flip.

PR head is now b16edb621 on mac/568-9p-pep-wiring. The follow-up rejects a repeated Tattach that presents a different valid ticket / Subject, while allowing idempotent repeat attach for the same Subject.

Validation I ran locally on the #767 worktree:
- CARGO_TARGET_DIR=/tmp/hyprstream-target-767 cargo test -p hyprstream-9p --lib: 25/25 pass.
- CARGO_TARGET_DIR=/tmp/hyprstream-target-767 cargo test -p hyprstream-9p --tests: unit + socket/tcp/uds integration pass.
- CARGO_TARGET_DIR=/tmp/hyprstream-target-767 cargo clippy -p hyprstream-9p --all-targets -- -D warnings: clean.

I am not starting new unrelated work. Waiting on GitHub checks/review for #767.

ewindisch:
## 🔴 BLOCKING REVIEW (1:7 coordination sweep) — one high-severity issue; seam design is otherwise sound

Verified airtight per the review asks: ticket verified exactly once at Tattach, ctx cached in `FidTable`; fid-clone (`nwname=0` walk) inherits source ctx (`walk_clone_inherits_source_context`); `getattr` is gated (PR body is stale — the diff does gate it with a test); per-op params carry only the fid (opaque handle), no plaintext MAC fields; deny-on-unlabeled is deliberately opt-in and documented. The b16edb621 single-subject fix is real. But:

**1. (HIGH) SecurityContext cache is fid-scoped but NOT connection-scoped.** The `Translator` (one `FidTable`, one cached-ctx store, one backend subject cell) is a single `Arc<Self>` shared across **all** accepted connections (`serve_listener` ~L215-233). This violates the ratified interface policy's requirement to cache the derived SubjectCtx **connection-scoped** inside the TCB. Once a real `AttachAuthenticator` is wired: client B connects to the same listener, **skips Tattach entirely**, and issues `Twalk`/`Tread` on fid numbers established by connection A (client-chosen small ints, trivially guessable) — `deny()` finds A's cached elevated ctx and authorizes B's ops under A's clearance. Conversely B's `Tattach fid 0` clobbers A's fid-0 ctx. Attach-time verification is fully bypassable across connections. Pre-existing structural sharing, but **this PR is what makes it security-load-bearing.** Fix before any real authenticator lands: per-connection `FidTable` (or connection-id-keyed fids).

**2. (med) The single-subject invariant lives only in `MountBackend::attach`, not the Translator.** `Backend::attach` has a default no-op impl (`backend.rs:66`), so any non-`MountBackend` backend (the capnp `Backend`, `MemoryBackend`) lets a second Tattach with a different credential proceed and re-insert fids with a second ctx → decider-ctx / backend-subject divergence. The b16edb621 protection silently doesn't cover the new `mac_seam` path. Enforce the invariant at the Translator layer.

**3. (low) `AuditedDecider` (`mac_seam.rs` ~L201-218) can't actually fail closed.** `AuditSink::record` is infallible, so the S7 "cannot durably audit ⇒ deny" contract isn't implementable through this trait — it only does decide→record, never downgrade. When a real WAL sink is wired without widening the trait, audit-write failures silently permit. The doc's "mirrors S7's fail-closed contract" phrasing overstates what the signature allows.

**4. (low) `OnceCell::set` race** (`mount_backend.rs` b16edb621 hunk): two racing attaches with the *identical* ticket can spuriously fail one with EIO instead of the correct errno. Fail-closed, availability nit only.

Item 1 blocks. Flagging @1:2's owning session.

ewindisch:
## ✅ Blocker verified resolved (1:7 re-review of d0363661b)

Traced both holes against the commit — both **FIXED**:

**Hole 1 (HIGH — per-connection state):** the accept loop now does `Arc::new(this.connection_scoped())` per accepted connection (`translator.rs:335`), and `connection_scoped()` (252-260) builds a fresh `FidTable` + fresh `attach_ctx` OnceCell + a fresh backend from the factory. No `static`/`OnceLock`/`thread_local` global fid map remains anywhere, so the process-global cache is gone. Cross-connection reuse is denied: conn B's `deny(fid)` reads its own empty table → `anonymous_floor` → EPERM under any real decider. Test-backed (`connection_scoped_translators_do_not_share_fid_context`).

**Hole 2 (MED — single-subject on no-op backends):** the invariant moved out of `MountBackend::attach` up into the Translator — `handle_attach` calls `bind_attach_context(&ctx)` for *every* backend (474-485), so a conflicting second Tattach → EACCES even on `MemoryBackend`'s default no-op attach. Test-backed (`translator_rejects_conflicting_attach_security_context`).

**One residual worth a follow-up (not blocking, out of scope of the ctx-cache blocker):** on the generic `Translator::new(backend)` path the `backend_factory` clones the same `Arc`, so backend-level fid state (`MemoryBackend.fids`, `ModelBackend`) is still shared across connections. This is *not* a ctx-escalation — every unknown-fid op is floored to `anonymous_floor` and the translator `deny()` gate closes it under any real decider — and the mount paths (`from_mount`/`with_authorizer`) already build a new `MountBackend` per connection so they're fully isolated. But once enforcement is live, a shared non-mount backend under `AllowAllDecider` lets two connections clobber bare-fid ints at the backend layer. Suggest tracking a per-connection backend on the generic path too, for defense-in-depth. Nice work — this closes the HIGH item.

PR #768, opened 2026-07-04
deps: crypto/authz bump — casbin 2.20, aws-lc 1.17/0.42

ewindisch:
**Draft** — isolated security-sensitive dependency bumps, pending e2e validation.

| crate | from | to | role |
|---|---|---|---|
| `casbin` | 2.17.0 | 2.20.0 | MAC/policy authorization engine (the just-merged MAC enforcement depends on it) |
| `aws-lc-rs` | 1.14.0 | 1.17.1 | crypto backend |
| `aws-lc-sys` | 0.31.0 | 0.42.0 | crypto FFI (large jump — new build scripts/FFI) |

Isolated from the bulk registry update (#TBD) so the crypto/authz behavior can be reviewed and test-validated on its own. Lockfile-only, 3 registry bumps, no transitive churn, git deps untouched.

**Validation gate before ready:** full compile (CI Clippy `--all-targets`) + local `cargo nextest` of the authz/MAC/crypto crates (the PR-level CI does not run the test suite — `build`+`nextest` are push/merge_group-only).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1070,9 +1070,9 @@ dependencies = [
 
 [[package]]
 name = "aws-lc-rs"
-version = "1.14.0"
+version = "1.17.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "94b8ff6c09cd57b16da53641caa860168b88c172a5ee163b0288d3d6eea12786"
+checksum = "4342d8937fc7e5dd9b1c60292261c0670c882a2cd1719cfc11b1af41731e32ad"
 dependencies = [
  "aws-lc-sys",
  "zeroize",
@@ -1080,15 +1080,15 @@ dependencies = [
 
 [[package]]
 name = "aws-lc-sys"
-version = "0.31.0"
+version = "0.42.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0e44d16778acaf6a9ec9899b92cebd65580b83f685446bf2e1f5d3d732f99dcd"
+checksum = "6d9ceb1da931507a12f4fccea479dccd00da1943e1b4ae72d8e502d707361444"
 dependencies = [
- "bindgen",
  "cc",
  "cmake",
  "dunce",
  "fs_extra",
+ "pkg-config",
 ]
 
 [[package]]
@@ -1363,9 +1363,7 @@ dependencies = [
  "bitflags 2.11.0",
  "cexpr",
  "clang-sys",
- "itertools 0.13.0",
- "log",
- "prettyplease",
+ "itertools 0.10.5",
  "proc-macro2",
  "quote",
  "regex",
@@ -1928,9 +1926,9 @@ dependencies = [
 
 [[package]]
 name = "casbin"
-version = "2.17.0"
+version = "2.20.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7c2588e0a621ab3f658b445bbe4e4fa05ed2182eb81a6b906c3dcf4d42c9cde5"
+checksum = "c53f7476c2d0d9cd7ccc88c16ffc5c7889a0497b3462b10b12b5329adde69665"
 dependencies = [
  "async-trait",
  "fixedbitset 0.4.2",
@@ -2259,7 +2257,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "117725a109d387c937a1533ce01b450cbde6b88abceea8473c4d7a85853cda3c"
 dependencies = [
  "lazy_static",
- "windows-sys 0.48.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -17997,7 +17995,7 @@ version = "0.1.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c2a7b1c03c876122aa43f3020e6c3c3ee5c05081c9a00739faf7503aeba10d22"
 dependencies = [
- "windows-sys 0.48.0",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]


ewindisch (commented):
**Verdict: Approve with changes** — nothing blocking in the diff itself, but two behavior notes to acknowledge and one description fix before merge.

## Scope check
Lockfile-only; no `Cargo.toml` change, so the `[patch.crates-io]` moq-net pin is untouched. No packages added or removed. All three target bumps are within their declared semver ranges (`aws-lc-sys` 0.31→0.42 is a 0.x minor jump, but it's an internal detail of `aws-lc-rs` 1.17.1, which pins it). However, the PR body's "no transitive churn" claim is not accurate: the diff also re-points a dozen deps from `windows-sys` 0.48/0.52 to 0.59/0.60, drops `bindgen` from `aws-lc-sys` (0.42 ships pre-generated bindings; `pkg-config` added), thins `bindgen`'s own resolved deps (itertools 0.13→0.10.5, log/prettyplease dropped), and moves `tempfile` to `getrandom` 0.4. All of this is inert at runtime on Linux (cfg(windows)-gated or build-script-only) and re-uses versions already in the tree, but the description should say so rather than claim zero churn.

## casbin 2.17.0 → 2.20.0 (checked release notes 2.18.0–2.20.0)
The only functional change across the range is casbin-rs#407: `FileAdapter::save_policy` now joins rule fields with ", " instead of "," and strips the trailing newline. **No changes to policy matching, model evaluation, or RBAC semantics** — 2.18.1/2.19.x/2.20.0 are CI/docs-only releases. Relevance here: `auth/policy_manager.rs` uses `FileAdapter` for both load (line 303) and persist (line 725), so persisted policy CSVs will be rewritten in the new format after the first save. Load-side parsing trims whitespace, so round-trips are semantically stable — expect only cosmetic diffs in on-disk policy files (and in any tests that string-compare persisted policy output).

## aws-lc-rs 1.14.0 → 1.17.1 / aws-lc-sys 0.31 → 0.42 (checked 1.15–1.17.1 notes)
No TLS or FIPS default changes; the FIPS module is still opt-in via the `fips` feature, which we don't enable, and the new `legacy-des` support is feature-gated off. Two things worth knowing:
1. **1.16.3 tightened key-length validation at runtime** in `UnboundCipherKey::new()` and streaming cipher constructors (previously debug_assert-only). That's fail-closed hardening — a wrong-length key that silently worked in release builds now errors. We reach aws-lc only through rustls (our COSE/JWT stack deliberately pins jsonwebtoken's `rust_crypto` backend), so exposure is via the QUIC/TLS provider where key sizes are protocol-fixed; low risk, but it's a real behavior change.
2. **1.17.1 added system-AWS-LC auto-detection**: the build script can now link a system-installed AWS-LC discovered via pkg-config or `OPENSSL_*` environment variables instead of building from source. Our build environments set `OPENSSL_NO_VENDOR`/`OPENSSL_*` for openssl-sys — worth a one-time check that CI and the AppImage build don't accidentally pick up a system libcrypto here (it gates on a declared minimum AWS-LC version, so a plain OpenSSL install shouldn't match, but the env-var overlap is new). Also note aws-lc-sys 0.42 tracks upstream AWS-LC v5.1.0 (was 1.73.0 lineage) with new ML-DSA native assembly — none of it on our COSE signing path, which uses the RustCrypto `ml-dsa` crate.

## Asks before merge
- The PR body defined a validation gate (local `cargo nextest` of the authz/MAC/crypto crates, since PR CI is compile-only — the `build`/test job is merge_group-only). The PR is now marked ready: please confirm that gate actually ran, and note the result in the PR.
- Update the body's "no transitive churn" line to reflect the (inert) windows-sys/bindgen/getrandom re-pointing.

## Merge order
All three open dep PRs (#763, this, #769) edit `Cargo.lock` and will conflict pairwise. Suggest #763 → this → #769; this PR will need a trivial rebase after #763, and #769 rebases/regenerates last.

ewindisch:
## 🟡 Review (1:7 coordination sweep) — draft; sharpen the validation gate before undrafting

Lock consistency good: checksums present, `casbin = \"2\"` makes this a legal lockfile-only bump, `deny.toml` has no casbin/aws-lc pins to update, aws-lc-rs is transitive-only (rustls 0.23) with no direct API usage.

**1. (med) casbin 2.18.0 (inside this 2.17→2.20 jump) changed `FileAdapter` CSV parsing to "preserve spaces after commas".** hyprstream uses `FileAdapter` for on-disk policy load *and* persistence (`auth/policy_manager.rs:303` load, `:725` save). Whitespace handling of existing `policy.csv` rule tokens can change across the upgrade — a pre-2.18-written policy.csv whose rules round-trip differently yields rule atoms with different whitespace, silently failing (or newly matching) enforcement. **Your validation gate should explicitly load a pre-2.18-written `policy.csv` and diff enforcement decisions**, not just run unit tests against `MemoryAdapter`.

**2. (low) aws-lc-sys 0.31→0.42 drops the `bindgen` dep and adds `pkg-config`** (ships pre-generated bindings + pkg-config discovery) — which libcrypto gets built/linked can now vary per host toolchain. Verify CI, AppImage (`build-appimage.sh`), and dev hosts all still take the vendored/static path before undrafting.

**3. (low) PR body says "no transitive churn" but two `windows-sys` pins moved** (0.48→0.59/0.60) in the same lock edit — cosmetic only (Windows-only, Linux target), worth a one-line body correction so the "isolated crypto/authz bump" framing stays literally true.

Item 1 is the one that could silently change authorization behavior — please validate against a real pre-upgrade policy.csv. Flagging @1:2.

ewindisch:
Coordination/review note: I reviewed the rebased lockfile-only diff after #769. Scope is limited to casbin 2.17.0 -> 2.20.0, aws-lc-rs 1.14.0 -> 1.17.1, aws-lc-sys 0.31.0 -> 0.42.0, plus expected Cargo.lock transitive metadata. CI is green. GitHub rejected my formal approval as author-equivalent, so this needs a non-author approval and auto-merge enablement.

PR #769, opened 2026-07-04
deps: curated registry update (67 direct deps, semver-safe; ratatui/arrow held)

ewindisch:
**Draft** — curated registry dependency update, validated by CI full-compile.

## What this is
67 direct workspace dependencies bumped to their latest **semver-compatible** versions (respecting the `Cargo.toml` ranges). Notable: `clap 4.5.47→4.5.60`, `bitflags`, `chrono`, `bytes`, `config`, `zeroize 1.8.2→1.9.0`, `rustls`, `reqwest`, etc. Git deps untouched (rev-pinned). `cargo metadata --locked` passes.

## Why curated instead of a blanket `cargo update`
A full `cargo update` **breaks the build** in two spots that need code migration, so they are deliberately **held at main's versions** here and tracked for separate follow-up:
- **ratatui 0.30.0→0.30.2** now pulls `critical-section` for its layout cache; with no platform impl linked it fails: `undefined symbol: _critical_section_1_0_acquire` (breaks `hyprstream-tui`). Fix: enable a `critical-section` impl feature.
- **arrow 56.2.0→56.2.1** (nominally a patch) introduces `E0271`/`E0308` type errors in `hyprstream-metrics`. Fix: arrow-56.2.1 API migration.

These can't be held via lockfile edits alone (arrow sub-crates are pinned to 56.2.1 by other bumped dependents), so they belong in a dedicated PR with the code changes.

## Validation
CI Clippy (`cargo clippy --workspace --all-targets`) is a full compile and gates this. The runtime test suite runs on merge (push/merge_group), not PRs; the two pre-existing XET-merkle test failures on main are unrelated to this change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -229,9 +229,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.102"
+version = "1.0.103"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
+checksum = "2a4385e2e34eb35d6b3efe798b9eb88096925d87726c0798709bf56d9ed84af3"
 
 [[package]]
 name = "api_client"
@@ -1680,9 +1680,9 @@ checksum = "8f1fe948ff07f4bd06c30984e69f5b4899c516a3ef74f34df92a2df2ab535495"
 
 [[package]]
 name = "bytes"
-version = "1.11.1"
+version = "1.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1e748733b7cbc798e1434b6ac524f0c1ff2ab456fe201501e6497c8417a4fc33"
+checksum = "8ae3f5d315924270530207e2a68396c3cc547f6dca3fbdca317cfb1a51edb593"
 
 [[package]]
 name = "bytes-utils"
@@ -1741,7 +1741,7 @@ dependencies = [
  "cap-primitives",
  "cap-std",
  "io-lifetimes 2.0.4",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -1758,7 +1758,7 @@ dependencies = [
  "maybe-owned",
  "rustix 1.1.4",
  "rustix-linux-procfs",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
  "winx",
 ]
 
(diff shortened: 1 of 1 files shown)


ewindisch (commented):
**Verdict: LGTM — safe to automerge, but merge LAST of the three open dep PRs (after #763 and #768) and regenerate the lock on rebase.**

## Scope check
Lockfile-only; `Cargo.toml` untouched, so all requirement ranges and the `[patch.crates-io]` moq-net pin are unchanged. No git-dependency source lines changed anywhere in the lock. The held crates are genuinely held: zero `ratatui` or `arrow` lines in the diff, matching the stated rationale (critical-section / arrow-56.2.1 migrations deferred to code-change PRs).

## Semver discipline
I enumerated every version transition in the diff. All direct-dep bumps respect their declared ranges. The apparent "major" bumps in the batch are all transitive and driven by semver-compatible parents:
- `clap_lex` 0.7.5→1.1.0 (pulled by clap 4.5.60)
- `env_logger` 0.8.4→0.11.9 (pulled by quickcheck 1.0→1.1, dev-only)
- `tiff` 0.10→0.11, `zune-core`/`zune-jpeg` 0.4→0.5, `moxcms` 0.7→0.8 (pulled by image 0.25.10)

None of these is a direct workspace dependency crossing a major boundary.

## Notes
1. **New transitive packages appear** (console, env_filter, indicatif, memo-map, shake, sponge-cursor, symlink, unit-prefix, zmij) and a few drop out (dirs, dirs-sys, iri-string, redox_users). All are pulled by the semver-safe parent bumps (env_logger→env_filter, minijinja 2.21→memo-map, ml-dsa 0.1.1→shake/sponge-cursor, serde_json 1.0.150→zmij, etc.). cargo-deny (bans + licenses) is green, which covers the vetting, but the PR body's framing as pure version bumps slightly undersells this — worth a one-line mention.
2. **`ml-dsa` 0.1.0-rc.11 → 0.1.1** is the one crypto-relevant bump hiding in the batch: this is the RustCrypto crate behind our ML-DSA-65 half of the hybrid COSE composite signatures (hyprstream-rpc `crypto/pq.rs` / `cose_sign.rs`). Going from a release candidate to the stable release is exactly the right direction, but signature/key-encoding stability between rc.11 and 0.1.1 is the thing to watch: any previously persisted or exchanged ML-DSA artifacts (signed policies, audit checkpoints, minted grants) must still verify. The test suite runs at merge_group, so the queue will exercise the round-trip tests — just don't bypass the queue for this one, and if there are cross-version fixtures, all the better.
3. Also quietly in here: `quinn`/`quinn-proto` patch bumps (QUIC transport), `tokio` 1.48→1.50, `zeroize` 1.8→1.9. All routine.

## Merge order for the three open dep PRs
#763, #768, and this all edit `Cargo.lock` and will conflict pairwise. Recommended: **#763 first** (four-line hand edit), **#768 second** (isolated crypto/authz bumps with their own validation gate), **this one last** — it has the largest lock churn but is the easiest to regenerate, since the curated update can simply be re-run on top of the new main. Expect to rebase/re-resolve this branch twice.

Issue #527, opened 2026-06-27
XC: Jobs as scheduled workloads — wire WorkflowService jobs → scheduler (honor runs_on/resources)


ewindisch:
# XC — Jobs as scheduled workloads (wire WorkflowService → scheduler)

**Epic:** Worker Scheduler (cross-cut). **Fixes:** #520. **Relates:** #349/#487 (WorkflowService), #521 (dead config). **Depends on:** P2 (admission/placement). **Note:** WorkflowService is the GitHub-Actions-style engine and is **not currently an active feature**.

## Goal
Make a workflow **job** a scheduled workload: `runs_on:` becomes a placement directive, the job is admitted + placed via the scheduler (P2) onto a sandbox (or in-proc backend), honoring resources + quotas. Brings the GHA model (jobs run on runners) to hyprstream correctly, from the start.

## Why
Audit found `WorkflowRunner` executes steps via the VFS/Tcl namespace (`uses:`→`/bin/{action}` ctl, `run:`→Tcl on `spawn_blocking`), **never touching `SandboxPool`**; `runs_on:` is parsed and **ignored**; and `WorkflowService`'s doc-comment falsely claims it "spawns containers via WorkerService" (#520). So jobs have **no isolation**. Since the feature isn't active, there's **no migration cost** — design it unified now.

## Scope
1. **Job → scheduled workload** — a job declares backend/resources (from `runs_on:` + needs); dispatch routes it through the **P2 scheduler** (admission + placement) onto a sandbox or in-proc backend (via #516 registry), instead of fire-and-forget `tokio::spawn`.
2. **Honor `runs_on:`** — map the runner label → group/backend/resource constraint for placement.
3. **Run-level concurrency + timeouts** — wire #521's dead `max_concurrent_runs` + `job_timeout_secs`/`step_timeout_secs` (or supersede with the P2 admission/queue). Add retry policy if desired.
4. **Fix the false doc-comment** (#520) regardless of how far the wiring goes.
5. **Finish the stubs as needed** — `scan_repo`/`initialize`/`handle_subscribe` (#349/#487) are TODO stubs; sequence with this.

## Design notes
- A "job" = a node in the `needs:` DAG containing steps; a "run" = a workflow execution (tracked Queued→InProgress→Success/Failure in-memory today).
[...]

PR #770, opened 2026-07-04
XC (#527): jobs as scheduled workloads — wire WorkflowService through the P2 scheduler

### LINKED ISSUE #520: bug(workflow): jobs don't run in sandboxes; runs_on ignored; doc claims containers but wiring absent
`WorkflowRunner` executes steps through the VFS namespace — `uses:` → `/bin/{action}` ctl writes, `run:` → Tcl on `spawn_blocking` (`workflow/runner.rs` ~271-292,439-462). It **never touches `SandboxPool`**, and the parsed `runs_on:` (`RunsOn::Label`) is **completely ignored** — no sandbox acquisition, no placement.

Meanwhile `WorkflowService`'s doc comment claims it 'spawns containers via WorkerService' (`workflow/service.rs` ~81-82) — **that wiring does not exist**. Also `scan_repo`/`initialize`/`handle_subscribe` are TODO stubs (~150-155,266-272,561-572).

Impact: workflow jobs have **no isolation** today (run in-process via Tcl), contradicting the documented model. This is the workflow↔worker integration gap.

Decision needed (ties to scheduler epic): should jobs be scheduled onto sandboxes (job → `SandboxPool.acquire` honoring `runs_on`/resources)? Fix the false doc comment regardless.

Found during the worker-scheduling audit. Relates to #349, #508.

### LINKED ISSUE #521: bug(workflow): WorkflowConfig.max_concurrent_runs + job/step timeouts are dead config (read nowhere)
`WorkflowConfig.max_concurrent_runs: 10`, `job_timeout_secs`, `step_timeout_secs` (`workflow/config.rs` ~198-213) are **never read anywhere**. The only live concurrency control is a hardcoded `Semaphore::new(64)` on Tcl script steps in `WorkflowRunner` (~runner.rs:113); `dispatch()` always `tokio::spawn`s immediately with no run-level concurrency cap or timeout enforcement.

Impact: operators setting these expect them to take effect; they silently don't. No run-level concurrency limit, no job/step timeout enforcement, no retry.

Fix: wire `max_concurrent_runs` (run-admission semaphore) + `job_timeout_secs`/`step_timeout_secs` (tokio::time::timeout around job/step execution) into the runner; or remove the dead fields if superseded by the scheduler epic's admission model.

Found during the worker-scheduling audit. Relates to #349, #508.

ewindisch:
## Summary

Implements #527 (cross-cutting phase XC of epic #523): workflow jobs become
scheduled workloads instead of a fire-and-forget `tokio::spawn` that ignored
`runs_on:` entirely. Base branch is `ewindisch/525-p2-scheduler-engine`
(PR #761, still open at time of writing) since this ticket wires jobs
through that PR's new admission engine.

Closes #520. Part of epic #523 (cross-cutting phase XC, #527). Resolves #521
(wired for real — see below, not superseded).

## What was built

1. **`workflow/scheduler.rs` (new)** — a `JobScheduler` that maps a job's
   `runs_on:` label onto either:
   - **in-proc** (today's VFS `/bin/` ctl + TclShell path, unchanged) for a
     small reserved alias set (`IN_PROC_LABELS`: `local`, `self-hosted`,
     `in-process`, `tcl`), or
   - a **real sandbox reservation** acquired via `SandboxPool::acquire`
     (#525's P2 admission engine) for any other label, with the label
     recorded as the admission `group` annotation (`ANN_GROUP`) so distinct
     runner labels get independent per-group quota via
     `AdmissionConfig::max_per_group`.

   `WorkflowRunner` now takes an optional `Arc<JobScheduler>`
   (`with_scheduler`); `None` (the default via `WorkflowRunner::new`)
   preserves the exact pre-#527 in-proc-only behavior — no regression.
   `WorkflowService::set_job_scheduler` exposes the same opt-in wiring at
   the service layer.

2. **`resources:` job field (`JobResources`: `cpu_millis`/`memory_bytes`/
   `gpu`)** — a hyprstream extension to the GHA job schema (GHA has no
   native resources key), `#[serde(default)]` so it never affects existing
   workflow YAML. Maps onto `PodSandboxConfig.linux.resources` +
   `ANN_GPU_REQUEST`, the same vocabulary `runtime::admission::derive_demand`
   already reads.

3. **#521 — `max_concurrent_runs`/`job_timeout_secs`/`step_timeout_secs`
   wired for real** (not superseded by P2): `WorkflowRunner::with_config`
   sizes a run-admission semaphore from `max_concurrent_runs` (acquired for
   the whole `run()` call — a different axis from `script_semaphore`, which
   bounds concurrent script-executing *jobs*), and wraps job/step execution
   in `tokio::time::timeout` using `job_timeout_secs`/`step_timeout_secs`. A
   job's own `timeout-minutes:` (parsed since day one, previously never
   read) overrides the default job timeout. I chose to wire this for real
   rather than treat it as superseded, because P2's admission governs
   resource *capacity/quota*, while this governs wall-clock *run/step
   duration* — a genuinely different axis, not a duplicate control.

4. **Doc-comment fix** (`workflow/mod.rs`, `workflow/service.rs`) — both
   claimed the engine "spawns containers via WorkerService"; that
   dependency doesn't exist. Corrected to describe the actual
   `WorkflowRunner` → `JobScheduler` → `SandboxPool` path.

5. **`scan_repo`/`initialize`/`handle_subscribe` stubs (#349/#487)** —
   verified these are **already fully implemented** on this base branch
   (not TODO stubs as the original #520/#527 issue text describes — that
   text appears stale relative to current `main`). No work was needed here;
   I did not expand into #349/#487's broader scope.

## Judgment calls flagged for reviewer sign-off

- **`runs_on` → group, not `runs_on` → backend.** `JobScheduler` wraps
  exactly **one** pre-resolved `SandboxPool` (bound to one `SandboxBackend`
  at construction — this mirrors `pool.rs`'s own documented invariant that a
  pool never re-selects its backend per-request). So `runs_on: kata` and
  `runs_on: nspawn` do **not** pick different backends within one scheduler;
  both go to that scheduler's one backend, tagged with a different admission
  group. Routing different labels to genuinely different backends would mean
  a router over multiple backend-specific pools — the same architectural
  change `pool.rs` already flags as a follow-up, not something this ticket
  re-derives. I picked this mapping because it's directly testable against
  the existing admission engine (quota/fit) without inventing new
  multi-backend plumbing in a crate that currently has zero external
  callers. Full write-up in `scheduler.rs`'s module docs.
- **What "runs in a sandbox" means today.** Placement acquires a *real*
  reservation — admission is checked and the backend's `start()` actually
  runs (a real Kata VM boots when the pool is Kata-backed), closing #520's
  "no isolation" gap structurally. Step *content* execution (`uses:`/`run:`
  bodies) still runs through the existing host-side VFS `/bin/` ctl +
  TclShell bridge, not inside the guest's own namespace — routing individual
  step exec into the guest (via `kata_agent::ExecProcess`, `nspawn`, etc.)
  needs the streaming/fd plane `exec_mount.rs` itself documents as not yet
  built (`/exec/instances/<id>/fd/` — "depends on the streaming plane").
  That's a materially larger follow-up, not attempted here.
- **Subject identity for job placement.** Reused the identity already
  threaded through `WorkflowRunner::run()`/`WorkflowService::dispatch()` —
  the RPC caller's verified `ctx.subject()` for manual dispatch, or the
  pre-existing placeholder `Subject::new("workflow-event-dispatch")` for
  event-triggered auto-dispatch (not new — already present on this base
  branch). I did not invent a new identity. Worth an operator decision on
  whether the event-triggered placeholder should eventually carry a real
  per-repo service identity/DID for quota purposes — flagged, not solved
  here.
- **Only the first `runs_on` label drives placement.** GHA's
  `runs-on: [a, b]` means "match ALL labels" against a runner's capability
  set; hyprstream has no such capability set to intersect against under the
  group mapping above, so multi-label AND-composition has no defined
  meaning yet. Additional labels are accepted (not a parse error) but
  otherwise ignored — documented in `scheduler::primary_label`, not a silent
  drop.
- **No capnp/wire-format changes.** Everything here is internal to
  `hyprstream-workers` (the `resources:` field is parsed YAML, not an RPC
  schema addition), so no reviewer sign-off is needed on that front per the
  issue's own "human sign-off" note.

## Test coverage (acceptance criteria)

- `workflow::scheduler::tests::isolated_runs_on_acquires_a_real_sandbox` /
  `in_proc_runs_on_never_touches_the_backend` — a job's `runs_on` value
  really does decide whether the backend's `start()` is invoked (closes
  #520's "no isolation" gap).
- `workflow::scheduler::tests::runs_on_label_influences_admission_via_group_quota`
  — two jobs with the *same* `runs_on` label hit a per-group quota and the
  second is rejected, while a job with a *different* label is unaffected —
  proving `runs_on`'s value concretely influences the scheduler's placement
  decision.
- `workflow::runner::tests::step_timeout_fails_the_step_and_job` /
  `job_timeout_fails_the_whole_job` — real `tokio::time::timeout` firing
  against a mount that genuinely pends (not a synchronously-ready future),
  so the timeout race is exercised for real.
- `workflow::runner::tests::max_concurrent_runs_serializes_concurrent_runs`
  — two concurrent `run()` calls with `max_concurrent_runs: 1` never
  overlap their job execution (tracked via a shared atomic peak counter).

`cargo build -p hyprstream-workers`: clean.
`cargo test -p hyprstream-workers`: 147/147 passed (144 pre-existing + 3 new
timeout/concurrency tests; scheduler.rs adds 7 of its own within that count).
`cargo clippy -p hyprstream-workers --all-targets`: clean, and also clean
under `--features wasm`, `--features oci`, `--features cri`, and
`--features kata-vm` (the one pre-existing warning under `--features wasm`,
an unused import in `wasm_backend.rs`, predates this change and isn't
touched by it).

## Test plan

- [x] `cargo build -p hyprstream-workers`
- [x] `cargo test -p hyprstream-workers` (147/147)
- [x] `cargo clippy -p hyprstream-workers --all-targets` (clean)
- [x] `cargo clippy -p hyprstream-workers --all-targets --features {wasm,oci,cri,kata-vm}` (clean)

## Rebased onto merged #761 + B′ alignment (2026-07-09)

Rebased onto `main` post-#761 (clean rebase, no textual conflicts) and aligned
with the ratified B′ group semantics (#921 decision 5, merge-path annotation
on this PR):

- `runs_on:` → `hyprstream.io/group` is a **capacity-partition selector**,
  not an authoritative quota key: the scheduler's only consumer of the label
  is the `ANN_GROUP` annotation read back by `admission::derive_group_selector`
  inside `SandboxPool::acquire`, so it flows through #761's
  `GroupSelectorValidator` seam (deny-unknown by default) — no parallel
  counter, no bypass. A new test
  (`unverified_runs_on_group_is_rejected_at_admission`) proves an
  unverifiable `runs_on` selector is rejected at admission under the
  production default, with the backend never touched.
- Added `SandboxPool::with_group_validator` (mirrors
  `AdmissionTracker::with_group_validator`) so a membership-backed validator
  (`PlacementIndex::is_member`, when wired) or a test double can be injected
  at pool construction; `SandboxPool::new` keeps the fail-closed
  `DenyUnknownGroupValidator`. No membership resolution is built here.
- Scheduler tests updated to B′: isolated placement + the group-quota test
  run against a `StaticGroupMembership` the test Subject belongs to;
  `max_per_group` updated for #761's `Option<usize>` quota fields.

`cargo test -p hyprstream-workers`: 160/160; workspace release clippy
(`-D warnings`): clean.

### DIFF
--- a/crates/hyprstream-workers/src/workflow/runner.rs
+++ b/crates/hyprstream-workers/src/workflow/runner.rs
@@ -88,6 +88,14 @@ pub enum RunnerError {
     #[error("cancelled")]
     Cancelled,
 
+    /// Job placement (scheduler decide + admission) failed. Carries the typed
+    /// [`WorkerError`] so callers/retry-policy can distinguish "retry later,
+    /// queue pressure" (`QueueFull`/`QueueTimeout`) from "terminal, over-quota
+    /// or unauthorized" (`AdmissionDenied`/`Unauthorized`/`AdmissionInfeasible`)
+    /// instead of a flattened string.
+    #[error("job placement failed: {0}")]
+    PlacementFailed(#[from] crate::error::WorkerError),
+
     #[error("vfs error: {0}")]
     VfsError(String),
 
@@ -302,11 +310,22 @@ async fn run_job(
     // #527: decide + (if isolation is requested) acquire a sandbox
     // reservation before running any steps. `None` scheduler ⇒ always
     // in-proc (pre-#527 behavior, unchanged).
+    //
+    // The placement sits outside `job_timeout` and any step-level cancel race,
+    // so it would otherwise wait up to `queue_timeout_secs` in the admission
+    // queue even for an already-cancelled workflow. Race it against the
+    // cancellation token so a cancel aborts the queue wait promptly, and
+    // propagate the typed `WorkerError` (via `PlacementFailed`) instead of
+    // flattening it into a string.
     let placement = match &scheduler {
-        Some(scheduler) => scheduler
-            .place(job_name, &job.runs_on, &job.resources, subject)
-            .await
-            .map_err(|e| RunnerError::VfsError(format!("job placement failed: {e}")))?,
+        Some(scheduler) => {
+            let place_fut = scheduler.place(job_name, &job.runs_on, &job.resources, subject);
+            tokio::select! {
+                biased;
+                _ = cancel.cancelled() => return Err(RunnerError::Cancelled),
+                result = place_fut => result?,
+            }
+        }
         None => Placement::InProc,
     };
 
--- a/crates/hyprstream-workers/src/workflow/runner.rs
+++ b/crates/hyprstream-workers/src/workflow/runner.rs
@@ -6,9 +6,19 @@
 //! - Resolves `uses:` action steps through VFS `/bin/` ctl files
 //! - Evaluates `run:` script steps via TclShell on `spawn_blocking`
 //! - Propagates CancellationToken for cooperative cancellation
+//! - Routes each job through a [`JobScheduler`] (#527) — `runs_on:` +
+//!   `resources:` decide in-proc vs. a real P2-admitted sandbox reservation
+//!   (see `workflow::scheduler`); no scheduler configured (`None`, the
+//!   default) preserves the pre-#527 in-proc-only behavior exactly.
+//! - Honors `WorkflowConfig::max_concurrent_runs` (a run-admission semaphore,
+//!   distinct from `script_semaphore`'s per-job script concurrency) and
+//!   `job_timeout_secs`/`step_timeout_secs` (#521 — wired for real, not
+//!   superseded by P2: admission governs resource capacity/quota, this
+//!   governs wall-clock run/step duration, a different axis)
 
 use std::collections::{HashMap, HashSet, VecDeque};
 use std::sync::Arc;
+use std::time::Duration;
 
 use tokio::sync::Semaphore;
 use tokio::task::JoinSet;
@@ -18,7 +28,10 @@ use hyprstream_workers_tcl::TclShell;
 // VFS proxy no longer needed — TclShell awaits namespace operations directly.
 use hyprstream_vfs::{Namespace, Subject};
 
+use crate::config::WorkflowConfig;
+
 use super::parser::{Job, Step, Workflow};
+use super::scheduler::{JobScheduler, Placement};
 
 // ─────────────────────────────────────────────────────────────────────────────
 // Result types (local to runner, not wire-format)
@@ -96,22 +109,63 @@ type RunnerResult<T> = std::result::Result<T, RunnerError>;
 /// Resolves action steps through VFS `/bin/` ctl files and evaluates
 /// script steps via TclShell on `spawn_blocking` threads. Concurrent
 /// script-executing jobs are capped by `script_semaphore`.
+///
+/// # Scheduling (#527)
+///
+/// Each job is routed through an optional [`JobScheduler`]: `runs_on:` +
+/// `resources:` decide in-proc execution (unchanged) vs. a real sandbox
+/// reservation acquired through the #525 P2 admission engine. With no
+/// scheduler configured (the default), every job runs in-proc exactly as
+/// before this change — no regression for existing callers.
+///
+/// # Run-level concurrency + timeouts (#521)
+///
+/// `run_semaphore` bounds concurrent workflow *runs* (`WorkflowConfig::
+/// max_concurrent_runs`) — a different axis from `script_semaphore`, which
+/// bounds concurrent script-executing *jobs* within/across runs.
+/// `job_timeout`/`step_timeout` (`job_timeout_secs`/`step_timeout_secs`) wrap
+/// job and step execution in `tokio::time::timeout`; a job's own
+/// `timeout-minutes:` overrides `job_timeout` when present.
 #[derive(Clone)]
 pub struct WorkflowRunner {
     ns: Arc<Namespace>,
     script_semaphore: Arc<Semaphore>,
+    run_semaphore: Arc<Semaphore>,
+    job_timeout: Duration,
+    step_timeout: Duration,
+    scheduler: Option<Arc<JobScheduler>>,
 }
 
 impl WorkflowRunner {
-    /// Create a new WorkflowRunner.
+    /// Create a new WorkflowRunner with `WorkflowConfig::default()` timeouts/
+    /// concurrency and no job scheduler (every job runs in-proc).
     ///
     /// * `ns` — the root namespace (mounts for `/bin/`, `/env/`, `/srv/`, etc.)
     /// * Default concurrency cap: 64 concurrent script-executing jobs.
     pub fn new(ns: Arc<Namespace>) -> Self {
+        Self::with_config(ns, &WorkflowConfig::default())
+    }
+
+    /// Create a WorkflowRunner honoring `config`'s `max_concurrent_runs` +
+    /// `job_timeout_secs`/`step_timeout_secs` (#521).
+    pub fn with_config(ns: Arc<Namespace>, config: &WorkflowConfig) -> Self {
         Self {
             ns,
             script_semaphore: Arc::new(Semaphore::new(64)),
+            // A zero-permit semaphore would deadlock every run forever;
+            // treat a misconfigured 0 as "at least 1" rather than a silent hang.
+            run_semaphore: Arc::new(Semaphore::new(config.max_concurrent_runs.max(1))),
+            job_timeout: Duration::from_secs(config.job_timeout_secs),
+            step_timeout: Duration::from_secs(config.step_timeout_secs),
+            scheduler: None,
         }
+    }
+
+    /// Attach a [`JobScheduler`] (#527): jobs whose `runs_on:` requests
+    /// isolation are placed onto its sandbox pool instead of running in-proc.
+    pub fn with_scheduler(mut self, scheduler: Arc<JobScheduler>) -> Self {
+        self.scheduler = Some(scheduler);
+        self
     }
 
     /// Run a complete workflow.
@@ -119,6 +173,9 @@ impl WorkflowRunner {
     /// Jobs are topologically sorted by `needs:` into execution waves.
     /// Independent jobs in each wave run concurrently. The `cancel` token
     /// is checked between waves and propagated to individual jobs.
+    ///
+    /// Bounded by `run_semaphore` (#521's `max_concurrent_runs`): a run that
+    /// can't get a permit waits here, honoring `cancel` while waiting.
     pub async fn run(
         &self,
         workflow: &Workflow,
@@ -128,6 +185,14 @@ impl WorkflowRunner {
     ) -> RunnerResult<WorkflowRunResult> {
         // Validate inputs: reject _-prefixed keys (provenance injection prevention).
         validate_inputs(&inputs)?;
+
+        // #521: bound concurrent workflow RUNS. Held for the whole run.
+        let _run_permit = tokio::select! {
+            permit = Arc::clone(&self.run_semaphore).acquire_owned() => {
+                permit.map_err(|_| RunnerError::Cancelled)?
+            }
+            _ = cancel.cancelled() => return Err(RunnerError::Cancelled),
+        };
 
         // Topological sort jobs into execution waves.
         let waves = topological_sort(&workflow.jobs)?;
@@ -150,6 +215,12 @@ impl WorkflowRunner {
                 };
                 let ns = Arc::clone(&self.ns);
                 let sem = Arc::clone(&self.script_semaphore);
+                let scheduler = self.scheduler.clone();
+                let job_timeout = job
+                    .timeout_minutes
+                    .map(|m| Duration::from_secs(u64::from(m) * 60))
+                    .unwrap_or(self.job_timeout);
+                let step_timeout = self.step_timeout;
                 let subject = subject.clone();
                 let cancel = cancel.clone();
                 let env = merge_env(&workflow.env, &job.env);
@@ -157,7 +228,10 @@ impl WorkflowRunner {
                 let name = job_name.clone();
 
                 join_set.spawn(async move {
-                    let result = run_job(ns, sem, &job, &name, env, &inputs, &subject, cancel).await;
+                    let result = run_job(
+                        ns, sem, scheduler, &job, &name, env, &inputs, &subject, cancel,
+                        job_timeout, step_timeout,
+                    ).await;
                     (name, result)
                 });
             }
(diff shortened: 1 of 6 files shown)


crates/hyprstream-workers/src/workflow/runner.rs
```diff
@@ -209,19 +283,72 @@ impl WorkflowRunner {
...
+    step_timeout: Duration,
+) -> RunnerResult<JobRunResult> {
+    // #527: decide + (if isolation is requested) acquire a sandbox
+    // reservation before running any steps. `None` scheduler ⇒ always
+    // in-proc (pre-#527 behavior, unchanged).
+    let placement = match &scheduler {
+        Some(scheduler) => scheduler
+            .place(job_name, &job.runs_on, &job.resources, subject)
+            .await
+            .map_err(|e| RunnerError::VfsError(format!("job placement failed: {e}")))?,
```
ewindisch: This flattens the admission engine's typed failure modes (`AdmissionDenied` vs `QueueFull` vs `QueueTimeout` vs `Unauthorized`) into a generic `VfsError` string — a caller (or a future retry policy) can't tell "retry later, queue pressure" apart from "terminal, over quota / unauthorized" without parsing text. A `RunnerError::PlacementFailed(WorkerError)`-style variant would preserve the class. Related: `place()` here sits outside both the `job_timeout` wrapper below and any `cancel` race, so a cancelled workflow can still wait up to `queue_timeout_secs` in the admission queue — consider `tokio::select!` with `cancel.cancelled()` around the placement.

crates/hyprstream-workers/src/workflow/scheduler.rs
```diff
@@ -0,0 +1,428 @@
...
+    /// e.g. via `runtime::resolve_backend` — the same #516 fail-closed
+    /// registry `worker.backend` uses) as the target for isolated jobs.
+    pub fn new(pool: Arc<SandboxPool>) -> Self {
+        Self { pool }
+    }
+
+    /// Decide placement for a job and, if it requests isolation, acquire a
+    /// sandbox reservation for it (admitted via `subject`'s identity/quota —
+    /// fail-closed if `subject` is anonymous, see `runtime::admission`).
+    pub async fn place(
```
ewindisch, line 200: Heads-up on the identity that arrives here via RPC dispatch: `handle_dispatch` (workflow/service.rs, pre-existing) constructs it as `Subject::new(ctx.subject().to_string())`. Because `Subject::anonymous()` Displays as `"anonymous"`, that round-trip turns an anonymous caller into a *named* subject `"anonymous"` (`is_anonymous() == false`), which then passes the fail-closed check in `AdmissionTracker::reserve` that this PR newly puts on the path. The `authorize()` gate is the current backstop, but the one-line fix — pass `ctx.subject()` through directly, it already returns a `Subject` — restores the fail-closed property here and keeps per-subject quota accounting honest. Worth doing in this PR since this PR is what makes the string round-trip security-relevant.

ewindisch (commented):
**Verdict: Approve with changes** — the wiring is the right shape (jobs route through #761's admission engine, not a bespoke queue; failure paths surface rather than drop), but one identity bug should be fixed in this PR because this PR is what makes it reach the admission layer, and there is a hard merge-order requirement.

**Merge order (required)**

This branch textually stacks on `ewindisch/525-p2-scheduler-engine`: PR #761's commit is an ancestor of this head, and both PRs target `main` independently. #761 must merge first. If #761 is squash-merged, this branch must then be rebased onto the new `main` before merging, or its merge will re-apply #761's entire diff as part of this PR. Do not put both into the merge queue simultaneously. (No file overlap with open PRs #838/#839.)

**What checks out**

- Jobs go through the P2 scheduler, not a parallel path: `JobScheduler::place` calls `SandboxPool::acquire` with the caller's `Subject`, `resources:` maps onto the exact vocabulary `derive_demand` reads (`linux.resources` + `ANN_GPU_REQUEST`), and `runs_on` → `ANN_GROUP` reuses the admission engine's existing group-quota axis. No new scheduling vocabulary, no framework ceremony.
- Admission rejection is surfaced, not dropped: a placement failure fails the job with an error carried into the run result; the sandbox reservation is released after job completion, timeout, or failure alike (`placement.release()` runs on every exit of the timeout race).
- `None`-scheduler default preserves pre-#527 behavior exactly; the #521 wiring (run semaphore + job/step timeouts, with `timeout-minutes:` override) is a genuinely different axis from admission, and the tests exercise real pending futures rather than instantly-ready ones. The `max(1)` guard on a zero `max_concurrent_runs` avoids a silent permanent hang.
- The false "spawns containers via WorkerService" doc-comment is fixed.

**Findings**

1. *Medium — anonymity laundering into admission (fix in this PR).* `handle_dispatch` builds the runner identity as `Subject::new(ctx.subject().to_string())` (workflow/service.rs, pre-existing on main). `Subject::anonymous()` Displays as `"anonymous"`, so the round-trip produces a **named** subject `"anonymous"` with `is_anonymous() == false` — which now flows into #761's fail-closed admission check and passes it. Before this PR that string only reached VFS calls; this PR is what makes it an admission credential. The `authorize()` gate mitigates when configured (and bails when not), but the identity fed to per-subject quotas is still wrong. The fix is one line: pass `ctx.subject()` through directly — it already returns a `Subject`; the stringify/re-wrap loses the anonymous state (and would also mangle nothing else, since federated subjects survive as their `iss:sub` string either way, though passing the value through is strictly better).
2. *Flagged for operator sign-off — placeholder identity for event-triggered dispatch.* `Subject::new("workflow-event-dispatch")` means every auto-dispatched job shares one per-subject quota bucket and counts as authenticated to the admission engine. The PR flags this honestly; fine for now, but worth an explicit operator decision (per-repo service identity?) before event-triggered workflows are load-bearing.
3. *Low — error class collapsed at the placement boundary.* `RunnerError::VfsError(format!("job placement failed: {e}"))` flattens `AdmissionDenied` / `QueueFull` / `QueueTimeout` / `Unauthorized` into one string-typed error, so callers can't distinguish retryable (queue pressure) from terminal (quota/authz) without parsing text. A dedicated `RunnerError::PlacementFailed` variant preserving the `WorkerError` would keep the class intact for future retry policy — no retry semantics exist today, which is fine (each `dispatch` mints a fresh run; there's no double-submission hazard), but the first retry implementation will want this distinction.
4. *Low — placement is outside both the job timeout and the cancel race.* `scheduler.place()` runs before the `tokio::time::timeout(job_timeout, ...)` wrapper and is not raced against `cancel`, so a cancelled workflow can still sit in the admission queue for up to `queue_timeout_secs`, and queue wait doesn't count against the job's wall-clock budget. Wrapping `place()` in a `tokio::select!` with `cancel.cancelled()` would make cancellation prompt.

**Expectation-setting (acknowledged, must not be oversold)**

A `runs_on: kata` job acquires and holds a *real* sandbox reservation (admission checked, backend `start()` runs), but step content still executes host-side via the VFS/Tcl bridge — isolation is a reservation, not containment. The PR documents this prominently and correctly; the operator should decide whether this satisfies #520's "jobs requesting isolation run in a sandbox (not in-proc Tcl)" criterion or whether #520 stays open until step exec routes into the guest. Nothing in the code claims otherwise, which is the right posture.

ewindisch (commented):
Approve with changes — remaining: all prior findings, including the identity round-trip bug. Follow-up check for post-review fixes.

This head (07a4e16e0) is the same commit the earlier review (#pullrequestreview-4631374264) was submitted against — no new commits have been pushed. Verified against the current code:

1. NOT FIXED — handle_dispatch (workflow/service.rs:721) still builds `Subject::new(ctx.subject().to_string())`; an anonymous context Displays as "anonymous" and round-trips into a named Subject, defeating #761's fail-closed is_anonymous() admission check. Fix remains: pass ctx.subject() through directly.
2. UNCHANGED — event-triggered dispatch still uses the placeholder `Subject::new("workflow-event-dispatch")` (service.rs:301). Still pending the operator decision flagged in the prior review.
3. NOT FIXED — placement errors still collapse to a VfsError string (runner.rs:309), and place() (runner.rs:307) still runs outside the job timeout at runner.rs:315.

Stacking confirmed: this branch is still a descendant of #761's head (a49d179fc), so merge order #761 → #770 holds. No new defects — the code is unchanged since the prior review. Finding 1 should land in this PR before merge, since this PR is what routes dispatch identity into the admission layer.

ewindisch:
## 🟡 Review (1:7 coordination sweep) — minor issues; MERGEABLE. ⚠️ merge-ordering caveat below

Verified positives: `runs_on`→`ANN_GROUP` reuses the #525/#628 vocabulary (not a parallel one) — `derive_demand` reads back exactly what `job_pod_sandbox_config` writes, cpu round-trip verified; manual dispatch threads the real verified `ctx.subject()` (`service.rs:686`) into placement, so RPC-initiated runs preserve authorization through the scheduler; wave-join drains before returning.

**⚠️ Merge ordering:** #770's base was retargeted to `main` but its branch *contains all of #761's commits*, so **merging #770 alone merges both changesets.** Land #761 first, then rebase/re-diff #770 — otherwise #761's review (see its blocking comments) is bypassed.

**1. (med) Timed-out jobs keep executing after their quota is released.** On job/step timeout (`runner.rs` ~L1935-1955, ~L2008-2043) the future is dropped but the TclShell `spawn_blocking` thread keeps running the script, while the sandbox placement (reservation + quota) and the script-semaphore permit are released immediately. The script continues mutating host VFS state after its quota is returned and its sandbox is back in the warm pool — the 64-script cap and per-subject/group quotas undercount actual concurrent execution.

**2. (med, partially disclosed) Event-triggered auto-dispatch uses a synthetic `Subject::new(\"workflow-event-dispatch\")`** (`scheduler.rs::place` ~L2667-2685) — passes #761's fail-closed anonymous check but carries no verified principal, so all event-triggered jobs from all repos/tenants share one quota bucket under an unauthenticated identity. Admission's identity gate is structurally satisfied but semantically void for this path.

**3. (low) Group key is workflow-author-controlled** — `runs-on: <someone-else's-label>` consumes that group's quota, or fresh labels get unlimited fresh groups. **(low) empty `runs_on` → in-proc host execution** (fail-open toward *less* isolation in an isolation feature). **(low) `Placement` has no `Drop` guard** — a future caller that drops/aborts a run mid-wave leaks quota+capacity (latent; current sole caller is a never-aborted detached spawn). **(low) "Closes #520" overstates** — `set_job_scheduler`/`JobScheduler::new` have zero non-test callers, so isolation stays dormant until a prod site attaches the scheduler.

Items 1–2 + the merge-ordering caveat should be resolved before merge. Flagging the owning session.

ewindisch:
## Ratified: group-authority decision (B′) — merge path for this PR

Per #921 open decision 5 (ratified 2026-07-09, full text on #761): the `runs_on:` → `ANN_GROUP` mapping in this PR is acceptable **only under the demoted semantics** — the group value is a *capacity-partition selector*, validated against the verified Subject's bidirectional-consent placement-group membership (`PlacementIndex::is_member`), never an authoritative quota/trust/billing key. A workflow naming a group its Subject is not a member of must be rejected at admission, not silently accepted. Authoritative quota arrives via the ledger path (#922/#925).

This PR merges after #761's fix wave lands (it inherits the validated-selector semantics from the scheduler side); no independent rework required beyond rebasing onto the realigned #761.

ewindisch:
Both review threads addressed by the fix commit fdd8d81 (kiro-opus review: PASS). Thread 1 (anonymous subject): `ctx.subject()` now passed directly, round-trip removed, anonymity preserved (regression test + admission's anonymous_subject_rejected_fail_closed cover it). Thread 2 (runner.rs): typed PlacementFailed(WorkerError) + cancel-aware placement select added. No regression to #761 admission logic (delta doesn't touch admission.rs/scheduler.rs).

Issue #725, opened 2026-07-04
ts_codegen: regen is unsafe — unresolved struct refs emitted as comments, dropped fields (EndpointInfo.pubkey/selfProof), barrel collisions


ewindisch:
## Problem

A fresh TS client regen against current schemas (attempted as W1 of the browser moq streaming work, www-cyberdione-ai#11) parses everything but produces **48 tsc errors** in the frontend, and inspection shows the generator — not the schemas — is at fault in at least three ways:

1. **Unresolved struct references emitted as comments.** The generated `streaming.ts` contains lines like `// certHashes: List(Data) — struct definition not found` at several accessor sites, while emitting a correct `getDataList(2)` for the *same field* at others. The generator is failing to resolve a struct/field in some emission contexts and silently degrading to a comment — a silent-drop failure mode.

2. **Dropped fields on regen.** The regenerated `discovery.ts` loses `EndpointInfo.pubkey` / `EndpointInfo.selfProof` — fields present in the schema and depended on by the frontend (identity/self-proof verification). A regen must never silently narrow a type.

3. **Barrel collisions between union types.** The `events` / `streaming` QoS unions collide in the generated barrel exports, so a full regen can't even typecheck without hand-editing.

Plus consumer-visible drift (`model.ts` `ChatTemplateRequest` breaking `useChatCore` / `HyprstreamRpcContext`) that is legitimate catch-up work, but is currently indistinguishable from generator bugs in the error pile.

## Why it matters

The generated clients are several schema-versions behind hyprstream because regen is effectively unsafe: every regen risks silently dropping fields (2) or emitting comment-holes (1). The frontend has been coping by *not* regenerating, and the pending browser-streaming wiring (www#11 W3) is blocked on a trustworthy regen. Hand-patching generated files is not an option — the fix must be in the generator.

## Acceptance

- [ ] Generator resolves `List(Data)` (and any other currently-unresolved shapes) in **all** emission contexts — zero `struct definition not found` comments in output; make that an assertion/build failure, not a comment.
- [ ] Regen of `discovery.ts` preserves `EndpointInfo.pubkey` / `selfProof` (add a golden/round-trip test over the current schema set).
- [ ] Barrel generation disambiguates colliding union type names across schemas (`events` vs `streaming` QoS).
- [ ] A clean full regen against current schemas typechecks in www with only *consumer-side* fixes remaining (real API drift, e.g. `ChatTemplateRequest`).

## Relationship

[...]

PR #772, opened 2026-07-04
feat(ts_codegen): support List(List(T)) nested-list fields (#758)

### LINKED ISSUE #758: ts_codegen: List(List(T)) nested-list fields fail to generate (blocks inference.ts/model.ts regen)
Follow-up to #725/PR #728. The generator now correctly hard-fails on unresolved list/struct references instead of silently degrading — but that surfaced a real gap: nested lists (`List(List(Float32))`, used by embedding fields in `inference.capnp`/`model.capnp`) have no codegen support at all, so a full regen of `inference.ts`/`model.ts` currently aborts.

**Impact:** blocks www-cyberdione-ai#11 child item #12 (W1 service-client regen for the browser moq inference streaming work). Current mitigation there: the hand-checked pre-#320 `GenerationRequest` builder is verified wire-correct against the live schema (same 16 fields/ordinals), so W3 streaming is unblocked without the regen — but consumers cannot pick up any other schema changes (e.g. `serviceJwt` flag-day, `StreamInfo` reach fields) until this lands.

**Ask:** extend `ts_codegen` (`crates/hyprstream-rpc-build/src/ts_codegen/{builders,parsers,message}.rs`) to emit `List(List(T))` getters/setters (reader: nested `StructList`/pointer-list-of-lists; builder: nested `initXList`/`setXList`), with a golden test covering an embeddings-shaped field, before flipping the failure back into working codegen for this shape.

cc #719, #722 (browser moq subscriber), www-cyberdione-ai#11/#12

ewindisch:
## Summary

Follow-up to #725/PR #728. `ts_codegen` now hard-fails on unresolved list/struct references instead of silently degrading, which surfaced a real gap: `List(List(T))` (used by the `embeddings` field of `EmbedImagesResponse` in `inference.capnp`/`model.capnp`) had no codegen support at all, so a full regen of `inference.ts`/`model.ts` aborted.

This extends `ts_codegen` to emit real `List(List(T))` getters/setters for primitive-scalar inner types (`Bool`/`UIntN`/`IntN`/`FloatN`), and flips the previous hard-panic back into working codegen for this shape.

## What changed

- **`crates/hyprstream-rpc-build/src/ts_codegen/message.rs`** (the generated `capnp.ts` runtime): added `get{Bool,Uint8,Int8,Uint16,Int16,Uint32,Int32,Uint64,Int64,Float32,Float64}ListList` to `CapnpReader`/`StructReader`, and matching `set*ListList` to `CapnpArena`/`StructBuilder`. The outer list is encoded as a plain pointer list (`element_size=6`) whose elements are themselves primitive lists — both the canonical pointer-list and inline-composite outer encodings are handled, mirroring the existing `List(Text)`/`List(Data)` support. `allocPrimitiveList` was refactored into an offset-based `allocPrimitiveListAt` so nested-list setters can write inner lists at a computed pointer slot instead of a fixed `ptrIndex`.
- **`crates/hyprstream-rpc-build/src/ts_codegen/mod.rs`**: added `nested_list_getter_method`/`nested_list_setter_method`, mapping a doubly-nested inner type name to its runtime method (mirrors `list_getter_method`/`list_setter_method`).
- **`crates/hyprstream-rpc-build/src/ts_codegen/parsers.rs`** and **`builders.rs`**: every place that previously matched `inner.starts_with("List(")` and hard-panicked now recognizes `List(List(<inner>))` and dispatches to the new runtime methods for primitive inner types, still hard-failing (never silently degrading) for inner shapes the runtime doesn't support yet (`Text`/`Data`/struct/triple-nested).

## Golden test

Added a test to both `parsers.rs` and `builders.rs` using an embeddings-shaped schema (`embeddings @0 :List(List(Float32)); dimensions @1 :UInt32;`, mirroring `EmbedImagesResponse`) asserting the generated TS uses the real `getFloat32ListList`/`setFloat32ListList` calls and never falls back to the old panic path.

## Verification

- `cargo test -p hyprstream-rpc-build` — 13/13 pass (including the two new golden tests).
- `cargo clippy -p hyprstream-rpc-build --all-targets` — clean.
- **End-to-end regen**: built `hyprstream-rpc-std` to produce live CGR files, then ran `cargo run -p hyprstream-rpc-build --bin hyprstream-ts-codegen -- --input-dir <hyprstream-rpc-std OUT_DIR> --output-dir <tmp>` against them directly. All 9 service files generated successfully with no panic, including `inference.ts` and `model.ts`. Confirmed the generated output uses the new runtime calls at every `embeddings` call site (standalone parser/builder, and the nested scoped-response read path for `model.ts`'s `infer.embed` variant):
  - `inference.ts`: `embeddings: number[][]`, `_s.getFloat32ListList(0)`, `msg.setFloat32ListList(0, p.embeddings)`, `reader.getFloat32ListList(0)`
  - `model.ts`: same shape plus the nested scoped-response read `_p0.getFloat32ListList(0)` inside the `infer.embed` variant parse.

Closes #758.

### DIFF
--- a/crates/hyprstream-rpc-build/src/ts_codegen/message.rs
+++ b/crates/hyprstream-rpc-build/src/ts_codegen/message.rs
@@ -311,7 +311,15 @@ export class CapnpArena {
    * per-element byte width (ignored for bit-packed bool lists).
    * Returns the byte offset of the first element, or -1 for an empty list.
    */
-  private allocPrimitiveList(ptrIndex: number, count: number, elemSize: number, byteWidth: number): number {
+  /**
+   * Same as `allocPrimitiveList`, but takes the absolute byte offset of the
+   * pointer word to write (rather than deriving it from `ptrIndex` against
+   * the root pointer section). Shared by the root-level list setters and by
+   * `List(List(_))` nested-list setters, whose inner-list pointer lives at a
+   * computed offset inside the allocated outer pointer list, not at a
+   * `ptrIndex` slot.
+   */
+  private allocPrimitiveListAt(ptrOffset: number, count: number, elemSize: number, byteWidth: number): number {
     if (count === 0) return -1;
 
     this.alignAlloc();
@@ -320,7 +328,6 @@ export class CapnpArena {
     this.ensureCapacity(wordLen * WORD_SIZE);
 
     const targetOffset = this.allocOffset;
-    const ptrOffset = this.rootPtrOffset + ptrIndex * WORD_SIZE;
     const offsetWords = (targetOffset - ptrOffset - WORD_SIZE) / WORD_SIZE;
     const lo = ((offsetWords << 2) | 1) >>> 0;
     const hi = ((count << 3) | elemSize) >>> 0;
@@ -329,6 +336,37 @@ export class CapnpArena {
 
     this.allocOffset = targetOffset + wordLen * WORD_SIZE;
     return targetOffset;
+  }
+
+  private allocPrimitiveList(ptrIndex: number, count: number, elemSize: number, byteWidth: number): number {
+    return this.allocPrimitiveListAt(this.rootPtrOffset + ptrIndex * WORD_SIZE, count, elemSize, byteWidth);
+  }
+
+  /**
+   * Allocate the outer pointer-list structure for a `List(List(<inner>))`
+   * field at `ptrIndex`: a plain pointer list (element_size=6) with one
+   * pointer-word slot reserved per outer element. Each slot is later filled
+   * in by `allocPrimitiveListAt` when the corresponding inner list is
+   * written. Returns the byte offset of the first outer-element pointer
+   * slot, or -1 for an empty outer list (leaving the field's pointer null).
+   */
+  private allocNestedListOuter(ptrIndex: number, outerCount: number): number {
+    if (outerCount === 0) return -1;
+
+    this.alignAlloc();
+    this.ensureCapacity(outerCount * WORD_SIZE);
+
+    const outerBase = this.allocOffset;
+    this.allocOffset = outerBase + outerCount * WORD_SIZE;
+
+    const ptrOffset = this.rootPtrOffset + ptrIndex * WORD_SIZE;
+    const offsetWords = (outerBase - ptrOffset - WORD_SIZE) / WORD_SIZE;
+    const lo = ((offsetWords << 2) | 1) >>> 0;
+    const hi = ((outerCount << 3) | 6) >>> 0; // element_size=6: pointer list
+    this.buf.setUint32(ptrOffset, lo, true);
+    this.buf.setUint32(ptrOffset + 4, hi, true);
+
+    return outerBase;
   }
 
   setBoolList(ptrIndex: number, values: boolean[]): void {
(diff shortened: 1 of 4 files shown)


crates/hyprstream-rpc-build/src/ts_codegen/message.rs
```diff
@@ -753,6 +947,132 @@ export class StructBuilder {
     for (let i = 0; i < values.length; i++) this.msg._getBuf().setFloat64(base + i * 8, values[i], true);
   }
 
+  // --- Nested List(List(<primitive>)) setters (see CapnpArena's for docs) ---
+
+  setBoolListList(ptrIndex: number, values: boolean[][]): void {
+    const outerBase = this.allocNestedListOuter(ptrIndex, values.length);
+    if (outerBase < 0) return;
+    const raw = this.msg._getRaw();
```
ewindisch: **Bug (silent wire corruption, latent): stale buffer capture.** `raw` is captured once here, but each loop iteration calls `allocPrimitiveListAt` → `_ensureCapacity`, which on growth reallocates the arena to a *new* `Uint8Array`. After a growth, every subsequent bit write lands in the detached old buffer — the wire ends up with all-false inner lists and no error.

Compare the safe siblings: `CapnpArena.setBoolListList` writes through `this.raw[...]` (re-read per access), and the existing `StructBuilder.setBoolList` captures `raw` only *after* its single alloc. Every other `set*ListList` here re-fetches `this.msg._getBuf()` per write, which is also safe — this method is the one outlier.

Nothing uses `List(List(Bool))` today (only `embeddings :List(List(Float32))`), so nothing breaks yet — but this is exactly the silent-degradation class #725/#728 exist to kill, and initial arena slack is small enough (~1KB) that a real payload would cross the growth threshold. One-line fix: move the `_getRaw()` read inside the loop, after the `allocPrimitiveListAt` call:

```ts
const base = this.allocPrimitiveListAt(outerBase + i * WORD_SIZE, inner.length, 1, 0);
const raw = this.msg._getRaw();
for (let j = 0; j < inner.length; j++) {
  if (inner[j]) raw[base + (j >> 3)] |= 1 << (j & 7);
}
```

ewindisch (commented):
**Second-opinion review (independent trace of the wire arithmetic against the capnp encoding spec).** Verdict: **approve with one required fix** (inline comment below) plus follow-ups.

What checks out:
- Outer list encoding is correct: plain pointer list (`element_size=6`), `lo=(offsetWords<<2)|1` with the offset counted from the end of the pointer word, shell reserved zero-initialized so unwritten slots are null pointers.
- All 22 inner-setter element-size/byte-width pairings are correct; inner data word-aligned with `wordLen=ceil(bytes/8)` padding; absolute offsets survive buffer growth in the alloc path.
- Read path exactly mirrors the field-proven `readPrimitiveList` (signed offset handling, far pointers via `resolvePtr`, bool bit addressing).
- `allocPrimitiveList` → `allocPrimitiveListAt` refactor is a byte-identical delegate; all existing single-level `set*List` call sites unaffected.
- Hard-fail invariant from #725/#728 preserved at all 5 panic sites: `List(List(Text))`, `List(List(Data))`, `List(List(struct))`, and triple nesting all still panic at generation time.

Findings beyond the inline one:
1. **Tests can't catch pointer-arithmetic bugs** — both new tests assert generated *source strings* only (`out.contains(".setFloat32ListList(0,")`); there is no JS runtime harness in `hyprstream-rpc-build`, so an off-by-one anywhere in the ~400 new lines of TS runtime passes CI green. The #725 `getDataList` bug was exactly this class. Filed as a follow-up issue (cross-impl round-trip: Rust `capnp` encode → generated `capnp.ts` decode under node, and reverse).
2. **Low**: `readNestedPrimitiveList`'s composite-outer (`element_size=7`) branch trusts the tag's element count but ignores dataWords/ptrWords, hard-coding a 1-word stride with the pointer at word 0. Unreachable from conformant writers (including this one), but consider validating the tag shape or degrading to `[]`.
3. **Low/informational**: empty inner lists encode as null pointers rather than zero-length list pointers — decodes identically everywhere; only matters if capnp bytes are ever canonicalized/hashed (envelopes sign bytes-as-sent, so no current impact).

ewindisch:
Follow-up from the review: the executable cross-impl round-trip test harness is now filed as #807. The inline `setBoolListList` fix is the only pre-merge ask.

ewindisch:
Pushed c71f50c69 addressing the one pre-merge ask from the review: `StructBuilder.setBoolListList` now re-fetches `_getRaw()` after each inner-list alloc instead of capturing it once before the loop, matching the other `set*ListList` variants. That was the only blocking item — enabling automerge. The executable Rust⇄TS round-trip harness remains tracked as #807.

ewindisch:
✅ **Reviewed (1:7 sweep) — ready.** Encoder/decoder verified symmetric (outer pointer-list elemSize=6, inner primitive lists at computed offsets; all elemSize/byteWidth capnp codes correct; empty inner → null ptr → decodes `[]`). TS `number[][]` emitted via recursive `capnp_to_ts_type`. The prior **setBoolListList stale-buffer ask is addressed** — `StructBuilder.setBoolListList` re-fetches `this.msg._getRaw()` after each inner alloc, and the `CapnpArena` path is safe (`ensureCapacity` reassigns `this.raw`, setter reads per access). Two low decode-tolerance notes (composite-outer elemSize=7 assumes 1 pointer word; wire skew yields `[]`) are defensible for your own encoder and non-blocking. Arming auto-merge.

Issue #441, opened 2026-06-25
Epic: Authoritative key binding — one identity→key model, fail-closed, no silent re-derivation


ewindisch:
## Problem
Key identity is established implicitly and divergently across the system. A principal's key is sometimes *registered*, sometimes *derived-from-CA by the consumer*, sometimes *bound*, sometimes *skipped* — with no single authoritative record every party agrees on. When registration is skipped, consumers silently fall back to a different derivation, and the two disagree. There is no fail-closed invariant that a principal's advertised key == its actual signing key.

## Four failures observed (same disease)
1. **CLI → anonymous** (#438): the wizard registers a user's access-control policy but never binds the user's signing-key pubkey to the identity → the CLI's signed JWT can't be verified → falls back to `anonymous`.
2. **No bring-your-own-key path** (#439): only a TEST env var (`HYPRSTREAM__OAUTH__USER_SIGNING_KEY`) or overwriting a fixed file; no first-class enrollment.
3. **Service key never registered → 'Response signed by unexpected key'** (the live chat failure):
   - `factories.rs:99` WARNs and continues when it can't register the model key (`trust store has no JWT — skipping registerServiceKey service="model"`). **Fail-open.**
   - The TUI then asks PolicyService for the model key; `policy.rs:1368` **derives an 'expected' key from the root CA** (the fallback). The inference service actually signs with its own derived signing key. Derived-expected ≠ actual-signer → `envelope.rs:1509 "Response signed by unexpected key"`.
4. **Trust-store first-writer-wins churn**: a wall of `equal-expiry attestation discarded (first-writer wins)` warnings — services race to register keys; outcome is order-dependent across restarts.

## The invariant (this epic enforces it)
**Every principal — user OR service — has exactly one authoritative key binding, established at enrollment/startup, fail-closed if it can't be established, never silently re-derived by consumers.**

Three principles:
1. **One source of truth per identity→key binding.** A key is authoritative because it is *registered and attested*, never because a consumer re-derived it.
2. **Registration is a hard precondition, not best-effort.** A service that cannot register its signing key MUST NOT come up serving signed responses (every client would mis-verify it). Fail closed with a clear error — not a silent skip that surfaces as a cryptic 'unexpected key' three layers away.
3. **Derive once, deterministically, register that exact key.** The signer registers the verifying key it will actually use; consumers read the registered key; no independent re-derivation.

## Decisions (made)
[...]

PR #774, opened 2026-07-04
fix: multi-service startup resolves secrets dir/keys inconsistently, causing key-mismatch failures (#759)

### LINKED ISSUE #759: bug: multi-service --ipc start fails "Failed to resolve model service key: Response signed by unexpected key" (blocks browser moq e2e)
Discovered while trying to run the browser moq inference-streaming e2e (www-cyberdione-ai#11 child item #13, Playwright `moq-inference` spec against a model-loaded server).

**Symptom:** starting hyprstream with multiple services over `--ipc` on `0.4.0+main.g2a46326` fails before reaching `Services ready:`:

```
Failed to create tui service: Failed to resolve model service key: Response signed by unexpected key
```

This looks like a service-to-service key-resolution mismatch on the IPC/UDS plane — possibly related to the identity/key-binding work in #441 (Epic: Authoritative key binding) or the now-closed #446 (service-to-service IPC callers resolving as anonymous on the AnySigner plane). Needs someone with server-startup/service-trust-store context to reproduce and bisect; I have not root-caused it, just hit it as a hard blocker.

**Impact:** blocks www-cyberdione-ai#11 child item #13 (W4 — the runtime proof for the browser moq subscriber path, #719/#722). W2/W3 are otherwise code-complete and green (tsc clean, prod build green, Playwright smoke suite passes); only the model-loaded e2e is blocked on this.

cc #719, #722, www-cyberdione-ai#11/#13, #441

ewindisch:
## Root cause (confirmed via code-path tracing, not a guess)

Two related divergences in `crates/hyprstream/src/bin/main.rs`'s service-startup code caused multi-service `--ipc` startup to fail with `Failed to resolve model service key: Response signed by unexpected key`:

**1. `secrets_dir` was hand-recomputed instead of using the authoritative resolver.**
The `--ipc` and single-process branches built `secrets_dir` via `dirs::config_dir().join("hyprstream").join("credentials")`, checking only `HYPRSTREAM__SECRETS__PATH` directly, while other call sites in the same startup function use `hyprstream_core::config::HyprConfig::resolve_secrets_dir()`, which loads full config (config file `[secrets] path`, plus env, plus `HYPRSTREAM_INSTANCE` namespacing) via `cfg.secrets.resolve_dir(cfg.config_dir())`. Under a configured `[secrets] path`, `HYPRSTREAM_INSTANCE`, or non-default config-dir resolution, the two computations could point at different directories — a process could load different key material than what bootstrap actually wrote.

*Correction (review finding):* `bootstrap_manager::do_bootstrap` itself does **not** use `resolve_secrets_dir()` — it still computes `config_dir().join("credentials")` directly (`cli/bootstrap_manager.rs`), so a configured `[secrets] path` still diverges bootstrap-vs-runtime after this PR. This PR unifies the two `main.rs` startup branches on the authoritative resolver; the remaining divergent sites (do_bootstrap, `factories::credentials_dir`, unit generation, policy JWT persist) are tracked in #804.

**2. PolicyService's own key resolution was inconsistent between deployment modes — this is the deterministic part.**
- Bootstrap (`do_bootstrap`) writes the root/CA key **flat** at `{secrets_dir}/signing-key` via `load_or_generate_node_signing_key`, and registers `bootstrap_pubkeys["policy"] = root_key.verifying_key()`. PolicyService's identity **is** the CA key — it never gets an independent `{secrets_dir}/policy/signing-key`.
- In **systemd mode**, the old code correctly loaded `load_or_generate_node_signing_key(&secrets_dir)` for every service (each service's systemd unit scopes `secrets_dir` to its own `LoadCredential` dir already).
- In **standalone (non-systemd) `--ipc` mode**, the old code instead called `load_or_generate_service_signing_key(&secrets_dir, "policy")` for PolicyService — which reads/writes `{secrets_dir}/policy/signing-key`, a file bootstrap never populates. On first standalone `--ipc` startup this **silently minted a fresh, unrelated key** for PolicyService. Every peer's trust store still expects `bootstrap_pubkeys["policy"]` (the root key's verifying key), so PolicyService's real responses fail verification — deterministically, every time, not intermittently.

This is a sibling gap to #441 in the same area: #441's "kill the CA-derivation fallback in `resolve_service_key`" and "fail-closed `registerServiceKey`" fixes are already live on `main` and are correct/working — this bug is a *different* divergence (PolicyService's own identity resolution at process-startup time), not the CA-derivation-fallback failure mode #441 described.

## Fix

- Both the `--ipc` and single-process branches now call `HyprConfig::resolve_secrets_dir()`.
- New `resolve_service_signing_key(secrets_dir, service_name, systemd_mode)` centralizes the policy-vs-non-policy, systemd-vs-standalone logic: `systemd_mode || service_name == "policy"` → the flat root/CA key; otherwise → the per-service subdirectory key.

## Verification

**Live end-to-end reproduction, not just static analysis.** In an isolated environment (dedicated `HOME`/`XDG_*`/`XDG_RUNTIME_DIR`, matching the isolation pattern `scripts/aaa-e2e.sh` uses), I:
1. Ran `hyprstream wizard --non-interactive` to bootstrap credentials.
2. Configured `[services] startup = ["policy", "discovery", "registry", "model", "tui"]`.
3. Ran `hyprstream service start --daemon` — this forces the standalone `ProcessSpawner` (bypassing systemd even though it was available in the sandbox), spawning each service as its own subprocess with `service start <name> --foreground --ipc`, i.e. exactly the multi-service `--ipc` deployment mode the issue names.

With the fix applied, all 5 services — including `tui`, whose `create_tui_service` factory calls `build_chat_vfs_namespace`, the exact function named in the issue's error (`crates/hyprstream/src/tui/vfs.rs`) — reached `Services ready:` cleanly. The daemon log shows the `resolveServiceKey` RPC (`resource=policy:ResolveServiceKey action=query`) being authorized and answered with no "unexpected key" failure anywhere in the run.

I also attempted the same repro against an unmodified `origin/main` baseline to positively confirm the failure disappears *because of* the fix (not just that it never manifested in this sandbox), but the baseline build was killed by a shared-host disk-quota exhaustion (tmpfs quota, unrelated to compilation correctness) before it could run — an environment limitation, not a finding about the code. Given:
- the file-path divergence in the root cause is deterministic and traceable to the byte level (two different, provably distinct on-disk paths for the same logical key), and
- the fixed code's live run exercises precisely the previously-diverging path and succeeds,

confidence remains high despite not completing the A/B baseline comparison.

Static verification:
- `cargo build -p hyprstream` (workspace build, ~9 min cold) — clean.
- `cargo clippy -p hyprstream --all-targets` — clean, no warnings.
- `cargo test -p hyprstream --lib auth::` — 261 passed, 0 failed (includes the 4 new regression tests below).

## Tests added (`crates/hyprstream/src/auth/identity_store.rs`)
- `test_resolve_service_signing_key_policy_matches_bootstrap_root_key` — standalone mode, "policy" must equal the root key, and must NOT create an independent `policy/signing-key` file.
- `test_resolve_service_signing_key_policy_matches_root_key_in_systemd_mode`
- `test_resolve_service_signing_key_non_policy_uses_independent_subdir_key` — non-policy services keep their own independent key (no regression for the correct existing behavior).
- `test_resolve_service_signing_key_is_stable_across_calls`

Closes #759.

### DIFF
--- a/crates/hyprstream/src/auth/identity_store.rs
+++ b/crates/hyprstream/src/auth/identity_store.rs
@@ -405,6 +405,46 @@ pub fn load_or_generate_node_signing_key(secrets_dir: &std::path::Path) -> Resul
     Ok(key)
 }
 
+/// Resolve the Ed25519 signing key a service process should use to sign its
+/// own RPC responses in `--ipc` (multi-process) deployment.
+///
+/// # PolicyService is special
+///
+/// PolicyService's identity IS the root/CA key, not an independent
+/// per-service key: `bootstrap_manager::do_bootstrap` registers
+/// `bootstrap_pubkeys["policy"] = root_key.verifying_key()` (the same value
+/// persisted flat at `secrets_dir/signing-key`), rather than generating a
+/// `secrets_dir/policy/signing-key` like every other service gets.
+///
+/// Resolving "policy" via the generic per-service path
+/// (`load_or_generate_service_signing_key`) reads/generates a *different*
+/// file (`{secrets_dir}/policy/signing-key`) that bootstrap never populated —
+/// so on first `--ipc` startup it silently mints a fresh, unrelated key. Every
+/// peer that resolves "policy" from the trust store (seeded from
+/// `bootstrap-pubkeys`) still expects `root_key.verifying_key()`, so
+/// PolicyService's actual responses fail verification with "Response signed
+/// by unexpected key" (#759) — a sibling gap to the CA-derivation fallback
+/// #441 already removed from `resolve_service_key`.
+///
+/// - `systemd_mode`: each service's per-unit `LoadCredential` directory
+///   already scopes `secrets_dir` to that one service, so the flat top-level
+///   file is correct for every service, policy included.
+/// - Standalone (non-systemd) `--ipc`: `secrets_dir` is shared across every
+///   spawned service process, so non-policy services use a `{service_name}/`
+///   subdirectory to avoid collisions — but policy must still resolve to the
+///   flat root/CA key, matching what bootstrap registered.
+pub fn resolve_service_signing_key(
+    secrets_dir: &std::path::Path,
+    service_name: &str,
+    systemd_mode: bool,
+) -> Result<SigningKey> {
+    if systemd_mode || service_name == "policy" {
+        load_or_generate_node_signing_key(secrets_dir)
+    } else {
+        load_or_generate_service_signing_key(secrets_dir, service_name)
+    }
+}
+
 /// Decode the `exp` claim from a JWT without signature verification.
 /// Returns `None` on any parse error (invalid base64, not JSON, missing exp).
 pub fn decode_jwt_exp_raw(jwt: &str) -> Option<i64> {
@@ -954,4 +994,70 @@ mod tests {
         assert_ne!(h1, h2, "renewed cert should have different DER due to new timestamps");
     }
 
+    // ── resolve_service_signing_key (#759 regression) ────────────────────────
+
+    #[test]
+    fn test_resolve_service_signing_key_policy_matches_bootstrap_root_key() {
+        // Reproduces the bootstrap flow in `bootstrap_manager::do_bootstrap`:
+        // the root key is written flat as "signing-key" and its verifying key
+        // is what gets registered as `bootstrap_pubkeys["policy"]` — PolicyService
+        // never gets an independent `policy/signing-key` file.
+        let dir = TempDir::new().unwrap();
+        let root_key = load_or_generate_node_signing_key(dir.path()).unwrap();
+
+        // Standalone (non-systemd) `--ipc` startup, resolving policy's own key —
+        // this must equal the root key bootstrap registered, not a freshly
+        // minted independent key.
+        let resolved = resolve_service_signing_key(dir.path(), "policy", false).unwrap();
+        assert_eq!(
+            resolved.to_bytes(), root_key.to_bytes(),
+            "policy must resolve to the flat root/CA key, matching bootstrap_pubkeys[\"policy\"]"
+        );
+
+        // No independent `policy/signing-key` file should have been created.
+        assert!(
+            read_secret(&dir.path().join("policy"), "signing-key").unwrap().is_none(),
+            "resolving policy's key must not mint an independent per-service key file"
+        );
+    }
+
+    #[test]
+    fn test_resolve_service_signing_key_policy_matches_root_key_in_systemd_mode() {
+        let dir = TempDir::new().unwrap();
+        let root_key = load_or_generate_node_signing_key(dir.path()).unwrap();
+        let resolved = resolve_service_signing_key(dir.path(), "policy", true).unwrap();
+        assert_eq!(resolved.to_bytes(), root_key.to_bytes());
+    }
+
+    #[test]
+    fn test_resolve_service_signing_key_non_policy_uses_independent_subdir_key() {
+        let dir = TempDir::new().unwrap();
+        let root_key = load_or_generate_node_signing_key(dir.path()).unwrap();
+
+        let model_key = resolve_service_signing_key(dir.path(), "model", false).unwrap();
+        assert_ne!(
+            model_key.to_bytes(), root_key.to_bytes(),
+            "non-policy services must keep their own independent key, distinct from the root/CA key"
+        );
+
+        // Independent key file lives under the service subdirectory.
+        let from_disk = read_secret(&dir.path().join("model"), "signing-key").unwrap().unwrap();
+        assert_eq!(from_disk, model_key.to_bytes().to_vec());
+    }
+
+    #[test]
+    fn test_resolve_service_signing_key_is_stable_across_calls() {
+        // Simulates repeated resolution (e.g. across a restart) — the same
+        // key must come back every time, for both policy and non-policy.
+        let dir = TempDir::new().unwrap();
+        let _root_key = load_or_generate_node_signing_key(dir.path()).unwrap();
+
+        let policy_1 = resolve_service_signing_key(dir.path(), "policy", false).unwrap();
+        let policy_2 = resolve_service_signing_key(dir.path(), "policy", false).unwrap();
+        assert_eq!(policy_1.to_bytes(), policy_2.to_bytes());
+
+        let model_1 = resolve_service_signing_key(dir.path(), "model", false).unwrap();
+        let model_2 = resolve_service_signing_key(dir.path(), "model", false).unwrap();
+        assert_eq!(model_1.to_bytes(), model_2.to_bytes());
+    }
 }
--- a/crates/hyprstream/src/bin/main.rs
+++ b/crates/hyprstream/src/bin/main.rs
@@ -1979,17 +1979,19 @@ fn main() -> Result<()> {
 
                                 if ipc {
                                     // Multi-process mode: each service gets its own independent key.
-                                    let secrets_dir = std::path::PathBuf::from(
-                                        std::env::var("HYPRSTREAM__SECRETS__PATH")
-                                            .unwrap_or_else(|_| {
-                                                dirs::config_dir()
-                                                    .map(|d| d.join("hyprstream").join("credentials"))
-                                                    .unwrap_or_default()
-                                                    .to_str()
-                                                    .unwrap_or("")
-                                                    .to_owned()
-                                            }),
-                                    );
+                                    //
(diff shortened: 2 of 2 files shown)


crates/hyprstream/src/auth/identity_store.rs
```diff
@@ -405,6 +405,46 @@ pub fn load_or_generate_node_signing_key(secrets_dir: &std::path::Path) -> Resul
...
+/// - Standalone (non-systemd) `--ipc`: `secrets_dir` is shared across every
+///   spawned service process, so non-policy services use a `{service_name}/`
+///   subdirectory to avoid collisions — but policy must still resolve to the
+///   flat root/CA key, matching what bootstrap registered.
+pub fn resolve_service_signing_key(
+    secrets_dir: &std::path::Path,
+    service_name: &str,
+    systemd_mode: bool,
+) -> Result<SigningKey> {
+    if systemd_mode || service_name == "policy" {
```
ewindisch, line 441: **Sound-but-fragile — the two halves of this condition are true for different reasons, and one of them is an unverified assumption.** The `service_name == "policy"` half is correct and well-reasoned: it matches what bootstrap provably registers (`bootstrap_pubkeys["policy"] = root verifying key`, flat `signing-key` write).

The `systemd_mode` half assumes the caller's `secrets_dir` is a *per-unit-scoped* credentials dir (`LoadCredential` → `%d`). That holds only for units generated by `hyprstream service install`. If the flag is derived from `HYPRSTREAM__SECRETS__PATH` being set (as the main.rs call site does) and an operator points that env var at a **shared** secrets dir — hand-written unit, container, or the k8s Helm chart that #783 is about to build (a chart would naturally set exactly this env var) — then *every* service loads `{shared}/signing-key`, i.e. the **root/CA private key**, as its own signing key. Failure mode: root key handed to every service process, plus their responses mismatch `bootstrap_pubkeys[svc]` — reproducing the exact "Response signed by unexpected key" symptom this PR fixes.

Not introduced by this PR (it faithfully preserves the pre-existing behavior), but this helper enshrines it in a documented public API. Suggest at minimum a rustdoc warning that `systemd_mode=true` is only valid for per-service-scoped dirs; the real fix (deployment-profile enum + honest detection, e.g. systemd's own `CREDENTIALS_DIRECTORY` env) is being filed as a follow-up since #783/K1a plans to consume this exact model.

ewindisch (commented):
**Second-opinion security review.** Verdict: **approve with changes** — the fix is correct for the reported failure, the root-cause analysis of the policy-key mismatch is verifiably accurate, and this does **not** reintroduce a #441-class hole (verification-side resolution is untouched and still registered-or-error; this PR only changes which *private* key a process loads to sign its own responses — it never derives an *expected/trusted* key at verification time; `service_name` is operator CLI input and `validate_service_name` blocks traversal).

Two issues found beyond the inline comments:

1. **The PR's centralization claim is half-false: `do_bootstrap` — the write side — still diverges.** `cli/bootstrap_manager.rs` (`do_bootstrap`, plus `credentials_dir()`/`is_bootstrap_needed()` there) resolves via `HyprConfig::load().config_dir().join("credentials")`, ignoring `[secrets] path`. `register_local_identity` does use `resolve_secrets_dir()`, but the PR description's claim that `do_bootstrap` does is inaccurate. Concrete failure: operator sets `[secrets] path = "/srv/hyprstream/creds"` → wizard writes keys to `~/.config/hyprstream/credentials`, the fixed main.rs reads `/srv/hyprstream/creds`, `load_or_generate_node_signing_key` silently mints a *fresh root key* there, and the exact #759 symptom returns — now with a spurious second "root" key on disk. Please either fix `do_bootstrap`'s resolution in this PR or link the follow-up issue (being filed) and correct the description.
2. **No startup consistency check (missed cheap defense-in-depth).** In the `--ipc` branch, `load_ca_verifying_key`/`load_bootstrap_pubkeys` failures are swallowed (`if let Ok(...)`) and both key loaders generate-if-missing. Asserting `own_key.verifying_key() == bootstrap_pubkeys[name]` after load and failing loudly on mismatch would convert every residual divergence (including item 1 and the inline `systemd_mode` hazard) into an early, clear error — and would have caught #759 itself at startup instead of as a cryptic verify failure three layers away. That's the actual #441 lesson applied here.

Noted for follow-up issues (pre-existing, not blockers): PolicyService's identity being the *hot* root/CA key (root compromise = policy-process compromise; the new rustdoc normalizes this without flagging it), and `auth/policy.rs` still *deriving* service pubkeys from the root key for `cnf.jwk` in issued/refreshed service tokens — which cannot match the independent per-service keys bootstrap actually generates. Test coverage note: tests 1 and 3 are meaningful (pin the #759 contract and the non-regression); the secrets-dir-resolution half of the fix and the riskiest branch (`systemd_mode=true`, non-policy → node key) have no tests.

crates/hyprstream/src/bin/main.rs
```diff
@@ -2017,12 +2019,12 @@ fn main() -> Result<()> {
                                     }
 
                                     // C3 fix: systemd uses flat %d/signing-key, standalone uses subdirectory.
+                                    // #759: "policy" always resolves to the flat root/CA key regardless of
+                                    // deployment mode — see `resolve_service_signing_key` for why.
                                     let systemd_mode = std::env::var("HYPRSTREAM__SECRETS__PATH").is_ok();
```
ewindisch, line 2024: This detection heuristic is the fragile link: `HYPRSTREAM__SECRETS__PATH` is simultaneously (a) the `%d` marker emitted by generated systemd units and (b) the general-purpose "relocate the secrets dir" config-override knob (config env layering; `factories.rs::credentials_dir()` treats it as plain relocation). Setting it for reason (b) with a shared dir silently selects the flat-root-key branch for every service — see the comment on `resolve_service_signing_key`. systemd exports `CREDENTIALS_DIRECTORY` itself, which would be an honest signal; alternatively make the mode explicit config. Fine to defer to the follow-up issue, but worth a `// NOTE:` here so the next reader doesn't take the variable name at face value.

ewindisch:
Follow-up issues from the review are now filed: #804 (remaining secrets-dir resolution divergence, incl. `do_bootstrap` — review finding 1), #805 (`systemd_mode` bool → explicit `SecretsProfile`, incl. the detection-heuristic hazard from the inline comments), #806 (`cnf.jwk` derived-from-root residue). Related from the same audit: #802 (install-time CA-key fallback in unit generation), #803 (JWT rotation persistence under systemd). With #804 linked, the remaining pre-merge asks are: correct the PR description's `do_bootstrap` claim, and (recommended) the `bootstrap_pubkeys[name]` startup consistency check.

ewindisch:
Description corrected per the review: the claim that `do_bootstrap` uses `resolve_secrets_dir()` was inaccurate — it still computes `config_dir().join("credentials")` directly; the residual divergent sites are tracked in #804. That was the required pre-merge ask, so enabling automerge. The recommended `own_key.verifying_key() == bootstrap_pubkeys[name]` startup consistency check remains a good hardening follow-up — folding it into #804/#805 scope rather than blocking this fix.

ewindisch:
## 🟡 Review (1:7 coordination sweep) — minor issues; fix narrows #759 but doesn't close it

**1. (med) The unification is partial.** The fix routes only the two `main.rs` startup branches through `resolve_secrets_dir()`; `do_bootstrap`, `factories::credentials_dir`, unit generation, and policy JWT persist still recompute the dir independently (self-acknowledged, tracked #804). Under a configured `[secrets] path` or `HYPRSTREAM_INSTANCE`, bootstrap still writes credentials to a different directory than runtime reads — the #759 class of bug is narrowed, not eliminated. Worth stating clearly in the PR that #804 is required to actually close #759.

**2. (med) `HYPRSTREAM__SECRETS__PATH` still doubles as both "secrets location" and "systemd mode" signal** (`main.rs` ~2021), now more exposed since the authoritative resolver honors other path sources. An operator setting `HYPRSTREAM__SECRETS__PATH` in standalone `--ipc` flips every service onto the flat root/CA key; non-policy services then sign with the root key while the trust store expects per-service bootstrap keys — the exact "unexpected key" failure this PR fixes, via a different door. Decouple the mode signal from the path var.

**3. (low) First-boot key generation isn't cross-process atomic** — `load_or_generate_node_signing_key` (`identity_store.rs:377`) does read→generate→`write_secret` (last rename wins), not `write_secret_exclusive` (O_EXCL). Two concurrent unbootstrapped starts keep divergent in-memory root keys until restart. Pre-existing; the new "stable across calls" test only covers sequential calls.

Items 1–2 should be resolved (or explicitly deferred to #804 with the PR description updated). Flagging the owning session.

PR #775, opened 2026-07-04
superseded: route user-management commands through PolicyService RPC (#449)

### LINKED ISSUE #449: bug: hyprstream user CLI conflicts with live credential store; route through correct user-management plane
## Problem

`hyprstream user list` and related user/key management commands currently open the credential store directly from the CLI process. With the RocksDB-backed store, a live service can hold the exclusive writer lock and the CLI fails with:

```text
IO error: While lock file: /path/to/rocksdb/LOCK: Resource temporarily unavailable
```

## Corrected design direction

The fix should not move user management into `PolicyService`. That collapses two separate control planes:

- **Runtime user management** belongs to `OAuthService` / `oauth.capnp`, with authorization enforced on the runtime RPC path.
- **Runtime policy management** remains in `PolicyService`, scoped to policy/role operations.
- **Out-of-band operator/admin management** remains local-only for bootstrap, break-glass, and repair. It must not depend on the runtime policy plane it may need to repair.

The runtime user-management API must also be exposed through the Plan 9/VFS surface, not only through direct RPC or CLI wrappers.

## Expected behavior

- CLI user/key commands no longer open RocksDB directly when targeting a running deployment.
- Runtime user/key operations go through authorized `OAuthService` RPCs.
- The same user/key operations are available through Plan 9/VFS paths with equivalent authorization.
- Local out-of-band admin/bootstrap commands remain available for operator recovery and initial provisioning.
- The implementation avoids introducing another multi-process RocksDB writer conflict between policy/oauth services.

## Follow-ups

- Replace PR #775's PolicyService approach with the corrected split above.
- Add/complete OAuthService user-management RPC handlers and VFS annotations/dispatch.
- Preserve/document the local operator admin plane separately from runtime user management.
- Document OIDC/SCIM IdP integration expectations separately from this storage-lock bug.

## Workaround

Temporarily stop the service holding the credential-store writer lock, run the local admin command, then restart the service. This is only an operator workaround, not the runtime API design.

ewindisch:
## Problem

`hyprstream user list` (and register/remove/keys list/import/remove) opened the RocksDB credential store directly from the CLI process. When PolicyService is running, it holds an exclusive RocksDB writer lock, so the CLI failed with:

```
IO error: While lock file: /path/to/rocksdb/LOCK: Resource temporarily unavailable
```

## Fix

Added `listUsers` / `registerUser` / `removeUser` / `listUserKeys` / `addUserKey` / `removeUserKey` to `policy.capnp` (additive ordinals: request `@22`-`@27`, response `@23`-`@28`), implemented the handlers on `PolicyService` backed by the existing `UserStore` trait, and rewired every CLI `user`/`user keys` subcommand (`crates/hyprstream/src/cli/user_handlers.rs`) to go through `PolicyClient` instead of opening `RocksDbUserStore` directly. The `AddUserKey` RPC takes only the raw 32-byte Ed25519 public key — the CLI still does all SSH-wire-format parsing client-side; `PolicyService` never parses key wire formats.

The service factory (`create_policy_service`) now wires a configured user store (RocksDB or Valkey, per `CredentialsConfig`) into `PolicyService` via a new `with_user_store` builder + `crate::auth::open_configured_user_store` helper. This is **non-fatal**: if the store can't be opened, the affected RPCs return `NOT_CONFIGURED` instead of panicking or taking down the service.

**Known cross-process caveat (documented in `open_configured_user_store`'s rustdoc, not fixed here):** `crate::services::oauth` already opens its own `RocksDbUserStore` at the same `credentials_dir` for OAuth/SCIM. RocksDB allows only one open writer per directory — including across independent OS processes, which is exactly how `hyprstream service install` deploys each service. If `policy` and `oauth` are both deployed as separate systemd units with the default RocksDB backend, whichever starts first (today: `policy`, since `oauth` depends on it) gets the writer; the other's user-store-backed RPCs degrade to `NOT_CONFIGURED` rather than crashing. This is a pre-existing multi-process RocksDB-sharing limitation, not introduced by this PR, and is out of scope for #449's ask (which is specifically about the CLI-vs-service conflict).

## Test coverage

Three new tests in `crates/hyprstream/src/services/policy.rs::user_management_tests`:
- `list_users_returns_not_configured_without_a_store` — a `PolicyService` with no store wired returns `NOT_CONFIGURED`, not a panic/error.
- `register_list_and_remove_user_round_trip` — register two users via the RPC handler, confirm they list (sorted), remove one, confirm removing again returns `false` (not an error), confirm the list reflects the removal.
- `add_list_and_remove_user_key_round_trip` — add a key, reject wrong-sized key material with `INVALID_INPUT` (not a panic), list it back, remove it, confirm it's gone.

All three build a real `PolicyService` + real `RocksDbUserStore` in a fresh `TempDir` — the same store type the CLI used to open directly, now reached exclusively through these handlers.

## Verification
- `cargo build -p hyprstream` — clean
- `cargo test -p hyprstream` (policy user-management tests) — 3/3 pass
- `cargo clippy -p hyprstream --lib --all-targets` — clean

## Confirmation
`users.toml.age` / `LocalKeyStore` are **not referenced anywhere** in this diff (verified via `grep -rn "users.toml.age|LocalKeyStore"` over every changed file) — this fix is entirely scoped to the separate RocksDB/Valkey-backed `UserStore` trait, per #449's actual ask.

Closes #449.

### DIFF
--- a/crates/hyprstream-rpc-std/schema/policy.capnp
+++ b/crates/hyprstream-rpc-std/schema/policy.capnp
@@ -92,6 +92,26 @@ struct PolicyRequest {
     # Requires 'exchange' permission on 'policy:exchange-wit' in Casbin policy.
     exchangeWit @21 :ExchangeWit
       $scope(manage) $mcpDescription("Exchange the caller's envelope WIT for an OAuth at+jwt; identity from signed envelope");
+
+    # List all registered usernames (#449 — moves `hyprstream user list` off
+    # direct RocksDB access, which conflicted with a running service holding
+    # the credential store's exclusive lock).
+    listUsers @22 :Void $scope(query) $mcpDescription("List all registered usernames");
+
+    # Register a new user account (#449).
+    registerUser @23 :RegisterUser $scope(manage) $mcpDescription("Register a new user account");
+
+    # Remove a registered user and all their keys (#449).
+    removeUser @24 :RemoveUser $scope(manage) $mcpDescription("Remove a registered user and all their keys");
+
+    # List public keys registered for a user (#449).
+    listUserKeys @25 :ListUserKeys $scope(query) $mcpDescription("List public keys registered for a user");
+
+    # Add a public key for a user (#449).
+    addUserKey @26 :AddUserKey $scope(manage) $mcpDescription("Add a public key for a user");
+
+    # Remove a public key from a user by fingerprint (#449).
+    removeUserKey @27 :RemoveUserKey $scope(manage) $mcpDescription("Remove a public key from a user by fingerprint");
   }
 }
 
@@ -168,6 +188,42 @@ struct GetDiff {
   gitRef @0 :Text $optional;
 }
 
+# Register a new user account (#449)
+struct RegisterUser {
+  # Username to register
+  username @0 :Text;
+}
+
+# Remove a registered user and all their keys (#449)
+struct RemoveUser {
+  # Username to remove
+  username @0 :Text;
+}
+
+# List public keys registered for a user (#449)
+struct ListUserKeys {
+  # Username to list keys for
+  username @0 :Text;
+}
+
+# Add a public key for a user (#449).
+# The caller decodes SSH-wire / raw key formats client-side and submits the
+# raw 32-byte Ed25519 public key — PolicyService never parses key wire formats.
+struct AddUserKey {
+  username @0 :Text;
+  # Raw Ed25519 public key (32 bytes)
+  pubkeyRaw @1 :Data;
+  # Optional label (e.g. "laptop", "yubikey")
+  label @2 :Text $optional;
+}
+
+# Remove a public key from a user by fingerprint (#449)
+struct RemoveUserKey {
+  username @0 :Text;
+  # Key fingerprint, e.g. "SHA256:...."
+  fingerprint @1 :Text;
+}
+
 # Unified policy response (covers both check and token issuance)
 struct PolicyResponse {
   # Request ID this response corresponds to
@@ -242,6 +298,24 @@ struct PolicyResponse {
 
     # at+jwt from exchangeWit
     exchangeWitResult @22 :TokenInfo;
+
+    # Registered usernames (for listUsers)
+    listUsersResult @23 :UserList;
+
+    # Subject UUID of the newly-registered user (for registerUser)
+    registerUserResult @24 :Text;
+
+    # Whether the user existed and was removed (for removeUser)
+    removeUserResult @25 :Bool;
+
+    # Public keys registered for a user (for listUserKeys)
+    listUserKeysResult @26 :UserKeyList;
+
+    # Fingerprint of the newly-added key (for addUserKey)
+    addUserKeyResult @27 :Text;
+
+    # Whether the key existed and was removed (for removeUserKey)
+    removeUserKeyResult @28 :Bool;
   }
 }
 
@@ -400,4 +474,24 @@ struct ExchangeWit {
 
   # TTL override in seconds. Server clamps to configured [min, max].
   ttl @2 :Opt.OptionUint32;
+}
+
+# List of registered usernames (#449)
+struct UserList {
+  usernames @0 :List(Text);
+}
+
+# A single registered public key entry (#449)
+struct UserKeyEntry {
+  fingerprint @0 :Text;
+  # Empty means no label was set
+  label @1 :Text $optional;
+  createdAt @2 :Int64;
+  # None if the key has never been used to authenticate
+  lastUsedAt @3 :Opt.OptionInt64;
+}
+
+# List of public keys registered for a user (#449)
+struct UserKeyList {
+  keys @0 :List(UserKeyEntry);
 }
--- a/crates/hyprstream/src/cli/user_handlers.rs
+++ b/crates/hyprstream/src/cli/user_handlers.rs
@@ -1,19 +1,24 @@
 //! Handlers for `hyprstream user` CLI subcommands.
+//!
+//! These commands go through `PolicyService` RPC (`listUsers`/`registerUser`/
+//! `removeUser`/`listUserKeys`/`addUserKey`/`removeUserKey`, #449) instead of
+//! opening the RocksDB credential store directly — direct access from a
+//! second process conflicted with any already-running service holding the
+//! store's exclusive writer lock (`IO error: While lock file: .../LOCK:
+//! Resource temporarily unavailable`).
 // CLI handlers intentionally print to stdout/stderr for user interaction
 #![allow(clippy::print_stdout, clippy::print_stderr)]
 
 use anyhow::{Context, Result};
 use base64::{engine::general_purpose::STANDARD, Engine};
-use ed25519_dalek::VerifyingKey;
+use ed25519_dalek::{SigningKey, VerifyingKey};
 use std::io::{Read, Write};
-use std::path::Path;
 
-use crate::auth::{RocksDbUserStore, UserStore};
 use crate::cli::commands::UserKeysImportFormat;
-
-fn open_store(credentials_dir: &Path) -> Result<RocksDbUserStore> {
-    RocksDbUserStore::open(credentials_dir).context("Failed to open credential store")
-}
+use crate::cli::policy_handlers::create_policy_client;
+use crate::services::generated::policy_client::{
+    AddUserKey, RegisterUser, RemoveUser, RemoveUserKey, ListUserKeys,
+};
 
 /// Resolve key material: from -f <path>, -f -, or inline data.
 fn resolve_key_input(file: Option<&str>, data: Option<&str>) -> Result<String> {
@@ -35,6 +40,9 @@ fn resolve_key_input(file: Option<&str>, data: Option<&str>) -> Result<String> {
 ///
 /// Accepts: `ssh-ed25519 <base64> [comment]`
 /// Wire format: u32be(11) "ssh-ed25519" u32be(32) <32-byte-key> = 51 bytes total.
+///
+/// This is pure client-side parsing — no store access — so it stays local to
+/// the CLI; only the decoded raw key bytes are sent to PolicyService.
 fn parse_ssh_ed25519(line: &str) -> Result<VerifyingKey> {
     let line = line.trim();
     let b64_part = if let Some(rest) = line.strip_prefix("ssh-ed25519 ") {
@@ -71,17 +79,24 @@ fn parse_ssh_ed25519(line: &str) -> Result<VerifyingKey> {
 }
 
 /// Handle `user register <username>`
-pub async fn handle_user_register(credentials_dir: &Path, username: &str) -> Result<()> {
-    let store = open_store(credentials_dir)?;
-    store.register(username).await.context("Failed to register user")?;
+pub async fn handle_user_register(signing_key: &SigningKey, username: &str) -> Result<()> {
+    let client = create_policy_client(signing_key)?;
+    client
+        .register_user(&RegisterUser { username: username.to_owned() })
+        .await
+        .context("Failed to register user via PolicyService. Are services running?")?;
     println!("Registered user '{username}'");
     Ok(())
 }
 
 /// Handle `user list`
-pub async fn handle_user_list(credentials_dir: &Path) -> Result<()> {
-    let store = open_store(credentials_dir)?;
-    let mut users = store.list_users().await;
+pub async fn handle_user_list(signing_key: &SigningKey) -> Result<()> {
+    let client = create_policy_client(signing_key)?;
+    let mut users = client
+        .list_users()
+        .await
+        .context("Failed to list users via PolicyService. Are services running?")?
+        .usernames;
     if users.is_empty() {
         println!("No users registered.");
     } else {
@@ -95,7 +110,7 @@ pub async fn handle_user_list(credentials_dir: &Path) -> Result<()> {
 
 /// Handle `user remove <username> [--force]`
 pub async fn handle_user_remove(
-    credentials_dir: &Path,
+    signing_key: &SigningKey,
     username: &str,
     force: bool,
 ) -> Result<()> {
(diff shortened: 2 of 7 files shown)


crates/hyprstream/src/services/factories.rs
```diff
@@ -487,6 +487,21 @@ fn create_policy_service(ctx: &ServiceContext) -> anyhow::Result<Box<dyn Spawnab
     // access token is rejected by both the HTTP path and the RPC auth check.
     let _ = SHARED_JTI_BLOCKLIST.set(policy_service.jti_blocklist_arc());
 
+    // Wire the user credential store (#449) so `listUsers`/`registerUser`/
+    // `removeUser`/`listUserKeys`/`addUserKey`/`removeUserKey` work without
+    // the CLI opening RocksDB directly. Non-fatal if it can't be opened (see
+    // `crate::auth::open_configured_user_store` for the cross-process caveat).
+    let creds_dir = credentials_dir();
+    let user_store = tokio::task::block_in_place(|| {
```
ewindisch, line 495: Blocking: this open contends with `services/oauth/mod.rs:476`, which opens `RocksDbUserStore` at the same `credentials_dir` for OAuth login/SCIM/device flows. RocksDB allows one writer per directory — including two opens inside the same process (RocksDB keeps a per-process lock table, so the second in-process open fails just like a cross-process one). Since `oauth` has `depends_on = ["policy"]`, policy always opens first: in both the single-process daemon and the systemd multi-unit layout, oauth's user store now degrades to "not configured" whenever both services run, breaking OAuth login. That contention is introduced by this PR — before it, oauth was the only steady-state opener. Prefer a process-global shared store (the `global_es256_key_store` pattern) handed to both services, or make oauth consume user data through these new policy RPCs so there is exactly one owner.

crates/hyprstream-rpc-std/schema/policy.capnp
```diff
@@ -92,6 +92,26 @@ struct PolicyRequest {
     # Requires 'exchange' permission on 'policy:exchange-wit' in Casbin policy.
     exchangeWit @21 :ExchangeWit
       $scope(manage) $mcpDescription("Exchange the caller's envelope WIT for an OAuth at+jwt; identity from signed envelope");
+
+    # List all registered usernames (#449 — moves `hyprstream user list` off
+    # direct RocksDB access, which conflicted with a running service holding
+    # the credential store's exclusive lock).
+    listUsers @22 :Void $scope(query) $mcpDescription("List all registered usernames");
```
ewindisch, line 99: Blocking: `$scope(query)` puts this behind the same grant used for service key resolution. `SERVICE_BASE_POLICIES` gives `policy:* / query` to nine service identities (oauth, registry, model, worker, oai, tui, discovery, notification, metrics), all of which can now enumerate every registered username. User management is privileged — use `$scope(manage)` like the mutating methods, or a dedicated scope (e.g. `userAdmin`) so `policy:*` wildcard grants don't silently absorb user administration. Same applies to `listUserKeys` below (fingerprints, labels, last-used timestamps).

crates/hyprstream-rpc-std/schema/policy.capnp
```diff
@@ -92,6 +92,26 @@ struct PolicyRequest {
...
+
+    # Register a new user account (#449).
+    registerUser @23 :RegisterUser $scope(manage) $mcpDescription("Register a new user account");
+
+    # Remove a registered user and all their keys (#449).
+    removeUser @24 :RemoveUser $scope(manage) $mcpDescription("Remove a registered user and all their keys");
+
+    # List public keys registered for a user (#449).
+    listUserKeys @25 :ListUserKeys $scope(query) $mcpDescription("List public keys registered for a user");
```
ewindisch, line 108: Same scope concern as `listUsers`: key fingerprints, labels, and last-used timestamps are audit-sensitive metadata; `query` exposes them to every service holding the `policy:* / query` key-resolution grant.

crates/hyprstream/src/cli/user_handlers.rs
```diff
@@ -71,17 +79,24 @@ fn parse_ssh_ed25519(line: &str) -> Result<VerifyingKey> {
 }
 
 /// Handle `user register <username>`
-pub async fn handle_user_register(credentials_dir: &Path, username: &str) -> Result<()> {
-    let store = open_store(credentials_dir)?;
-    store.register(username).await.context("Failed to register user")?;
+pub async fn handle_user_register(signing_key: &SigningKey, username: &str) -> Result<()> {
+    let client = create_policy_client(signing_key)?;
```
ewindisch, line 83: Note (non-blocking): this path assumes PolicyService signs with the flat root key (`create_policy_client`'s bootstrap assumption). That is exactly what #759 reports as broken in standalone multi-service startup — until PR #774 lands, `hyprstream user *` will fail envelope verification in those modes. Worth merging #774 first and re-testing this command in standalone `--ipc` mode.

ewindisch (commented):
**Verdict: Blocking** — the CLI-side rework is correct and cleanly done, but two findings need to be addressed before merge: a startup lock collision that regresses OAuth login in default deployments, and read-scope selection on the user-enumeration RPCs that is broader than user management should be.

**What checks out**

- `users.toml.age` / `LocalKeyStore` are untouched anywhere in the diff; everything is scoped to the RocksDB/Valkey `UserStore` trait, as required.
- The CLI no longer opens RocksDB at all: `open_store()` is deleted from `user_handlers.rs` and every subcommand goes through `PolicyClient`. The only remaining direct opens in the CLI crate are `cli/bootstrap_manager.rs`, which runs before any service exists — pre-existing and out of scope for #449.
- Error behavior on a down PolicyService is a clear contextual error ("Are services running?"), with no silent fallback to direct DB access. Good — a fallback would have reintroduced the bug.
- capnp evolution is additive: request union previously ended at `@21`, response at `@22`; new ordinals are `@22`–`@27` / `@23`–`@28` with no reuse, and the new structs are appended.
- Server-side input validation is solid (32-byte length check before `VerifyingKey::from_bytes`, empty-username rejection, errors instead of panics), and keeping SSH-wire parsing client-side is the right call.
- The generated dispatch does gate these methods: `$scope` annotations produce `authorize(ctx, resource, action)` calls and `PolicyService` overrides `authorize` with the Casbin check, so enforcement is real — the issue below is the *choice* of scope, not missing enforcement.
- Tests exercise the real `RocksDbUserStore` through the real handlers, including the NOT_CONFIGURED and INVALID_INPUT paths.

**Blocking findings** (inline comments have details)

1. **PolicyService now steals the RocksDB writer lock from the OAuth service.** `create_policy_service` opens `RocksDbUserStore` at the same `credentials_dir` that `services/oauth/mod.rs:476` opens for OAuth login/SCIM/device flows. RocksDB is single-writer per directory — including two opens within one process (RocksDB's per-process lock table makes the second open fail), and `oauth` has `depends_on = ["policy"]`, so policy always opens first and wins. Result: in the default all-in-one deployment *and* the systemd multi-unit deployment, OAuth's user store degrades to "not configured" whenever both services run — i.e., this PR trades a CLI inconvenience for broken OAuth login. The PR description frames this as a pre-existing multi-process-only caveat, but before this change oauth was the only steady-state opener (the CLI conflict was transient); the permanent contention is introduced here. Suggested fixes, in rough order of preference: (a) a process-global shared user-store singleton in the pattern of `global_es256_key_store`, handed to both policy and oauth when co-located; (b) make oauth consume user data via the new policy RPCs so policy is the single owner; (c) at minimum, open the store lazily on first user-management RPC rather than at policy startup, so a deployment that never uses these RPCs doesn't lose OAuth.

2. **`listUsers`/`listUserKeys` under `$scope(query)` is user enumeration for nine service identities.** `SERVICE_BASE_POLICIES` grants `resource: policy:*, action: query` to oauth, registry, model, worker, oai, tui, discovery, notification, and metrics (added for service *key resolution*). With `$scope(query)`, all of them can now enumerate every registered username and each user's key fingerprints/labels/last-used timestamps. User management is privileged surface — these should be `$scope(manage)` like the mutating methods, or better, a dedicated scope (e.g. `$scope(userAdmin)`) so existing `policy:*` wildcard grants don't silently absorb user administration. Related: `service:mcp` holds `policy:* / *`, so the MCP service identity now gets full user register/remove capability as a side effect of the wildcard — a dedicated scope would contain that too.

**Non-blocking notes**

- Interaction with PR #774 (#759): the CLI path added here authenticates by loading the root key and verifying PolicyService against that same key (`create_policy_client`'s bootstrap assumption). #759 is precisely the bug where a standalone multi-service startup resolves policy's key to a different file — in those modes the new `hyprstream user` RPCs will fail envelope verification until #774 lands. Suggest merging #774 first and re-testing `hyprstream user list` in standalone `--ipc` mode. Both PRs also touch `bin/main.rs`, so expect a (likely trivial) rebase.
- PR #767 (9P MAC seam) touches only `hyprstream-9p`; no conflict.
- `block_in_place` + `Handle::current().block_on` in the factory matches the existing pattern in `factories.rs`, but note it panics on a current-thread runtime; since only the Valkey path is genuinely async, a sync `open_configured_user_store_blocking` for the RocksDB arm would remove the hazard entirely.
- `registerUser` validates empty usernames but `removeUser`/`listUserKeys`/`addUserKey` don't; harmless (store lookups just miss), but consistency is cheap.

ewindisch (commented):
Still blocking: no new commits have landed since the blocking review — the PR head is still 12759ff5b (committed 2026-07-04), which is the exact commit that review assessed. Re-verified both blocking findings against that head; both remain.

**1. RocksDB double-open lock contention — NOT FIXED.** `create_policy_service` (crates/hyprstream/src/services/factories.rs) still opens the user store via `crate::auth::open_configured_user_store(&creds_dir, ...)` → `RocksDbUserStore::open`, while `services/oauth/mod.rs` (~line 476) independently opens `RocksDbUserStore::open(&credentials_dir)` at the same directory. The doc comment on `open_configured_user_store` in `auth/mod.rs` describes this as a cross-process deployment limitation, but the collision applies equally in the default single-process daemon: RocksDB takes an exclusive LOCK per directory regardless of process boundaries, oauth `depends_on = ["policy"]` so policy always opens first, and the OAuth user store silently degrades to "not configured" — regressing OAuth login/SCIM in default deployments. Acceptable fixes remain as stated: a process-global shared store Arc (the `global_es256_key_store` pattern already used two lines up in the same factory), or single ownership with oauth resolving users through the new policy RPCs.

**2. Enumeration scope too broad — NOT FIXED.** `policy.capnp` still declares `listUsers @22 ... $scope(query)` and `listUserKeys @25 ... $scope(query)`, and the handlers (`handle_list_users` / `handle_list_user_keys` in services/policy.rs) perform no additional check beyond the generated scope gate. Every service identity holding `policy:*/query` from SERVICE_BASE_POLICIES can enumerate usernames plus key fingerprint/label/last-used metadata, and `service:mcp`'s `policy:*/*` wildcard still absorbs full registerUser/removeUser authority. These should be `$scope(manage)` (or a dedicated admin scope), matching the mutation methods.

Prior non-blocking notes (the #774 interplay for standalone `--ipc` envelope verification, `block_in_place` panicking on a current-thread runtime, and empty-username validation existing only on registerUser) also stand unchanged.

If a fix was pushed somewhere, it hasn't reached this PR branch — happy to re-review as soon as commits land.

ewindisch:
## 🔴 BLOCKING REVIEW (1:7 coordination sweep) — high-severity authorization bypass

The direct-RocksDB removal is clean and correct: `open_store` is deleted from `user_handlers.rs`, `main.rs` no longer derives `credentials_dir` for user commands, `users.toml.age`/`LocalKeyStore` untouched, the `services/oauth` direct open is a legitimate service-side path. But the new RPC surface has no authorization:

**1. (HIGH) None of the six new PolicyService handlers authorizes the caller.** `handle_list_users` / `handle_register_user` / `handle_remove_user` / `handle_list_user_keys` / `handle_add_user_key` / `handle_remove_user_key` (diff ~L583-745) all take `_ctx` unused with no `self.authorize(...)` call — unlike every other mutating PolicyService handler, which has an explicit "Unauthorized: {} cannot manage policy" check (policy.rs:461/:530/:581/:676/:763/:820). `dispatch_policy` is invoked bare from `handle_request` (policy.rs:1483) with no scope enforcement; the schema's `$scope(manage)` annotations are registry/MCP metadata, **not** enforced at dispatch. PolicyService is exposed over QUIC (`into_spawnable_quic(policy_service, config.policy.quic_port)`). **Any principal able to deliver a validly-signed envelope can call `addUserKey` to attach their own Ed25519 key to any account** — authentication bypass for every pubkey-authenticated surface — or `removeUser`/`registerUser` at will. The PR's own CLI change (self-minted signing key at `models_dir/.registry/keys`) demonstrates arbitrary callers are accepted.

Add the same `self.authorize(action, resource, &ctx)` guard the sibling handlers use, with an appropriate `manage:users` scope, before merge.

**2. (low, informational)** `block_in_place` + `Handle::block_on` in `create_policy_service` (`factories.rs` ~:504) panics on a current-thread runtime — matches ~10 pre-existing factory call-sites, so it's the established pattern, not a new regression. Noting only in case policy is ever spawned in Thread mode.

Item 1 blocks. Notifying the owning session.

ewindisch:
Superseding this PR direction based on the corrected control-plane split.

The underlying bug in #449 is real: runtime CLI user/key commands should not race a live service by opening the credential store directly. However, this implementation moves user management into `PolicyService`, which collapses an intentional boundary:

- Runtime user/key management belongs in `OAuthService` / `oauth.capnp`.
- Runtime policy/role management stays in `PolicyService`.
- Local operator/bootstrap/break-glass administration remains an out-of-band plane and must not depend on the runtime policy plane it may need to repair.
- The runtime user API must also be available through Plan 9/VFS, not only RPC/CLI.

I updated #449 with the corrected expected behavior and split the follow-up work into:

- #870 — runtime OAuth user management API over RPC and Plan 9/VFS
- #871 — local out-of-band user/policy admin plane
- #872 — external IdP matrix for OIDC/OAuth2/SCIM/SAML bridging

Recommendation: do not merge this PR as-is. Either close it in favor of the new tickets, or rewrite it so the runtime path goes through `OAuthService` plus VFS while preserving the local operator admin path.

ewindisch:
Closing per triage: superseded by the corrected control-plane split (#870/#871/#872 — user management belongs in OAuthService, not PolicyService), as noted in the author's own comment. Also carried an unaddressed authorization-bypass review on the mutating handlers. Reopen if any part is still wanted.

Issue #421, opened 2026-06-24
Epic: end-to-end validation of the consensus-free federated filesystem


ewindisch:
End-to-end validation of the full federated-FS lifecycle (#387). All structural pieces are merged — this epic wires them together and validates the complete flow.

## The full lifecycle to validate

```
register model → publish record → resolve from client → fetch weights → load → infer → train → promote → federate
```

### E1 — Content addressing + identity (CPU, no network)
- Register a model repo (git2db) → git commit OID
- Encode the OID as a CID (#395 cid module)
- Build an `ai.hyprstream.model` record (3 fields: repo, currentOid, createdAt) (#392 PDS record store)
- DAG-CBOR encode → deterministic CID → insert into MST → signed commit → CAR proof
- `verifyRecordProof(commit, key, path, record)` → tamper detection
- ModelRef prefix-dispatch: `at://did:web:.../ai.hyprstream.model/qwen3` parses → federated branch; `qwen3:main` → local branch

### E2 — Weight fetch + cas-serve (CPU)
- Model repo has safetensors behind LFS pointers with `xet-merkle`/`xet-shard` extensions (#386)
- UploadFile → CDC chunks → xorbs → mdb_shard (#390)
- GetFile → mdb_shard reconstruction → bytes match original
- The LFS pointer's `xet-merkle` resolves to the cas-serve hash

### E3 — 9P filesystem walk (CPU)
- Registry/worktree mounted via RemoteRegistryMount → real qids (#388 threading)
- `walk /srv/registry/{repo}/{worktree}/config.json` → real file bytes
- `create` + `write` lands in writable upper (copy-up from immutable floor) (#394)
- `promote` → stageAll + commit → new OID → deterministic

### E4 — Iroh transport (single host, two processes)
- Daemon binds Iroh endpoint (primary, #411) → both ALPNs
- Client dials Iroh (`EndpointType::Iroh{node_id, direct_addrs}`) → `Arc<dyn RpcClient>`
- RPC round-trip over Iroh (model.status, registry.list)
- moq stream over Iroh (`MoqStreamSession::Iroh`) → token stream

### E5 — MoQ event plane (single host)
- EventPublisher publishes to per-OID track (`local/events/publications/{oid_hash}`) (#393)
- EventSubscriber.subscribe_oid(oid) → receives only that OID's events (selectivity verified)
- Flat-track back-compat still works
- FirehoseBackfill mode → replays history before live (if firehose source available)

[...]

PR #776, opened 2026-07-04
fix(9p/git2db): Tstat name, register-path RPC, swallowed worktree error (#427)

### LINKED ISSUE #427: Follow-ups from E3 9P walk test: empty Tstat name, deprecated register RPC, swallowed git2db worktree error
Surfaced during #421-E3 (9P walk e2e, PR #426). Three real findings, out of scope for the test but worth fixing:

1. **Empty Tstat name** — `metadata_to_np_stat("", ..)` means the registry's Tstat handler never returns a file name; identity is carried by qid/qtype/size only. wstat rename is likewise deferred. 9P clients that rely on stat.name (some do) would see empty names.

2. **Deprecated `register` RPC path** — the `register` handler uses `register_repository`, which ignores the supplied path and hard-requires the repo at `base_dir/<random-uuid>`. The non-deprecated path is `Git2DB::register(...)`. The RPC surface should route to the non-deprecated builder.

3. **Swallowed error in `git2db/storage/vfs.rs::create_git_worktree`** — `let _ = tokio::task::spawn_blocking(...)` discards the inner `Git2DBError`; a libgit2 worktree-add failure surfaces later as a confusing "Worktree path does not exist / LFS fetch error" instead of the real cause. Propagate the error.

Also: the `clone` RPC routes through the global GitManager whose env default is `prefer_shallow`, and a shallow fetch over `file://` local transport is rejected by libgit2 — making both RPC registration paths unusable in a local-file test harness.

ewindisch:
Three real findings surfaced during #421-E3 (9P walk e2e, PR #426), out of scope for that test but worth fixing:

1. **Swallowed error in `create_git_worktree`** (`git2db/src/storage/vfs.rs`) — `let _ = spawn_blocking(...).await.map_err(...)?` discarded the inner `Git2DBResult`, so a libgit2 worktree-add failure (bad ref, missing source, etc.) silently returned `Ok(())`; the real cause only surfaced later as a confusing downstream "path does not exist"-style error. Now propagates both the join error and the inner result. Regression test: `create_git_worktree_propagates_inner_error` (forces an unresolvable ref_spec and asserts the failure surfaces).

2. **Deprecated `register` RPC path** (`hyprstream/src/services/registry.rs::handle_register`) — used `register_repository`, which ignores the supplied path and hard-requires the repo at `base_dir/<random-uuid>`. Now routes through the non-deprecated `Git2DB::register(...)` builder, honoring the caller-supplied path, and wires the previously-ignored `tracking_ref` parameter. Test: `test_register_honors_caller_supplied_path`.

3. **Empty Tstat name** — `metadata_to_np_stat("", ..)` meant the registry's Tstat handler never returned a file name (identity carried by qid/qtype/size only). Added `rel_path_file_name()` to derive the real name from the fid's tracked `rel_path`, threaded into `handle_np_stat`. Test: `test_tstat_returns_real_registry_repo_name`.

**Bonus** (matches the issue's secondary note): `git2db/src/manager.rs` — the `clone` RPC's shallow-fetch-by-default (`prefer_shallow`) is rejected by libgit2 over `file://` local transport, making both RPC registration paths unusable in a local-file test harness. Now detects `file://` and forces a full fetch there.

## Testing
- `cargo test -p git2db --lib` — 93/93 pass (incl. the new worktree-error test).
- `cargo test -p hyprstream --lib services::registry::` — 12/14 pass; both new #427 tests pass. The 2 unrelated failures (`test_xet_merkle_denied_for_planted_reference`, `test_xet_merkle_permitted_when_file_reconstructs_from_it`) are **pre-existing on main**, confirmed by running them against the unmodified base commit (same failures, same assertions) before this diff was applied.
- `cargo clippy -p git2db --all-targets` and `cargo clippy -p hyprstream --lib --all-targets` — both clean.
- Rebased onto current `main`, no conflicts.

Closes #427.

### DIFF
--- a/crates/git2db/src/manager.rs
+++ b/crates/git2db/src/manager.rs
@@ -483,7 +483,18 @@ impl GitManager {
         // Convert inputs to owned types for spawn_blocking
         let url = url.to_owned();
         let target_path = target_path.to_path_buf();
-        let options = options.unwrap_or_else(|| self.default_clone_options());
+        let mut options = options.unwrap_or_else(|| self.default_clone_options());
+
+        // libgit2's local transport rejects shallow fetch over `file://` —
+        // disable it here so callers that default to prefer_shallow (the
+        // global config default) don't get an opaque failure when cloning a
+        // local repository (e.g. RPC registration in local-file test
+        // harnesses, or same-host repo mirroring).
+        if options.shallow && url.starts_with("file://") {
+            debug!("Disabling shallow clone for local file:// transport: {}", url);
+            options.shallow = false;
+            options.depth = None;
+        }
 
         // Acquire operation permit
         let _permit =
--- a/crates/hyprstream/src/services/registry.rs
+++ b/crates/hyprstream/src/services/registry.rs
@@ -1145,18 +1145,16 @@ impl RegistryService {
     }
 
     /// Handle register operation
-    #[allow(deprecated)]
     async fn handle_register(
         &self,
         path: &str,
         name: Option<&str>,
-        _tracking_ref: Option<&str>,
+        tracking_ref: Option<&str>,
     ) -> Result<TrackedRepository> {
         let path = PathBuf::from(path);
         // I5: Validate path is within base_dir
         self.validate_path_within_base(&path)?;
         let name = name.map(std::borrow::ToOwned::to_owned);
-        // Note: tracking_ref is not yet used by register_repository
 
         // Register requires write lock
         let mut registry = self.registry.write().await;
@@ -1167,8 +1165,20 @@ impl RegistryService {
         // Derive URL from path (local file URL)
         let url = format!("file://{}", path.display());
 
-        // TODO: Migrate to registry.register(repo_id) builder API
-        registry.register_repository(&repo_id, name, url).await?;
+        // Use the non-deprecated registry.register(repo_id) builder so the
+        // caller-supplied path is honored, rather than register_repository's
+        // hard-coded base_dir/<uuid> location.
+        let mut builder = registry
+            .register(repo_id.clone())
+            .url(url)
+            .worktree_path(path);
+        if let Some(name) = name {
+            builder = builder.name(name);
+        }
+        if let Some(tracking_ref) = tracking_ref {
+            builder = builder.tracking_ref(GitRef::Branch(tracking_ref.to_owned()));
+        }
+        builder.exec().await?;
 
         // Get the tracked repository to return
         let repo = registry
@@ -1536,6 +1546,18 @@ impl RegistryService {
             qtype: qid.qtype,
             version: qid.version,
             path: qid.path,
+        }
+    }
+
+    /// Extract a file's own name (the final path component) from a fid's
+    /// tracked `rel_path` (relative to the contained root). Returns `""` for
+    /// the contained root itself (`"."` / empty) — a synthetic entry with no
+    /// name of its own — which callers may leave empty in `Tstat`.
+    fn rel_path_file_name(rel_path: &str) -> &str {
+        if rel_path.is_empty() || rel_path == "." {
+            ""
+        } else {
+            rel_path.rsplit('/').next().unwrap_or(rel_path)
         }
     }
 
@@ -2857,16 +2879,17 @@ impl WorktreeHandler for RegistryService {
         let entry = self.fid_table.get_verified(data.fid, &subject)
             .map_err(|e| anyhow!("{}", e))?;
 
-        let meta = match &entry.state {
-            FidState::Walked { walk_handle, .. } => {
-                walk_handle.metadata().map_err(|e| anyhow!("{}", e))?
-            }
-            FidState::Opened { file, .. } => {
-                file.lock().metadata()?
+        let (meta, rel_path) = match &entry.state {
+            FidState::Walked { walk_handle, .. } => (
+                walk_handle.metadata().map_err(|e| anyhow!("{}", e))?,
+                walk_handle.rel_path().to_owned(),
+            ),
+            FidState::Opened { file, rel_path, .. } => {
+                (file.lock().metadata()?, rel_path.clone())
             }
         };
 
-        let stat = Self::metadata_to_np_stat("", &meta);
+        let stat = Self::metadata_to_np_stat(Self::rel_path_file_name(&rel_path), &meta);
         Ok(RStat { stat })
     }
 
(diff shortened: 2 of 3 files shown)


crates/hyprstream/src/services/registry.rs
```diff
@@ -1167,8 +1165,20 @@ impl RegistryService {
...
 
-        // TODO: Migrate to registry.register(repo_id) builder API
-        registry.register_repository(&repo_id, name, url).await?;
+        // Use the non-deprecated registry.register(repo_id) builder so the
+        // caller-supplied path is honored, rather than register_repository's
+        // hard-coded base_dir/<uuid> location.
+        let mut builder = registry
+            .register(repo_id.clone())
+            .url(url)
+            .worktree_path(path);
```
ewindisch, line 1174: Non-blocking, latent: now that caller-supplied paths are honored, a nested path like `base_dir/a/b/repo` becomes reachable through this RPC for the first time — and `register_repository_internal` computes the submodule relative URL as `format!("../{worktree_name}")` from `file_name()` only, which is wrong for anything deeper than a direct child of `base_dir` (it would record `../repo` instead of `../a/b/repo`). `TrackedRepository.worktree_path` stores the full path so most operations survive, but anything that consults the submodule URL will be off. Pre-existing git2db bug, not introduced here — but this PR is what exposes it, so worth either a follow-up issue or restricting `register` to direct children of `base_dir` for now.

crates/hyprstream/src/services/registry.rs
```diff
@@ -1167,8 +1165,20 @@ impl RegistryService {
...
+        if let Some(name) = name {
+            builder = builder.name(name);
+        }
+        if let Some(tracking_ref) = tracking_ref {
+            builder = builder.tracking_ref(GitRef::Branch(tracking_ref.to_owned()));
```
ewindisch, line 1179: Non-blocking: `trackingRef` in the capnp schema is an undocumented free-form Text, and this maps it unconditionally to `GitRef::Branch` — a caller passing a tag (`v1.0`) or SHA gets it misclassified as a branch in the stored metadata. This does match the existing convention in this file (the `tracked_repository_from_parts` helper does the same), so it's consistent; but `GitRef::Revspec` exists and would be the faithful mapping for an arbitrary ref string. Alternatively, document in registry.capnp that `trackingRef` is branch-only.

crates/git2db/src/manager.rs
```diff
@@ -483,7 +483,18 @@ impl GitManager {
...
         let target_path = target_path.to_path_buf();
-        let options = options.unwrap_or_else(|| self.default_clone_options());
+        let mut options = options.unwrap_or_else(|| self.default_clone_options());
+
+        // libgit2's local transport rejects shallow fetch over `file://` —
+        // disable it here so callers that default to prefer_shallow (the
+        // global config default) don't get an opaque failure when cloning a
+        // local repository (e.g. RPC registration in local-file test
+        // harnesses, or same-host repo mirroring).
+        if options.shallow && url.starts_with("file://") {
```
ewindisch, line 493: Non-blocking coverage note: libgit2's local transport also serves scheme-less paths (`/abs/path/to/repo`), where a shallow fetch is rejected the same way. This check only catches the `file://` spelling. Since the registry RPC always builds `file://` URLs this fix is sufficient for the #427 harness, but a caller cloning by plain path still hits the opaque failure — consider `url.starts_with("file://") || !url.contains("://")` (or checking `Path::new(&url).is_absolute()`) if you want to close it fully.

ewindisch (commented):
**Verdict: LGTM — safe to automerge** (three non-blocking notes inline).

Reviewed each of the three fixes independently against the source at the head commit, plus the bonus shallow-clone change.

**1. Tstat name (registry.rs)** — Correct. `rel_path_file_name` returns the final `/`-separated component of the fid's tracked `rel_path`, which is exactly what the stat name field should carry (final path element, never the full path). `rel_path` is built by `WalkHandle::child_rel_path` with `/` joins and `.`/empty at the contained root, and both root spellings are handled. Both fid states (`Walked` via `walk_handle.rel_path()`, `Opened` via the pre-existing `rel_path` field) are covered, and `metadata_to_np_stat` has no other empty-name call sites left. Leaving the contained root's own name as `""` is a reasonable, documented choice (classic 9P servers vary between `""`, `"/"`, and the attach name here). The `.unwrap_or(rel_path)` after `rsplit('/').next()` is dead (rsplit always yields at least one element) but harmless.

**2. register RPC (registry.rs)** — Correct and the auth surface is unchanged and adequate: `register @5` already carries `$scope(write)` in registry.capnp and the coarse scope gate runs in generated dispatch (same pattern the `handle_put_blob` doc comment describes); no new capnp method or schema change in this PR. Path traversal is covered by the pre-existing `validate_path_within_base` (canonicalize + prefix for existing paths, `contained_join` otherwise), which still runs before anything else. Git operations go through the git2db `register(repo_id)` builder — no raw `Repository::open` introduced. Wiring the previously-ignored `tracking_ref` is a genuine fix. Two non-blocking notes inline.

**3. Swallowed worktree error (git2db/storage/vfs.rs)** — Correct. The closure's `Git2DBResult<()>` is now surfaced via `??` after the join-error map, and the smudge-progress hook cleanup still runs unconditionally before the closure returns (the `result` capture pattern is preserved), so no cleanup regression. The regression test drives the exact previously-swallowed path (unresolvable ref inside `spawn_blocking`) and asserts on the real libgit2 message. Nit: the file lost its trailing newline (`\ No newline at end of file`).

**Bonus (git2db/manager.rs shallow-over-file://)** — Matches the issue's secondary note; scoped to `file://` URLs with a debug log, clears `depth` too. One coverage note inline.

**Cross-reference #427**: this addresses all three numbered findings plus the secondary shallow/`file://` note, each with a dedicated regression test. The one thing the issue mentions that is not done here is wstat rename — but the issue itself already marked that "likewise deferred", so `Closes #427` is fair; consider a one-line tracking note if wstat rename should survive the close. No overlap with #838 (namespace_builder/tui/cli files), #839 (rpc-std stream mount), or #767 (hyprstream-9p translator — this PR never touches crates/hyprstream-9p): file sets are fully disjoint.

Review was static (source inspection at head commit 3513c24); I did not re-run the test suite — the PR body's pre-existing-failure verification for the two xet merkle tests is a reasonable methodology.

ewindisch:
✅ **Reviewed (1:7 sweep) — ready.** All three fixes match their descriptions: (1) `vfs.rs` `??` now propagates the inner `Git2DBResult` — the only callers "relying on Ok" then failed downstream with a misleading path error, so surfacing the real error is strictly better, and the regression test pins it; (2) `handle_register` routes through the `registry.register(repo_id)` builder honoring `worktree_path` + `tracking_ref`; (3) Tstat name derived from fid `rel_path`, root stays `""` by documented convention.

Two low, non-blocking notes:
- `git2db/src/manager.rs` shallow-disable guard is `url.starts_with("file://")` only; plain local paths (`/abs/path`) also use libgit2's local transport and equally reject shallow, so a local-path clone with `prefer_shallow` (default) still hits the same opaque error the bonus fix targets. Consider broadening the guard.
- `handle_register` wraps caller-supplied `tracking_ref` unconditionally as `GitRef::Branch` — a tag or commit SHA gets silently misclassified.

Neither blocks. Arming auto-merge.

Issue #809, opened 2026-07-05
Epic: 9P kernel completion — make the namespace load-bearing


ewindisch:
hyprstream already has the skeleton of a hosted 9P kernel in the Inferno/Wanix sense: a per-process `Namespace` with Plan 9 bind/union/fork semantics (ADR #651), a `Mount` device layer with 15+ implementations, and a bidirectional mount driver (`Remote9pMount` import, the 9P translator export, the Wanix bridge; internal 9P ops ride Cap'n Proto via `nine.capnp`). What it does not have is *obligatoriness*: the important data paths run beside the namespace instead of through it. One disease, three expressions:

1. **Storage silos.** Four content-addressed stores share no substrate and (except RAFS) no Mount: git objects (git2db, RPC-proxy-only), XetStorage (remote HTTPS via the hardcoded git filter — `filter.rs:63`; default endpoint HuggingFace), `cas_serve::CasStore` (local scaffolding behind registry putBlob — the retired cas-serve binary's leftover write core), RAFS blobs. `GittorrentStorage` (P2P) is a real backend stranded unreachable.
2. **The inference bypass.** Model weights load via raw `worktree_path()` PathBuf → `std::fs`/mmap (`inference.rs:778`, `model_factory.rs:362,391`). No Namespace, no Subject, no MAC on the hot path.
3. **The moq sibling plane.** Streaming is its own wire protocol beside 9P (opaque frames through the relay), which is why `/moq` needs a hardcoded wire selector — contradicting "inference streams are files".

Plus one boundary defect underneath all three: the fuse up-adapter drops `Subject` (neutral root `Context::default()`, `fuse_adapter.rs:79-83`), so even the one composed CAS (RAFS rootfs) loses its caller at the leaf.

**Completion criterion: no path to a resource except through a Mount; all cross-boundary IO is 9P.** Carriers (vsock/QUIC/WebTransport/iroh/moq/RDMA) are fixed at dial time and never app-visible; remaining wire planes are discovery-advertised under `/.well-known/{name}`.

[...]

PR #838, opened 2026-07-05
fix(services): make namespace_builder mount paths caller-supplied

ewindisch:
## Summary

An independent audit found that `crates/hyprstream/src/services/namespace_builder.rs::build_standard_namespace` (landed via #703, part of epic #632's namespace-composition-unification work) baked the six standard VFS mount paths — `/srv/model`, `/srv/registry`, `/worktree`, `/bin`, `/env`, `/lang/tcl` — as string literals directly inside the shared builder function.

This violates the mount-point convention ratified in epic #809 ("9P kernel completion"): a mount point is a per-namespace binding decision made by the entity actually composing the namespace, never something a shared library/server dictates. Centralizing the six literals inside `build_standard_namespace` put that decision in the wrong place, even though (as a hardcoded default) it happened to produce correct current behavior.

## Changes

- `crates/hyprstream/src/services/namespace_builder.rs`: `StandardNamespaceConfig` gains six required fields — `model_path`, `registry_path`, `worktree_path`, `bin_path`, `env_path`, `tcl_path` — replacing the string literals previously baked into `build_standard_namespace`. The function body now reads these fields for every `ns.mount(...)`/`ns.bind_mount(...)` call; none of the six literal path strings remain inside the builder.
- `crates/hyprstream/src/cli/shell_handlers.rs`: the CLI shell call site (`handle_shell_tui`) now supplies the six paths explicitly when constructing `StandardNamespaceConfig`.
- `crates/hyprstream/src/tui/vfs.rs`: the TUI ChatApp call site (`build_chat_vfs_namespace`) does the same.

Both call sites pass identical literal strings today (`/srv/model`, `/srv/registry`, `/worktree`, `/bin`, `/env`, `/lang/tcl`), so runtime behavior is unchanged — this is purely a relocation of where the path decisions live, from the shared builder to each caller, per #809's convention. No new shared "default paths" convenience was introduced, since that would just relocate the same violation one level up.

Out of scope (confirmed already correct, not touched): `mount_all_services` in `hyprstream-rpc-std/src/vfs_mount.rs` (per-service name-derived prefix, documented convention) and `CriBackend::ANN_MOUNT_PREFIX` (caller-supplied via annotations).

Refs #703, #632, #809.

## Test plan
- [x] `cargo test -p hyprstream --lib` — 885 passed, 0 failed, 1 ignored
- [x] `cargo clippy --workspace --all-targets --features download-libtorch -- -D warnings` — clean, no warnings
- [x] pre-commit clippy hook passed on commit

### DIFF
--- a/crates/hyprstream/src/services/namespace_builder.rs
+++ b/crates/hyprstream/src/services/namespace_builder.rs
@@ -16,6 +16,15 @@
 //! This is a pure assembly layer: it does not change any Subject/MAC
 //! scoping semantics of the individual mounts (`RemoteModelMount`,
 //! `RemoteRegistryMount`, `TclMount`) it wires together.
+//!
+//! Per epic #809's mount-point convention ("a mount point is a per-namespace
+//! binding decision... the server never dictates the path"), this builder
+//! does not itself choose where anything lands in the tree — every mount
+//! path below is a required field on [`StandardNamespaceConfig`] that the
+//! caller must supply. The CLI shell and TUI ChatApp callers both currently
+//! pass the same literal paths (`/srv/model`, `/srv/registry`, `/worktree`,
+//! `/bin`, `/env`, `/lang/tcl`), but that's a coincidence of current
+//! behavior, not something this module bakes in.
 
 use std::collections::HashMap;
 use std::sync::Arc;
@@ -32,21 +41,43 @@ use super::RegistryClient;
 /// Inputs that legitimately differ between the CLI shell and TUI ChatApp
 /// call sites. Everything else about the recipe is fixed by
 /// [`build_standard_namespace`].
+///
+/// The mount-point paths are required fields, not defaults this builder
+/// invents (epic #809) — every caller states its own paths explicitly, even
+/// when they happen to match another caller's.
 pub struct StandardNamespaceConfig {
     /// Subject the `/lang/tcl` driver operates as (the caller separately
     /// keeps its own copy of the `Subject` for VFS proxy / RPC use — this
     /// builder only consumes it for `TclMount::spawn`).
     pub subject: Subject,
-    /// Already-constructed model service RPC client, mounted at
-    /// `/srv/model`.
+    /// Already-constructed model service RPC client.
     pub model_client: ModelClient,
-    /// Registry service RPC client. When `Some`, mounts `/srv/registry`
-    /// (+ a `/worktree` alias, `BindFlag::After`, same backing mount).
+    /// Registry service RPC client. When `Some`, mounts `registry_path`
+    /// (+ a `worktree_path` alias, `BindFlag::After`, same backing mount).
     /// The CLI shell always has a registry client available; the TUI
     /// ChatApp namespace builder currently does not obtain one and passes
     /// `None` — this was already the pre-#634 behavior of each builder, now
     /// made an explicit parameter rather than silently unified.
     pub registry: Option<RegistryClient>,
+    /// Path at which `model_client`'s 9P filesystem is mounted (caller
+    /// convention today: `/srv/model`).
+    pub model_path: String,
+    /// Path at which `registry`'s worktree filesystem is mounted, when
+    /// `registry` is `Some` (caller convention today: `/srv/registry`).
+    pub registry_path: String,
+    /// Alias path bound (`BindFlag::After`) to the same backing mount as
+    /// `registry_path`, when `registry` is `Some` (caller convention today:
+    /// `/worktree`).
+    pub worktree_path: String,
+    /// Path for the static `/bin`-style command listing + discovered `.tcl`
+    /// tool scripts (caller convention today: `/bin`).
+    pub bin_path: String,
+    /// Path for the session-variable file tree (caller convention today:
+    /// `/env`).
+    pub env_path: String,
+    /// Path at which the Tcl shell driver is mounted (caller convention
+    /// today: `/lang/tcl`).
+    pub tcl_path: String,
 }
 
 /// Compose the standard VFS namespace shared by the CLI shell and TUI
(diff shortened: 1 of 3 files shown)


crates/hyprstream/src/services/namespace_builder.rs
```diff
@@ -178,14 +214,14 @@ pub fn build_standard_namespace(cfg: StandardNamespaceConfig) -> Namespace {
...
+    let _ = ns.mount(&env_path, Arc::new(env_tree));
 
-    // /lang/tcl — the driver gets a fork() snapshot of the namespace (taken
-    // before /lang/tcl is mounted, so it never observes itself) for
-    // `/bin/{cmd}` fallback resolution inside `/lang/tcl/eval`.
+    // tcl_path — the driver gets a fork() snapshot of the namespace (taken
+    // before tcl_path is mounted, so it never observes itself) for
+    // `{bin_path}/{cmd}` fallback resolution inside `{tcl_path}/eval`.
     let driver_ns = Arc::new(ns.fork());
     let tcl_mount = Arc::new(hyprstream_workers_tcl::TclMount::spawn(subject, driver_ns));
```
ewindisch, line 223: Medium: the driver spawned here does not respect the caller's `bin_path`. `hyprstream-workers-tcl/src/lib.rs:173` builds `format!("/bin/{cmd_name}")` for command fallback, and `builtins.rs` hardcodes `/bin` (line 193) and `/srv` (lines 223, 240). So while this builder no longer dictates paths, the Tcl mount it wires in still assumes the conventional ones — a caller choosing `bin_path: "/cmd"` gets a namespace where `{tcl_path}/eval` fallback silently resolves nothing. Suggest either passing `bin_path` (and ideally the srv-style prefixes) into `TclMount::spawn`, or adding a doc note on `bin_path`/`model_path` that the Tcl driver currently requires the conventional values, so the field docs don't overpromise.

crates/hyprstream/src/services/namespace_builder.rs
```diff
@@ -60,35 +91,40 @@ pub fn build_standard_namespace(cfg: StandardNamespaceConfig) -> Namespace {
...
 
     // Mount the model service's 9P filesystem via RPC proxy.
     // RemoteModelMount translates sync Mount trait calls to async ModelClient
-    // RPC requests, so `/srv/model/{model_ref}/status` etc. are served by the
-    // model service's SyntheticTree on the other end of the RPC channel.
+    // RPC requests, so `{model_path}/{model_ref}/status` etc. are served by
+    // the model service's SyntheticTree on the other end of the RPC channel.
     let remote_model_mount = RemoteModelMount::new(model_client);
-    let _ = ns.mount("/srv/model", Arc::new(remote_model_mount));
+    let _ = ns.mount(&model_path, Arc::new(remote_model_mount));
```
ewindisch, line 109: Low (pre-existing pattern, but now more load-bearing): `let _ =` discards the mount result throughout. With paths now caller input, a malformed path (empty, relative, duplicate) fails silently and the caller has no signal. Consider returning a Result from `build_standard_namespace` or logging failures — fine as a follow-up.

crates/hyprstream/src/services/namespace_builder.rs
```diff
@@ -32,21 +41,43 @@ use super::RegistryClient;
...
+    /// Path at which `registry`'s worktree filesystem is mounted, when
+    /// `registry` is `Some` (caller convention today: `/srv/registry`).
+    pub registry_path: String,
+    /// Alias path bound (`BindFlag::After`) to the same backing mount as
+    /// `registry_path`, when `registry` is `Some` (caller convention today:
+    /// `/worktree`).
+    pub worktree_path: String,
+    /// Path for the static `/bin`-style command listing + discovered `.tcl`
+    /// tool scripts (caller convention today: `/bin`).
+    pub bin_path: String,
```
ewindisch, line 74: Low: no test exercises non-conventional paths. A unit test that builds a namespace with e.g. `bin_path: "/x/bin"` and asserts the mount resolves there would pin the "caller decides" contract against future re-hardcoding. (It would also have surfaced the TclMount `/bin` coupling noted above.)

ewindisch (commented):
**Verdict: Approve with changes** (non-blocking follow-ups; nothing here is a correctness regression for current callers).

This is the right structural change, not a rename-to-a-constant. The builder genuinely no longer dictates any mount point: all six paths are required fields on `StandardNamespaceConfig`, both call sites (CLI shell in `cli/shell_handlers.rs`, TUI ChatApp in `tui/vfs.rs`) state their own paths explicitly, and the caller in both cases is the namespace owner — not the mounted service — which is exactly who the mount-point rule says should decide. I grepped the three touched files at the PR head: no server-side path literals remain in `namespace_builder.rs` code (only doc-comment mentions of the current caller convention, which the module doc correctly frames as "a coincidence of current behavior"). Deliberately *not* providing a `Default` keeps the builder from re-inventing canonical paths, which matches the intent.

Findings (details inline):

1. **Medium** — the `bin_path`/`tcl_path` choice is not honored by the Tcl driver this same builder spawns: `hyprstream-workers-tcl/src/lib.rs:173` resolves commands via a hardcoded `format!("/bin/{cmd_name}")`, and `builtins.rs` hardcodes `/bin` and `/srv` for `ls`/doc discovery. A caller supplying `bin_path != "/bin"` (or a non-`/srv` model/registry path) gets a namespace where `{tcl_path}/eval` command fallback and builtins silently miss. Either thread the paths into `TclMount::spawn` or document the coupling as a constraint on those fields until the driver is parameterized.
2. **Low** — mount errors are still swallowed (`let _ = ns.mount(&model_path, ...)`). That pattern predates this PR, but now that paths are caller input, an empty/relative/colliding path fails silently. Worth a follow-up to surface or at least log these.
3. **Low** — no test pins the new contract. A small unit test building a namespace with nonstandard paths and asserting the mounts land where specified would prevent a future "helpful" re-hardcoding.
4. **Info** — `hyprstream_vfs::STANDARD_NAMESPACE_PATHS` (`crates/hyprstream-vfs/src/lib.rs:92`) still exists as a library-side list of `/srv/model`, `/srv/registry`, `/worktree` and is referenced from `hyprstream-rpc-std/src/vfs_mount.rs`. It now reads as a leftover server-canonical path set; worth deciding whether it becomes a documented caller convention (and moving/renaming it accordingly) in a follow-up.

Cross-checked open PRs #839 (`stream_mount.rs`/`wasm_rpc_client.rs`/`stream_consumer.rs`) and #767 (`hyprstream-9p` only): no file overlap, no conflict risk.

ewindisch:
✅ **Reviewed (1:7 sweep) — ready, no findings.** Correctly implements the #809 convention: all six path literals removed from `build_standard_namespace`, decision relocated to the two composing call sites (`cli/shell_handlers.rs`, `tui/vfs.rs`), no re-centralizing "default paths" constant. No namespace-escape/injection surface — paths mount into an in-process `Namespace` the caller owns, `normalize_prefix`/`normalize_path` coerce leading `/`, both call sites pass compile-time literals. (Informational, not blocking: `hyprstream_vfs::STANDARD_NAMESPACE_PATHS` still hardcodes the browser-mirror paths the builder doc references — nothing ties the two together if a future caller diverges.) Arming auto-merge.

Issue #819, opened 2026-07-05
feat(9p): /stream/{topic} as 9P files — QoS vocabulary (moq fold-in, app half)


ewindisch:
The moq fold-in, **app/interface half** — not a data-plane change. Streams become files: 9P is the *interface* (naming, addressing, open-time authorization), **not** a byte-copy path. The stream tree's mount point is a per-namespace binding decision; `/stream/{topic}` in the title is illustrative shorthand for "streams are files", not a required path.

## Interface vs data plane (the load-bearing split — read #817 first)

- `open()` through the namespace authorizes (Subject / MAC) and returns a **handle / subscription**. The stream's bytes then flow over the **moq carrier**, QUIC-native delivery intact (per-object streams, drop-stale, datagram, relay rendezvous). moq is **not** replaced and frames are **not** memcpy'd through 9P.
- blocking `Tread` = push semantics (exactly Plan 9 `/net/tcp/n/data`) — this is the always-works **floor**, correct and sufficient for low-rate streams (tokens). It is **not the ceiling**: high-rate / bulk uses the carrier-backed handle, exactly as #817 does mmap-at-open for weights. An implementation that routes every frame through `Tread` and loses moq's delivery semantics is **wrong**.
- QoS that doesn't reduce to reliable-in-order gets a small vocabulary on `ctl` / open-mode — reuse StreamOpt's existing encoding (latest-wins, lossy-ok, drop-stale) moved onto the file. This vocabulary is *how* moq's (and iroh's) rich, **non-fungible** carrier semantics stay first-class **through** the interface, rather than being flattened to a lowest-common-denominator read.

## Grounded precedent

`StreamMount` already exists (`hyprstream-rpc-std/src/stream_mount.rs` — MoQ token streams as files). This story makes it the canonical surface rather than a side door, **including events** (group-keyed EventService, #600).

#169's at-least-once / offset-resume restates naturally: resume = `Tread` at offset (floor path) or subscription-resume on the carrier handle; dedup at the file layer (see the #169 rewrite).

Epic: #809 (W3, app/interface half). Prereq: #670 (StreamMount pipe-reader `Send`). Pairs with: #817 (same handle-at-open data-plane model). Feeds: #820, #169.

Issue #670, opened 2026-07-02
feat(9p-rpc): make StreamMount pipe-reader native — StreamHandle futures must be Send


ewindisch:
**Deferred from T3 (#542/PR #662), explicitly out of scope there.** T3 ported the generic `ServiceMount`/`DocMount` capnp→9p adapter to native `Send+Sync`, but **`StreamMount`** (the `/stream/{topic}` pipe reader) stayed `wasm32`-gated.

## Why it couldn't go native in T3
`StreamMount::read`/`write` await `StreamHandle::next_payload`/`cancel`, which return **`!Send` futures** (`#[async_trait(?Send)]` on the `StreamHandle` trait in `hyprstream-rpc`). A native `Send+Sync` Mount can't hold a `!Send` future across `.await`. Making `StreamHandle`'s futures `Send` is a transport-layer change, separate from the mount port.

## Current state
`StreamRegistry`/`StreamEntry` ARE target-agnostic (T3 ported them), so `GenericServiceMount` registers streams on any target — only the pipe **reader** is deferred. So native services can register streams but can't yet serve the `/stream/{topic}/{data,info,ctl}` read face natively.

## The fix
Make `StreamHandle`'s trait futures `Send` (drop `?Send`, or restructure so the reader doesn't hold a `!Send` future across await), then un-gate `StreamMount` for native.

## Refs
Epic #539 (T3 follow-up) · deferred from #542. Composes with the streaming plane (#170).

ewindisch:
[...]

Issue #651, opened 2026-07-02
ADR: overlay is a union bind-mount policy + ctl, NOT a new composition primitive (Plan 9 native)


ewindisch:
**Decision record.** Locks the VFS composition architecture after the 2026-07-01 design call.

## Decision

Overlay/copy-on-write is **NOT a new primitive** (no `OverlayMount` type, no `Overlay<FsMount>`). It is a **write/readdir policy** layered onto the existing Plan 9 `Namespace` bind-mount union (`bind_mount(prefix, target, BindFlag::Before|After)`), expressed via a `UnionPolicy` + ctl files.

The composition mechanic for the entire system — services over services, federation, deltas, snapshots, sandbox rootfs — is ONE primitive: **`Namespace::bind_mount` over `Arc<dyn Mount>`**, with `Subject` on every call. Overlay is the copy-up/whiteout policy on that union (#370), not a separate type.

## Why (the two rejected alternatives)

- **Fuse `OverlayFs` as the composition layer (the path PR #641 ships for sandbox rootfs):** composes at the fuse layer, which loses `Subject` (fuse `FileSystem` ops have no `&Subject` param). A service-as-9P-`Mount` lower would require a `Mount → fuse FileSystem → Mount` round-trip that drops the policy principal — structurally incompatible with per-layer authz in a federated/service composition. Fuse `OverlayFs` is acceptable as a **leaf backend** (`FuseFileSystemMount<Rafs>` → bind in as a `Mount`), NOT as the composition mechanism.
- **A native `OverlayMount<L: Mount, U: FsMount>` primitive:** would be a parallel special-case on top of bind-mounting — the same drift pattern that EventService consolidation and the scheduling substrate deliberately avoided. The namespace already stacks `Mount`s at a prefix with union read/readdir/write; inventing a second stacking type duplicates that mechanic.

## Consequence for the codebase

- `crates/hyprstream-vfs/src/overlay.rs` (the fuse `rootfs_overlay` path) stays as a **sandbox-rootfs-specific leaf backend** — clearly marked as "bind this in, do not extend composition here." It is not the composition layer.
- `crates/hyprstream-vfs/src/namespace.rs` is the composition layer. #370 adds `UnionPolicy` (copy-up/whiteout) to its `write`/`readdir`/`create` paths + a ctl surface.
- The composition vision (#632 universal namespace-composition, #549 federation) rides `bind_mount` — no new composition code beyond the #370 policy.

## Interop commitment

[...]

PR #839, opened 2026-07-05
feat(9p-rpc): make StreamHandle futures Send so StreamMount goes native (#670)

ewindisch:
## Summary

W3 mechanical prereq for epic #809 — unblocks #819.

The `/stream/{topic}` pipe reader (`StreamMount` in `hyprstream-rpc-std`) was `wasm32`-only because its `Mount::read`/`write` path awaits `StreamHandle::next_payload`/`cancel`, and the `StreamHandle` trait in `hyprstream-rpc` was `#[async_trait(?Send)]`. A native `Send + Sync` `Mount` cannot hold a `!Send` future across `.await`.

## Investigation — why was `StreamHandle` `?Send`?

The only implementors of the `StreamHandle` trait are:
- the blanket `impl<T: Transport> StreamHandle for StreamHandleImpl<T>`
- `impl StreamHandle for Box<dyn StreamHandle>`

(`WasmStreamHandle` in `wasm_rpc_client.rs` wraps `StreamHandleImpl<WtConnection>` but calls its **inherent** methods, not the trait — unaffected.)

`StreamHandleImpl<T>` is built entirely from the `Transport` associated types, which are already `Send`-bound (`Transport::Sub: Send`, `Transport::Pub: Send`). So on **native** the futures are genuinely `Send`.

**The `?Send` was load-bearing on wasm, though:** the WebTransport `WtPublisher::send_frames` future holds JS values (`web_sys` writer) and is genuinely `!Send`. A blanket drop of `?Send` breaks the wasm build (confirmed via `cargo check --target wasm32-unknown-unknown`).

## Approach chosen

Mirror exactly what the `hyprstream_vfs::Mount` trait already does — a per-target `cfg_attr` split — rather than a blanket drop:

- `StreamHandle`: `#[async_trait]` (Send) on native, `#[async_trait(?Send)]` on wasm.
- `cancel` changed from `&self` to `&mut self`: a `dyn StreamHandle` is `Send` but not `Sync` (its transport `Sub`/`Pub` are only `Send`-bound), so a `&self` future can never be `Send`. `&mut self` is `Send` because `Self: Send`, and it matches the state `cancel` mutates (the cancelled flag). `next_payload` was already `&mut self`.
- Un-gate `StreamMount` for native (`Send + Sync`), using the same `Mount` `cfg_attr` so wasm stays `?Send`.
- Drop a dead `is_completed` binding that the native compile exposed as an unused-var warning.

The constraining implementor is `WtPublisher` (its `!Send` `send_frames` future) — that's why the trait can't be unconditionally `Send`.

## Tests

Added native tests in `stream_mount.rs`:
- `stream_mount_is_native_send_sync` — `StreamMount`/`StreamRegistry` are `Send + Sync` and usable as `Arc<dyn Mount>`.
- `stream_mount_read_future_is_send` — the `Mount::read` future (which holds `Box<dyn StreamHandle>` across `next_payload().await`) is `Send`. This is the exact regression gate.

## Build results

- `cargo check -p hyprstream-rpc` — OK
- `cargo check -p hyprstream-rpc-std` — OK
- `cargo clippy -p hyprstream-rpc --all-targets -- -D warnings` — clean
- `cargo clippy -p hyprstream-rpc-std --all-targets -- -D warnings` — clean
- `cargo test -p hyprstream-rpc-std --lib stream_mount` — 2 passed
- `cargo test -p hyprstream-rpc --lib stream` — 69 passed
- `cargo check --target wasm32-unknown-unknown -p hyprstream-rpc` — OK
- `cargo check --target wasm32-unknown-unknown -p hyprstream-rpc-std` — OK (wasm path preserved)

Refs #670, epic #809.

### DIFF
--- a/crates/hyprstream-rpc-std/src/stream_mount.rs
+++ b/crates/hyprstream-rpc-std/src/stream_mount.rs
@@ -1,12 +1,40 @@
-//! VFS Mount for streaming data via named pipes.
+//! VFS Mount for streaming data — streams as 9P files (epic #809 W3, #819).
 //!
 //! `/stream/{topic}/data` — read stream data (sequential, blocks until data/EOF)
 //! `/stream/{topic}/info` — read stream metadata (JSON)
-//! `/stream/{topic}/ctl`  — write control commands (cancel)
+//! `/stream/{topic}/qos`  — read/select QoS (StreamOpt vocabulary as tokens)
+//! `/stream/{topic}/ctl`  — write control commands (`cancel`, `qos …`, `resume …`)
 //!
 //! Follows Plan 9 `/net/tcp/{n}/data` pattern. Streams are created by
 //! dispatching streaming RPC methods via `RpcClient::open_stream()` and
 //! registered in the StreamRegistry.
+//!
+//! # Interface, not a byte-copy path (#819)
+//!
+//! This mount is the *interface*: naming, addressing, and open-time
+//! authorization. It is **not** a memcpy path that replaces the moq carrier.
+//! The registered [`StreamHandle`] IS the moq carrier subscription established
+//! at open — QUIC-native delivery (per-object streams, drop-stale, datagram,
+//! relay rendezvous) stays intact underneath. Two consumption modes ride the
+//! same handle:
+//!
+//! - **Floor (`Tread` on `data`)**: blocking push semantics, exactly Plan 9
+//!   `/net/tcp/n/data`. Always works; correct and sufficient for low-rate
+//!   streams (tokens). This is the floor, **not** the ceiling.
+//! - **Carrier handle (bulk)**: high-rate / bulk consumers take the
+//!   carrier-backed handle directly ([`StreamRegistry::take_handle`]) and drive
+//!   the moq subscriber, exactly as #817 hands out mmap-at-open for weights.
+//!   The frames are **never** routed one-by-one through `Tread` — that would
+//!   flatten moq's non-fungible delivery semantics to a lowest-common-
+//!   denominator read, the anti-pattern this design exists to avoid.
+//!
+//! QoS that does not reduce to reliable-in-order is selected through a small
+//! vocabulary ([`StreamQos`]) on `qos` / `ctl` — StreamOpt's existing encoding
+//! (latest-wins, lossy-ok, drop-stale) moved *onto the file*, so the carrier's
+//! rich semantics stay first-class through the interface.
+//!
+//! Events (group-keyed EventService, #600) ride this same surface via
+//! [`StreamKind::Event`] rather than a separate plane — see [`StreamEntry`].
 
 use std::collections::HashMap;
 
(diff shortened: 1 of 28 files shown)


crates/hyprstream-rpc-std/src/stream_mount.rs
```diff
@@ -109,26 +109,26 @@ enum StreamFidState {
...
-/// generic service mount can register streams on any target.
-#[cfg(target_arch = "wasm32")]
+/// Native + wasm (#670): the `Mount` `read`/`write` path awaits
+/// [`StreamHandle`] methods, which now return `Send` futures (the
+/// `StreamHandle` trait dropped its `?Send`). That lets this mount be a
+/// genuinely `Send + Sync` `Mount` on native — served to a Cloud Hypervisor
+/// guest or any native namespace — while `#[async_trait(?Send)]` on wasm keeps
+/// the browser path working. `StreamRegistry`/`StreamEntry` above are
+/// target-agnostic so the generic service mount can register streams on any
+/// target.
```
ewindisch, line 436: Non-blocking, but worth a follow-up issue when this mount gets bound into a native namespace (#819): `StreamEntry.owner` is documented "for access control" but nothing enforces it — every method ignores `_caller`, `readdir` uses `list_topics(None)`, and any Subject can read any topic's decrypted payloads. On the wasm-only single-user browser namespace that was moot; a native `Send + Sync` mount served to a CH guest or multi-tenant namespace makes it a real cross-tenant surface. Per the ratified interface policy the PEP belongs at the AVC seam (#568/#767), so no change requested here — but an owner-vs-Subject filter in `readdir`/`read` would be cheap defense-in-depth, and at minimum the wiring PR that binds this natively should not ship without one or the other.

crates/hyprstream-rpc-std/src/stream_mount.rs
```diff
@@ -206,8 +206,7 @@ impl Mount for StreamMount {
                 // Read next verified payload (await without holding borrow)
                 let result = handle.next_payload().await;
 
-                // Put handle back
-                let is_completed = handle.is_completed();
+                // Put handle back (its completion state is tracked internally).
```
ewindisch, line 544: Pre-existing race that true parallelism now makes reachable: if a ctl `cancel` runs while a reader has the handle checked out, `registry.remove(topic)` returns the entry with `handle: None`, so `handle.cancel()` is never awaited and the producer never receives the cancel control message; the reader's put-back then finds no entry and silently drops the handle (transport cleanup via Drop only, no authenticated ctrl-cancel). Net effect is a leaked producer-side stream until it times out, not a safety issue — fine to leave for a follow-up, just flagging that the take/put-back window is no longer theoretical on native.

ewindisch (commented):
**Verdict: LGTM — safe to automerge.**

Reviewed for the four risk areas that matter for a `?Send` drop, plus scope and conflicts.

**Send soundness — verified, not papered over.** `StreamHandleImpl<T>` is composed entirely of `Send` state: `T::Sub`/`T::Pub` carry `Send` bounds on the `Transport` trait (`transport_traits.rs`), and the rest is a `StreamVerifier` (plain crypto state), `VecDeque<StreamPayload>` (byte buffers), `Arc<AtomicBool>`, strings, and key arrays. Nothing torch-adjacent — `hyprstream-rpc` has no tch dependency, so no `!Send` tensors can reach these futures. The `cfg_attr` split (Send native, `?Send` wasm) correctly preserves the one genuinely `!Send` implementor path (`WtPublisher::send_frames` holding JS values), and matches the existing `hyprstream_vfs::Mount` pattern. Crucially there are **no new `unsafe impl Send/Sync`** — the Send-ness falls out of real bounds. The pre-existing wasm32-only `unsafe impl`s (`WtConnection`, `JsSigner`, etc.) are untouched and irrelevant to the native path this PR enables.

**Lock discipline — clean.** `StreamRegistry` is a `parking_lot::Mutex`, and every `read`/`write` path uses take-handle → drop guard → await → re-lock → put-back. No guard is held across an await, so going multi-threaded introduces no deadlock or blocking-executor hazard. A concurrent second reader on the same topic gets a clean `Io("stream read in progress")` rather than blocking or corrupting state.

**`cancel(&self)` → `&mut self`** is the right call: `dyn StreamHandle` is `Send` but not `Sync` (Sub/Pub are only `Send`-bound), so a `&self` future could never be `Send`; `&mut self` also matches the mutation. All callers are updated (`wasm_rpc_client.rs`, `stream_mount.rs`); the only other implementors are the two in-crate blanket impls, confirmed by grep. The `WasmStreamHandle::cancel` JS-facing borrow semantics are effectively unchanged (an in-flight `nextPayload(&mut self)` already excluded a concurrent `cancel` under wasm-bindgen's runtime borrow check).

**Subject / MAC (ADR #651)** — every `Mount` method still takes `&Subject`; signatures are preserved and nothing MAC-relevant is dropped by the refactor. See the inline note on the pre-existing (now more relevant) unenforced `owner` field.

**Scope / conflicts.** Matches #670's stated fix exactly (make `StreamHandle` futures Send, un-gate `StreamMount`). No file overlap with #838 (namespace_builder) or #767 (9P attach-time MAC seam); #767 is complementary — it supplies the attach-time Subject derivation that a natively-served StreamMount would eventually consume.

Two non-blocking observations inline; both are pre-existing behavior that native un-gating makes more relevant, and both fit better as follow-up issues than as scope creep here.

ewindisch:
ℹ️ **Reviewed (1:7 sweep) — nothing to merge.** This branch's diff against main is empty; the #670 StreamHandle-Send content already landed on main as commit 899846813 ("feat(9p-rpc): make StreamHandle futures Send… (#670)"). Reviewed the landed content: **no findings** — Send-ness is compiler-proven via the async_trait cfg split (no `unsafe impl`, no lock held across await, no tch-rs capture; `cancel` `&self`→`&mut self` is correct, wasm `?Send` path preserved). Recommend the owner **close this PR as already-landed** rather than merging an empty diff (I'm not authorized to close it).

ewindisch:
Closing per coordination decision: this branch's diff against main is empty — the #670 StreamHandle-Send content already landed on main as commit 899846813. Reviewed the landed content: no findings (Send-ness compiler-proven via the async_trait cfg split; no unsafe impls, no lock-across-await, no tch-rs capture). Nothing to merge.

Issue #784, opened 2026-07-04
K1b: Helm observability — OTel Collector with Prometheus exposition (autoscaling signal source)


ewindisch:
Part of #778 (Phase 0). Sub-item of K1 (#779).

**K1b — the metrics path that makes autoscaling (K6a) possible.**

Verified gap: hyprstream has **no Prometheus exposition surface** — the `otel` feature is OTLP-push only (`opentelemetry-otlp`, `crates/hyprstream/Cargo.toml`). HPA/KEDA need a scrapeable/queryable signal.

### Deliverables
- Chart-deployed **OpenTelemetry Collector** (or values-wired external one): OTLP receiver ← hyprstream services; Prometheus exporter → scrape endpoint (kube-prometheus-stack compatible ServiceMonitor optional).
- Values plumbing: `OTEL_EXPORTER_OTLP_ENDPOINT` (or the config equivalent) pointed at the collector for every service.
- Document the signal set relevant to scaling (request rate, stream backpressure/queue depth, tokens/s) and which are exported today vs. need instrumentation follow-ups (file follow-ups, don't scope-creep here).
- Non-goal: a native `/metrics` endpoint in hyprstream (later nicety, explicitly out of scope).

### Acceptance
- On kind: OAI request traffic is visible via the collector's Prometheus endpoint; a sample HPA can consume one custom metric end-to-end (via prometheus-adapter or KEDA prometheus scaler).

### Depends on / consumes
- K1 (#779). Consumed by: K6a (HPA/KEDA).

Issue #783, opened 2026-07-04
K1a: Helm key/trust bootstrap — per-service signing keys, trust store, node DID in Kubernetes


ewindisch:
Part of #778 (Phase 0). Sub-item of K1 (#779).

**K1a — the identity/key story for a multi-pod deployment.** Every RPC request is a hybrid-PQC-signed envelope verified against a per-service trust store (`hyprstream-service::trust_store`); pods are ephemeral, so key material and trust must be provisioned, not accumulated on disk.

### Deliverables
- Per-service signing keys (Ed25519 + ML-DSA-65 seed material) as K8s Secrets / projected volumes; chart-level generation job or external-secrets hook; rotation story documented.
- Trust-store bootstrap: how each service learns its peers' verifying keys in-cluster — this MUST implement epic #441's invariant (registered-and-attested, fail-closed, never consumer-re-derived), consuming the #759/PR #774 resolution model rather than forking it.
- Node identity: the deployment's DID (`did:web` at the FQDN) served consistently; QUIC cert/secret naming per `QuicConfig` (`quic-cert`/`quic-key` secrets already distinct from HTTP TLS).
- Fail-closed: a service that cannot resolve a peer's key refuses, never accepts-unverified.

### Acceptance
- Fresh `helm install` on kind: all services come up with distinct keys, cross-service RPC verifies, killing and rescheduling a pod does not break trust (keys are provisioned, not pet state); no "Response signed by unexpected key" class failures (#441 failure #3 / #759).

### Depends on / consumes
- K1 (#779) chart skeleton.
- **Governed by epic #441 (authoritative key binding — one identity→key, fail-closed, no silent re-derivation).** #759 / PR #774 is the leaf fix for #441's failure #3 ("service key never registered → 'Response signed by unexpected key'") — consume its resolution model; do not track the symptom in isolation.
- Consumed by: K5b (operator's in-cluster RPC client), every service-to-service path in the chart.

ewindisch:
Cross-reference from the PR #774 second-opinion review + a systemd secrets audit — three things worth deciding **before** this chart consumes the #759/PR #774 resolution model as-is:

[...]

Issue #779, opened 2026-07-04
K1: Helm chart — service-per-Deployment over QUIC/iroh (no UDS topology requirement)


ewindisch:
Part of #778 (Phase 0).

**K1 — Helm chart: the deployment substrate every other K8s axis sits on.**

Translate the compose topology (event/registry/policy/streams/model/oai/flight/worker) to Kubernetes as **one Deployment per service**, communicating over the networked transport (QUIC + iroh, both already on by default via `QuicConfig` — self-provisioned certs confirmed in `crates/hyprstream/src/config/mod.rs`). **UDS is explicitly not a topology requirement** — it remains only an incidental same-host optimization, never assumed by the chart (ratified in #778 non-goals).

### Deliverables
- `charts/hyprstream/`: Deployments per service, Services, ConfigMap-driven `HYPRSTREAM__*` env, PVC for `/var/lib/hyprstream` (models via XET/CAS or PVC), GPU variants via image tag + `nodeSelector`/`resources` (nvidia device plugin; ROCm `/dev/kfd` via device plugin), no systemd (container builds already exclude it).
- Service dependency ordering expressed via readiness probes, not init hacks.
- Values toggles per service (flight/worker optional; worker documents its privilege needs).
- CI: chart lint + `helm template` golden test; kind-based smoke job if cheap.

### Acceptance
- `helm install` on a stock cluster brings up event/registry/policy/streams/model(cpu)/oai; `/v1/models` answers through the oai Service; no shared-volume IPC anywhere in the chart.

### Depends on / consumes
- Soft-gated by #759 / PR #774 (multi-service key resolution) — the same trust-bootstrap problem class; K1a (filed separately) owns the key/Secret story.
- Sub-items: K1a (key/trust bootstrap), K1b (observability) — separate issues under #778.

ewindisch:
**Blocker found: service-per-Deployment is structurally impossible on current `main`.**

This issue specifies *"one Deployment per service, communicating over the networked transport… UDS is explicitly not a topology requirement."* The code disagrees.

`worker` and `model` obtain their PolicyClient **only** via `PolicyClient::for_local_bootstrap` (`services/factories.rs:935`, `:1086`), and the generated constructor refuses network transports by construction:

```rust
anyhow::ensure!(
    matches!(&transport.endpoint, EndpointType::Inproc{..} | EndpointType::Ipc{..} | EndpointType::SystemdFd{..}),
[...]

Issue #778, opened 2026-07-04
Epic: Kubernetes integration — mutual namespace binding (Helm, SPIFFE→WIT, 9P CSI, k8s SandboxBackend, operator/CRDs, serving, /k8s VFS, RAFS substrate)


ewindisch:
Epic for **maximalist, loosely-coupled, well-integrated Kubernetes support across multiple axes**: hyprstream and Plan 9 systems can drive Kubernetes, and Kubernetes can drive hyprstream — deploy, identity, filesystem, compute, control plane, serving, and image axes, each independently useful, all sharing one primitive and one security spine.

> **Status: decomposition ratified; sub-issues filed below.** Explicitly complements (does not relitigate) epic #508's runtime-side K8s work (`CriBackend` #510). This epic is the north-south/control-plane half.

## 1. The primitive: mutual namespace binding

The K8s API and the hyprstream VFS are two namespaces; we integrate by binding each under the other, **subspaced by origin**, so names never collide:

- **hyprstream → K8s:** resources project as CRDs under hyprstream-owned API groups — `models.hyprstream.io/{Model,Adapter}`, `training.hyprstream.io/TrainingRun`, `serving.hyprstream.io/InferenceService`, `mesh.hyprstream.io/TenantBinding`. API groups are DNS-subspaced by construction: `serving.hyprstream.io/InferenceService` cannot collide with `serving.kserve.io/InferenceService`; KServe becomes an optional adapter target, never the owner of our nouns.
- **K8s → hyprstream:** cluster objects project as files under `/k8s/...` (the `/exec/instances` pattern, epic #608, over a kube-rs watch).

## 2. The security spine (one credential pipeline)

```
pod projected SA token / SPIFFE JWT-SVID
  → WIT (wit+jwt; docs/spiffe.md already treats WIT ≈ JWT-SVID)
  → verified Claims × VerifiedKeyMaterial
  → SecurityContext derived ONCE at 9P Tattach, cached fid-scoped (#568)
  → AVC enforces the MAC floor on every op
```

K8s supplies compute/scheduling/network isolation; hyprstream supplies **mandatory information-flow control on the data a pod touches** — the pod never opts in, because the PEP lives at the 9P attach, not the container boundary. Hybrid-PQC (#550) seals the channel at the app layer regardless of the mesh's TLS. `federation:register` is the single gate deciding whether a SPIFFE trust domain / cluster issuer is admitted at all.

[...]

PR #846, opened 2026-07-05
feat(k8s) K1: Helm chart — service-per-Deployment over QUIC/iroh (#779)

ewindisch:
Implements **K1 (#779)** under **Epic #778, Phase 0**: a Helm chart at `charts/hyprstream/` deploying hyprstream to Kubernetes as **one Deployment + one ClusterIP Service per service** ("service-per-Deployment"), addressing each other by Kubernetes DNS Service name over the networked transport (**QUIC over UDP / iroh**) — **no shared-volume UDS anywhere in the chart** (per the #778 non-goal; UDS stays an incidental same-host optimization only).

### Service set modeled (source of truth)
Mirrors the real inventory-registered `#[service_factory(...)]` set in `crates/hyprstream/src/services/factories.rs`. Map keys are the exact names passed to `hyprstream service start <name> --foreground`:

- **Default-enabled** (matches the #779 acceptance path + transitive `depends_on`): `policy`, `discovery`, `registry`, `model` (cpu), `event`, `streams`, `oai`. (`discovery` is on because `registry`/`model`/`oai` depend on it.)
- **Opt-in toggles** (`enabled: false`): `oauth`, `mcp`, `xet`, `flight`, `metrics`, `tui`, `worker`. `worker` documents its elevated-privilege needs (device access / userns / `/dev/kvm`) which the default hardened `securityContext` deliberately does not grant.

QUIC/RPC services expose a **UDP** port; HTTP surfaces (`oai`/`xet`/`flight`) expose **TCP**. `oai` gets a real readiness/liveness probe (`httpGet /health`, from `crates/hyprstream/src/server/mod.rs`).

### What's in the chart
- `Chart.yaml`, `values.yaml` (fully commented), `_helpers.tpl`, `serviceaccount.yaml`, `configmap.yaml` (shared `HYPRSTREAM__*` env), `pvc.yaml` (shared `/var/lib/hyprstream` data dir), and **DRY** `deployments.yaml` / `services.yaml` that `range` over `.Values.services` merging `.Values.defaults`.
- **GPU variants**: `services.model.variant: cpu|cuda|rocm` appends the image suffix (matching the Dockerfile VARIANT stages) and is the anchor for `resources.limits.nvidia.com/gpu` + `nodeSelector` (nvidia/ROCm device plugins).
- **Env scheme**: `HYPRSTREAM__<SECTION>__<FIELD>` (config crate, double-underscore) — e.g. `HYPRSTREAM__POLICY__QUIC_PORT`, `HYPRSTREAM__OAI__PORT`, `HYPRSTREAM__STORAGE__MODELS_DIR`.
- **Golden render** committed at `charts/hyprstream/tests/golden-default.yaml`.

### Extension hooks left (NOT implemented here — separate issues)
- `keyBootstrap.*` → **#783 (K1a)**: per-service signing-key Secrets, trust store, node DID. Chart reserves the shape + a stable `HYPRSTREAM__SECRETS__PATH`; a `volumeMounts` TODO marker sits in `deployments.yaml`.
- `observability.*` → **#784 (K1b)**: OTel Collector → Prometheus exposition. Chart reserves the shape + a ConfigMap TODO for `OTEL_EXPORTER_OTLP_ENDPOINT`.

### Image assumption
No published image reference exists in-repo yet — only a local multi-variant distroless `Dockerfile` that tags nothing. `image.repository` defaults to a **placeholder** `ghcr.io/hyprstream/hyprstream`; `image.tag` defaults to `Chart.appVersion`; variant suffixes (`-cpu`/`-cuda128`/`-rocm71`) match the Dockerfile. Override for your registry or wire CI to publish there.

### Known gap (flagged, not a chart defect)
Cross-pod QUIC service-to-service **peer resolution** — feeding K8s DNS names into the RPC resolver so `for_service("policy")` dials `hyprstream-policy` — is **not yet wired in hyprstream itself**. Today the multi-process model resolves peers to UDS paths in a shared runtime dir (same-host only; `EndpointRegistry::default_endpoint`, `crates/hyprstream-rpc/src/registry/mod.rs`), and QUIC is a WebTransport/announce plane with no static peer-address config. This chart delivers the **Phase-0 deployment substrate**; a working end-to-end install additionally needs that app-side resolver wiring plus the #783 key/trust bootstrap (consistent with #779 being "soft-gated by #759/#774"). Documented in the chart README + `NOTES.txt`.

### Validation performed
- `helm lint charts/hyprstream` — clean.
- `helm template` — renders (17 objects default; 31 with all services enabled + `model.variant=cuda`, image resolves to `...:0.1.0-cuda128`).
- `kubeconform -strict` (built-in kinds, k8s 1.31) — 17/17 and 31/31 valid.
- `helm install --dry-run=server` against a live **kind** cluster (`kind-hyprstream`) — renders against the API server without error.

Part of #779. Epic #778.

### DIFF
--- a/charts/hyprstream/templates/configmap.yaml
+++ b/charts/hyprstream/templates/configmap.yaml
@@ -24,8 +24,9 @@ data:
   {{- end }}
   {{- end }}
   {{- if .Values.observability.enabled }}
-  # TODO(#784): observability wiring populates OTEL_EXPORTER_OTLP_ENDPOINT here.
-  {{- with .Values.observability.otelExporterOtlpEndpoint }}
-  OTEL_EXPORTER_OTLP_ENDPOINT: {{ . | quote }}
-  {{- end }}
+  # K1b (#784): turn on the OTLP trace exporter and point it at the collector.
+  # hyprstream only exports OTLP when HYPRSTREAM_OTEL_ENABLE=true (main.rs); the
+  # collector derives Prometheus-scrapeable RED metrics from those spans.
+  HYPRSTREAM_OTEL_ENABLE: "true"
+  OTEL_EXPORTER_OTLP_ENDPOINT: {{ include "hyprstream.otlpEndpoint" . | quote }}
   {{- end }}
--- a/charts/hyprstream/templates/configmap.yaml
+++ b/charts/hyprstream/templates/configmap.yaml
@@ -1,0 +1,28 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "hyprstream.fullname" . }}-config
+  labels:
+    {{- include "hyprstream.labels" . | nindent 4 }}
+data:
+  # Shared HYPRSTREAM__* env for every service. Config crate maps
+  # HYPRSTREAM__<SECTION>__<FIELD> onto HyprConfig.
+  HYPRSTREAM__STORAGE__MODELS_DIR: {{ printf "%s/models" .Values.config.dataDir | quote }}
+  HYPRSTREAM__STORAGE__LORAS_DIR: {{ printf "%s/loras" .Values.config.dataDir | quote }}
+  HYPRSTREAM__STORAGE__CACHE_DIR: {{ printf "%s/cache" .Values.config.dataDir | quote }}
+  HYPRSTREAM__STORAGE__CONFIG_DIR: {{ printf "%s/config" .Values.config.dataDir | quote }}
+  HYPRSTREAM__SECRETS__PATH: {{ .Values.keyBootstrap.secretsPath | quote }}
+  HYPRSTREAM__QUIC__ENABLED: {{ .Values.config.quic.enabled | quote }}
+  HYPRSTREAM__QUIC__IROH: {{ .Values.config.quic.iroh | quote }}
+  HYPRSTREAM__QUIC__BIND_ADDR: {{ .Values.config.quic.bindAddr | quote }}
+  {{- with .Values.config.extraEnv }}
+  {{- range $k, $v := . }}
+  {{ $k }}: {{ $v | quote }}
+  {{- end }}
+  {{- end }}
+  {{- if .Values.observability.enabled }}
+  # TODO(#784): observability wiring populates OTEL_EXPORTER_OTLP_ENDPOINT here.
+  {{- with .Values.observability.otelExporterOtlpEndpoint }}
+  OTEL_EXPORTER_OTLP_ENDPOINT: {{ . | quote }}
+  {{- end }}
+  {{- end }}
--- a/charts/hyprstream/templates/configmap.yaml
+++ b/charts/hyprstream/templates/configmap.yaml
@@ -11,7 +11,10 @@ data:
   HYPRSTREAM__STORAGE__LORAS_DIR: {{ printf "%s/loras" .Values.config.dataDir | quote }}
   HYPRSTREAM__STORAGE__CACHE_DIR: {{ printf "%s/cache" .Values.config.dataDir | quote }}
   HYPRSTREAM__STORAGE__CONFIG_DIR: {{ printf "%s/config" .Values.config.dataDir | quote }}
+  HYPRSTREAM__TLS__ENABLED: {{ .Values.config.tls.enabled | quote }}
+  {{- if .Values.keyBootstrap.enabled }}
   HYPRSTREAM__SECRETS__PATH: {{ .Values.keyBootstrap.secretsPath | quote }}
+  {{- end }}
   HYPRSTREAM__QUIC__ENABLED: {{ .Values.config.quic.enabled | quote }}
   HYPRSTREAM__QUIC__IROH: {{ .Values.config.quic.iroh | quote }}
   HYPRSTREAM__QUIC__BIND_ADDR: {{ .Values.config.quic.bindAddr | quote }}
--- a/charts/hyprstream/templates/deployments.yaml
+++ b/charts/hyprstream/templates/deployments.yaml
@@ -48,11 +48,15 @@ spec:
           envFrom:
             - configMapRef:
                 name: {{ include "hyprstream.fullname" $root }}-config
-          {{- with $svc.env }}
+          {{- if or $svc.env (and $root.Values.keyBootstrap.enabled $root.Values.keyBootstrap.issuerUrl) }}
           env:
-            {{- range $k, $v := . }}
+            {{- range $k, $v := $svc.env }}
             - name: {{ $k }}
               value: {{ $v | quote }}
+            {{- end }}
+            {{- if and $root.Values.keyBootstrap.enabled $root.Values.keyBootstrap.issuerUrl }}
+            - name: HYPRSTREAM__OAUTH__ISSUER_URL
+              value: {{ $root.Values.keyBootstrap.issuerUrl | quote }}
             {{- end }}
           {{- end }}
           {{- with $svc.ports }}
@@ -83,7 +87,9 @@ spec:
             - name: data
               mountPath: {{ $root.Values.config.dataDir }}
             {{- if $root.Values.keyBootstrap.enabled }}
-            # TODO(#783): mount per-service signing-key Secret at secretsPath.
+            - name: credentials
+              mountPath: {{ $root.Values.keyBootstrap.secretsPath }}
+              readOnly: true
             {{- end }}
       {{- with (default $defaults.nodeSelector $svc.nodeSelector) }}
       nodeSelector:
@@ -105,5 +111,34 @@ spec:
           {{- else }}
           emptyDir: {}
           {{- end }}
+        {{- if $root.Values.keyBootstrap.enabled }}
+        - name: credentials
+          projected:
+            defaultMode: 0400
+            sources:
+              - secret:
+                  name: {{ include "hyprstream.trustSecretName" $root }}
+                  items:
+                    - key: ca-pubkey
+                      path: ca-pubkey
+                    - key: bootstrap-pubkeys
+                      path: bootstrap-pubkeys
+              - secret:
+                  name: {{ include "hyprstream.serviceCredentialSecretName" $ctx }}
+                  items:
+                    - key: signing-key
+                      path: signing-key
+                    {{- if ne $name "policy" }}
+                    - key: service-jwt
+                      path: service-jwt
+                    {{- end }}
+              {{- if eq $name "policy" }}
+              - secret:
+                  name: {{ include "hyprstream.policyCaSecretName" $root }}
+                  items:
+                    - key: ca-key
+                      path: ca-key
+              {{- end }}
+        {{- end }}
 {{- end }}
 {{- end }}
--- a/charts/hyprstream/templates/deployments.yaml
+++ b/charts/hyprstream/templates/deployments.yaml
@@ -48,11 +48,15 @@ spec:
           envFrom:
             - configMapRef:
                 name: {{ include "hyprstream.fullname" $root }}-config
-          {{- if or $svc.env (and $root.Values.keyBootstrap.enabled $root.Values.keyBootstrap.issuerUrl) }}
+          {{- if or $svc.env $root.Values.observability.enabled (and $root.Values.keyBootstrap.enabled $root.Values.keyBootstrap.issuerUrl) }}
           env:
             {{- range $k, $v := $svc.env }}
             - name: {{ $k }}
               value: {{ $v | quote }}
+            {{- end }}
+            {{- if $root.Values.observability.enabled }}
+            - name: OTEL_SERVICE_NAME
+              value: {{ include "hyprstream.svcName" $ctx | quote }}
             {{- end }}
             {{- if and $root.Values.keyBootstrap.enabled $root.Values.keyBootstrap.issuerUrl }}
             - name: HYPRSTREAM__OAUTH__ISSUER_URL
--- a/charts/hyprstream/templates/deployments.yaml
+++ b/charts/hyprstream/templates/deployments.yaml
@@ -114,7 +114,7 @@ spec:
         {{- if $root.Values.keyBootstrap.enabled }}
         - name: credentials
           projected:
-            defaultMode: 0400
+            defaultMode: 0440
             sources:
               - secret:
                   name: {{ include "hyprstream.trustSecretName" $root }}
--- a/charts/hyprstream/templates/deployments.yaml
+++ b/charts/hyprstream/templates/deployments.yaml
@@ -1,0 +1,109 @@
+{{- $root := . -}}
+{{- $defaults := .Values.defaults -}}
+{{- range $name, $svc := .Values.services }}
+{{- if $svc.enabled }}
+{{- $ctx := dict "root" $root "name" $name }}
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "hyprstream.svcName" $ctx }}
+  labels:
+    {{- include "hyprstream.labels" $root | nindent 4 }}
+    app.kubernetes.io/component: {{ $name }}
+spec:
+  replicas: {{ $svc.replicas | default $defaults.replicas }}
+  selector:
+    matchLabels:
+      {{- include "hyprstream.selectorLabels" $root | nindent 6 }}
+      app.kubernetes.io/component: {{ $name }}
+  template:
+    metadata:
+      labels:
+        {{- include "hyprstream.selectorLabels" $root | nindent 8 }}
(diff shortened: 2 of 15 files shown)


charts/hyprstream/templates/configmap.yaml
```diff
@@ -0,0 +1,28 @@
...
+  labels:
+    {{- include "hyprstream.labels" . | nindent 4 }}
+data:
+  # Shared HYPRSTREAM__* env for every service. Config crate maps
+  # HYPRSTREAM__<SECTION>__<FIELD> onto HyprConfig.
+  HYPRSTREAM__STORAGE__MODELS_DIR: {{ printf "%s/models" .Values.config.dataDir | quote }}
+  HYPRSTREAM__STORAGE__LORAS_DIR: {{ printf "%s/loras" .Values.config.dataDir | quote }}
+  HYPRSTREAM__STORAGE__CACHE_DIR: {{ printf "%s/cache" .Values.config.dataDir | quote }}
+  HYPRSTREAM__STORAGE__CONFIG_DIR: {{ printf "%s/config" .Values.config.dataDir | quote }}
+  HYPRSTREAM__SECRETS__PATH: {{ .Values.keyBootstrap.secretsPath | quote }}
```
ewindisch, line 16: **Blocking.** This is set unconditionally on every service (all pods consume this ConfigMap via `envFrom`), and it points inside `config.dataDir` — the shared RW PVC.

Per `main.rs:2020`, mere *presence* of `HYPRSTREAM__SECRETS__PATH` flips `systemd_mode`, and in the `--ipc` path that makes **every** service load the flat root/node `signing-key` via `load_or_generate_node_signing_key` instead of its per-service key (#805). On a shared writable volume, the first pod to boot mints the root key and all services then share one signing identity; the policy CA key lives in the same dir writable by every pod (#802 pattern). Even with today's non-`--ipc` args it isn't inert: `services/factories.rs::credentials_dir()` honors this var for service-JWT resolution while the key path doesn't — the #804 split-brain.

Please gate this line on `{{- if .Values.keyBootstrap.enabled }}` (off by default, owned by #783) so the default install sets no secrets path at all.

charts/hyprstream/values.yaml
```diff
@@ -0,0 +1,338 @@
...
+# Per-service signing-key Secrets, CA/trust store, node DID. Intentionally NOT
+# implemented here (separate issue). When keyBootstrap.enabled is true, #783
+# will project per-service Secret volumes into HYPRSTREAM__SECRETS__PATH and set
+# the node DID. This chart only reserves the shape + a stable secrets mount path.
+# -----------------------------------------------------------------------------
+keyBootstrap:
+  enabled: false
+  # Directory each service reads signing keys / bootstrap pubkeys / JWTs from
+  # (HYPRSTREAM__SECRETS__PATH). #783 owns populating it from Secrets.
+  secretsPath: /var/lib/hyprstream/credentials
```
ewindisch, line 97: This default path is inside `config.dataDir`, i.e. on the shared PVC mounted read-write by every service pod. When #783 populates it, per-service signing keys and the CA key would be readable/writable across all services — defeating per-service keys entirely, and clashing with `load_or_generate_*`'s write-if-missing behavior (any pod can mint into it). Suggest defaulting to a non-PVC path (e.g. `/etc/hyprstream/credentials`) where #783 projects per-service Secrets read-only, and adding a comment cross-referencing #805 (the presence-of-this-env-var = systemd-mode hazard) so K1a inherits the constraint.

charts/hyprstream/templates/deployments.yaml
```diff
@@ -0,0 +1,109 @@
...
+      {{- end }}
+      containers:
+        - name: {{ $name }}
+          image: {{ include "hyprstream.image" (dict "root" $root "svc" $svc) }}
+          imagePullPolicy: {{ $root.Values.image.pullPolicy }}
+          args:
+            - service
+            - start
+            - {{ $name }}
+            - --foreground
```
ewindisch, line 47: Worth documenting precisely in the README known-gap section: `service start <name> --foreground` **without** `--ipc` takes main.rs's single-process branch — `EndpointMode::Inproc`, secrets from `dirs::config_dir()/hyprstream/credentials` (this branch ignores `HYPRSTREAM__SECRETS__PATH`), and a hard requirement on wizard-created CA credentials ("CA key not found — run 'hyprstream wizard' first"). In a distroless pod none of that exists, so every default pod CrashLoopBackOffs before the peer-resolution gap is even reached. And note that switching these args to `--ipc` later must not happen while the ConfigMap sets `HYPRSTREAM__SECRETS__PATH` globally, or every service loads the flat root key (#805).

charts/hyprstream/values.yaml
```diff
@@ -0,0 +1,338 @@
...
+  policy:
+    enabled: true
+    kind: rpc              # QUIC/RPC (informational)
+    ports:
+      - name: quic
+        port: 7001
+        protocol: UDP
+    env:
+      HYPRSTREAM__POLICY__QUIC_PORT: "7001"
+      HYPRSTREAM__QUIC__SERVER_NAME: "hyprstream-policy"
```
ewindisch, line 246: These `HYPRSTREAM__QUIC__SERVER_NAME` values hardcode `hyprstream-<svc>`, but the actual Service DNS names are `{{ fullname }}-<svc>` — for any release not named exactly `hyprstream` (or with `nameOverride`/`fullnameOverride` set) they'll be e.g. `myrel-hyprstream-policy`, so the TLS server name won't match the DNS name peers dial. Since values.yaml can't template, consider moving `SERVER_NAME` out of per-service `env` and rendering it in `deployments.yaml` via the `hyprstream.svcName` helper (same source of truth as the Service objects). Applies to all services in this map.

ewindisch (commented):
**Verdict: Blocking** — one change required before merge (drop/gate the unconditional `HYPRSTREAM__SECRETS__PATH` in the shared ConfigMap); everything else is documented-gap or minor. The chart structure itself (service-per-Deployment, QUIC/UDP Services, no shared-volume UDS, DRY range templates, honest known-gap section) is good K1 substrate work.

### Blocking: the ConfigMap sets `HYPRSTREAM__SECRETS__PATH` on every service, pointing into the shared RW PVC

`configmap.yaml` emits `HYPRSTREAM__SECRETS__PATH: /var/lib/hyprstream/credentials` unconditionally (even with `keyBootstrap.enabled: false`), and every Deployment consumes it via `envFrom`. That directory sits inside `config.dataDir`, i.e. on the shared PVC mounted read-write by **all** pods. This walks straight into #805/#802/#804:

- **#805**: `main.rs` treats *presence* of `HYPRSTREAM__SECRETS__PATH` as "systemd mode" (`main.rs:2020`). In the multi-process path (`--ipc`, the mode this per-pod topology is the k8s analog of), `systemd_mode=true` makes **every** service — not just policy — call `load_or_generate_node_signing_key(secrets_dir)` (`main.rs:2021-2025`), i.e. the flat root/node `signing-key`. Under this chart that file lives on the shared PVC: whichever pod boots first **mints the root key**, and every service then signs with the same root identity. The policy CA key (`ca-key`) sits in the same flat dir, readable and writable by every service pod. That's the #802 collision pattern realized cluster-wide.
- **#804**: even with today's args (no `--ipc`, so the main.rs key path ignores the var), the var is *not* inert: `services/factories.rs::credentials_dir()` honors `HYPRSTREAM__SECRETS__PATH` for service-JWT resolution while the single-process key path reads `dirs::config_dir()/hyprstream/credentials`. Setting it globally activates exactly the split-brain #804 describes (keys from one dir, JWTs from another).
- The PR body calls this "reserving the shape + a stable `HYPRSTREAM__SECRETS__PATH`" — but this env var is live, semantics-bearing config (it is the deployment-profile switch #805 wants to replace with an explicit `SecretsProfile`). A Helm chart freezing "this var is set on every pod, pointing at a shared writable volume" is precisely the deployment-model decision K1 must not freeze.

**Fix**: emit `HYPRSTREAM__SECRETS__PATH` only under `{{- if .Values.keyBootstrap.enabled }}` (which is off and owned by #783), and change the default `keyBootstrap.secretsPath` to a path **not** on the shared PVC (e.g. `/etc/hyprstream/credentials`, where #783 will project per-service Secrets read-only). With that, the default install sets no secrets path at all and the landmine is defused until #783/#805 define the real contract.

### Per-service key-loading trace (what actually happens under this chart)

Default install, args `service start <name> --foreground` (no `--ipc`), for **all seven** default services (policy, discovery, registry, model, event, streams, oai):

1. `main.rs` takes the **single-process** branch (`ipc=false`) → `EndpointMode::Inproc`.
2. `secrets_dir = dirs::config_dir()/hyprstream/credentials` (`HYPRSTREAM__SECRETS__PATH` ignored here) — nothing exists there in a distroless nonroot container.
3. `load_ca_verifying_key` fails → pod exits with "CA key not found — run 'hyprstream wizard' first" (`main.rs:2066-2067`).

So the default render **CrashLoopBackOffs on every pod** — no service loads any key. The README/NOTES known-gap section describes the missing piece as "peer resolution wiring", but the gap is earlier: the launch args select inproc single-process mode with wizard-required credentials, so pods never come up at all. Worth stating that plainly (and it means the #779 acceptance criterion — `/v1/models` answering through the oai Service — is not met by this PR alone, which the body already implies via the soft-gate on #759/#774).

If the args later become `--ipc` **without** the ConfigMap fix above: every service (policy and non-policy alike) loads the same flat shared-PVC root key — the #805 scenario. That's why the ConfigMap change is blocking now, not deferred to #783.

### Other findings

- **#803 (JWT rotation keys)** — not acknowledged in chart or PR body. Under this chart oauth's rotation keys land on the shared RW PVC (writable, so they persist — but shared with every other service). Once #783 moves credentials to read-only Secret mounts, oauth needs an explicitly writable rotation-key location or a pod restart invalidates all tokens. Please note it in the README extension-hooks section so #783 inherits the requirement.
- **Shared RWO PVC across all Deployments** — with `ReadWriteOnce` all seven pods must co-schedule on the node holding the volume attachment; the "service-per-Deployment over QUIC" topology silently collapses to single-node until RWX is configured. The pvc.yaml comment covers this — fine for Phase 0, just noting it's also the thing #783/#705 (CAS out-of-band) should eventually remove.
- **Probes** — QUIC/UDP RPC services get no probes, with an honest comment explaining why (no TCP/HTTP surface; Cap'n Proto over QUIC isn't probe-able by kubelet). Reasonable Phase 0 call. #779 asked for "dependency ordering via readiness probes"; that's only real for oai — worth a followup note that RPC services may want a minimal HTTP/TCP health listener or exec probe when the resolver wiring lands (distroless has no shell, so exec is out too).
- **No VFS mount-path hardcoding** — checked; the chart only sets container filesystem paths (`dataDir`), not VFS namespace bindings. Good.
- **CI** — #779 acceptance mentions chart lint + golden `helm template` in CI; the golden file is committed but no workflow consumes it. Fine as a followup, just don't lose it.

### Overlap / merge order

- **PR #847 (K5a CRDs)**: zero file overlap (`crates/hyprstream-k8s/` vs `charts/hyprstream/`); no ordering constraint.
- **PR #774 (still OPEN)**: this chart doesn't consume #774's resolver code, but #774 + #805 (SecretsProfile) define the secrets-resolution contract #783 must consume. Merge-order implication: this chart should avoid encoding any `HYPRSTREAM__SECRETS__PATH` semantics (the blocking fix above achieves that); #783 should land only after #774/#805 settle the profile model, per #783's own text.
- **#805 is open**: until the `systemd_mode` presence-check is replaced, *any* chart revision that sets `HYPRSTREAM__SECRETS__PATH` on non-policy pods is unsafe. Suggest a comment in values.yaml keyBootstrap block cross-referencing #805 so the constraint survives into K1a.

ewindisch:
## 🔴 BLOCKING REVIEW (1:7 coordination sweep) — two high-severity runtime blockers

Templates render cleanly and the key-bootstrap material is public-only with `required` guards — no ZMQ, no minted secrets. But as shipped the default install won't come up:

**1. (HIGH) oai probes use `httpGet` (HTTP) but the server serves HTTPS.** `TlsConfig::default()` has `enabled: true` self-signed (`config/mod.rs:236-288`) and the OAI path serves HTTPS when TLS resolves (`server/mod.rs:154-165`, `factories.rs:929`). The chart never sets `HYPRSTREAM__TLS__ENABLED=false` or probe `scheme: HTTPS`, so with a real image oai serves HTTPS on 6789, the HTTP probe fails the TLS handshake, the pod never goes Ready — and the README's `curl http://localhost:6789/v1/models` also fails. Pick one: disable TLS in-cluster via env, or set the probe scheme to HTTPS.

**2. (HIGH) `HYPRSTREAM__SECRETS__PATH` is rendered unconditionally, even when `keyBootstrap.enabled=false`.** `main.rs` treats the mere presence of that env var as flat/systemd credential mode (`systemd_mode = env::var(\"HYPRSTREAM__SECRETS__PATH\").is_ok()` → `load_or_generate_node_signing_key`, `main.rs:2020-2025`). In a default install every service load-or-generates the *same* flat `signing-key` on the shared PVC → per-service identity collapses to one key (plus a first-boot generation race), and since `bootstrap-pubkeys` doesn't exist the trust store is empty, so oai fails at "trust store has no model key" (`factories.rs:884-891`) and crash-loops. Gate `SECRETS__PATH` behind `keyBootstrap.enabled`.

**3. (med) Projected Secret volume `defaultMode: 0400` with `runAsUser: 65532` and no `fsGroup`** — secret files are root-owned 0400, so with `keyBootstrap.enabled=true` the non-root process can't read `signing-key`/`service-jwt`/`ca-key`; the whole bootstrap path fails at runtime despite rendering cleanly.

**4. (med) No `fsGroup` anywhere** (`defaults.podSecurityContext: {}`) with the PVC mounted at `/var/lib/hyprstream` — on most provisioners the fresh volume root is root:root, so services get EACCES creating `models/`/`loras/`/`cache/`.

**5. (low) streams service declares port 7011 with no listener env** and the config key would be `HYPRSTREAM__STREAMING__QUIC_PORT` (defaults `None`) — the Service is dead weight as shipped (acknowledged "nominal" in values).

Items 1–4 must be resolved before this can merge. Flagging @1:1.

ewindisch:
## ✅ Both HIGH blockers + both med verified resolved (1:7 re-review of 30a7f1e08)

All four findings correctly fixed; env-var→config mappings traced and exact:

**1. (HIGH) TLS probe mismatch — FIXED.** `configmap.yaml:14` emits `HYPRSTREAM__TLS__ENABLED` from `config.tls.enabled: false`. Mapping verified: `config/mod.rs:2147` `with_prefix(\"HYPRSTREAM\").separator(\"__\")` → `tls` → `TlsConfig.enabled`, `try_parsing` parses `\"false\"`. Flag actually flips the protocol (`server/tls.rs:104` → `oai.rs:143` serves plain HTTP), so the `httpGet` probe (defaults HTTP) and README curl now match. No mesh regression — QUIC/iroh TLS 1.3 and COSE signing are separate from this flag.

**2. (HIGH) SECRETS__PATH flat-key mode — FIXED.** `configmap.yaml:15-17` gates the var behind `{{- if .Values.keyBootstrap.enabled }}`; default `keyBootstrap.enabled: false` → var absent (golden-default.yaml:33 confirms it's gone). `main.rs:2020` `systemd_mode` is then false → per-service `load_or_generate_service_signing_key` instead of the flat node key; no collapse, no empty-trust-store crash-loop. Credentials volume mount is under the same guard — internally consistent.

**3. (med) Secret unreadable by non-root — FIXED.** `defaultMode: 0440` + `defaults.podSecurityContext.fsGroup: 65532` / `fsGroupChangePolicy: OnRootMismatch`. Mechanism sound: kubelet group-owns the projected tmpfs to 65532, 0440 sets group-read, container uid 65532 is in the fsGroup → readable.

**4. (med) PVC EACCES — FIXED.** Same pod-wide fsGroup applies to the `data` PVC at `/var/lib/hyprstream`, so uid 65532 can create `models/loras/cache`.

**Two low operator caveats (non-blocking, worth a note in values/README):**
- `config.tls.enabled=false` is cluster-wide across all HTTP surfaces — enabling `oauth`/`mcp`/`xet`/`flight` later serves them plain HTTP unless the operator re-enables TLS and switches those probes to `scheme: HTTPS`.
- The fsGroup fix relies on the CSI driver honoring fsGroup ownership (default RWO block does; NFS/hostPath ignore it → would still EACCES).

Default `helm install` is now bootable with matching probe/protocol, no flat-key crash-loop, readable secrets, writable PVC. Nice.

ewindisch:
Coordination/rebase note: rebased the Helm chart branch onto current main. Merge-tree was clean; final diff remains chart-only (charts/hyprstream). Pushed 9066bdea2 for fresh CI. Formal approval still needs a non-author identity.

Issue #788, opened 2026-07-04
K5a: CRD schemas under *.hyprstream.io — Model, Adapter, TrainingRun, InferenceService, TenantBinding


ewindisch:
Part of #778 (Phase 1). The **hyprstream → K8s** direction of the mutual namespace binding: hyprstream resources as declarative K8s objects, **subspaced under hyprstream-owned API groups** so nouns never collide (`serving.hyprstream.io/InferenceService` ≠ `serving.kserve.io/InferenceService` by construction — KServe is an optional adapter target (K6b), never the owner of our nouns).

### Deliverables (schemas + conversion/validation only; controller is K5b)
- `models.hyprstream.io/v1alpha1 {Model, Adapter}` — Model spec = git ref intent (repo, ref, STEP stage: staged/promoted); Adapter spec = file-based adapter intent (`adapters/NN_name.safetensors`).
- `training.hyprstream.io/v1alpha1 TrainingRun` — TTT/adapter training job intent (model ref, dataset mount path, resources/runs_on).
- `serving.hyprstream.io/v1alpha1 InferenceService` — serving intent (model:branch, autoscaling bounds, statefulness class).
- `mesh.hyprstream.io/v1alpha1 TenantBinding` — **the confused-deputy fix (#778 threat table)**: explicit, admin-created mapping K8s namespace ↔ hyprstream tenant, gated by `federation:register`; controllers refuse any cross-binding reference.
- **Single-writer rule baked into every schema's docs:** git owns weights — `spec` is desired STEP intent, `status` is observed git/runtime truth, the operator is a translator and never a second writer of weights (promotion = merge, rollback = checkout, via RegistryService).
- CEL validation rules; `hyprstream.io/managed-by` + ownerReference conventions documented.

### Acceptance
- CRDs install cleanly; invalid specs rejected by CEL; a `Model` with a cross-tenant reference is rejected at admission when the TenantBinding doesn't cover it; conversion-ready (v1alpha1 with upgrade path noted).

### Depends on / consumes
- Naming/claims vocabulary shared with K2a (#780). Consumed by: K5b/K5c, K6a/K6b.

Issue #787, opened 2026-07-04
K4c: placement→PodSpec vocabulary map — scheduling.rs decisions emitted as nodeSelector/affinity/tolerations/resources


ewindisch:
Part of #778 (Phase 1). Sub-item of the `k8s` backend (K4a, #781); the two-level-scheduling seam.

**K4c — hyprstream's scheduler decides *what runs with what constraints*; Kubernetes decides *which node*.** This is the Argo/Ray meta-scheduler pattern, and the substrate was built speaking K8s's language already: `hyprstream-discovery/src/scheduling.rs` (#523 P0) uses k8s-quantity strings (`"512Gi"`, `"100m"`) and label-selector match-expressions (`In/NotIn/Exists/Gt/Lt`) natively.

### Deliverables
- A translation module: `ResourceRequest` → `PodSpec.resources` (requests/limits incl. extended resources like `nvidia.com/gpu`); `LabelSelector`/match-expressions → `nodeSelector`/`nodeAffinity`; taints policy → `tolerations`; placement priority → `priorityClassName` (optional).
- The `k8s` backend consumes a placement decision (P2 `SandboxPool::acquire` / #525 engine output) and stamps the emitted Pod — the native scheduler's decision is *narrowed*, never widened, by the map (a constraint the tests assert).
- Round-trip explainability: the `SelectionReport` trace references the emitted K8s constraints so a placement can be explained end-to-end.

### Acceptance
- Property tests: every native predicate that rejected a candidate corresponds to a K8s constraint that would also exclude it (no widening); GPU quantity requests land as extended resources; golden PodSpec fixtures.

### Depends on / consumes
- K4a (#781); #523 P0 substrate (landed), #525 P2 engine (in flight — consume its decision type when it lands, stub until then).

Issue #781, opened 2026-07-04
K4a: `k8s` SandboxBackend core — kube-rs client emitting real Pods/Jobs (API server, not the CRI socket)


ewindisch:
Part of #778 (Phase 1; core has no hard deps so may start in Phase 0). Complements epic #508's backend taxonomy.

**K4a — a new `SandboxBackend` that drives the Kubernetes API server via kube-rs, creating real `Pod`/`Job` objects.**

This is deliberately distinct from `CriBackend` (#510), which drives a node's containerd **socket** and produces sandboxes invisible to Kubernetes ("shadow pods", same-node only). API-server pods are scheduler-placed, quota'd, GPU-accounted, `kubectl`-visible, and cross-node. hyprstream stays the client/orchestrator — we never implement the kubelet-facing CRI contract (#778 non-goal).

### Deliverables
- `k8s` backend under the existing seam: `inventory::submit!{ BackendRegistration { name: "k8s", auto_selectable: false, ... } }` — **explicit-name-only**, same auth-surface rationale as `cri` (#510): auto must never silently hand a workload to a cluster.
- kube-rs (`kube`, `k8s-openapi`) deps gated behind a `k8s` cargo feature; no tonic/prost duplication concerns (kube-rs is hyper-based).
- `to_k8s_podspec` translation sibling of `to_cri_*`: capnp `PodSandboxConfig`/`ContainerConfig` → `PodSpec` (resources, security context, labels, `runtimeClassName` passthrough so Kata-via-RuntimeClass still works — now scheduler-placed).
- Lifecycle mapping: `start`→create Pod + watch until Running; `stop`/`destroy`→delete (grace/force); `exec_sync`→`AttachedProcess` (blocking-capture shape); `get_pids`→`Ok(vec[])` (same rationale as cri); logs via `log_stream`.
- **Origin markers**: `hyprstream.io/managed-by` label + instance identity on every emitted object; the backend refuses to adopt objects it did not create (two-writer guard, #778 threat table).
- In-cluster (`ServiceAccount`) and kubeconfig auth; RBAC manifest for exactly pods/jobs/exec/log in a namespace.

### Acceptance
- `worker.backend = "k8s"` runs a workload as a real Pod on a kind cluster: visible to `kubectl`, exec/stop/destroy work, origin labels present, `auto` never selects it. Fail-closed when the API server is unreachable.

### Depends on / consumes
- Namespace delivery is **out of scope here** → K4b (needs K3). Placement mapping → K4c.

Issue #792, opened 2026-07-04
K6a: serving.hyprstream.io/InferenceService controller — Deployments/HTTPRoute + HPA/KEDA (TTT-stateful-aware)


ewindisch:
Part of #778 (Phase 2). The serving axis, under our own noun (`serving.hyprstream.io` — subspaced, no KServe collision; KServe adapter is K6b).

**K6a — reconcile an `InferenceService` CR into runnable serving infrastructure.**

### Deliverables
- Reconciler (in the K5b runtime): `InferenceService` → model-service Deployment (GPU variant by values/spec), Service, Gateway API `HTTPRoute` for the OpenAI-compat surface; status carries endpoint + readiness.
- **Statefulness classes (the TTT rule):** hyprstream inference is stateful — per-tenant LoRA deltas (`DeltaPool` LRU), KV cache, TTT warm state. Spec declares the class:
  - `stateful` (default): HPA/KEDA with **minReplicas ≥ 1**, scale on load signals via the K1b OTel→Prometheus path; **never scale-to-zero** (a cold start pays libtorch + model load + warm-state loss).
  - `stateless`: scale-to-zero permitted (KEDA http-add-on or equivalent).
- Session/tenant affinity documented (consistent routing so per-tenant deltas stay warm); autoscaling signals: request rate, stream backpressure, tokens/s (from K1b's documented set).
- Model reference resolves through `models.hyprstream.io/Model` (`model:branch` worktree naming, matching `/v1/models`).

### Acceptance
- On kind (CPU): applying an `InferenceService` yields a routable OpenAI-compat endpoint; load drives HPA scale-out on the K1b metric; `stateful` class never scales below 1; deletion garbage-collects via ownerRefs.

### Depends on / consumes
- K5a (#788), K5b runtime, K1b (#784) metrics, K1 (#779). Consumed by: K6b (KServe adapter).

ewindisch:
## Direction decisions from the PR #865 review (operator-directed, 07-05)

Full review: https://github.com/hyprstream/hyprstream/pull/865#issuecomment-4887201904. The load-bearing decisions for K6a, grounded in the TTT delta mechanics and the DID/discovery layer:

[...]

PR #847, opened 2026-07-05
feat(k8s) K5a: CRD schemas under *.hyprstream.io (#788)

ewindisch:
## Summary

New crate **`crates/hyprstream-k8s`** — the foundation for the kube-rs track of the Kubernetes epic (**#788**, Epic **#778**). It implements the hyprstream -> Kubernetes projection: hyprstream resources as declarative CRDs, under hyprstream-owned API groups so nouns never collide with third-party CRDs (`serving.hyprstream.io/InferenceService` != `serving.kserve.io/InferenceService`).

This branch also currently carries the stacked kube runtime foundation that landed on top of the CRD work before the review queue caught up:

- **K4a / #781**: `K8sBackend` Pod/Job sandbox backend, non-auto-selectable, with managed-by + sandbox-id two-writer guard.
- **K4c / #787**: placement -> PodSpec mapping with golden/property coverage.
- **RBAC/manifests** needed by the kube backend/operator path.

## CRDs (all `v1alpha1`, all with a `status` subresource)

| Group | Kind | Scope | Purpose |
|---|---|---|---|
| `models.hyprstream.io` | `Model` | Namespaced | Git-ref intent (repo, gitRef, STEP `stage`: Staged/Promoted) |
| `models.hyprstream.io` | `Adapter` | Namespaced | File-based adapter intent (`adapters/NN_name.safetensors`) |
| `training.hyprstream.io` | `TrainingRun` | Namespaced | TTT/adapter training intent (modelRef, dataset mount, runsOn, resources) |
| `serving.hyprstream.io` | `InferenceService` | Namespaced | Serving intent (`model:branch`, autoscaling bounds, statefulness class) |
| `mesh.hyprstream.io` | `TenantBinding` | **Cluster** | Explicit admin-created namespace <-> tenant binding, the confused-deputy guard |

## Security / epic invariants

- Single-writer rule in resource rustdoc + CRD descriptions: `spec` is desired STEP intent, `status` is observed git/runtime truth.
- `TenantBinding` is cluster-scoped so a tenant confined to its namespace cannot forge its own binding.
- K8s workloads carry `hyprstream.io/managed-by`, `hyprstream.io/sandbox-id`, and `hyprstream.io/instance`. Delete ownership is now guarded by managed-by + sandbox-id; the per-process instance UUID remains audit metadata so a restarted backend can reap its own sandbox workload without adopting another sandbox's objects.
- `TenantBinding.spec.namespace` now enforces the DNS-1123 63-character label bound as well as shape.

## Validation

- kubeconform: all 5 emitted CRDs validate against the apiextensions/v1 meta-schema.
- `cargo test -p hyprstream-k8s`: 10 offline tests pass, including CEL checks and committed YAML regeneration parity.
- `cargo test -p hyprstream-workers --features k8s runtime::k8s_backend`: 15 tests pass.
- `cargo clippy -p hyprstream-workers --features k8s --all-targets -- -D warnings`: clean.
- `cargo clippy -p hyprstream-k8s --all-targets --all-features -- -D warnings`: clean.
- Commit hook full release clippy passed on the latest push.

## Notes

- CRD/controller runtime reconciliation remains split across the later K5b/K5c work.
- `k8s-openapi` uses the repo-supported `v1_32` feature.

Refs #788, #781, #787. Epic #778.


crates/hyprstream-k8s/src/training.rs
```diff
@@ -0,0 +1,97 @@
...
+
+use kube::CustomResource;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+/// Coarse resource request for a training run. Lowered to a concrete PodSpec by
+/// the placement→PodSpec vocabulary map (K4c); kept minimal and portable here.
+#[derive(Serialize, Deserialize, Clone, Debug, Default, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct ResourceSpec {
```
ewindisch, line 16: **Medium — forks the scheduling-substrate resource vocabulary.** The substrate (#523 P0, `hyprstream-discovery/src/scheduling.rs`) already defines the one resource shape every scheduling surface is supposed to compose on: `Resource { name, quantity }` / `ResourceRequest { name, min_quantity }` with k8s-quantity strings for everything, and its module doc says that vocabulary lives in ONE place. This `ResourceSpec` introduces a parallel fixed-field shape (`gpus: u32`, `cpu: String`, `memory: String`) with mixed typing — `cpu`/`memory` are k8s quantities but `gpus` is a bare integer, so the K4c lowering to both the substrate's `ResourceRequest` list and #761's `hyprstream.io/gpu-request` annotation needs a bespoke translation instead of a 1:1 map.

Suggestion while this is still v1alpha1: either make `resources` a named-quantity list mirroring the substrate (`[{name: "gpu", quantity: "2"}, ...]` — also closest to native k8s `resources` semantics), or keep the ergonomic fixed fields but use k8s-quantity strings uniformly and document the exact mapping to the substrate names and to `hyprstream.io/gpu-request` in the rustdoc so K4c is mechanical. Changing this later means a conversion.

crates/hyprstream-k8s/src/training.rs
```diff
@@ -0,0 +1,97 @@
...
+    /// Name of the output file-based adapter to produce, e.g.
+    /// `00_finetune.safetensors`. Defaults are chosen by the reconciler when
+    /// omitted.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub adapter_name: Option<String>,
+
+    /// Placement class / scheduling hint (`runs_on`) resolved by the scheduling
+    /// substrate. Free-form label consumed by K4c.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub runs_on: Option<String>,
```
ewindisch, line 71: Minor: `runsOn` as a free-form string is per the issue text, but the substrate's selection language is `LabelSelector { key, op, values }` (In/NotIn/Exists/Gt/Lt). If K4c is going to resolve this against `queryCandidates`, a single opaque string will end up growing an ad-hoc mini-syntax. Fine to keep for v1alpha1, but consider noting in the rustdoc that the v1beta upgrade path is a match-expression list aligned with `hyprstream_discovery::scheduling::LabelSelector`, so nobody starts encoding selectors into this string.

crates/hyprstream-k8s/src/serving.rs
```diff
@@ -0,0 +1,95 @@
...
+    validation = "self.spec.model != ''",
+    validation = "self.spec.minReplicas <= self.spec.maxReplicas",
+    validation = "self.spec.maxReplicas >= 1",
+    cel
+)]
+#[serde(rename_all = "camelCase")]
+pub struct InferenceServiceSpec {
+    /// Model to serve, in `model:branch` form (branch optional, defaults to the
+    /// promoted ref).
+    pub model: String,
```
ewindisch, line 56: **Medium — two different model-reference vocabularies across the CRD set.** `Adapter.modelRef` and `TrainingRun.modelRef` name a `Model` CR in the same namespace, while `InferenceService.model` is a raw `model:branch` string that bypasses the `Model` object entirely. The issue text does say "model:branch" for serving, but the K5a acceptance criterion — cross-tenant references rejected when no `TenantBinding` covers them — needs references the admission layer can resolve and attribute to a tenant; a free-form `model:branch` string is the hardest of the three shapes to check.

Consider `modelRef` (CR name) + optional `branch` here for consistency, with the `model:branch` string derived by the reconciler — or, if serving deliberately addresses the git namespace directly (skipping the Model CR), say so explicitly in the rustdoc and note how admission will tenant-scope it in K5b. Either is fine; the silent inconsistency is the problem.

crates/hyprstream-k8s/src/mesh.rs
```diff
@@ -0,0 +1,79 @@
...
+    cel
+)]
+#[serde(rename_all = "camelCase")]
+pub struct TenantBindingSpec {
+    /// The Kubernetes namespace this binding authorizes.
+    pub namespace: String,
+
+    /// The hyprstream tenant identifier (e.g. an atproto DID) the namespace is
+    /// bound to.
+    pub tenant: String,
```
ewindisch, line 57: Minor: identity should be DID-shaped, and this field only documents it as an example ("e.g. an atproto DID"). Since `tenant` is the trust anchor for every cross-tenant check, a loose string invites host-location or ad-hoc tenant ids creeping in. If DIDs are the intended form, add a CEL rule (`self.spec.tenant.startsWith('did:')`, or a fuller `did:method:...` regex) now — tightening validation later is a breaking change for existing objects. If non-DID tenant ids are intentionally allowed, document the accepted forms instead.

crates/hyprstream-k8s/src/mesh.rs
```diff
@@ -0,0 +1,79 @@
...
+    // to be referenced with the `__namespace__` escape (same rule as the Go
+    // apiserver's `apiserver/schema/cel/model` escaping).
+    validation = "self.spec.__namespace__.matches('^[a-z0-9]([-a-z0-9]*[a-z0-9])?$')",
+    validation = "self.spec.tenant != ''",
+    cel
+)]
+#[serde(rename_all = "camelCase")]
+pub struct TenantBindingSpec {
+    /// The Kubernetes namespace this binding authorizes.
+    pub namespace: String,
```
ewindisch, line 53: Minor (doc-level): the `federation:register` gate is documentation-only in this PR — until the K5b admission webhook exists, the only thing standing between a caller and a forged binding is Kubernetes RBAC on the cluster-scoped `tenantbindings` resource. The cluster-scoping itself is the right call and does defeat the namespace-confined tenant. Suggest one sentence in the CRD `description` (not just rustdoc) telling operators to RBAC-restrict `create`/`update` on `tenantbindings.mesh.hyprstream.io` to cluster admins until K5b lands, since the CRD description is what `kubectl explain` shows the person installing this.

crates/hyprstream-k8s/src/models.rs
```diff
@@ -0,0 +1,139 @@
...
+    singular = "model",
+    shortname = "hsmodel",
+    namespaced,
+    status = "ModelStatus",
+    category = "hyprstream",
+    doc = "A hyprstream model as a git-ref intent (Stage → Train → Evaluate → Promote). spec is desired intent; status is observed git truth. The operator never writes weights.",
+    printcolumn = r#"{"name":"Repo","type":"string","jsonPath":".spec.repo"}"#,
+    printcolumn = r#"{"name":"Ref","type":"string","jsonPath":".status.observedRef"}"#,
+    printcolumn = r#"{"name":"Stage","type":"string","jsonPath":".spec.stage"}"#,
+    validation = "self.spec.repo != ''",
```
ewindisch, line 46: Minor, applies to all five resources: the CEL rules have no `message`/`messageExpression`, so users get the raw rule text (`failed rule: self.spec.file.matches('^[0-9]{2}_...')`) as their error. kube's `validation` attribute supports attaching a message; short human strings ("adapter file must be NN_name.safetensors", "minReplicas must not exceed maxReplicas") would make first-contact UX much better for one line each. Related nit: the rules sit at the root of `openAPIV3Schema` and prefix everything with `self.spec.` — attaching them to the spec schema instead is the conventional placement and keeps them from being re-evaluated on status-subresource writes (harmless today, but it spends the CEL cost budget).

ewindisch (commented):
**Verdict: Approve with changes** — no blocking findings; the schemas honor the epic's security invariants, but the resources block forks the scheduling-substrate vocabulary and a few validation/consistency gaps are worth fixing while the API is still v1alpha1.

Checked against the review criteria:

**Group/version naming** — good. All five groups are hyprstream-owned subspaces (`models`/`training`/`serving`/`mesh` `.hyprstream.io`), all `v1alpha1`, single served+stored version each, `hyprstream` category and short names throughout. `models.models.hyprstream.io` as a CRD name is inherently awkward but correct and unavoidable given plural.group; no action.

**Schema quality** — solid for a first cut: required fields declared, CEL `x-kubernetes-validations` on every resource, status subresource everywhere, printer columns on every kind, statuses carry `observedGeneration`. The kubeconform pass plus client-side `validate_cel()` round-trips in `tests/offline.rs` are a good offline harness, and the committed-YAML-matches-generated test is the right drift guard (one weakness noted inline).

**MAC interface policy** — clean. No CRD carries a security label or clearance field as an authoritative input; there is nothing resembling a caller-supplied `SecurityContext`. `TenantBinding` is the right shape under the ratified policy: an admin-authored, cluster-scoped policy-store object (analogous to the Casbin store), not a wire label. One doc-level caveat inline about the `federation:register` gate being documentation-only until K5b.

**Git-shaped model refs** — yes: `Model` is repo + gitRef + STEP stage with status as observed git truth, `Adapter` is file-based (`NN_name.safetensors`, regex matches the `adapters/00_name.safetensors` convention), `InferenceService.model` is `model:branch`. The single-writer rule (git owns weights, operator is a translator) is stated in every rustdoc and CRD description. One consistency gap between `modelRef` and `model:branch` referencing noted inline.

**Identity as DID** — `TenantBinding.spec.tenant` documents "e.g. an atproto DID" but doesn't validate it; suggestion inline. No host-location identity fields anywhere (`status.url` is observed runtime truth, which is fine).

**Cross-PR consistency**
- **#846 (Helm chart)**: the chart does not install or reference these CRDs — no name/group conflict, but also no delivery path yet besides the feature-gated `install.rs` helper. Before K5b lands, K1/K5 need an explicit decision: chart `crds/` directory (helm's install-once semantics) vs. operator-boot server-side apply via `apply_all_crds`. Worth a line in the epic so the two tracks don't each assume the other installs them.
- **#761 (P2 scheduler)**: the `hyprstream.io/managed-by` label constant is consistent with the `hyprstream.io/*` annotation namespace #761 establishes (`hyprstream.io/gpu-request`, `hyprstream.io/group`). The `resources` block, however, does not reuse the substrate vocabulary — see the inline comment on `ResourceSpec`; that's the main change I'd like before this hardens.

Severity summary: 2 Medium (vocabulary fork; model-reference inconsistency), 5 Minor (tenant DID validation, CEL messages, federation:register doc caveat, YAML drift test uses contains, cel feature in default deps).

crates/hyprstream-k8s/tests/offline.rs
```diff
@@ -0,0 +1,261 @@
...
+fn committed_yaml_matches_generated() {
+    // Guards against forgetting to re-run `gen-crds` after a schema change.
+    for (crd, stem) in hyprstream_k8s::crd_manifests() {
+        let path = concat!(env!("CARGO_MANIFEST_DIR"), "/crds/");
+        let file = std::path::Path::new(path).join(format!("{stem}.yaml"));
+        let on_disk = std::fs::read_to_string(&file)
+            .unwrap_or_else(|e| panic!("read {}: {e}", file.display()));
+        let regenerated = serde_yaml::to_string(&crd).unwrap();
+        assert!(
+            on_disk.contains(regenerated.trim()),
```
ewindisch, line 284: Minor: `on_disk.contains(regenerated.trim())` catches the common staleness direction (schema changed, YAML not regenerated) but passes if the committed file contains the regenerated output plus stale extra content (e.g. a hand-appended rule, or two concatenated documents). Since the generator's output is exactly header + `serde_yaml::to_string`, strip the two known header lines and assert equality — same cost, strictly stronger guard.

crates/hyprstream-k8s/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+# CRD *type* definitions + YAML emission compile with no features and no cluster.
+default = []
+# `k8s` gates anything that needs a live client / controller-runtime.
+k8s = ["kube/client", "kube/runtime", "kube/rustls-tls"]
+
+[dependencies]
+# `derive` gives the CustomResource derive + CustomResourceExt (kube-core, no client).
+# `cel` gives client-side CEL validation (kube-core, no client) so `validate_cel()`
+# and `x-kubernetes-validations` emission work offline in tests.
+kube = { version = "4.0", default-features = false, features = ["derive", "cel"] }
```
ewindisch, line 38: Minor: the `cel` feature drags the `cel` 0.13 interpreter and its `antlr4rust` parser tree into the crate's *default* build, but the only consumer of `validate_cel()` is `tests/offline.rs` — the `x-kubernetes-validations` emission is done by the derive/`CustomResourceExt` and shouldn't need the interpreter at runtime. If that holds (worth a quick check), keep the lib dependency at `features = ["derive"]` and add a `[dev-dependencies] kube = { ..., features = ["derive", "cel"] }` entry so the interpreter is test-only. Keeps the workspace's default dep graph slimmer, since Cargo.lock already grew noticeably for this. If emission genuinely requires `cel`, ignore this.

ewindisch:
## 🟡 Review (1:7 coordination sweep) — minor issues; resolve/ack before merge

CRD schemas check out: groups correctly subspaced, status subresources present, TenantBinding genuinely cluster-scoped, CEL rules match the documented invariants and are tested, `__namespace__` CEL escape correct, committed YAML asserted equal to generated output, RBAC Role namespace-scoped.

**1. (med — scope/provenance) This branch contains far more than the CRD crate.** The stacked K4a k8s SandboxBackend (`hyprstream-workers/src/runtime/k8s_backend.rs`, ~1170 lines), K4c podspec map (`hyprstream-discovery/src/podspec.rs`) and the RBAC manifest were merged into this branch via #851/#853, but the PR description covers only the CRD crate. Merging #847 lands ~2/3 of its change surface under a CRD-only description — anyone reviewing/bisecting by PR description misses it. Please expand the description to cover the K4a/K4c content actually riding along.

**2. (med) `K8sBackend` orphans workloads across restart.** `instance_id` is a fresh UUID per process, and `stop`/`destroy` refuse to delete any object whose `hyprstream.io/instance` label ≠ the current UUID. After a restart/crash, every Pod/Job the previous instance created can never be reaped by the backend — orphans accumulate with no in-band cleanup. The `MANAGED_BY` + sandbox-id labels would be a sufficient guard without pinning to the ephemeral instance UUID.

**3. (low) TenantBinding `namespace` DNS-1123 CEL rule has no 63-char length bound** and no `maxLength` in the schema — a binding naming an over-length (nonexistent) namespace passes admission.

Item 1 (description) and item 2 (orphan reaping) should be resolved; item 3 is a nice-to-have. Flagging @1:1.

ewindisch:
Review follow-up pushed as `1e7fd181b`.

Fixes the remaining CRD-lane review items:
- Expanded the PR body to describe the stacked K4a/K4c/RBAC content carried on this foundation branch.
- Changed K8sBackend delete ownership from ephemeral instance UUID to `managed-by + sandbox-id`; `instance` remains audit metadata, so restarted backends can reap their own sandbox workloads.
- Added a regression test proving different instance IDs still pass for the same sandbox, while a different sandbox fails.
- Tightened TenantBinding namespace CEL to enforce the DNS-1123 63-character bound and added an offline test; regenerated committed CRD YAML.

Validation:
- `cargo test -p hyprstream-k8s` -> 10 passed
- `cargo test -p hyprstream-workers --features k8s runtime::k8s_backend` -> 15 passed
- `cargo clippy -p hyprstream-workers --features k8s --all-targets -- -D warnings` -> clean
- `cargo clippy -p hyprstream-k8s --all-targets --all-features -- -D warnings` -> clean
- commit hook release clippy passed

ewindisch:
## ✅ All three blockers verified resolved (1:7 re-review of 1e7fd181b)

**1. (med — orphaned workloads) FIXED, restart-safe AND not over-broad.** `labels_are_ours` (`k8s_backend.rs:290-294`) now matches `managed-by == \"hyprstream-workers\"` **&&** `sandbox-id == <this sandbox>`; the per-process `INSTANCE_LABEL` UUID is removed from the predicate and kept only as audit metadata (`:504`, rustdoc `:287-289`). Restart trace confirmed: both call sites look up by `sandbox_id` (`:803,:824`, and `K8sHandle.name == sandbox_id`), so after a fresh-UUID restart `stop`/`destroy` matches and reaps the prior instance's Pod/Job — test `two_writer_guard_rejects_foreign_sandbox_labels` exercises exactly this (foreign `INSTANCE_LABEL`, same sandbox-id → still ours). Not over-broad: sandbox-id specificity is still required, so cross-sandbox/cross-tenant deletes stay blocked (tests cover different sandbox-id, missing managed-by, and no-labels → all rejected). Good balance.

**2. (med — description scope) FIXED.** PR body now has a dedicated stacked-work section naming K4a/#781 (SandboxBackend + two-writer guard), K4c/#787 (placement→PodSpec map), and the RBAC manifests, plus a security-invariants bullet on the delete guard.

**3. (low — CEL 63-char bound) FIXED.** CRD YAML (`mesh.hyprstream.io_tenantbindings.yaml:86`) and source annotation (`mesh.rs:46`) now enforce `size() <= 63 && matches(...)`; offline test `too_long_ns` (`tests/offline.rs:224-234`) asserts a 64-char namespace fails validation.

#847 cleared. That's the full k8s-foundation lane (#846/#847/#849/#850) verified clean.

ewindisch:
## 🔴 Blocking (operator-directed): `Statefulness` must default to `TttStateful`, and status needs a message/conditions field

Two CRD-layer changes surfaced by the #865 serving-controller review (mechanism evidence there: https://github.com/hyprstream/hyprstream/pull/865#issuecomment-4887201904):

**1. `#[default] Stateless` on `Statefulness` in `serving.rs` is wrong.** Issue #792 specifies stateful as the default, and the implementation agrees: the per-tenant TTT delta register lives only in an in-memory `VarStore` (`training/tenant_delta.rs`); `DeltaPool::get_or_create` always constructs a zero-initialized delta and never rehydrates from `.snapshots/` (`training/delta_pool.rs`) — so losing a replica silently loses all uncommitted per-tenant adaptation. With `Stateless` as the serde default, an unannotated `InferenceService` opts into scale-to-zero and free replica shuffling, i.e. data loss by default. Move `#[default]` to `TttStateful`.

**2. `InferenceServiceStatus` has no `message`/conditions field.** The controller (#865) observes a message on both success and failure paths but has nowhere to persist it; healthy scaled-to-zero vs never-came-up are also indistinguishable via `phase` alone. Add a `message` (or conditions list) so the serving controller can surface it.

Since this stack is unmerged, both fix at the source here rather than as follow-ups.

ewindisch:
Coordination/rebase note: rebased the CRD/k8s foundation branch onto current main and resolved the Cargo.lock conflict by keeping the k8s-added deps while preserving current main transitive resolutions where applicable. cargo metadata --locked --no-deps passes. Pushed 42905f2e8 for fresh CI. Formal approval still needs a non-author identity.

Issue #804, opened 2026-07-04
Secrets-dir resolution still diverges outside PR #774's two sites (do_bootstrap, factories::credentials_dir, unit generation, policy JWT persist)


ewindisch:
## Summary

PR #774 centralizes secrets-dir resolution onto `HyprConfig::resolve_secrets_dir()` for the two `bin/main.rs` service-spawn sites — but several other sites still hand-compute the path and diverge whenever `[secrets] path` (config file) or instance namespacing is set. Each one is a latent #759: keys written to one directory, read from another, with `load_or_generate_*` silently minting divergent keys on the read side.

## Divergent sites (verified on main / PR #774 head)

| Site | What it does today |
|---|---|
| `cli/bootstrap_manager.rs:731-738` (`do_bootstrap`) | `HyprConfig::load().config_dir().join("credentials")` — **the write side of bootstrap** ignores `[secrets] path`. Operator sets `[secrets] path` → wizard writes to `~/.config/...`, fixed main.rs reads the configured dir, a fresh root key is silently minted there → #759 symptom returns with two "root" keys on disk |
| `cli/bootstrap_manager.rs:142,190` (`credentials_dir()`, `is_bootstrap_needed()`) | same hand-rolled path |
| `services/factories.rs:85-97` (`credentials_dir()`) | honors `HYPRSTREAM__SECRETS__PATH` env but not file-config `[secrets] path` → JWT registration lookup fails closed even when signing-key load succeeds |
| `hyprstream-service/src/service/manager/units.rs:165-185` + `manager/systemd.rs:357-366,447-450` | raw `dirs::config_dir()` for `LoadCredential` source paths and credstore checks — ignores both `[secrets] path` and `HYPRSTREAM_INSTANCE` namespacing (bootstrap writes instance-namespaced paths via `StoragePaths`; unit generation looks in the non-namespaced location) |
| `auth/policy.rs:1385-1392` | persists renewed JWTs to a hand-rolled `config_dir()/credentials` |

Note: PR #774's description claims `do_bootstrap` uses `resolve_secrets_dir()` — that's true of `register_local_identity` but not of `do_bootstrap` itself (flagged in review).

## Fix direction

[...]

Issue #802, opened 2026-07-04
systemd unit generation: missing per-service signing key silently falls back to ImportCredential of the CA root key


ewindisch:
## Summary

In the systemd unit generator, if a **non-policy** service's per-service signing key file is missing at install time, the credentials loop falls through to `ImportCredential=signing-key` — which is the **node/CA root key** (written flat into the encrypted credstore by `encrypt_credentials_if_available`). The service then boots, loads `%d/signing-key` as its *own* signing key (`bin/main.rs` systemd branch), and silently signs everything with the root key.

## Where

- `crates/hyprstream-service/src/service/manager/units.rs:201-214` — per-service `LoadCredential` emission for `signing-key`/`service-jwt`; on missing plaintext file the loop falls through to the `ImportCredential={name}` branch, and `signing-key` collides with the flat node/CA credential name.
- `crates/hyprstream-service/src/service/manager/systemd.rs:~472` — `encrypt_credentials_if_available` writes the node signing key into the credstore under the unprefixed name `signing-key` (per-service keys get `{service}-signing-key` prefixed names, so the collision is only with the root key).
- Consumer: `crates/hyprstream/src/bin/main.rs` (systemd mode loads flat `%d/signing-key` as own key).

## Impact

- Root/CA private key material handed to an arbitrary service process — violates the independent-per-service-key design and the #441 one-identity→one-key invariant.
- The service's responses then fail `bootstrap_pubkeys[svc]` verification ("Response signed by unexpected key", the #759 symptom) — so the failure *is* eventually visible, but only as a cryptic runtime verify error, after the key has already been exposed to the wrong process.

## Fix direction

Missing per-service key at install time should be a **hard install error** ("re-run bootstrap / repair"), never a fallthrough to a different credential — mirroring `identity_store::missing_in_readonly()`'s fail-closed posture. The name collision should also go away: the flat node credential should not share the name `signing-key` with the per-service credential slot.

## Related

#441 (invariant), #759 / PR #774 (runtime half of the same disease), #783 (K1a must not inherit this pattern into the Helm chart). Found during a systemd secrets audit prompted by the PR #774 review.

PR #855, opened 2026-07-05
fix(auth): stop deriving service cnf.jwk from the CA root key

### LINKED ISSUE #806: policy.rs still derives service pubkeys from the root key for cnf.jwk in issued/refreshed service tokens (#441 residue)
## Summary

`auth/policy.rs` still **derives** per-service public keys from the root key when populating `cnf.jwk` in issued and refreshed service tokens: `derive_purpose_key(&self.signing_key, "service:{name}")` at `policy.rs:364` and `:1372`. But bootstrap generates **independent random** per-service keys (`load_or_generate_service_signing_key`) — the derived pubkeys cannot match the keys services actually hold.

This is the closest living relative of the #441 disease left in the tree: #441 removed CA-derivation from the *verification/resolution* side (`resolve_service_key` is now registered-or-error), but the *issuance* side still fabricates key bindings from the CA instead of using the registered key set.

## Impact

Any consumer that actually enforces the `cnf.jwk` sender-binding on these service tokens will reject legitimate holders (fail-closed mismatch), or — worse — a consumer might be tempted to "fix" it by re-deriving on their side, reintroducing exactly the #441 hole. Either way the field is wrong data signed by the CA.

## Fix direction

`cnf.jwk` in service tokens must come from the **registered** key (trust store / `bootstrap_pubkeys` / `registerServiceKey` state), never re-derived. If no registered key exists for the service at issuance time, error — same registered-or-error posture as post-#441 `handle_resolve_service_key` (`policy.rs:1258`).

## Related

#441 (invariant; this is residue it didn't reach), #759 / PR #774 (surfaced during its second-opinion review), #804 (the adjacent hand-rolled persist path at `policy.rs:1385-1392`).

ewindisch:
## Summary

Fixes #806, closing out the last living residue of epic #441 on the issuance side.

`crates/hyprstream/src/services/policy.rs` populated `cnf.jwk` on issued and refreshed
service tokens by **deriving** a pubkey from the root/CA signing key
(`derive_purpose_key(&self.signing_key, "service:{name}")`, at what were `policy.rs:364`
and `:1372`). But bootstrap actually generates **independent random** per-service keys
(`load_or_generate_service_signing_key`) — the derived pubkey can never match the key a
service actually holds. #441 already fixed the *verification/resolution* side
(`resolve_service_key`/`handle_resolve_service_key` is registered-or-error, no CA-derivation
fallback); this residue was on the *issuance* side.

## What changed

- `PolicyService::handle_issue_token` (service-token branch): instead of deriving
  `service_key_bytes` via `derive_purpose_key`, it now resolves the service's verifying key
  from the global trust store (`hyprstream_service::global_trust_store().resolve_one(svc_name)`).
  If the service has no registered key yet, `handle_issue_token` now errors instead of
  signing a fabricated key binding — mirroring the exact registered-or-error posture already
  used by `handle_resolve_service_key`.
- `PolicyService::handle_refresh_service_token`: same fix — the refreshed token's `cnf.jwk`
  now comes from the trust store lookup instead of `derive_purpose_key`, with the same
  fail-closed error on an unregistered key.

Both call sites named in the issue are fixed; no new fallback/compat path was introduced.

## registerServiceKey fail-closed (paired #441 decision)

The issue also asked me to check whether the paired #441 decision — "`registerServiceKey`
must be fail-closed (`factories.rs:99`): no JWT / cannot register → the service does not
serve signed responses, don't WARN-and-continue" — had already landed.

It has: `register_service_key()` in `crates/hyprstream/src/services/factories.rs` already
propagates every failure via `?` (all 11 call sites, one per service factory, use
`register_service_key(ctx, "...", &sk)?;`), and `resolve_registration_jwt` bails with a
descriptive error (naming the service and expected credentials path) when no CA-signed JWT
is found on disk or in the trust store — with an existing regression test
(`resolve_registration_jwt_fails_closed_when_missing`) covering exactly this. No
WARN-and-continue path remains, so **no change was needed there**; I left it alone rather
than bundling an unrelated diff into this PR.

## Scope discipline

Per the issue, I did **not** touch:
- #804 (secrets-dir resolution divergence, `policy.rs:1385-1392`) — adjacent hand-rolled
  disk-persist path, explicitly a separate ticket.
- #802/#803/#805/#808 (systemd credential-generation cluster) — separate layer, separate
  tickets.

## Judgment calls

- Used `anyhow!` + `?` (matching `handle_resolve_service_key`'s existing style) rather than
  building a new `PolicyResponseVariant::Error` variant, since the issue explicitly asked to
  mirror the already-fixed handler's error shape, not invent a new one.
- This is a behavior change for any deployment where a service was previously issued a token
  with a CA-derived (and effectively unusable) `cnf.jwk` before ever calling
  `registerServiceKey` — token issuance/refresh for that service will now fail until the
  service registers its real key. That's the intended fail-closed behavior per the issue;
  I did not add a compatibility shim to paper over it.

## Test plan

- [x] `cargo test -p hyprstream --lib` — 885 passed, 0 failed, 1 ignored
- [x] `cargo clippy --workspace --all-targets --features download-libtorch -- -D warnings` — clean, no warnings
- [x] Pre-commit hook (`cargo clippy --workspace --all-targets --release -- -D warnings`) passed

### DIFF
--- a/crates/hyprstream/src/services/policy.rs
+++ b/crates/hyprstream/src/services/policy.rs
@@ -356,16 +356,19 @@ impl PolicyHandler for PolicyService {
         let audience = data.audience.as_ref().filter(|s| !s.is_empty()).cloned()
             .or_else(|| self.default_audience.clone());
 
-        // Service tokens: derive the Ed25519 key bytes from the root CA key.
-        // The CA (PolicyService) is authoritative — the pubkey is not caller-provided.
+        // Service tokens: cnf.jwk must be the service's REGISTERED key, never a
+        // CA-derived guess (#441/#806) — bootstrap generates independent random
+        // per-service keys (`load_or_generate_service_signing_key`), so a derived
+        // pubkey here would not match the key the service actually holds. If no
+        // registered key exists yet, error rather than sign a wrong key binding.
         // User tokens: decode the caller-provided pubkey (from OAuth consent page).
         let service_key_bytes: Option<[u8; 32]> = if is_service_token {
             let svc_name = &subject["service:".len()..];
-            let svc_signing_key = hyprstream_rpc::node_identity::derive_purpose_key(
-                &self.signing_key,
-                &format!("service:{svc_name}"),
-            );
-            Some(*svc_signing_key.verifying_key().as_bytes())
+            let trust = hyprstream_service::global_trust_store();
+            let vk = trust.resolve_one(svc_name).ok_or_else(|| {
+                anyhow!("service key '{svc_name}' not registered; refusing to issue a service token with a fabricated cnf.jwk")
+            })?;
+            Some(*vk.as_bytes())
         } else {
             use base64::{engine::general_purpose::URL_SAFE_NO_PAD, Engine as _};
             data.user_pub_key.as_deref().and_then(|s| {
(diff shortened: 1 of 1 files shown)


ewindisch:
✅ **Reviewed (1:7 sweep) — ready.** Fix-completeness verified: after this diff no issuance path derives per-service key material from the CA root — remaining `derive_purpose_key` sites use the legit JWT-CA purpose or MESH node-identity purposes; all `with_cnf_jwk` mint sites now bind real signer keys. Sender-binding is *strengthened* (cnf.jwk == the key the service signs envelopes with). Migration: previously-minted tokens carried a cnf.jwk that could never match a real signer, so no working deployment depended on them.

Three low, non-blocking notes:
- `trust.resolve_one(svc_name)` returns the *first* key for a scope (HashMap iteration) — during a rotation window with two live keys the minted cnf.jwk is nondeterministic. Pre-existing in `handle_resolve_service_key`; this PR extends the pattern. Worth a deterministic tiebreak eventually.
- `Claims::cnf` doc comment (`claims.rs` ~L134) still says "derived by the CA from the root key" — now factually wrong; please fix so nobody reintroduces CA-derivation as "the documented contract."
- Fail-closed on unregistered key is an ordering-sensitive behavior change (issue/refresh before `registerServiceKey` now errors) — intended per #806, worth a release note.

None block. Arming auto-merge.

Issue #550, opened 2026-06-29
Epic: Complete hybrid-PQC confidentiality — N-KEM hybrid key agreement, COSE_Encrypt0, fail-closed, transport PQ


ewindisch:
Epic to **complete the started hybrid post-quantum cryptography work** so that **confidentiality**, not just authenticity, is hybrid-PQ across every plane (RPC envelopes, streams, notifications, transport). Greenfield — no backwards-compat constraint. This is the confidentiality counterpart to the already-landed hybrid **signature** stack (EdDSA + ML-DSA-65 nested COSE), and supplies the crypto floor that epic #547 (MAC authz) and #549 (federation perimeter) depend on.

**Branch:** `feature/pqc-hybrid-confidentiality` (off `main`). Sub-PRs target the epic branch; the epic PR targets `main`. Schema breakage in scope.

> **Status: DRAFT FOR REVIEW.** Decomposition S0–S6 below; sub-issues filed as a task list.

---

## 1. Why (the gap)

Audit finding: **signatures are genuinely hybrid-PQ and well-built, but every confidentiality key-agreement is classical** → harvest-now-decrypt-later (HNDL) exposure on all recorded traffic.

- **Streams** (`crypto/key_exchange.rs::derive_stream_keys`) derive `mac_key`/`enc_key` from a single **classical Ristretto255** DH. The #321 AES-256-GCM stream sealing is active but classically keyed. This protects the inference token stream — the highest-value HNDL target.
- **Envelope hybrid encryption is ORPHANED:** `encrypt_envelope_hybrid` (X25519 + ML-KEM-768) is implemented and wired through `envelope.rs`, but **no identity derives an ML-KEM key, no DID `keyAgreement` publishes one, and no production caller invokes it** (abandoned when #152 moved from the `pq-hybrid` cargo feature to runtime `CryptoPolicy` — only signatures carried forward).
- **Notifications/events** (`crypto/notification.rs`, `crypto/event_crypto.rs`) are classical end-to-end; publisher attestation there is plain Ed25519.
- **Transport** channel KEX is classical X25519 (rustls **ring** provider; no `kx_groups` override; iroh defaults).
- **Errors:** envelope reuses the Ed25519 **signing** key as a **static** X25519 KEX key (`to_montgomery()`) → key-reuse + no forward secrecy; non-standard Blake3 KEM combiner; fixed zero AEAD nonce; `epoch` hardwired to 0 (no rekey).

## 2. Principles (decided)

[...]

Issue #553, opened 2026-06-29
S2: canonical COSE_Encrypt0 sealing substrate for hybrid stream/group lifecycles


ewindisch:
Part of #550. Relates to #468. The original S2 implementation landed in #560/#747; this issue remains open for the standards and lifecycle hardening audit required before #554/#555 consume it.

**S2 — reusable canonical COSE_Encrypt0 sealing substrate over Hyprstream's pinned hybrid-KEM suite.**

### Standards boundary

- The wire object is `COSE_Encrypt0` as defined by RFC 9052, using registered COSE algorithm 24 (ChaCha20/Poly1305) for content encryption.
- Hyprstream's suite-identified N-KEM combiner is a project construction. It is not an implementation of HPKE and this issue makes no `draft-ietf-cose-hpke` interoperability claim. In particular, the project's composite encapsulation material is not the draft's HPKE `enc`/`ek` value and its private-use headers are not draft registry assignments.
- PQUIP/RFC 9794 supplies transition terminology and guidance, not wire semantics or a combiner specification.
- Do not reserve or advertise an XChaCha20-Poly1305 COSE algorithm identifier until a complete, policy-selected profile and interoperability evidence exist.

### Deliverables

- A bounded `seal`/`open` API producing and consuming deterministic, canonical `COSE_Encrypt0` bytes and rejecting duplicate/non-canonical headers, truncation, unknown/extra headers, and oversized inputs before expensive crypto.
- Protected, authenticated binding of the pinned suite, recipient identity, selected recipient key id, nonce-domain id, epoch, and sequence; bind the caller's external AAD/transcript through the RFC 9052 `Enc_structure`.
- Reuse the existing N-KEM combiner and canonical suite-ordered material. Reject wrong suite/key/recipient/domain, missing or ignored PQ shares, reordered/cross-paired component material, tampering, and Cartesian pairing of otherwise valid component keys.
- A bounded owning lifecycle state that rejects outbound nonce reuse and inbound replay. Membership, admission, reach resolution, epoch installation/retirement, and grace policy remain with #554/#555/#726/#1004 -> #1039 -> #1027.
- Keep a raw pre-derived-key primitive only as a narrow internal seam for the later stream/group lifecycle owners; it must carry the same authenticated context and cannot itself claim replay safety.

### Acceptance

[...]

PR #857, opened 2026-07-05
feat(#552): wire #mesh-kem keyAgreement into the DID document

### LINKED ISSUE #552: S1: PQ keyAgreement identity + DID #mesh-kem publication + trust anchoring
Part of #550. Relates to #280, #441.

**S1 — PQ keyAgreement identity + DID `#mesh-kem` publication + trust anchoring.**

The orphaned envelope KEM has no key-distribution path. Build the KEM identity parallel to the existing `#mesh-pq` ML-DSA signing key.

### Deliverables
- `derive_mesh_kem_key` in `node_identity.rs`: separated HKDF purpose label (key-separation per PQUIP), deterministic from the node Ed25519 root (survives restarts), distinct from the Ed25519 and `#mesh-mldsa` keys.
- Publish a DID **`keyAgreement`** verification method `#mesh-kem` (Multikey; extend #280's multicodec work to the ML-KEM-768 / suite encapsulation key). For the future McEliece leg, an additional static/rotated `keyAgreement` key.
- Trust anchoring: a `KemTrustStore` keyed by signer identity (mirrors `KeyedPqTrustStore` for ML-DSA), so the recipient KEM key is resolved out-of-band, never self-asserted. Rides the #441 authoritative-key-binding model (registered-or-fail).
- **Rotation / prekeys** for one-shot envelope forward secrecy (short-lived rotated server KEM keys, X3DH-style), since the envelope encapsulates to a published recipient key.

### Acceptance
- A peer can resolve another peer's `#mesh-kem` key via DID and anchor it; unanchored → fail closed (no classical fallback, per #550 principle 1).

ewindisch:
Closes #552, part of epic #550.

## Context

S1 (#552) landed earlier as PR #562 (`derive_mesh_kem_recipient`, the `KemTrustStore` trust store, `KemPrekey` rotation seam, and `did_document.rs` Multikey helpers for `#mesh-kem`), but that PR explicitly deferred three items pending a libtorch-capable build environment (see its body / the code comments it left behind):

1. Wire the `#mesh-kem` helpers into `build_did_document`'s `keyAgreement` set (cross-cutting — touches the signature and every caller).
2. A libtorch compile pass to confirm the wiring actually builds.
3. Confirm the **provisional** ML-KEM-768 multicodec code point against the multiformats registry.

This PR does all three, closing out #552.

## What's implemented

- `build_did_document` (`crates/hyprstream/src/services/oauth/did_document.rs`) now takes `mesh_kem_public: Option<&hyprstream_rpc::crypto::hybrid_kem::RecipientPublic>` and emits a `keyAgreement` array with `#mesh-kem-x25519` / `#mesh-kem-mlkem768` Multikey verification methods. These are embedded directly in `keyAgreement` (did:key convention), not cross-referenced from `verificationMethod`/`authentication`/`assertionMethod` — they're key-agreement-only, never used for signing.
- `OAuthState` (`state.rs`) grows a `mesh_kem_public` field, derived in `with_signing_key` alongside the existing `mesh_pq_verifying_key` derivation, via `hyprstream_rpc::node_identity::derive_mesh_kem_recipient`. A derivation failure is logged and leaves the field unset — the DID document then simply omits `keyAgreement` rather than publish a key that doesn't actually work (fail-closed, consistent with epic #550's "no in-band downgrade / never publish something unusable" posture).
- `root_did_document` passes `state.mesh_kem_public.as_ref()` through; the per-user/per-client DID document endpoints pass `None` (no mesh identity there).

## Judgment call flagged: the multicodec bug

The prior pass's `MULTICODEC_ML_KEM_768_PUB` constant was `0x120b` → bytes `[0x8b, 0x24]`, marked `PROVISIONAL... confirm before production`. I checked the upstream multiformats `multicodec/table.csv`:

```
mlkem-512-pub,  key, 0x120b, draft
mlkem-768-pub,  key, 0x120c, draft
mlkem-1024-pub, key, 0x120d, draft
```

`0x120b` is actually **`mlkem-512-pub`**, not 768. I fixed the constant to the correct `0x120c` → `[0x8c, 0x24]` and updated the doc comment to state this as confirmed (not provisional) against the registry, with a note on what the mixup would have caused (peers silently anchoring the wrong key-size code point). `x25519-pub` (`0xec` → `[0xec, 0x01]`) and the already-shipped `ml-dsa-65-pub` (`0x1211` → `[0x91, 0x24]`) both checked out correct against the same table.

## Deferred (later stages, not this PR)

Per scope discipline in #552/#550: no `cose_encrypt.rs` (S2/#553), no stream hybrid wiring (S3/#554), no envelope/notification hybrid-by-default (S4/#555 — actually already merged separately as #577/`3172d4faf`), no fail-closed posture changes (S5/#556), no transport PQ (S6/#557 — already merged as `eb1a231b5`). This PR is purely: derive → publish → (trust anchoring/rotation already existed in `hybrid_kem.rs` from the earlier S1 pass, untouched here).

## Testing

- `cargo test -p hyprstream --lib services::oauth::did_document`: **20 passed, 0 failed** (18 existing + 2 new: `build_did_doc_omits_key_agreement_when_mesh_kem_absent`, `build_did_doc_publishes_mesh_kem_key_agreement`).
- `cargo test -p hyprstream-rpc --lib node_identity`: **18 passed, 0 failed**.
- `cargo test -p hyprstream-rpc --lib hybrid_kem`: **20 passed, 0 failed**.
- `cargo test -p hyprstream-rpc --lib mesh_kem`: **4 passed, 0 failed** (includes the existing envelope hybrid-encrypt/decrypt roundtrip, unaffected by this change).
- `cargo clippy --workspace --all-targets --features download-libtorch -- -D warnings`: clean, 0 warnings.

## Test plan

- [x] `derive_mesh_kem_recipient` deterministic + key-separated (pre-existing, still green)
- [x] `build_did_document` emits `keyAgreement` with correct shape when a `#mesh-kem` key is present
- [x] `build_did_document` emits an empty `keyAgreement` array (not an omitted field) when absent
- [x] Published multibase bytes round-trip to the exact `RecipientPublic.eks` bytes, with the corrected multicodec prefix
- [x] Full workspace clippy clean

### DIFF
--- a/crates/hyprstream/src/services/oauth/did_document.rs
+++ b/crates/hyprstream/src/services/oauth/did_document.rs
@@ -255,9 +255,9 @@ fn mesh_kem_key_agreement_methods(did: &str, x25519_ek: &[u8], mlkem768_ek: &[u8
 /// Validates the suite shape defensively (exactly the 2 components of
 /// `SuiteId::HyKemX25519MlKem768`, X25519 then ML-KEM-768) before emitting —
 /// `recipient_public` only ever comes from this node's own
-/// `derive_mesh_kem_recipient`, but a shape mismatch must fail closed (omit
-/// `keyAgreement`) rather than publish a malformed or misordered key that a
-/// peer would silently mis-anchor.
+/// `derive_mesh_kem_recipient`, but a shape mismatch must fail closed (publish
+/// an empty `keyAgreement` relationship) rather than publish a malformed or
+/// misordered key that a peer would silently mis-anchor.
 fn mesh_kem_key_agreement(
     did: &str,
     recipient_public: Option<&hyprstream_rpc::crypto::hybrid_kem::RecipientPublic>,
--- a/crates/hyprstream/src/services/oauth/did_document.rs
+++ b/crates/hyprstream/src/services/oauth/did_document.rs
@@ -186,30 +186,23 @@ fn mldsa65_verification_method(did: &str, key_id: &str, vk_bytes: &[u8]) -> Valu
 
 // ============================================================================
 // #mesh-kem hybrid keyAgreement identity (S1 / #552, epic #550)
-//
-// NOTE: this crate requires libtorch to compile, which was unavailable in the
-// implementing fork — the helpers below are faithful mirrors of the working
-// `#mesh-pq` / p256 multibase helpers above but are NOT compile-verified here.
-// Integration into `build_did_document` (push these into a `keyAgreement` array)
-// is a deliberate follow-up: it touches the builder signature + every caller, so
-// it is left to a pass that can compile the `hyprstream` crate.
 // ============================================================================
 
 /// multicodec `x25519-pub` unsigned-varint prefix (`0xec` → bytes `0xec 0x01`).
 const MULTICODEC_X25519_PUB: [u8; 2] = [0xec, 0x01];
 
-/// multicodec `ml-kem-768-pub` unsigned-varint prefix.
+/// multicodec `mlkem-768-pub` unsigned-varint prefix.
 ///
-/// PROVISIONAL: code point `0x120b` (ML-KEM-768 / FIPS 203 encapsulation key) as
-/// an unsigned varint → bytes `0x8b 0x24` (same shape as `ml-dsa-65-pub`
-/// `0x1211` → `0x91 0x24`). **Confirm against the multiformats multicodec
-/// registry / draft-ietf-cose-mlkem before production** — the exact ML-KEM code
-/// point must match what peers decode.
-const MULTICODEC_ML_KEM_768_PUB: [u8; 2] = [0x8b, 0x24];
+/// Code point `0x120c` (ML-KEM-768 / FIPS 203 encapsulation key) from the
+/// multiformats multicodec registry (`table.csv`, status: draft) — confirmed
+/// against the upstream registry (NOT `0x120b`, which is `mlkem-512-pub`; a
+/// one-code-point mixup there would silently anchor peers to the wrong key
+/// size). As an unsigned varint, `0x120c` encodes to the two bytes `0x8c 0x24`
+/// (same shape as `ml-dsa-65-pub` `0x1211` → `0x91 0x24`).
+const MULTICODEC_ML_KEM_768_PUB: [u8; 2] = [0x8c, 0x24];
 
 /// Multibase z-encode (base58btc, multicodec-prefixed) a raw X25519 encapsulation
 /// key (32 bytes) as a `Multikey` `publicKeyMultibase`.
-#[allow(dead_code)] // wired into build_did_document's keyAgreement set by the #552 follow-up
 fn x25519_to_multibase(ek_bytes: &[u8]) -> String {
     let mut payload = Vec::with_capacity(MULTICODEC_X25519_PUB.len() + ek_bytes.len());
     payload.extend_from_slice(&MULTICODEC_X25519_PUB);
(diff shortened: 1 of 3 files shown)


ewindisch:
## 🟡 Review (1:7 coordination sweep) — one med (fail-closed posture) + doc fix

Encoding verified by hand: `0x120c`→`0x8c 0x24` correct for `mlkem-768-pub`; the old `0x120b` was `mlkem-512-pub` and its constant was `#[allow(dead_code)]`/never published, so no wire-compat break; `x25519-pub` `0xec`→`0xec 0x01` correct. Suite-shape pinning (`SuiteId::HyKemX25519MlKem768`, 2-component destructure) correctly prevents misordered publication. No production sender consumes `kem_recipient_for` yet, so no live downgrade path today.

**1. (med) `with_signing_key` (`oauth/state.rs` ~648) logs-and-continues on `derive_mesh_kem_recipient` failure**, publishing a DID doc with no `keyAgreement` without consulting `CryptoPolicy`. Under `CryptoPolicy::Hybrid` the epic-#550 posture ("missing PQ key fails closed, never silently classical") argues this should be **fatal at startup**, not one log line. When S2 (#553) sender wiring lands, a peer resolving this keyAgreement-less DID doc has nothing to hybrid-encrypt to — unless every future sender independently fails closed, this becomes a silent classical fallback this node caused. Gate the behavior on `CryptoPolicy`: fail closed under Hybrid.

**2. (low) Three doc comments say the DID doc will "omit `keyAgreement`", but `build_did_document` unconditionally emits `\"keyAgreement\": []`** (the PR's own test asserts empty-not-omitted). Fix the docs to match the emitted shape.

Item 1 should be resolved before merge. Flagging @1:2.

ewindisch:
## ✅ Blocker verified resolved (1:7 re-review of 8a8f31dc1)

The med fail-closed finding is fully fixed:
- **Policy-gated + propagates + aborts startup:** new `mesh_kem_public_for_policy` (`state.rs:486-506`) branches on `policy.uses_pq()` — Hybrid → `Err(...).context(\"...under Hybrid policy\")`, Classical → warn + `Ok(None)`. `with_signing_key` now returns `Result` and `?`s it; the caller `?`s inside `run()` (`mod.rs:387-391`) **before** any listener binds or `/.well-known/did.json` is served (TLS resolve/thread-spawn/publish all happen later in the same block). So a Hybrid derive failure genuinely short-circuits startup — not a swallowed `return Err`.
- **Policy source is real:** `envelope_policy_from_env()` defaults to Hybrid (fail-closed); only literal `HYPRSTREAM_ENVELOPE_POLICY=classical` downgrades, unknown warns→Hybrid. Not an illusory Classical default.
- **Test-backed:** `mesh_kem_derivation_failure_fails_closed_under_hybrid` asserts `unwrap_err()` + `"Hybrid policy"` in the message; Classical counterpart asserts `is_none()`. (The end-to-end `?`-aborts-`run()` wiring is verified by code-reading, not a test — fine.)
- **Doc finding fixed:** all three "omit keyAgreement" prose comments corrected to "empty keyAgreement"; code emits `[]`.

**One non-blocking residual for optional defense-in-depth:** `mesh_kem_key_agreement` (`did_document.rs:261-279`) returns an empty keyAgreement on a suite-shape mismatch *without* consulting CryptoPolicy — unreachable by construction today (suite id is fixed by the derivation, so `Ok` can't yield a wrong shape), but if that ever changes a Hybrid node could publish empty without failing. Worth a policy check there too, someday. Blocker cleared — thanks.

ewindisch:
Coordination/review note: reviewed the #mesh-kem DID publication changes. The builder now emits keyAgreement only for HyKemX25519MlKem768 with the expected two-component shape, root DID wiring passes the derived material, user/client DID docs stay empty, and Hybrid policy now fails closed on KEM derivation failure. CI is green. No blocking issues found. GitHub treats my account as author-equivalent, so a non-author approval is still required for branch protection.

Issue #791, opened 2026-07-04
K5b: operator controller runtime — kube-rs reconcilers + in-cluster RPC client for *.hyprstream.io CRDs


ewindisch:
Part of #778 (Phase 1). The controller half of the operator; schemas are K5a (#788).

**K5b — a kube-rs controller Deployment that reconciles `*.hyprstream.io` CRDs into hyprstream RPC calls.**

### Deliverables
- Controller runtime (kube-rs `Controller`) with per-CRD reconcilers wired to typed hyprstream RPC clients (RegistryService/ModelService/WorkerService) over QUIC from inside the cluster — envelope-signed, keys via the K1a story.
- **Translator, never a second writer:** reconcile = read spec intent → issue hyprstream API calls → reflect observed git/runtime truth into `status`. Weights move only by RegistryService git operations (promotion = merge, rollback = checkout). Drift between spec and observed state surfaces as a status condition, not a forced write.
- `TenantBinding` enforcement in every reconciler: object's K8s namespace must be bound to the tenant its spec references; unbound → status error, no action (confused-deputy guard, #778).
- Origin markers + ownerReferences on anything the operator creates; finalizers for deregistration; leader election; audit-relevant actions logged on the S7 pattern.
- `Model`/`Adapter` reconcilers land here as the proof pair; `TrainingRun`/`InferenceService` reconciliation are K5c/K6a.

### Acceptance
- On kind: applying a `Model` CR registers/checks out the repo via RegistryService and status reports the observed ref; deleting it deregisters via finalizer; a cross-tenant spec is refused with a clear condition; operator pod restart resumes cleanly (no duplicate registration — idempotent by managed-by markers).

### Depends on / consumes
- K5a (#788) schemas; K1 (#779)/K1a (#783) deploy + keys. Consumed by: K5c, K6a.

PR #863, opened 2026-07-05
feat(k8s): add operator controller runtime (#791)

ewindisch:
## Summary
- add the K5b kube-rs operator runtime scaffold for Model and Adapter CRDs
- enforce TenantBinding before RPC reconciliation and surface rejected/ready phases in status
- apply managed-by metadata via the main resource API and write generation-aware status through the status subresource
- define an RPC abstraction for the future in-cluster Registry/Model clients without binding this PR to transport construction

## Validation
- OPENSSL_NO_VENDOR=1 cargo test -p hyprstream-k8s --features k8s
- OPENSSL_NO_VENDOR=1 cargo clippy -p hyprstream-k8s --features k8s --all-targets -- -D warnings

## Notes
- Stacked on #847 / ewindisch/788-crd-schemas.
- Functional in-cluster RPC e2e still depends on #859/#848 landing or stabilizing; this PR keeps the transport behind HyprstreamOperatorRpc so unit coverage remains offline and deterministic.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7097,7 +7097,9 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_yaml",
+ "thiserror 1.0.69",
  "tokio",
+ "tracing",
 ]
 
 [[package]]
--- a/crates/hyprstream-k8s/Cargo.toml
+++ b/crates/hyprstream-k8s/Cargo.toml
@@ -19,6 +19,9 @@ default = []
 # `ws` adds the websocket upgrade path kube uses for pod `exec`/`attach`
 # (`AttachedProcess`), consumed by the workers `k8s` sandbox backend.
 k8s = [
+    "dep:futures",
+    "dep:tokio",
+    "dep:tracing",
     "kube/client",
     "kube/runtime",
     "kube/rustls-tls",
@@ -26,9 +29,7 @@ k8s = [
     "dep:hyprstream-vfs",
     "dep:hyprstream-rpc",
     "dep:async-trait",
-    "dep:tokio",
     "dep:parking_lot",
-    "dep:futures",
 ]
 
 [dependencies]
@@ -41,14 +42,16 @@ serde = { workspace = true }
 serde_json = { workspace = true }
 serde_yaml = { workspace = true }
 schemars = "1"
+futures = { workspace = true, optional = true }
+thiserror = { workspace = true }
+tokio = { workspace = true, optional = true }
+tracing = { workspace = true, optional = true }
 
 # `/k8s` VFS projection (K7, #782) — all optional, pulled in by the `k8s` feature.
 hyprstream-vfs = { path = "../hyprstream-vfs", optional = true }
 hyprstream-rpc = { path = "../hyprstream-rpc", optional = true }
 async-trait = { version = "0.1", optional = true }
-tokio = { version = "1", features = ["sync", "rt"], optional = true }
 parking_lot = { workspace = true, optional = true }
-futures = { version = "0.3", optional = true }
 
 [dev-dependencies]
 tokio = { version = "1", features = ["macros", "rt", "rt-multi-thread", "sync"] }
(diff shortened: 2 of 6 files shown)


ewindisch:
## 🟡 Review (1:7 coordination sweep) — minor issues (no high); reconcile-loop hardening

**1. (med) Status-write hot-loop.** Every reconcile unconditionally SSA-writes status and the Controller watches the same resource with no generation predicate, so any status write whose content differs (notably `Err(error).to_string()` in `status_for_model_outcome`, or a varying `observed.message`) bumps resourceVersion → watch event → immediate re-reconcile → another RPC call, bypassing the 300s requeue. An RPC error whose Display varies per attempt (addresses, retry detail) hot-loops the reconciler at RPC latency. Add a `predicates::generation` filter or an unchanged-status short-circuit.

**2. (med) `tenant_for_namespace` does a full cluster `list` of all TenantBindings on every reconcile** (no reflector cache) and `find`s the first namespace match in unspecified order. N resources × requeue hammers the API server; two bindings for one namespace → nondeterministic tenant selection.

**3. (low) No exponential backoff** — `model_error_policy`/`adapter_error_policy` use a fixed 300s and each failed pass still does two API writes; a permanently broken RPC endpoint generates steady-state write load forever.

**4. (low) `tokio::select!` in `run_model_adapter_operator` cancels the surviving controller** when either future completes (both return `Ok(())` because `for_each` swallows errors) — one stream terminating silently exits the whole operator `Ok` and drops the other controller.

**5. (low) `status_for_adapter_outcome` discards both the error and `observed.message`** (AdapterStatus carries no message, unlike ModelStatus) — an Adapter stuck `Rejected` gives the user no reason string.

**6. (low) No finalizer** — deletes are never reconciled, so hyprstream-side registry state is never torn down when a Model/Adapter CR is deleted.

Items 1–2 are the ones to resolve before this contract is consumed. Flagging @1:4 / @1:1.

ewindisch:
## ✅ Blocking meds verified resolved (1:7 re-review of a82d84a53)

**1. (med) Status-write hot-loop — FIXED (the load-bearing one).** `model_status_observed_current`/`adapter_status_observed_current` compare `status.observed_generation` to `meta().generation` (`:385-399`); both reconcilers short-circuit to a bare requeue before any status write when observed-current (`:201-203`, `:222-224`). Critically the **error path also stamps `observed_generation`** (`:416`), so even a varying error Display can't re-trigger — one write per generation, loop broken.

**2. (med) `tenant_for_namespace` LIST-per-reconcile + unspecified-order — addressed (partial).** LIST is now amortized behind a 60s TTL cache (`TenantBindingCache`, `:107-116,:300-321`), and the unspecified-order first-match is replaced by deterministic duplicate detection → `TenantBindingDuplicate` (`:353-383`). Both concrete complaints resolved. It's a periodic-LIST cache, not a reflector/watch — acceptable; a reflector is a nice-to-have follow-up, not a blocker.

**Remaining lows (non-blocking, none regressed):** no exponential backoff (`model/adapter_error_policy` fixed interval `:244,:255`); `run_model_adapter_operator` `select!` still cancels the surviving controller if one stream ends (`:149-153`) — worth a `try_join`/restart eventually since a silent one-controller exit is a real availability nit; `AdapterStatus` still has no `message` field so the Err arm drops the reason (`:431-435`); no finalizer for Model/Adapter. All optional follow-ups. The two blocking meds are cleared.

ewindisch:
Coordination update from 1:5.0: after stacked dependents merged into their bases, rebased ewindisch/791-operator-runtime onto updated ewindisch/788-crd-schemas (70b17901a, includes #850). Pushed with force-with-lease from b9efd160f to 889735e07. Conflict resolution kept both K7 /k8s VFS dependencies and operator runtime dependencies. Local validation passed on this branch: cargo test -p hyprstream-k8s --features k8s; cargo clippy -p hyprstream-k8s --features k8s --all-targets -- -D warnings. GitHub checks are running.

PR #864, opened 2026-07-05
Add TrainingRun STEP operator loop

ewindisch:
## Summary
- extend the kube operator runtime to watch TrainingRun resources alongside Model and Adapter
- add a TrainingRun RPC observation contract with STEP phases through promotion/rollback/failure
- persist TrainingRun status through the training API group with produced adapter and failure messages

## Validation
- OPENSSL_NO_VENDOR=1 CARGO_TARGET_DIR=/tmp/hyprstream-794-target cargo test -p hyprstream-k8s --features k8s
- OPENSSL_NO_VENDOR=1 CARGO_TARGET_DIR=/tmp/hyprstream-794-target cargo clippy -p hyprstream-k8s --features k8s --all-targets -- -D warnings
- pre-commit release clippy hook passed during commit

Refs #794
Stacked on #863 / #791.

Issue #794, opened 2026-07-04
K5c: STEP-loop reconciliation — TrainingRun → workers → adapter/weights; promotion = git merge


ewindisch:
Part of #778 (Phase 2). The payoff of the operator: **the STEP loop (Stage → Train → Evaluate → Promote) as a Kubernetes reconciliation loop — GitOps for weights.**

**K5c — `training.hyprstream.io/TrainingRun` reconciliation end-to-end.**

### Deliverables
- Reconciler (K5b runtime): `TrainingRun` spec → placed workload via WorkerService — native backends first (kata/oci/nspawn); `worker.backend = "k8s"` (#781/#793) once landed makes the run itself a scheduler-placed pod with its dataset/output namespace delivered over 9P. Honor `runs_on`/resources via the scheduling vocabulary (#787 map when on the k8s backend).
- STEP stages projected into status conditions: `Staged` (branch/worktree created) → `Training` (workload running; progress from the stream/event plane) → `Evaluated` (quality-filter result recorded) → `Promoted` (RegistryService merge) / `RolledBack` (checkout). Every transition is a git operation via git2db — **the CRD never writes weights; it requests transitions and observes refs** (single-writer rule, #788).
- Promotion gating: spec can require manual approval (`spec.promote: Manual|Auto`) — Manual leaves an `AwaitingPromotion` condition; approval is a spec edit (declarative, auditable).
- Failure semantics: training failure → condition + retained branch for inspection (never auto-deleted); retry policy in spec.
- Relationship to WorkflowService documented: TrainingRun is the *declarative* K8s-facing intent; existing GH-Actions-style workflows remain the imperative path — both drive the same WorkerService, neither wraps the other.

### Acceptance
- On kind (CPU-scale fixture): applying a TrainingRun takes a toy model through Staged→Training→Evaluated→Promoted with a real git merge observable in the model repo; Manual mode halts at AwaitingPromotion; failed run retains its branch.

### Depends on / consumes
- K5a (#788), K5b (#791); optional K4a (#781)/K4b (#793) for pod-executed runs; TTT/training stack, RegistryService.

PR #865, opened 2026-07-05
Add InferenceService serving controller

ewindisch:
## Summary
- extend the kube operator runtime to watch serving.hyprstream.io/InferenceService resources
- build deterministic native serving plans: Deployment, Service, Gateway API HTTPRoute, and HPA/KEDA autoscaler manifests
- enforce the TTT statefulness rule: stateful services floor minReplicas at 1 and use session affinity; stateless services may scale to zero through KEDA
- persist serving status under serving.hyprstream.io with ready replicas and OpenAI-compatible endpoint URL

## Validation
- OPENSSL_NO_VENDOR=1 CARGO_TARGET_DIR=/tmp/hyprstream-792-target cargo test -p hyprstream-k8s --features k8s
- OPENSSL_NO_VENDOR=1 CARGO_TARGET_DIR=/tmp/hyprstream-792-target cargo clippy -p hyprstream-k8s --features k8s --all-targets -- -D warnings
- pre-commit release clippy hook passed during commit

Refs #792
Stacked on #864 / #794.

### DIFF
--- a/crates/hyprstream-k8s/src/operator.rs
+++ b/crates/hyprstream-k8s/src/operator.rs
@@ -17,11 +17,13 @@ use kube::runtime::{watcher, Controller};
 use kube::{Api, Client, Resource, ResourceExt};
 use serde::Serialize;
 use serde_json::json;
+use serde_json::Value;
 use thiserror::Error;
 
 use crate::{
-    Adapter, AdapterStatus, Model, ModelStatus, TenantBinding, TrainingRun, TrainingRunStatus,
-    API_VERSION, MANAGED_BY_LABEL, MANAGED_BY_VALUE,
+    Adapter, AdapterStatus, InferenceService, InferenceServiceStatus, Model, ModelStatus,
+    Statefulness, TenantBinding, TrainingRun, TrainingRunStatus, API_VERSION, MANAGED_BY_LABEL,
+    MANAGED_BY_VALUE,
 };
 
 /// Field manager used by the operator when writing status.
@@ -30,7 +32,9 @@ pub const OPERATOR_FIELD_MANAGER: &str = "hyprstream-operator";
 const MODEL_KIND: &str = "Model";
 const ADAPTER_KIND: &str = "Adapter";
 const TRAINING_RUN_KIND: &str = "TrainingRun";
+const INFERENCE_SERVICE_KIND: &str = "InferenceService";
 const TRAINING_RUN_FINALIZER: &str = "training.hyprstream.io/finalizer";
+const SERVING_APP_LABEL: &str = "hyprstream.io/serving-app";
 
 /// Runtime configuration shared by every reconciler.
 #[derive(Clone, Debug)]
@@ -39,6 +43,14 @@ pub struct OperatorConfig {
     pub requeue: Duration,
     /// Maximum age of the namespace -> tenant cache.
     pub tenant_binding_cache_ttl: Duration,
+    /// Container image used by native serving Deployments.
+    pub serving_image: String,
+    /// Gateway API parentRef name for generated HTTPRoutes.
+    pub gateway_parent_ref: String,
+    /// Container port exposed by the OpenAI-compatible serving process.
+    pub serving_port: u16,
+    /// Prometheus base URL used by generated KEDA ScaledObjects.
+    pub prometheus_server: String,
 }
 
 impl Default for OperatorConfig {
@@ -46,6 +58,10 @@ impl Default for OperatorConfig {
         Self {
             requeue: Duration::from_secs(300),
             tenant_binding_cache_ttl: Duration::from_secs(60),
+            serving_image: "ghcr.io/hyprstream/hyprstream:latest".to_owned(),
+            gateway_parent_ref: "hyprstream".to_owned(),
+            serving_port: 8080,
+            prometheus_server: "http://prometheus-server.monitoring.svc:9090".to_owned(),
         }
     }
 }
@@ -73,6 +89,25 @@ pub struct AdapterObservation {
 pub struct TrainingRunObservation {
     pub phase: TrainingRunPhase,
     pub produced_adapter: Option<String>,
+    pub message: Option<String>,
+}
+
+/// Kubernetes objects projected for a native hyprstream serving endpoint.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ServingPlan {
+    pub deployment: Value,
+    pub service: Value,
+    pub http_route: Value,
+    pub autoscaler: Value,
+    pub prunes: Vec<Value>,
+    pub url: String,
+}
+
+/// Result observed after reconciling an InferenceService serving plan.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct InferenceServiceObservation {
+    pub ready_replicas: u32,
+    pub url: Option<String>,
     pub message: Option<String>,
 }
 
@@ -143,6 +178,13 @@ pub trait HyprstreamOperatorRpc: Send + Sync + 'static {
     ) -> BoxFuture<'a, Result<(), OperatorError>> {
         Box::pin(async { Ok(()) })
     }
+
+    fn reconcile_inference_service<'a>(
+        &'a self,
+        tenant: &'a str,
+        service: &'a InferenceService,
+        plan: &'a ServingPlan,
+    ) -> BoxFuture<'a, Result<InferenceServiceObservation, OperatorError>>;
 }
 
 /// Shared controller state.
@@ -198,7 +240,7 @@ pub enum OperatorError {
     Rpc(String),
 }
 
-/// Run Model, Adapter, and TrainingRun controllers until Ctrl-C.
+/// Run Model, Adapter, TrainingRun, and InferenceService controllers until Ctrl-C.
 pub async fn run_operator<R>(
     client: Client,
     rpc: Arc<R>,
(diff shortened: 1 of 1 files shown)


ewindisch:
## 🔴 BLOCKING REVIEW (1:7 coordination sweep) — resolve before merge

Plan generation is deterministic and well-tested; the blocking items are in what the plan encodes. Do **not** arm auto-merge until these are resolved.

**1. (med) `serving_plan` — Deployment carries `spec.replicas` unconditionally even when an HPA/KEDA autoscaler owns the same Deployment.** With server-side apply + `.force()` on the 300s requeue, every reconcile re-applies `replicas=min_replicas`, reverting autoscaler scale-ups back to the floor — operator and autoscaler fight indefinitely. Omit `spec.replicas` from the applied manifest whenever an autoscaler is planned for the Deployment.

**2. (med) `serving_plan` autoscaler branch — the plan flips between `HorizontalPodAutoscaler` and `keda.sh/ScaledObject` (same name, different GVK) with no prune signal for the abandoned kind.** Editing a spec from stateless/min=0 to min=1 leaves the old ScaledObject alive; KEDA's internal HPA and the native HPA then both target the Deployment → replica thrash. The plan needs an explicit delete/prune entry for the not-chosen kind.

**3. (med) `service_session_affinity` — TTT stickiness via Service `sessionAffinity: ClientIP` doesn't work behind the Gateway.** The "client IP" seen by the Service is the gateway pod's IP, so all traffic through one gateway pod pins to a single backend replica and per-end-user TTT stickiness isn't achieved. Needs `sessionPersistence` on the HTTPRoute (or consistent-hash at the gateway), not Service-level ClientIP.

**4. (low) `keda_scaled_object` — Prometheus `serverAddress` hardcoded** to `http://prometheus-server.monitoring.svc:9090`; belongs in `OperatorConfig`. Otherwise stateless min=0 services on clusters with Prometheus elsewhere never scale up from zero.

**5. (low) `status_for_inference_service_outcome` — `message` never surfaced into status and the Err branch discards the error**; also `ready_replicas == 0` maps to `Pending`, so healthy scaled-to-zero services report perpetual Pending. Distinguish "scaled to zero by design" from "never came up", and surface the message.

**6. (low) Name/URL hygiene** — `hs-serve-{name}` / `{name}.{namespace}.hyprstream.local` not length-validated against 63-char limits; status URL asserts `https://` with no TLS terminating on the route.

*Cross-cutting:* `reconcile_inference_service` is only a trait method with no in-tree applier yet, so items 1–2 are defects in the plan's contract with the future applier — acceptable to resolve as documented plan semantics + a test, but they must be resolved before this contract is consumed.

ewindisch:
## 🔴 Consolidated review (1:9 deep review, operator-directed) — kick back to stack owner

This confirms the 1:7 sweep's six blocking items and adds mechanism-grounded findings from the TTT delta implementation and the DID/discovery layer. **Do not arm auto-merge.** Items below marked (op) reflect operator direction from this review session.

### Sweep items 1–6: confirmed, with refinements

- **Item 1 is worse for stateless-zero than stated**: the plan sets `replicas: 0` on the Deployment when stateless/min=0; a force-apply on the 300s requeue scales a KEDA-scaled-up service *back to zero*, not just to the floor.
- **Item 3 undershoots** — see "Stickiness key" below: cookie `sessionPersistence` is also wrong, not just `ClientIP`.
- **Item 5** needs a status-schema change: `InferenceServiceStatus` has no message/conditions field at all, so surfacing errors requires a field added in the CRD layer (#847 / `serving.rs`), not just operator.rs.

### New findings

**7. (high) The container command doesn't exist.** The Deployment runs `["serve", "--openai-compatible", "--model", …, "--tenant", …]`, but the hyprstream CLI has no `serve` subcommand — the OpenAI-compat surface is `hyprstream service start oai`. Every planned pod CrashLoops on an unknown subcommand; #792's kind-based acceptance test fails immediately. Align the args with the real CLI (or add a real entrypoint first).

**8. (med) No resource requests.** No `nvidia.com/gpu`, no memory sizing for libtorch + weights. #792 calls for "GPU variant by values/spec"; the spec has no field for it. GPU scheduling can't work as written.

**9. (med) No readiness/liveness probes.** Model load takes minutes; without a readinessProbe, `ready_replicas → phase: Ready` counts pods that can't answer, and the Service/HTTPRoute route to them.

### Mechanism-grounded corrections (op)

**Statefulness default (upstream, #847).** `Statefulness` defaults to `Stateless` in `serving.rs`; #792 and the implementation both say stateful is the norm. The per-tenant TTT delta lives only in an in-memory `VarStore` (`training/tenant_delta.rs`), `DeltaPool::get_or_create` always builds a zero delta and **never rehydrates** from `.snapshots/` (`training/delta_pool.rs`), so replica loss = silent loss of all uncommitted per-tenant adaptation. Statefulness is a *correctness* constraint today, not a warm-cache optimization. Default must be `TttStateful`. (Fix belongs on #847; separate comment posted there.)

**Stickiness key.** The delta register is keyed by `Subject` (`DashMap<Subject, …>` in `delta_pool.rs`). Both `sessionAffinity: ClientIP` (wrong key, and broken behind the gateway — the Service sees the gateway pod IP) and HTTPRoute cookie `sessionPersistence` (per-browser-client, not per-Subject; same tenant from two clients forks its register) are incorrect. The plan needs **Subject-header consistent hashing** at the gateway. Portability of header-hash policy (implementation-specific `BackendLBPolicy`) is the one genuinely open design question — encode the requirement in the plan and document the gateway-specific realization.

**Drain-before-scale.** The `TttStateful` rustdoc already names it. Scale-in of a stateful replica must force a delta snapshot/export on drain (preStop / termination-grace semantics in the pod template); today the plan encodes nothing and scale-in is equivalent to mass LRU eviction with no reload path. Tracked app-side as the rehydration issue (linked from #792); until that exists, drain-export + Subject stickiness are mandatory.

**Endpoint naming (op, hard rule): never mint hostnames.** `https://{name}.{namespace}.hyprstream.local/v1` must go, including from tests, and moving the domain into `OperatorConfig` is *also* wrong. Endpoints are resolved via the PDS/DID mechanism: nodes bind operated TLS-DNSSEC domains to their root key via `compute_tls_endorsement` and publish through `DiscoveryService`; Iroh endpoints pin Ed25519 keys; records resolve as `at://did:…` with signed CAR proofs. Concretely: drop `url` from `ServingPlan` (a plan cannot know the endpoint — it's observed truth), have `status.url` come from the applier's observation of what the node announces through discovery, and derive HTTPRoute hostnames from the DID-operated domain.

### Structural recommendation (resolves items 1–2)

Make `ServingPlan` an explicit **apply-set + prune-set**: manifests to server-side-apply (omitting `spec.replicas` whenever an autoscaler owns the Deployment) plus object refs to delete (the not-chosen autoscaler kind on a statefulness/min flip). Self-describing contract instead of documented semantics the future applier may not honor.

### Also still open
- Hardcoded Prometheus `serverAddress` (sweep item 4) → config.
- Hardcoded HPA thresholds (10 req/s, 100 tok/s, backpressure 1) — spec-level class vs values-level defaults per #792's "by values/spec".
- `hs-serve-{name}` 63-char validation (sweep item 6).
- `max_replicas.max(min_replicas.max(1))` silently raises a declared max — at least surface in status.

ewindisch:
## 🔴 Re-review of current #865 head (`3ad9c690`) — serving pods still won't stay healthy

Two runtime blockers remain in the current pushed plan:

1. **Container command starts the service in background mode, not foreground.** The Deployment args are `service start oai --model ... --tenant ...` (`crates/hyprstream-k8s/src/operator.rs:851-857`). In this CLI, foreground service startup is gated by `--foreground`; without it, `service start` goes through the normal start handler that spawns/manages services and then returns. In a Kubernetes container that means PID 1 exits after launching, so the pod restarts instead of running the OAI server. Existing standalone spawner code uses `service start <service> --foreground --ipc` for exactly this reason (`crates/hyprstream/src/cli/service_handlers.rs:227-228`). Add `--foreground` to the planned container args.

2. **The readiness/liveness probe paths don't exist on the OAI HTTP server.** The plan probes `/healthz/ready` and `/healthz/live` (`operator.rs:863-878`), but the HTTP app only registers `/` and `/health` as public health routes (`crates/hyprstream/src/server/mod.rs:49-52`, `:105-111`); I don't see any `/healthz/*` routes in the service code. Kubelet will get 404s and keep the serving pods unready/restarting. Point probes at `/health`, or add the `/healthz/ready` + `/healthz/live` routes before relying on them.

The earlier command/probe findings were directionally addressed, but this current shape still fails at runtime before it can satisfy #792's serving-controller acceptance path.

ewindisch:
Coordination update from 1:5.0: rebased ewindisch/792-serving-controller onto updated ewindisch/794-step-loop (6608f924b). Skipped already-applied base commits and replayed the serving-controller delta. Pushed with force-with-lease from 800035df7 to db63bb52e. Cumulative top-of-stack validation passed on #866; GitHub checks for this PR are running.

ewindisch:
**Triage flag (07-08): not merge-ready despite green CI.** A re-review at head (3ad9c690) found two unresolved runtime blockers: (1) container args are missing `--foreground`, so PID 1 exits and the pod restart-loops; (2) health probes point at nonexistent `/healthz/ready`/`/healthz/live` routes vs the real `/health`. "Clean green" is true for CI but not for review. Holding out of the merge pass — flagging to the operator-stack owner (pane 1:5.0).

ewindisch:
Prereq #869 (delta rehydration + `export_all`/`drain_export`) merged as PR #945. Two review follow-ups land in this controller's rework: (1) `export_all` is exposed but NOT auto-wired into shutdown — the drain hook (call `drain_export()` before scale-in/termination) is this PR's job, since `LocalServiceBridge` has no post-shutdown seam today; (2) the FsOps/9P snapshot path is untested — only the sync std::fs branch is covered; exercise the 9P path when wiring drain.

Issue #795, opened 2026-07-04
K6b: optional KServe adapter — translate serving.hyprstream.io/InferenceService → serving.kserve.io when installed


ewindisch:
Part of #778 (Phase 2, optional). Interop without noun surrender: **KServe as an adapter target, never the owner** — our `InferenceService` lives at `serving.hyprstream.io` (K5a), so there is no CRD collision by construction; this issue makes the two cooperate where KServe is present.

**K6b — a values-gated adapter mode for clusters standardized on KServe.**

### Deliverables
- Discovery-gated: adapter activates only when `serving.kserve.io` CRDs exist (feature-detect via the discovery API); otherwise K6a's native path is used — zero KServe dependency in the default install.
- `ClusterServingRuntime` manifest registering hyprstream as a KServe custom serving runtime (OpenAI-compat container, model `model:branch` naming threaded through runtime args).
- Translation reconciler: `serving.hyprstream.io/InferenceService` → `serving.kserve.io/InferenceService` referencing the hyprstream runtime; ownerRef from ours to theirs; status reflected back. Origin markers so neither controller re-adopts the other's objects (two-writer guard, #778).
- **Statefulness caveat carried through:** KServe/Knative scale-to-zero is only permitted for `stateless`-class services (the K6a TTT rule); `stateful` maps to minReplicas ≥ 1 in the KServe predictor spec.
- Non-goal: reimplementing KServe features (explainers, transformers) — passthrough only.

### Acceptance
- On kind + KServe: applying our InferenceService yields a working KServe InferenceService routed to hyprstream's OpenAI surface; deleting ours garbage-collects theirs; on a cluster without KServe the adapter stays dormant and K6a serves.

### Depends on / consumes
- K6a (#792), K5a (#788), K5b (#791).

PR #866, opened 2026-07-05
Add optional KServe adapter plan

ewindisch:
## Summary
- add a values/discovery-gated KServe adapter config, default disabled
- extend serving plans with optional KServe ClusterServingRuntime + translated serving.kserve.io/InferenceService manifests
- preserve the K6a statefulness rule: stateful minReplicas floors at 1, stateless may scale to zero
- mark translated KServe objects with hyprstream origin metadata and ownerReferences for two-writer guard/GC

## Validation
- OPENSSL_NO_VENDOR=1 CARGO_TARGET_DIR=/tmp/hyprstream-795-target cargo test -p hyprstream-k8s --features k8s
- OPENSSL_NO_VENDOR=1 CARGO_TARGET_DIR=/tmp/hyprstream-795-target cargo clippy -p hyprstream-k8s --features k8s --all-targets -- -D warnings
- pre-commit release clippy hook passed during commit

Refs #795
Stacked on #865 / #792.

### DIFF
--- a/crates/hyprstream-k8s/src/operator.rs
+++ b/crates/hyprstream-k8s/src/operator.rs
@@ -10,6 +10,7 @@ use std::time::{Duration, Instant};
 
 use futures::future::BoxFuture;
 use futures::StreamExt;
+use k8s_openapi::apiextensions_apiserver::pkg::apis::apiextensions::v1::CustomResourceDefinition;
 use kube::api::{ListParams, Patch, PatchParams};
 use kube::core::NamespaceResourceScope;
 use kube::runtime::controller::Action;
@@ -63,6 +64,22 @@ pub struct OperatorConfig {
     pub gpu_count: u32,
     /// Memory request for serving pods.
     pub serving_memory_request: String,
+    /// Optional KServe adapter settings. Disabled unless values/config enable it
+    /// and discovery confirms the `serving.kserve.io` CRDs exist.
+    pub kserve_adapter: KserveAdapterConfig,
+}
+
+/// Values-gated KServe adapter configuration.
+#[derive(Clone, Debug)]
+pub struct KserveAdapterConfig {
+    /// Whether values/config requested KServe adapter mode.
+    pub enabled: bool,
+    /// Whether startup discovery observed KServe CRDs in the cluster.
+    pub discovered: bool,
+    /// Name of the KServe ClusterServingRuntime registered for hyprstream.
+    pub runtime_name: String,
+    /// Image used by the KServe runtime container.
+    pub runtime_image: String,
 }
 
 impl Default for OperatorConfig {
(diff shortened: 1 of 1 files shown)


ewindisch:
## 🔴 BLOCKING REVIEW (1:7 coordination sweep) — resolve before merge

One high-severity multi-tenancy defect plus several contract gaps. Do **not** arm auto-merge until these are resolved.

**1. (HIGH) `kserve_adapter_plan_if_discovered` — cross-tenant leakage through the shared ClusterServingRuntime.** The cluster-scoped, single-named runtime (`config.kserve_adapter.runtime_name`) bakes `HYPRSTREAM_TENANT = <tenant of whichever InferenceService is being reconciled>` into its env and `--tenant $(HYPRSTREAM_TENANT)` into args. With two tenants' InferenceServices, each reconcile rewrites the shared runtime to the last tenant reconciled — tenant B's KServe pods can start with `--tenant tenant-a` (cross-tenant model serving/data exposure), plus permanent write-flapping of the shared object. Move the tenant binding out of the ClusterServingRuntime — put it on the per-isvc predictor (env/args on the translated InferenceService), keeping the runtime tenant-neutral.

**2. (med) The dual gate exists structurally but discovery is unimplemented.** `kserve_adapter.discovered` is a plain bool defaulting false that nothing sets — no code on this branch performs CRD discovery or constructs `OperatorConfig` (verified via git grep across the branch tree). Either implement the discovery probe (and decide startup-time vs per-reconcile re-check — startup-only means KServe installed later is never used, uninstalled-later keeps receiving applies) or re-scope the PR description to say discovery is a follow-up.

**3. (med) Dual serving stacks.** When the adapter is active, the KServe plan is emitted *in addition to* the full native plan (Deployment/Service/HTTPRoute/autoscaler all still built) — one InferenceService produces two independent stacks loading the same model: doubled GPU/memory, two competing endpoints. The adapter should suppress the native serving plan (or be explicitly documented as shadow/migration mode with the native path authoritative).

**4. (med) `minReplicas: 0` + `deploymentMode: RawDeployment` — KServe RawDeployment doesn't support scale-to-zero** (needs Serverless/Knative or KServe's KEDA integration, neither configured). The K6a "stateless may scale to zero" rule silently doesn't hold on the KServe path; it clamps to 1, diverging from the native plan for the same spec.

**5. (low) PR description claims ownerReferences GC for translated objects, but the ClusterServingRuntime can't be owned by a namespaced CR** — deleting the last hyprstream InferenceService orphans it forever. Document the orphan or add explicit cleanup.

ewindisch:
Coordination update from 1:5.0: rebased ewindisch/795-kserve-adapter onto updated ewindisch/792-serving-controller (db63bb52e). Skipped already-applied base commits and replayed the KServe adapter delta. Pushed with force-with-lease from 7f0aaf456 to d46ffbe24. Cumulative local validation passed: cargo test -p hyprstream-k8s --features k8s; cargo clippy -p hyprstream-k8s --features k8s --all-targets -- -D warnings. GitHub checks are running.

ewindisch:
**Triage flag (07-08): not merge-ready despite green CI.** Unresolved HIGH-severity finding: the shared cluster-scoped ClusterServingRuntime is baked with one tenant's `HYPRSTREAM_TENANT`/`--tenant` env, so it leaks cross-tenant model serving across reconciles. Also gated behind #865. Holding out of the merge pass — flagging to the operator-stack owner (pane 1:5.0).

PR #913, opened 2026-07-08
feat(pds) 910a: wire PDS record publication + durable RocksDB store

ewindisch:
## #910a — PDS record publication, correct atproto sign-once model

Wires the `ai.hyprstream.model` PDS record publish/resolve path onto a durable RocksDB store, following the **correct atproto model**: the commit is signed **once at write time** and persisted; reads are a **keyless walk** that serve a proof CAR. The resolver holds **no private key**.

### The bug this fixes
The earlier revision kept records keyed by `(did, collection, tid)` and, on **every `getRecord`/`getRepo` read**, rebuilt the MST and **re-signed the commit** (`Commit::sign` on the read path). That is why the resolver had to hold the `#atproto` private key at read time — the root of this PR's security problems. atproto never re-signs on read.

### The model now implemented
- **Sign once, at write.** `PdsPublisher::publish` rebuilds the MST over the repo's full record set, builds the `UnsignedCommit { did, data = MST root, rev, prev }`, **signs it once** with the `#atproto` P-256 key, and persists the record **and** the signed-commit bytes atomically (one `WriteBatch`). Advancing `rev`/`prev` is the atproto pointer advance (`prev` = previous commit CID, `rev` strictly increases). Only the writer holds the private key.
- **Keyless reads.** `resolve_record`/`resolve_repo` rebuild the deterministic, keyless MST (`Node::from_records`), **load the stored signed commit**, and assemble the proof CAR from `{stored signed commit + MST proof blocks + record}`. Every `Commit::sign` is gone from the read path.
- **The MST is not stored.** It is a pure function of the records, so only the *signature* needs persisting. Storage schema:
  - `rk\0{did}\0{collection}\0{tid}` → canonical DAG-CBOR record bytes
  - `commit\0{did}` → signed-commit DAG-CBOR bytes (new)

### Key consistency (the other half of the fix)
The writer signs with the **active key of the shared `Es256SigningKeyStore`** — the *same* P-256 key `oauth::did_document` publishes as the `#atproto` verification method. Signer and published key are now one source of truth. The duplicate `load_or_generate_atproto_signing_key` loader and the separate `atproto-signing-key` secret are deleted.

### The resolver holds no key
`PdsRecordStore::open`/`open_readonly` and `RepoState` no longer carry a signing key; the discovery factory no longer loads one. Reads are keyless by construction.

### Preserved from the prior revision
RocksDB substrate (records + now signed-commit bytes), restart durability, fail-closed `pds_store_dir()`/`credentials_dir()` (no `/tmp` fallback), 0700 store-dir hardening, single-collection `ai.hyprstream.model` (multi-collection is #910b), deterministic `tid_for_repo` (SHA-256).

### Tests
`cargo test -p hyprstream --lib pds_store` (6 pass):
- `keyless_read_serves_writer_signed_commit` — publish (writer signs), then a resolver built with **no key** returns a proof whose commit `Commit::verify`s against the published P-256 key.
- `publish_read_full_proof_verifies` — publish → getRecord → full `verify_record_proof` (commit sig + MST path + record CID) passes keyless.
- `pointer_advance_bumps_rev_and_links_prev` — re-publish advances `rev` and links `prev` to the prior commit CID.
- `readonly_store_survives_restart_and_sees_writer_updates` — signed commit + records survive reopen and still verify.
- `readonly_store_rejects_writes`, `tid_for_repo_is_deterministic_and_distinct`.

`cargo build -p hyprstream` and `cargo clippy -p hyprstream --lib -- -D warnings` both clean.

---

## Adversarial-review fixes (follow-up commit)

An adversarial review of the reworked store found two HIGH-severity defects; both are fixed in this PR.

### Bug 1 — concurrent publish was an unserialized read-modify-write
All repos publish under the node's single DID, sharing one `commit\0{did}` pointer and one MST. `PdsPublisher::publish` is synchronous and did `load_repo → build_tree → sign → put_record_and_commit` with **no lock**. On a multi-threaded runtime, two concurrent publishes each loaded the same base, added *their* record, and signed a commit over *their* view — the later write clobbered the earlier record while persisting a `commit.data` the stored MST no longer matched, so a reader's rebuilt tree diverged and the served proof failed `verify_record_proof`.

**Fix (both parts):**
- **Serialize the critical section per node DID.** A `parking_lot::Mutex<()>` on `PdsRecordStore` (`publish_lock`) is held across the *entire* `load → build → sign → write` sequence in `publish`. `publish` is sync (no `.await` under the guard), so a plain mutex is correct; since all repos share one DID this is effectively a single publish lock.
- **Read-side fail-closed consistency check.** `resolve_record` and `resolve_repo`, after rebuilding the MST, assert `tree.root_cid() == commit.data` and **refuse to serve** (return an error) on divergence — defence in depth so an inconsistency is detected, never served as a bad proof.
- **Regression test** `concurrent_publishes_serialize_and_stay_consistent`: 8 threads publish distinct records concurrently against one store; asserts all records survive, the stored `commit.data` equals the MST root over the full record set, and every record's proof verifies. **Verified it fails without the lock** (lost update → `commit.data` mismatch) and passes with it.

### Bug 2 — publisher captured the #atproto key once at construction (frozen)
The factory cloned `es256_store.active_key()` into a frozen `SigningKey`. The ES256 store rotates (~14 days); after the first rotation the publisher kept signing with the stale key while `did_document` published the new active key, so new commits failed verification until restart.

**Fix:** `PdsPublisher` now holds an `Es256KeyGetter` (a getter closure wired by the factory to the live `Es256SigningKeyStore`) and resolves the **current active key inside `publish`** — every commit is signed with the same key `did_document` advertises right now. No frozen key is captured.

### Out of scope — tracked in #918
Commits already signed by a now-rotated key stop verifying, because `did_document` publishes only the single active `#atproto` verification method (no drain/lead slots). That is atproto-rotation-log-scale (drain-slot publication or re-sign-on-rotation) and is deferred to **#918**, documented as a `KNOWN LIMITATION` comment on `Es256KeyGetter`.

### Tests / build
`cargo test -p hyprstream --lib pds_store` — now **7 pass** (6 existing + the new concurrency test). `cargo clippy -p hyprstream --lib -- -D warnings` clean (pre-commit hook passed).

### DIFF
--- a/crates/hyprstream-pds/src/tid.rs
+++ b/crates/hyprstream-pds/src/tid.rs
@@ -32,6 +32,15 @@ impl Tid {
         Tid(bits)
     }
 
+    /// The raw 64-bit integer value of this TID.
+    ///
+    /// Used by callers that need to advance a revision monotonically (the
+    /// atproto commit `rev` must strictly increase), e.g. bumping past a
+    /// previous commit's rev when the wall clock has not moved.
+    pub const fn to_raw(self) -> u64 {
+        self.0
+    }
+
     /// Build a TID from a microsecond timestamp (high 53 bits) and a 10-bit
     /// clock id (low 10 bits).
     ///
--- a/crates/hyprstream/src/auth/identity_store.rs
+++ b/crates/hyprstream/src/auth/identity_store.rs
@@ -445,43 +445,12 @@ pub fn resolve_service_signing_key(
     }
 }
 
-/// Load or generate this node's `#atproto` commit-signing key (P-256/ES256).
-///
-/// PDS commits are signed classically (ES256) per atproto — there is no PQ
-/// variant of this key (#910a). Mirrors [`load_or_generate_node_signing_key`]:
-///
-/// 1. Read `secrets_dir/atproto-signing-key` → return if present.
-/// 2. If writable: generate a new key, write, return.
-/// 3. If read-only and missing: hard error.
-pub fn load_or_generate_atproto_signing_key(
-    secrets_dir: &std::path::Path,
-) -> Result<p256::ecdsa::SigningKey> {
-    const NAME: &str = "atproto-signing-key";
-
-    if let Some(bytes) = read_secret(secrets_dir, NAME)? {
-        let seed: [u8; 32] = bytes
-            .as_slice()
-            .try_into()
-            .map_err(|_| anyhow!("secret '{}' must be 32 bytes (P-256 seed)", NAME))?;
-        let sk = p256::ecdsa::SigningKey::from_bytes(seed.as_slice().into())
-            .map_err(|e| anyhow!("secret '{}' is not a valid P-256 key: {}", NAME, e))?;
-        tracing::info!("Loaded #atproto (P-256) signing key from '{}'", secrets_dir.display());
-        return Ok(sk);
-    }
-
-    if !is_writable(secrets_dir) {
-        return Err(missing_in_readonly(secrets_dir, NAME));
-    }
-
-    let key = p256::ecdsa::SigningKey::random(&mut rand::rngs::OsRng);
-    write_secret(secrets_dir, NAME, key.to_bytes().as_slice())?;
-    tracing::info!(
-        "Generated new #atproto (P-256) signing key → '{}/{}'",
-        secrets_dir.display(),
-        NAME
-    );
-    Ok(key)
-}
+// The `#atproto` commit-signing key is NOT loaded here. It is the *active* key
+// of the shared `Es256SigningKeyStore` (`auth::key_rotation`), the same P-256
+// key `oauth::did_document` publishes as the `#atproto` verification method —
+// one source of truth for signer and published key. The PDS writer
+// (`services::discovery::PdsPublisher`) sources it from that store; there is no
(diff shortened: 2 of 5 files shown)


ewindisch:
**Fixup pushed** addressing the review blockers (commit `25b3d0450`):
- **H1** — the `#atproto` private key no longer enters RocksDB. Both the registry (writer) and discovery (resolver) load it from the secrets dir into memory; the store signs CAR proofs from that in-memory key and the DB holds records only. Store dir hardened to 0700.
- **H2** — `pds_store_dir()`/`credentials_dir()` fail closed instead of falling back to `/tmp`.
- **M1** — `tid_for_repo` is now a pure `SHA-256(repo_id)` derivation (no mint, no race, no aliasing).
- Also preserved the original error in the RO bootstrap path (M3).

Build + clippy `-D warnings` clean; the 4 store tests pass (round-trip, restart-durability, read-only rejection, TID determinism/distinctness).

**Deploy note (k8s/systemd):** the `#atproto` key must be a **shared, pre-provisioned** secret (read-only mount / systemd credential) — every replica must load the *same* key or CAR proofs won't verify consistently. Worth a Helm follow-up so the key is provisioned into the Secret rather than generated per-pod.

Issue #924, opened 2026-07-09
feat(pds): ai.hyprstream.ledger.* lexicons + holder-controlled inventory write path (D7)


ewindisch:
Part of #922 (design: #921). Decision **D7**, invariant **D8-1**; plan §4. Depends on #908 (lexicon home) and **hard-requires #910b** (multi-collection MST).

The `ai.hyprstream.ledger.*` lexicons + the holder-controlled PDS write path (the inventory):
- [ ] `allocation` — issuer-signed grant record (UCAN CID, unit, amount, epoch, issuer DID)
- [ ] `receipt` — dual-signed usage record (both principals per #681, grant CID, transfer_id, quantum)
- [ ] `tranche` — cross-cell pre-funded lease record (Phase 2, with sub-issue 4)
- [ ] `checkpoint` — signed ledger head (seq, head_hash, state roots, optional `anchor` update)
- [ ] PDS record structs + canonical round-trip tests in the service layer (WB 1.5, M, mechanical)

Constraints: credits are issuer liabilities (D8-1) — the unit names its issuer; records sign via composite COSE; deterministic rkey = transfer_id for at-least-once receipt idempotency.

ewindisch:
B′ ratification rider: add a group-owned allocation variant — `allocation` records may be held by a placement-group owner (team budgets). Account owner dimension = group URI/owner DID; membership fact = bidirectional-consent lists (`ai.hyprstream.placement.{group,groupItem}`). See #925 comment for the spend-authorization composition; #932 for the ingest-verification gate.

ewindisch:
P0 constraints from the #928 charter (ratified 2026-07-09): (1) **grant classes** — `prepaid` (bearer-like, lease-mode only, issuable to a bare did:key, no identity) vs `underwritten` (issuer relationship, detect-mode eligible); the class↔mode coupling is a protocol rule. (2) Grant binding must support presentation to a **pairwise per-cell identifier** (issuer-linkable, cell-unlinkable). (3) **No legal-identity fields** anywhere in the lexicons — audit binds pseudonyms only.

ewindisch:
## Follow-up resource proof schemas

#1064/#1065 will decide whether dual-attested resource manifests and encrypted detailed receipts require new `ai.hyprstream.resource.*` lexicons or a versioned profile alongside the completed ledger collections. This does not reopen this issue or change its completed ledger-lexicon scope.

Issue #923, opened 2026-07-09
feat(ledger): hyprstream-ledger crate — embedded double-entry credit engine (D1/D3)


ewindisch:
Part of #922 (design: #921). Decisions **D1 D3 D8**; plan §2.0–§2e, §3 (comments 1–2 of #922).

The `hyprstream-ledger` crate: embedded double-entry credit engine over RocksDB, deliberately mirroring TigerBeetle's data model. No RPC/VFS deps; WASM-friendly core; `hyprstream-crypto` is an **unconditional** dependency (no feature gate — signed checkpoints are load-bearing).

Work items (plan §7, Phase 1):
- [ ] 1.1 Crate skeleton: `types.rs`, `backend.rs` (sync `LedgerBackend` trait), errors, `MemLedger` + `engine.rs` state machine; conservation + idempotency proptests vs MemLedger oracle (M, judgment — trait surface is load-bearing)
- [ ] 1.2 `RocksLedger`: CF schema, serialized commit loop (one WriteBatch per logical commit), group commit, crash recovery, expiry sweep (L, judgment)
- [ ] 1.3 Hash-chained journal + `SignedCheckpoint` (commits to seq, head_hash, balances_root, pending_root) + `CheckpointSigner` injection seam + `verify.rs` offline tool (M, mechanical — pattern from `mac/audit.rs`)
- [ ] 1.4 Transactional outbox + at-least-once drain contract (`outbox_peek/ack`), kill-torture crash tests (M, mechanical)
- [ ] 1.9 Property/crash/concurrency suites to plan §6.1–6.3 incl. INV-2 zero-I/O-admit latency regression (M, mechanical)
- [ ] 1.10 `hyprstream ledger verify` CLI + `docs/ledger-architecture.md` (S, mechanical)
- [ ] (Phase 2) 2.5 `TigerBeetleLedger` behind `tigerbeetle` cargo feature (D2) + sidecar journal mirror + nightly equivalence CI (L, mechanical — model maps 1:1 by construction)

Key PLAN DECISIONS to honor: account id = blake3-128 of `(ledger_id, owner_did, unit, purpose)` with issuer baked into the unit; expiry-vs-post race resolved by logical commit clock, first terminal commit wins; outcome-index retention max(30d, 2×max timeout, settlement+7d), post-prune replay = `IdTooOld` never re-execution.

## Resource-entitlement consumer

[...]

Issue #922, opened 2026-07-09
Epic: cellular ledgers — self-sovereign quota (implementation of #921)


ewindisch:
Implementation epic for the cellular-ledger / self-sovereign quota model. **Design theory + reconciliation: #921** (read it first). Technical decisions D1–D8 and the full implementation plan (architecture, core algorithms, failure modes, test plan) are in the first two comments below.

## What this builds

Resource entitlement becomes a capability the identity **holds** (an issuer-signed UCAN grant in its own PDS collection), spent against a **per-cell double-entry credit ledger** (embedded RocksDB engine mirroring TigerBeetle's data model), producing **dual-signed receipts** on the atproto proof plane (triple-entry accounting, PQC-hybrid signed). Cross-cell spend = pre-funded tranches + async net settlement (correspondent banking, ILP/Rafiki as cited pattern — no chain, no coin, no consensus).

**Two invariants (D8, testable):**
1. Credits are **issuer liabilities, not bearer tokens** — cross-issuer exchange only via conversion/settlement, never implicit fungibility.
2. **No ledger tier on the inference hot path** — hot path = local enforcer against a locally-materialized balance; receipts/settlement/anchoring all async.

## Phases

- **Phase 1 — single-cell credits (enterprise-internal quota).** `hyprstream-ledger` crate + lexicons + the local-enforcer contract (the #761 realign): allocate employees/contractors metered inference budgets, hard caps, auditable chargeback. Ships product value with zero federation machinery. Everything lands behind a `ledger` config flag, default off, until the Phase-1 test suites pass.
- **Phase 2 — intra-company cellular.** Exchange capnp schema, tranche state machines, net settlement, TigerBeetle optional backend, duplicity-detection hook (#886 wiring).
- **Phase 3 — inter-org federation.** OpenTimestamps anchoring, fraud-proof verifier + bond registry (gated on slash-economics ratification), BYOC settlement-adapter trait (fiat/invoice exporter as first impl).

## Sub-issues

- [ ] #923 — hyprstream-ledger crate (D1/D3, Phase 1)
- [ ] #924 — ai.hyprstream.ledger.* lexicons + inventory write path (D7, Phase 1; needs #908/#910b)
- [ ] #925 — Phase-1 local-enforcer contract: CreditGate + ledger service + #761 realign (D8)
- [ ] #926 — cross-cell exchange: tranches + net settlement (D4, Phase 2)
[...]

Issue #925, opened 2026-07-09
feat(scheduler): Phase-1 local-enforcer contract — CreditGate + ledger service + #761 realign (D8)


ewindisch:
Part of #922 (design: #921). Invariant **D8-2**; plan §5 (the #761 realign — **highest-leverage slice**). Depends on sub-issues 1 and 2.

The Phase-1 local-enforcer contract: the scheduler stops owning quota and becomes *"verify presented capability → spend a credit → emit a receipt."*

Work items (plan §7):
- [ ] 1.6 Ledger service (`services/ledger/`): actor + `LedgerHandle`, `#[service_factory]`, `CoseCheckpointSigner` (composite, `CryptoPolicy`-aware), receipt emitter draining outbox → PDS, **ReceiptDebt breaker** (sustained receipt debt >15 min / >10k flips enforcer fail-closed for new spends — AuditedAvc shape) (M, judgment)
- [ ] 1.7 `CreditGate`: grant cache + balance cells + generation-counter revocation (AVC amortization model) + spend-authorization verification (M, judgment — security boundary)
- [ ] 1.8 #761 realign: scheduler admission calls CreditGate; reserve/post/void wiring; transfer_id in admission result; **reject-don't-queue** on insufficient balance (retryable 429-style; cold gate / anonymous / unknown unit all deny) (M, judgment)

PLAN DECISION: hold-then-reserve admission — in-memory CreditGate gates admit synchronously (zero I/O, INV-2), durable reserve follows async; bounded over-admission optimism, ledger overdraft check remains the hard floor. Lands behind `ledger` config flag, default off, until 1.9 suites pass. Gated on the #761 group-authority call (#921 open decision 5).

ewindisch:
B′ ratification rider (#921 open decision 5, 2026-07-09): group accounts compose here. A grant may be issued **to a placement group** (the group's ledger account: owner = the group URI/owner DID in the account-id formula). A member spends from the group account with authorization = verified member DID × `PlacementIndex::is_member(did, group)` (bidirectional consent) × the group's held grant; the receipt names both the spending member and the group account (dual-principal vocabulary, #681). Gate: group-account spend authorization requires verified-by-construction ingest (#932).

ewindisch:
## Follow-up resource enforcement consumer

[...]

PR #941, opened 2026-07-09
feat(ledger): hyprstream-ledger crate — types, LedgerBackend trait, MemLedger oracle (#923 item 1.1)

ewindisch:
Part of #923 (epic #922, plan on the epic comments). Item **1.1** — the
accounting-plane skeleton the whole cellular-ledger stack builds on. The trait
surface is load-bearing (every backend and the enforcer depend on it), so this
is the judgment-tier slice; RocksLedger / journal persistence / outbox emitter /
TigerBeetle are later items and are intentionally **not** built here.

## What's in it

New crate `crates/hyprstream-ledger` (MIT), added to the workspace:

- **`types.rs`** — `AccountId` = blake3-128 over `(ledger_id, owner, unit,
  purpose)` (PLAN DECISION 1; issuer baked into the unit per INV-1 and the #924
  P0 pseudonymity rule — no legal-identity fields, DID-only principals),
  `UnitId` (names its issuer), `TransferId` (client-supplied u128), the
  four-counter `Account` quadruple, `Purpose` (incl. `Bond`), `IssueTransfer` /
  `Transfer`, `Outcome`, `BalanceView`.
- **`errors.rs`** — the deterministic error vocabulary. Data-only (no wrapped
  sources) so `Outcome` is `Clone + Eq` and a replay is asserted byte-identical.
- **`engine.rs`** — the pure, backend-agnostic `stage` state machine (plan §2b):
  two-phase reserve/post(full|partial)/void/expire, overdraft check counts
  pending inside the commit, on-touch + sweep expiry judged by the ledger's
  logical commit clock, first-terminal-commit-wins race resolution (PLAN
  DECISION 2), transparent idempotent replay (PLAN DECISION 3).
- **`journal.rs`** — journal / checkpoint / outbox types, the `CheckpointSigner`
  injection seam, and hybrid-composite (`EdDSA + ML-DSA-65`) checkpoint
  verification via `hyprstream-crypto` — an **unconditional** dependency, no
  `cose` feature gate (PLAN DECISION 8, ratified).
- **`backend.rs`** — the synchronous `LedgerBackend` trait (PLAN DECISION 8:
  sync + actor facade owned by the service layer, keeps the core tokio-free).
- **`mem.rs`** — `MemLedger`: the complete in-memory reference implementation
  **and** the proptest oracle. WASM-friendly — no tokio, no I/O, no wall clock;
  the logical clock is advanced explicitly via `advance_clock`, which is what
  makes the expiry-vs-post race deterministic and reproducible.

## Trait-surface decisions (these gate downstream items)

- **Sync `LedgerBackend`** with a `&mut self` single-writer contract; the full
  §3.2 surface (open/credit/debit/reserve/post/void/balance/checkpoint/tick/
  outbox_peek/outbox_ack/journal_range/head) so 1.2/1.3/1.4 just implement it.
- **`AccountId::derive` is the one canonical id function** every backend calls,
  using a domain-separated length-prefixed encoding (not raw DAG-CBOR) so the
  derive stays infallible while remaining canonical across backends.
- **Transparent idempotent replay** is the contract (stored `Outcome` returned
  verbatim, including original errors); `DuplicateTransferId` / `IdTooOld` exist
  in the vocabulary for backends/retention that surface replays explicitly.
- **Same-account transfers are rejected** (`AccountsMustDiffer`, the TigerBeetle
  `accounts_must_be_different` rule) — found via a conservation proptest failure
  where a self-transfer collapsed two deltas onto one row.
- **`Op::Expire`** is an internal, non-idempotency-keyed op emitted by `tick`, so
  the hash chain records every expiry as a real journal entry.

## Tests

`cargo test -p hyprstream-ledger` green: proptests for **conservation** (Σposted
and Σpending debits == credits after every op), **inline idempotency** (replay
yields identical `Outcome`, zero state change), **two-phase overdraft safety**
(no debit-constrained account ever exceeds its credits across
reserve→post/void/expire interleavings), and **expiry-vs-post race determinism**
(never an `Applied` post at/after the deadline; second phase succeeds at most
once). Plus an end-to-end hybrid-composite checkpoint sign/verify roundtrip
proving the `hyprstream-crypto` dependency is load-bearing.

Verified: `cargo build -p hyprstream-ledger --target wasm32-unknown-unknown`
passes; `cargo clippy -p hyprstream-ledger --all-targets` clean; full-workspace
clippy pre-commit hook passed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7125,6 +7125,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "hyprstream-ledger"
+version = "0.1.0"
+dependencies = [
+ "bitflags 2.11.0",
+ "blake3",
+ "ciborium",
+ "ed25519-dalek 2.2.0",
+ "hyprstream-crypto",
+ "proptest",
+ "serde",
+ "thiserror 1.0.69",
+]
+
+[[package]]
 name = "hyprstream-metrics"
 version = "0.1.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ members = [
     "crates/hyprstream-workers-wasmtime",
     "crates/hyprstream-workers-python",
     "crates/hyprstream-k8s",
+    "crates/hyprstream-ledger",
 ]
 # bitsandbytes-sys is standalone - only built when bnb feature is enabled.
 # The wasm guests (hyprstream-workers-python-guest = wasm32-unknown-unknown cdylib; the
--- a/crates/hyprstream-ledger/Cargo.toml
+++ b/crates/hyprstream-ledger/Cargo.toml
@@ -1,0 +1,33 @@
+[package]
+name = "hyprstream-ledger"
+version = "0.1.0"
+edition.workspace = true
+license = "MIT"
+repository.workspace = true
+description = "Embedded double-entry credit engine (accounting plane): LedgerBackend trait, MemLedger reference/oracle, two-phase transfer state machine (#923)"
+
+# The core crate carries no tokio, no FFI, no RPC/VFS/PDS deps (D3): it must
+# build for wasm32-unknown-unknown so a browser cell / Wanix guest can run a real
+# ledger for local-first quota. `hyprstream-crypto` (#920, WASM-safe) is an
+# UNCONDITIONAL dependency — signed checkpoints are load-bearing, so a build that
+# cannot verify them must not exist (no `cose` feature gate, ratified).
+#
+# The `rocks` (RocksLedger, item 1.2) and `tigerbeetle` (item 2.5) backends land
+# behind their own cargo features in later work items; they are intentionally
+# absent here so the skeleton stays FFI-free and wasm-clean.
+
+[dependencies]
+hyprstream-crypto = { path = "../hyprstream-crypto" }
+
+blake3 = { workspace = true }
+bitflags = { workspace = true, features = ["serde"] }
+ciborium = { workspace = true }
+serde = { workspace = true }
+thiserror = { workspace = true }
+ed25519-dalek = { workspace = true }
+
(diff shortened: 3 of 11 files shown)


ewindisch:
⏸️ **Held for human review (Erica).** This is the ledger foundation: the `LedgerBackend` trait surface and the `AccountId`/`UnitId`/`Purpose` model become a long-lived cross-implementation contract (a browser/Wanix cell re-derives account IDs independently). Opus PASS-WITH-FOLLOWUPS; one structural fix (issuer-authority enforced by account identity, not a settable flag) is landing on this branch now. Two contract points worth your eyes before downstream (#923 RocksLedger, #925 enforcer) build on it: (1) `AccountId::derive` uses a domain-separated length-prefixed blake3-128 encoding instead of the plan's `blake3(dag_cbor(...))` — better (infallible) but now normative across implementations; (2) the sync single-writer `&mut self` trait shape. Not auto-merging until you sign off.

ewindisch:
✅ **Fix confirmed** (delta-check of e65fd9317 by the original reviewer). Issuance authority is now structural — `Account` stores `Purpose`, bound into `AccountId::derive`; `stage_credit` rejects any source whose purpose ≠ `IssuerLiability`, so an `Available`/empty-flags account can no longer mint supply. New test `issuance_rejects_available_account_even_with_empty_flags` pins it (would fail pre-fix). Conservation/two-phase/idempotency invariants untouched; wasm constraint intact; the idempotency proptest now also pins no-journal-growth. **Still held for @ewindisch's review** — the two contract points from the hold comment (AccountId encoding as normative cross-impl contract; sync single-writer trait shape) are the open questions for you. Three low follow-ups tracked in a separate issue.

ewindisch:
@claude provide humanized grammar for the decision point with clear examples of the contract for review

Issue #902, opened 2026-07-08
refactor(at9p) F4: rename gittorrent -> hyprstream-p2p (RATIFIED)


ewindisch:
**Ratified rename.** crate dir + `Cargo.toml` package name; workspace members; feature `gittorrent`→`p2p` and **drop from `default`**; `git2db/src/gittorrent_integration.rs` + its `gittorrent-transport` feature; `git-xet-filter/src/gittorrent_storage.rs` + feature; `hyprstream/Cargo.toml:199,205`; CLAUDE.md crate map.

**Land WITH F1–F3 (gut-and-rename in one PR), not before** — renaming a crate about to be gutted is throwaway churn.

- Seam: workspace-wide
- Satisfies: epic F4 (ratified)
- Depends on: F1, F2, F3
- Acceptance: crate is `hyprstream-p2p`; no `gittorrent` references; not in `default`; build + tests green.

---
Parent epic: #880 · Design: #879

Issue #901, opened 2026-07-08
refactor(at9p) F3: retire the libp2p Kademlia stack


ewindisch:
Retire the libp2p stack (kad/mdns/tcp/noise/yamux/identify + the dead `request-response` feature); drop ~15 transitive crates. Removes the git-transport stub path — **obviates #811** (`hash_size=20` bug) in that path; confirm consciously rather than assume.

- Seam: `crates/gittorrent` (→ `hyprstream-p2p`)
- Satisfies: epic F3
- Depends on: F2
- Acceptance: libp2p removed from the tree; build green; #811's path gone (or the fix migrated if any tree-parsing survives).

---
Parent epic: #880 · Design: #879

Issue #900, opened 2026-07-08
feat(at9p) F2: reimplement put/get_object on iroh-blobs + locator (facade preserved)


ewindisch:
Reimplement `{put,get}_object` against {iroh-blobs bytes + locator find}, **keeping the object-API facade** so `git-xet-filter` / `git2db` are untouched. `git-xet-filter/src/gittorrent_storage.rs` (the XET `StorageBackend`, #654) becomes the `hyprstream-p2p` backend.

- Seam: `hyprstream-p2p` service, `git-xet-filter/src/gittorrent_storage.rs`
- Satisfies: epic F2; #654
- Depends on: F1, C2
- Acceptance: XET backend works over the new impl; consumers compile unchanged.

---
Parent epic: #880 · Design: #879

ewindisch:
F1 (#943) review follow-ups to carry into F2, before iroh usage broadens: (1) iroh-blobs 0.101.0 pins iroh-base =1.0.0-rc.0, converging the workspace onto rc.0 (single coherent tree, but confirm nothing depended on an rc.1-only fix); (2) blob GC/tag lifetime — `add_bytes` returns a TempTag dropped right after reading `.hash`; persistence-across-reopen is test-proven, but if iroh-blobs GC is ever enabled, confirm the dropped TempTag doesn't make objects reclaimable.

Issue #899, opened 2026-07-08
feat(at9p) F1: iroh-blobs content-addressed byte transport


ewindisch:
Introduce/evaluate **iroh-blobs** as the bytes transport (content model is already SHA256-CA immutable). Part of the gittorrent→`hyprstream-p2p` convergence.

- Seam: `hyprstream-p2p`
- Satisfies: epic §2 F1
- Depends on: C2
- Acceptance: blob put/get over iroh-blobs, verified by content hash.

---
Parent epic: #880 · Design: #879

ewindisch:
F1 delivered by PR #943 (merged) — iroh-blobs content-addressed byte transport behind the iroh-blobs feature.

PR #943, opened 2026-07-09
feat(gittorrent): iroh-blobs object store spike behind iroh-blobs feature (F1)

ewindisch:
Part of #899 / #880 Track F.

## Phase 1 — evaluation verdict: SUITABLE

**iroh-blobs 0.101.0** (n0-computer, MIT OR Apache-2.0, active monthly cadence — 0.103.0 shipped 2026-06-15):

- **Version fit**: 0.101.0 pins `iroh =1.0.0-rc.0` — the *exact* version this workspace already pins via `web-transport-iroh 0.5.x`, so **no duplicate iroh tree** enters the graph. 0.102/0.103 track `=1.0.0-rc.1` / `^1.0.0` stable; co-bump iroh-blobs with the iroh/web-transport-iroh pair (comment left in `crates/gittorrent/Cargo.toml`).
- **API stability**: pre-1.0 but the post-0.90 "blobs2" store API (`MemStore`/`FsStore` deref to `api::Store`; `blobs().add_bytes()` / `blobs().get_bytes()`) has been stable across 0.94→0.103. One sharp edge found: `FsStore` holds its db lock until an explicit `Store::shutdown()` — dropping is not enough (surfaced as a test hang; the spike exposes `shutdown()`).
- **SHA256 ↔ BLAKE3** (the critical fit question): iroh-blobs is BLAKE3-addressed, the consumer contract (git-xet-filter `StorageBackend`, git2db) is SHA256-keyed. **Decision: keep the facade, bridge with a sha256→blake3 index maintained at put time** — re-keying consumers would ripple into XET pointer formats (`{"xet":"gittorrent","sha256":...}`) and `MerkleHash` interop for zero benefit. The index is a *locator only, never a trust root*: reads are BLAKE3-verified by iroh-blobs, then re-verified against the requested SHA256, so a poisoned/corrupt index entry can fail a lookup but can never serve wrong bytes (covered by a test).
- **Provider announcement**: iroh-blobs deliberately has no global content discovery — providers serve via `BlobsProtocol` on an iroh Endpoint; discovery is external (tickets or a locator). That is exactly the epic's split: **bytes = iroh-blobs, find = at9p mainline locator (C2/#890)**. `blake3_of()` is the exposed seam for F2.
- **wasm**: explicit `wasm_browser` cfg support upstream (mem store; `fs-store` is native-only) — compatible with the Wanix direction.
- **Dependency weight**: +24 lockfile entries, mostly small (bao-tree, irpc, genawaiter, redb for fs-store), vs the 19 `libp2p*` crates plus their tree that F3 (#901) retires — net reduction once libp2p goes, and the substrate (iroh) is already in-tree.

## Phase 2 — what this adds

- `gittorrent` gains an **off-by-default `iroh-blobs` feature** (optional dep, `default-features = false, features = ["fs-store"]`).
- New `gittorrent::blobs::IrohBlobStore`: local (in-process, **no network**) `put_object(Vec<u8>) -> Sha256Hash` / `get_object(&Sha256Hash) -> Option<Vec<u8>>` against a mem or fs iroh-blobs store, with the sha256→blake3 index (append-only sidecar file for the fs backend) and SHA256 re-verification on every read.
- 7 unit tests, no network: mem/fs round-trips, reopen persistence, idempotent put, unknown-hash → `None`, malformed index lines skipped, and poisoned-index-never-serves-wrong-bytes.

**Not in scope here** (per the story split): networked fetch via the locator and wiring `GitTorrentService::{put,get}_object` onto this store is F2 (#900); retiring libp2p is F3 (#901); the `hyprstream-p2p` rename is F4 (#902, lands with F1–F3).

## Test results

`cargo test -p gittorrent --features iroh-blobs`: 52 lib tests passed (7 new), 0 failed; doctests green; clippy clean; default (no-feature) build unchanged.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1039,6 +1039,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "atomic-polyfill"
+version = "1.0.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8cf2bce30dfe09ef0bfaef228b9d414faaf7e563035494d7fe092dba54b300f4"
+dependencies = [
+ "critical-section",
+]
+
+[[package]]
 name = "atomic-waker"
 version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1274,6 +1283,25 @@ dependencies = [
 ]
 
 [[package]]
+name = "bao-tree"
+version = "0.16.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "06384416b1825e6e04fde63262fda2dc408f5b64c02d04e0d8b70ae72c17a52b"
+dependencies = [
+ "blake3",
+ "bytes",
+ "futures-lite 2.6.1",
+ "genawaiter",
+ "iroh-io",
+ "positioned-io",
+ "range-collections",
+ "self_cell",
+ "serde",
+ "smallvec",
+ "tokio",
+]
+
+[[package]]
 name = "base-x"
 version = "0.2.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


ewindisch:
@coderabbitai Resolving: valid Major catch — the TempTag/GC lifetime concern is real. It's non-blocking for THIS PR because F1 is a feature-gated (`iroh-blobs`, off by default), local-only spike with no GC enabled; persistence-across-reopen is test-proven. The TempTag-drop-vs-GC hardening is tracked as an F2 follow-up on #900 (before iroh usage broadens to the networked path). Independently matches the opus review's flag on the same line.

Issue #880, opened 2026-07-08
Epic: did:at9p PQC-hybrid identity + mainline locator + P2P convergence (impl of #879)


ewindisch:
> **Status reconciliation (2026-07-15):** the body below is historical planning context. A linked authoritative comment supersedes stale child status, DID-URL grammar, NodeId identity, PR #1036, and dependency-order claims.

# Epic: `did:at9p` PQC-hybrid identity + mainline locator + P2P convergence

**Design source of truth:** #879 (self-certifying `did:at9p`, mainline DHT as untrusted locator, adversarially reviewed).
**Related:** #550 (hybrid-PQC confidentiality), #705 (`cid.rs` / content-addressed blobs), #547 (native MAC — labels as hints), #385/#361 (iroh pkarr metadata leak / oblivious relay), #910 (PDS-record publication into discovery — the real gate, under #387 Phase 2; #394 was a mis-cite), #873 (resolver profiles), #392 (PDS).

This epic implements #879 and, in the same stroke, **converges hyprstream onto a single P2P substrate** by retiring the redundant libp2p stack in `gittorrent`. It is scoped for implementation: tracks, stories, dependency order, and the exact code seams each story touches.

---

## 1. The integration thesis (read this first)

hyprstream already has a self-certifying peer identity: the iroh **NodeId is an Ed25519 pubkey = `did:key`** (`iroh_peer.rs`, `did_key.rs`). It is self-certifying but **classical-only and un-rotatable**, and its discovery bottoms out in **classical Ed25519-signed pkarr records published to the real BitTorrent mainline DHT** (`iroh_substrate.rs` `presets::N0`, threat-noted at #385).

`did:at9p` is, structurally, **a PQC-hybrid, rotatable wrapper around that iroh NodeId**:

```
did:at9p:<cid512>
   └─ GATE pipeline (H512 + composite verify) ─▶ capsule
        ├─ subject-keys.ed25519  ── IS the iroh NodeId  ──▶ existing dial path (unchanged)
        ├─ subject-keys.mldsa65  ── the hybrid upgrade   ──▶ admission gate key-binding (extended)
        ├─ endpoints["iroh"]     ── NodeId + relays      ──▶ TransportConfig::iroh (existing shape)
        └─ next-key-commitments  ── KERI rotation         ──▶ new (no analog today)
```

[...]

PR #951, opened 2026-07-09
feat(at9p) A4: GATE verification pipeline — canon→hash→sig (#884)

### LINKED ISSUE #884: feat(at9p) A4: GATE verification pipeline (pure fn)
`canon-gate → hash-gate(H512==cid512) → sig-gate` as a **pure, I/O-free** function. Property tests incl. A1/A16.

- Seam: new at9p verify module
- Satisfies: #879 §5.6
- Depends on: A1, A2, A3
- Acceptance: accepts valid capsule; rejects tampered / non-canonical / classical-only; no I/O in the pipeline.

---

ewindisch:
Closes #884, part of #880 Track A

### DIFF
--- a/crates/hyprstream-pds/src/at9p_gate.rs
+++ b/crates/hyprstream-pds/src/at9p_gate.rs
@@ -1,0 +1,453 @@
+//! GATE verification pipeline for `did:at9p` genesis capsules (#884, design
+//! #879 §5.6).
+//!
+//! Given the claimed self-certifying identifier `did:at9p:<cid512>` and the raw
+//! bytes fetched for it from some untrusted locator, this module decides — as a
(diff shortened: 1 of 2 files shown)


ewindisch:
@coderabbitai Resolving: the classical-only rejection is covered where it's enforced — A4's sig-gate delegates to `verify_capsule` (#939), whose suite includes `classical_only_signature_rejected_under_classical_policy` (constructs a valid-classical + invalid/absent-ML-DSA composite and asserts rejection). A4's own `valid_canon_bad_sig_fails_sig_gate` proves the sig-gate is reachable; a duplicate gate-level classical-only test would be redundant. Good catch though — it's a real property, just already pinned upstream.

PR #952, opened 2026-07-09
fix(rpc-derive): satisfy clippy::question_mark in client codegen

ewindisch:
unblocks CI clippy for all PRs; 3 question_mark sites in codegen/client.rs


ewindisch:
Superseded by #955, which fixed these same 3 `client.rs` question_mark sites plus the full Rust 1.97 lint set (useless_borrows_in_formatting, manual_filter, for_kv_map) and the ethnum 1.5.3 compile fix, and pinned the toolchain. #956 then advanced the pin to 1.97.

PR #999, opened 2026-07-10
feat(at9p): persist and surface duplicity alarms

### LINKED ISSUE #888: feat(at9p) B4: duplicity-alarm plumbing (resolves #879 Q6)
Decide + wire where the fail-closed duplicity signal surfaces — S7 audit store (#547 `mac::audit`) and/or operator notification. Resolves #879 Q6.

- Seam: `mac::audit` / notification path
- Satisfies: #879 Q6
- Depends on: B2
- Acceptance: a duplicity event is durably recorded (tamper-evident) and surfaced to an operator.

---

ewindisch:
Closes #888

## Summary
- add an fsync-on-write, BLAKE3 hash-chained duplicity alarm WAL
- checkpoint the journal head to detect tail truncation
- emit structured operator alarms through tracing/OTel
- include the affected did:at9p subject in each alarm

## Verification
- `cargo build -p hyprstream-pds`
- `cargo test -p hyprstream-pds` (166 passed)
- `cargo clippy -p hyprstream-pds --lib -- -D warnings`
- repository pre-commit CI-equivalent clippy hook

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7148,6 +7148,7 @@ dependencies = [
  "tempfile",
  "tokio",
  "tokio-test",
+ "tracing",
  "tracing-subscriber",
 ]
 
--- a/crates/hyprstream-pds/Cargo.toml
+++ b/crates/hyprstream-pds/Cargo.toml
@@ -25,6 +25,9 @@ ed25519-dalek = { workspace = true }
 blake3 = { workspace = true }
 # Non-poisoning Mutex for the in-memory duplicity watermark store (#886).
 parking_lot = { workspace = true }
+# Operator-visible duplicity alarms; the structured event is also exported by
+# the process tracing/OTel subscriber when configured.
+tracing = { workspace = true }
 # Note: DAG-CBOR + CIDv1 + varints are hand-rolled in src/dag_cbor.rs and
 # src/cid.rs to enforce the determinism invariants (sorted keys, minimal ints,
 # tag-42 links) that serde-driven CBOR cannot guarantee. No CBOR/CID crate dep.
--- a/crates/hyprstream-pds/src/at9p_duplicity.rs
+++ b/crates/hyprstream-pds/src/at9p_duplicity.rs
@@ -65,10 +65,48 @@
 //!
 //! This layer only *raises* the duplicity signal, as a typed [`DuplicityAlarm`]
 //! delivered to a pluggable [`DuplicityAlarmSink`]. [`WalDuplicityAlarmSink`]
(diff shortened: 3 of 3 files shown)


ewindisch:
Addressed all three BLOCK findings (commit `eb29e3e2a`), scoped to `crates/hyprstream-pds/src/at9p_duplicity.rs`. Full-workspace release clippy (`-D warnings`) and `cargo test -p hyprstream-pds` (19/19 in the module) pass.

**1 — Crash-window recovery (durability).** `record` now advances the in-memory head immediately after the entry fsync, *before* persisting the checkpoint; a checkpoint-write failure is non-fatal (the WAL entry is already durable) and only logged. `open` accepts an exact checkpoint match **or** a verified one-entry WAL-ahead lag (the durable WAL head is authoritative) and re-anchors the checkpoint — a two-entry lag still fails closed. Regression tests: `wal_ahead_of_checkpoint_by_one_entry_recovers` (recovers + next append still chains), `wal_ahead_of_checkpoint_by_two_entries_fails_closed`.

**2 — Tamper-evidence (security).** Every WAL entry **and** the head checkpoint are now signed with the same pinned hybrid composite (EdDSA + ML-DSA-65, `require_pq`) the at9p record path and `mac::audit` use. A local writer without the signing key can no longer rewrite the chain to a shorter/forged valid journal or roll the checkpoint back to a truncated head — both fail signature verification on `open` (`wal_alarm_rejects_forged_checkpoint`, `wal_alarm_rejects_rewrite_without_signing_key`). The module docs state the guarantee honestly, including the residual limit (total erasure of journal + checkpoint needs an off-host anchor), mirroring `WalAuditStore`'s honest limit — the "tamper-evident" claim is no longer false. Parent-dir fsync added after the checkpoint rename. WAL format bumped to `AT9PALM2`.

**3 — Production wiring.** `WalDuplicityAlarmSink::open` now requires the node's audit signing keys, and `DuplicityGuard::with_durable_alarm(store, wal_path, ed_sk, pq_sk)` is the production construction path — a guard built that way routes **every** `admit_successor` duplicity detection to the signed durable sink by construction. `new()`/`NoopAlarmSink` is documented as the explicit **test/opt-out** path, never the accidental production default. `end_to_end_fork_routes_to_durable_signed_sink` drives a real same-epoch fork through the production guard and asserts the detection is durably recorded and reopens+verifies.

**Honest scoping (no false close):** this makes the durable, tamper-evident route real and the default construction path. The live mesh **PEP that drives `admit_successor`** in the running daemon (the update-record admission flow, still `not wired beyond tests` per `admission.rs`) remains a separate epic step and is **not** claimed live here.

Issue #892, opened 2026-07-08
feat(at9p) C4: R1 no-names + R7 add-only-hints enforcement (resolves #879 Q3)


ewindisch:
**R1** — no string-name path anywhere in the DHT client (API takes `cid512` only; a name→cid cache in the DHT re-opens A5). **R7** — if the BEP44 deterministic-mailbox channel ships, hints may only **add** lookup targets, never prune/replace the watermark-anchored scan (A14). Resolves #879 Q3 (ship the mailbox or BEP5-only for v1).

- Seam: `hyprstream-p2p` locator API surface
- Satisfies: #879 R1/R7, A5/A14
- Depends on: C2
- Acceptance: no name path compiles in; mailbox decision recorded; if shipped, hint channel is provably add-only.

---
Parent epic: #880 · Design: #879

Issue #890, opened 2026-07-08
feat(at9p) C2: mainline locator API + rendezvous


ewindisch:
`async fn providers(cid512) -> Vec<PeerContact>`; k-derived rendezvous with fixed-width `k-index` + `epoch-be64` (§5.2). **TRUNC20 confined to the lookup call site** — the full cid512 keys everything internal (R2); the 20-byte infohash never escapes the DHT client.

- Seam: `hyprstream-p2p` locator module
- Satisfies: #879 §5.2/R2
- Depends on: C1
- Acceptance: announce/get_peers over mainline works; R2 enforced by API shape; infohash never used as an internal key.

---
Parent epic: #880 · Design: #879

PR #1000, opened 2026-07-10
feat(p2p) C4: enforce R1 (no name path) + R7 (add-only hints) #892

### LINKED ISSUE #879: Design: did:at9p self-certifying identity — classical mainline DHT as untrusted locator under PQC-hybrid trust
> **Normative reconciliation (2026-07-15):** implementation semantics were updated after this design closed. See the authoritative reconciliation comment linked here; it supersedes conflicting grammar, NodeId, rendezvous, current-state, live-admission, and PR-status statements below. The historical threat analysis is retained intentionally.

# Design: `did:at9p` self-certifying identity with classical mainline DHT as an untrusted locator (PQC-hybrid trust rider)

Status: **DRAFT / spike write-up** — design proposal, adversarially reviewed (§12). **Implementation planned in epic #880.**
Origin: PQC-DHT spike discussion (2026-07-08).
Related: **#880 (implementation epic)**; **#905 (addressing & composition design — DID URL / at:// / gateway, sibling to this method spec)**; #550 (hybrid-PQC confidentiality epic), #705 (content-addressed blob Mount / `cid.rs`), #809 (9P kernel completion), #547 (native MAC — labels ride in capsules as hints per D1), #385/#361 (iroh pkarr discovery-metadata leak / oblivious relay), atproto PDS (#392), `hyprstream-discovery`.
Identity/discovery neighbors: #579 (algorithm-agile ML-DSA multikey + `did:key` — shared key-binding mechanism), #185 (bind QUIC cert to DID key / two-trust-roots — closed for `did:at9p` by construction), #282 (iroh substrate + pkarr — the foundation; its pkarr demoted to liveness-only), #470 (browser did:key-via-DHT — reshaped, shares the wasm no-raw-mainline constraint), #873 (resolver profiles), #548/#549 (crypto-assurance clearance / federation perimeter binding). Codec: #812 (unified multihash CAS — `cid.rs` alignment), #566 (DAG-CBOR canonical form — R3 dependency).

---

## 1. Summary

We want federated hyprstream nodes to locate and authenticate each other's **authority material** (hybrid signing keys, endpoints, label hints, delegation material) without trusting any resolution infrastructure — and we want the trust to be **PQC-hybrid end to end**, even when the discovery substrate is a classical, SHA-1-era system like the BitTorrent mainline DHT.

The claim this design supports is deliberately narrow:

> **Given `did:at9p:<cid512>`, a client will accept only the unique byte string whose BLAKE3-512 hash matches the cid, and will derive all authority exclusively from PQC-hybrid signatures inside those bytes.**

Under that claim, the mainline DHT is demoted to a **pure, untrusted locator**: it can help you *find* bytes, it can never make you *accept* bytes. A fully adversarial DHT degrades availability, latency, and lookup privacy — never integrity or authenticity. That is what makes the construction **standalone-safe**: it needs no PQ-trusted namespace above it to be sound (it needs one to be *useful for names* — see Non-goals).

**We already ride mainline today, worse.** Iroh node discovery publishes pkarr records to the real BitTorrent mainline DHT (`hyprstream-rpc/src/transport/iroh_substrate.rs`), and those records are Ed25519-signed BEP44-mutable — classical-only trust in the discovery path, with a documented metadata-leak concern (#385, mitigation #361). This design replaces "trust a classical signature from the DHT" with "trust nothing from the DHT"; it is an upgrade over the status quo, not a new exposure.

## 2. Motivation

- Federation addresses atproto DIDs via `hyprstream-discovery` + resolver profiles; `did:web` depends on WebPKI/DNS (classical, revocable, censorable), `did:key` has no rotation and no location.
- Epic #550's posture is fail-closed hybrid-PQC for signatures; but *discovery* of peer key material is currently outside that boundary — today it bottoms out in classical pkarr records on mainline (§1). A HNDL-grade adversary, or just a hostile resolver, sits exactly there.
- We want an identity layer whose verification chain contains **no classical-only trust link**, while piggybacking on massive existing availability infrastructure (mainline: ~10M nodes) instead of building a PQ DHT first.
- The content-commitment machinery is being built anyway for #705 (`cid.rs`, self-describing digests, ratified dual-hash direction), so `did:at9p` is a thin layer over infrastructure we already need.

## 3. Non-goals (explicitly out of scope)

1. **Name resolution.** `name/service/query → cid512` is NOT provided by this design and MUST NOT be answered by the DHT. Semantic lookup requires a PQ-signed `NameRecord` from a trusted namespace authority (PDS / discovery / MoQ group). A DHT that answers name queries with trust is the "toxic mold" failure mode; this design structurally excludes it.
2. **PQ availability.** The DHT can censor, eclipse, delay, and drop. This design does not fix that; it bounds the damage to liveness.
3. **Lookup privacy.** Querying a rendezvous key announces interest in that identity to DHT participants. This is the same leak class as #385 (pkarr publication metadata); mitigations there (#361 oblivious relay, self-hosted `iroh-dns-server`) apply in spirit here (§8, A9) but the leak is not solved at this layer.
4. **Transport confidentiality.** Out of scope; that is #550's COSE_Encrypt0 / N-KEM track. This design covers identity/authority *authenticity* only.
5. **Replacing `hyprstream-discovery`.** The DHT locator slots in beside the existing `RecordResolver` path as one capsule/record source among several; PDS/discovery remain the trusted name path and the witness layer (§7.3).

### 4.1 The trust ladder

```
did:at9p:<cid512>                     (self-certifying; obtained out-of-band or
        │                              from a PQ-signed NameRecord — NOT from the DHT)
        ▼
mainline DHT rendezvous               (UNTRUSTED — contributes zero authority)
        │  returns peer contacts / hints
        ▼
fetch capsule (genesis: whole, size-capped) from N independent providers
        │
        ▼
GATE 1: canonical-form check, then H512(bytes) == cid512   ← load-bearing, PQ-safe
        │
        ▼
GATE 2: verify_composite (EdDSA + ML-DSA-65 COSE) over capsule body,
        Hybrid verification PINNED (§4.4), record-type context string checked
        │
        ▼
GATE 3: rotation-chain validation back to genesis + duplicity check (§7.2)
        │
        ▼
derive: hybrid keys, endpoints, label hints (D1: hints only), delegations
        ▼
dial endpoint → normal RPC SignedEnvelope verification (existing layer)
```

Every security property lives at or below GATE 1. Everything above it is availability plumbing.

### 4.2 Why the classical DHT adds zero attacker power

The decisive observation: **announcing at an infohash is permissionless.** Any DHT participant can `announce_peer` at any 20-byte key without breaking anything. Therefore an attacker who fully breaks SHA-1 (collisions, second preimages) at the rendezvous layer gains *exactly nothing* beyond what the protocol already hands every honest participant: the ability to appear in a provider list and serve bytes. Whatever they serve dies at GATE 1.

Equivalently: this design would be exactly as sound over a DHT with **no cryptography at all**. SHA-1 (BEP5/BEP44 keys) and Ed25519 (BEP44 mutable signatures) at the locator layer are inherited for **liveness only, never trust**.

### 4.3 Self-certification and its contract

`did:at9p` is self-certifying in the `did:key` / onion-address / magnet-link family: the identifier *is* the trust root. Consequences, stated as contract rather than fine print:

- Bootstrapping ("how did you get the DID") is out-of-band **by definition**. That is not a gap to be patched later by the DHT.
- Identifier compromise = trust compromise. There is no recovery of a DID whose genesis capsule embedded attacker keys; you get a new DID.
- The DID is stable across key rotation (§7); the cid names the **genesis** capsule, not the current key state.
- **Freshness assumes a roughly synchronized local clock** (`expires-at` checks). An attacker who controls the victim's time source stretches the rollback bound (§7.2) accordingly. This is a disclosed trust dependency (§8.1), shared with every expiry-based system in the stack (JWT, TLS).

### 4.4 Capsule verification pins Hybrid — independent of node `CryptoPolicy`

`CryptoPolicy::{Hybrid, Classical}` is per-node runtime configuration for the general RPC plane. If capsule verification honored it, a node configured `Classical` would accept an Ed25519-only-valid capsule — and an attacker holding only a stolen classical key could forge identities for every such node. Therefore:

> **at9p GATE 2 requires both composite components (EdDSA AND ML-DSA-65) to verify, always.** Classical-only verification of capsules, update-records, or manifests is forbidden, not configurable. The grammar has no classical-only production and none may be added.

On the signing side, note the actual enforcement point in the code: `sign_composite(ed, None, ..)` will happily emit a classical-only composite — fail-closed lives in the policy-gated caller path and the verify side (`envelope.rs` policy plumbing), not inside `sign_composite`. Capsule code MUST go through the policy-gated path and MUST NOT call `sign_composite` with `pq_sk = None` (review rule R4); a violation is caught by every verifier (availability bug, not a security hole), but should never ship.

### 5.1 Identifier

The cid is a standard **CIDv1** following the repo's `cid.rs` convention (version ‖ multicodec ‖ multihash), with a private-use `at9p-capsule` multicodec in the established xet-xorb/xet-shard LOCAL_CONVENTION pattern. `cid.rs` currently implements sha1/sha2-256/blake3(-256); this work uses **BLAKE3 at 512-bit output** (multihash `0x1e`, length 64 — no new codec), staying in-family with the CAS default (BLAKE3-256, #812). **`H512` below denotes BLAKE3-512.**

BLAKE3-512, not SHA-2/SHA-3: Bernstein/ChaCha lineage (BLAKE3 ← BLAKE2 ← BLAKE ← ChaCha), an XOF so 512-bit is free, no length-extension. The full 512-bit width gives **256-bit collision resistance** for the witness/endorsement path (§7.3) — where a holder-ground collision pair could otherwise transplant a witness endorsement from an inspected capsule onto an unseen twin (a collision, not a preimage, attack; the pure self-certifying path needs only preimage strength, which 256 bits already covers post-Grover). Note `BLAKE3-512(x)[0:32] == BLAKE3-256(x)`, so a capsule's cid contains its CAS address as a prefix; dedup domains must therefore key on `(algo=blake3, len=64)`.

```ebnf
did              = "did:at9p:" , cid512 ;

cid512           = multibase-prefix , cid-v1 ;
multibase-prefix = "b" ;                      (* base32-lower *)
cid-v1           = %x01 , multicodec , multihash ;
multicodec       = at9p-capsule-code ;        (* private-use, LOCAL_CONVENTION,
                                                 registered beside xet-xorb 0x71 /
                                                 xet-shard 0x72 *)
multihash        = mh-code , mh-length , digest ;
mh-code          = %x1e ;                     (* blake3, CAS-aligned; H512 = BLAKE3-512 *)
mh-length        = %x40 ;                     (* 64 bytes *)
digest           = 64 * OCTET ;               (* H512(canonical genesis capsule) *)

did-url          = did , [ path ] , [ "?" , query ] ;
path             = "/" , segment , { "/" , segment } ;   (* 9P path into namespace *)
query            = query-param , { "&" , query-param } ;
query-param      = "service" , "=" , service-id       (* select a capsule service entry, default "#ns" *)
                 | "relativeRef" , "=" , path          (* path relative to the selected service *)
                 | "versionId" , "=" , cid             (* content pin — e.g. git-OID CID *)
                 | "hl" , "=" , cid ;                  (* hashlink content-integrity commitment *)
```

Future digest migration (e.g. SHA-3-512) is a new `mh-code`, not a new DID method.

### 5.1a DID URL dereference (method-defined path semantics)

A `did:at9p` DID URL is a **W3C DID URL**, not an AT-URI and not an authority-bearing URL (the DID is the scheme prefix; the path hangs off it natively — RFC-3986 opaque-path, parses everywhere including WHATWG/browser like `mailto:`). Because DID Core leaves path *meaning* to the method, this method defines it:

```
did:at9p:<cid>/<path>?service=<id>&relativeRef=<rel>&versionId=<pin>
  1. resolve <cid>            → GATE-verified capsule (§4)
  2. select service <id>      → NinePExport service entry (default "#ns")  → TransportConfig
  3. dial + admit             → SecurityContext (assurance created here, §4.4 / #894)
  4. Tattach(aname = export)  → the selected export tree (#877)
  5. Twalk(path / relativeRef)→ walk; path components are SELECTORS, MAC-checked, assert nothing
  6. versionId / hl           → client verifies dereferenced bytes against the pin, else reject
```

For **foreign methods** we don't own (`did:web`, `did:plc`), we use only the registered-parameter form (`?service=&relativeRef=`) and read the endpoint from the resolved DID document — we never assume their bare-path semantics.

**Durable/live rule (load-bearing).** `serviceEndpoint`s are durable, signed capsule content — changing one is a new signed capsule state. Runtime objects (tasks, sessions, live streams) are therefore **addressable only via `path` (walk), never via a service entry**: they appear in the walked namespace (`/proc`-style) and require no capsule mutation to become addressable. Instance-scoped path segments (e.g. `task/<id>`) are **non-aliasing pins** — a re-dereference reaches the same instance or fails (ENOENT), never a different one. See #905 for the full addressing/composition design and #834 for the task `/proc` projection.

### 5.2 Rendezvous (locator layer — untrusted)

```ebnf
infohash        = TRUNC20( H512( context , cid512-raw , k-index ) ) ;
context         = "at9p:provider:v1" | update-context ;
update-context  = "at9p:update:v1" , epoch-be64 ;   (* fixed-width big-endian epoch *)
k-index         = 1 * OCTET ;                        (* fixed-width redundancy index,
                                                        0x00 for the primary key *)

(* BEP5: providers announce_peer(infohash); clients get_peers(infohash).
   Returned values are bare peer contacts — they carry NO authority. *)
dht-value       = peer-contact , { peer-contact } ;
peer-contact    = ip , port ;
```

All variable components inside the hash input (`epoch-be64`, `k-index`) are fixed-width, so no two `(context, cid, k)` triples can collide by concatenation ambiguity.

### 5.3 Genesis capsule (the bytes cid512 commits to)

Canonical encoding: **deterministic DAG-CBOR** (the hand-rolled codec in `hyprstream-pds/src/dag_cbor.rs`; Phase 0 decides whether it moves to a shared crate or the verifier depends on `hyprstream-pds` — see §10). Consistent with the CBOR-durable rule.

```ebnf
capsule          = capsule-body , signatures ;

capsule-body     = version , subject-keys , next-key-commitments ,
                   services , [ label-hints ] , [ delegations ] , [ witnesses ] ;

version          = "at9p-capsule/1" ;         (* doubles as the COSE context string,
                                                 §4.4 / A18 *)
subject-keys     = hybrid-keypair , { hybrid-keypair } ;
hybrid-keypair   = ed25519-pub , mldsa65-pub ;
next-key-commitments = { key-digest } ;       (* MAY be empty: declared-immutable
                                                 identity, rotation frozen forever *)
key-digest       = H512( hybrid-keypair ) ;

services         = service-entry , { service-entry } ;   (* named service MAP, DID-Core
                                                            serviceEndpoint-as-map; §5.1a
                                                            ?service= selects by id *)
service-entry    = service-id , service-type , endpoint ;
service-id       = "#" , token ;                         (* e.g. "#ns", "#pds" — default "#ns" *)
service-type     = "NinePExport" | "AtprotoPDS" ;        (* NinePExport.endpoint ≙ TransportConfig *)
endpoint         = transport , address ;
transport        = "iroh" | "quic" | "moq" | "https" ;

witnesses        = witness-ref , { witness-ref } ;      (* freshness anchors, §7.3 *)
witness-ref      = did | endpoint ;

signatures       = cose-composite ;    (* policy-gated composite signing path —
                                          EdDSA + ML-DSA-65 nested COSE; the
                                          record-type context string is bound in
                                          the COSE protected header; Ed25519-only
                                          is NOT accepted (§4.4) *)
```

`label-hints`: MAC labels in capsules are **hints, never authoritative PDP inputs**, per the ratified interface policy and D1 (#698/#699) — trusted only from services we operate, clamped at import boundaries (`import_label` / `bind_time_label` restrict-only clamps) otherwise.

**Genesis size cap:** the genesis capsule is fetched **whole**, with a hard length cap (kilobytes, exact number an open question — §11). No sharding for genesis in v1: a shard manifest for the genesis fetch would be signed by keys *inside the object being fetched* (circular), so its signature is unverifiable at that moment and an attacker-crafted manifest with a huge `total-len` becomes an unbounded bandwidth-waste amplifier. Sharding (§5.4) is for large objects the capsule *references* by their own cid512.

### 5.4 Shard manifest and shards (post-genesis objects only — never DHT-stored)

```ebnf
manifest        = subject-cid512 , codec , total-len ,
                  shard-entry , { shard-entry } , cose-composite ;
shard-entry     = index , shard-len , shard-digest ;
shard-digest    = H512( shard-bytes ) ;

shard           = subject-cid512 , index , shard-bytes ;
```

Acceptance: `total-len` is checked against a policy cap before any fetch; the manifest signature is verified against **already-established** capsule keys (meaningful precisely because this path is post-genesis); each shard is checked against its manifest digest (early rejection); the final gate is always `H512(reassembled canonical bytes) == subject-cid512`.

### 5.5 Rotation record

```ebnf
update-record   = subject-cid512 , epoch , prev-record-digest ,
                  new-capsule-body , expires-at , cose-composite ;
                  (* context string "at9p-update/1" in the COSE protected header;
                     signed by keys whose H512 appears in the predecessor's
                     next-key-commitments *)

chain-valid     = genesis-check , { successor-check } , duplicity-check ;
genesis-check   = "H512(genesis capsule) == cid512" ;
successor-check = "composite sig verifies (both components)
                   AND signer key-digest ∈ predecessor.next-key-commitments
                   AND prev-record-digest == H512(predecessor record)
                   AND epoch strictly increasing
                   AND now < expires-at" ;
duplicity-check = "new segment chains through the client's persisted
                   (epoch, H512(accepted record)) watermark;
                   a VALID record divergent from the persisted digest at
                   epoch <= watermark is DUPLICITY: hard fail-closed + alarm,
                   never treated as a mere lower-epoch reject" ;   (* §7.2 *)
```

### 5.6 The whole thing in one production

```ebnf
accept    = did , dht-lookup , fetch , canon-gate , hash-gate , sig-gate , chain-valid ;
canon-gate = "reserialize(decode(bytes)) == bytes" ;   (* strict canonical form *)
hash-gate  = "H512(bytes) == cid512" ;                 (* load-bearing *)
sig-gate   = "verify_composite, Hybrid pinned, context string checked" ;
dht-lookup = "?" ;                                      (* contributes zero trust *)
```

### 6.1 You cannot choose BEP44 immutable keys

BEP44 immutable puts derive the key as `SHA1(bencoded v)`; you cannot rendezvous at a chosen `H(prefix || cid512)` with them. Options, in preference order:

1. **BEP5 announce (chosen).** Treat the §5.2 infohash as a torrent infohash; providers `announce_peer`, clients `get_peers`, then fetch the capsule over a real transport (iroh/QUIC/HTTPS) from returned peers. Zero novel DHT machinery; blends with torrent traffic.
2. **BEP44 mutable with a deterministic keypair.** Derive an Ed25519 seed as `H(cid512 || "rendezvous-v1")` — the keypair is public by construction, so the record is a *public mailbox* anyone can overwrite. Acceptable because values are untrusted hints; overwrite wars are availability-only. **Hint rule (load-bearing):** mailbox hints may only **add** lookup targets — they may never prune, replace, or terminate the client's sequential update scan from its persisted watermark (§7.2). A client that lets a bogus "current epoch" hint skip the scan converts an availability channel into a rollback assist (A13).

### 6.2 PQ signatures physically do not fit in DHT values

BEP44 caps values at 1000 bytes. ML-DSA-65 signatures are 3,309 bytes (ML-DSA-44: 2,420). Therefore **no PQ-signed record can live in the DHT**, period. This kills the naive "signed ProviderAdvert in the DHT" design — and that forcing function is *healthy*: it makes "the DHT proves nothing" structural rather than disciplinary. All PQ-signed material is fetched from providers; DHT values are contacts and small unsigned hints only.

### 6.3 Redundancy for availability

- Publish/announce at `k` derived rendezvous keys via the fixed-width `k-index` (§5.2).
- Query multiple bootstrap paths; require/prefer multiple independent providers.
- Providers re-announce on the standard mainline cadence (announce tokens are honored for ~10 minutes per BEP5 guidance; re-announce every ~15–30 minutes).

### 6.4 Existing DHT footprints in this repo (and why they aren't this)

- **`gittorrent`** has a full DHT stack — but it is **libp2p Kademlia** (`crates/gittorrent/src/dht/`), a disjoint overlay network, not mainline BEP5. Its domain-separated 256-bit key derivation (`GitHash::to_dht_key()`) is a reusable *pattern*; the network is not reusable for "meet the whole world" rendezvous. **Per #880 this crate is renamed `hyprstream-p2p` and its libp2p Kademlia is retired** — the at9p mainline locator + iroh-blobs transfer replace it; "gittorrent" was doubly wrong (drops git-specificity, drops libp2p pseudo-BitTorrent, adds *real* mainline).
- **iroh pkarr discovery** (`hyprstream-rpc/src/transport/iroh_substrate.rs`) already publishes Ed25519-signed BEP44-mutable records to real mainline — classical trust in the discovery path, plus the #385 metadata leak. at9p subsumes this role for federated identity with strictly weaker DHT trust (zero) and the same-or-better privacy posture (#361 applies to both).

### 7.1 KERI-style pre-committed rotation

Prior art: **KERI** (key event logs with pre-rotation). We adopt the core mechanism, not the full stack: the genesis capsule commits (by H512 digest) to the *next* key set; a rotation is an `update-record` signed by those pre-committed keys, itself committing to the following set. The chain verifies back to genesis with no external trust. The DID stays stable; key state evolves.

Compromise of a *current* key does not let the attacker rotate — rotation authority lives in the pre-committed next keys (held offline/cold). Compromise of the *next* keys before use is the KERI-known catastrophic case (A11); §7.2's duplicity detection turns it from "silent takeover" into "loud, detectable fork." Operational mitigations: cold storage; a future `next-key-commitments` entry may be a threshold structure.

**Edge cases (specified, not implied):**
- **Empty `next-key-commitments` = declared-immutable identity.** Legal and useful (e.g. ephemeral or fixed-function identities). The chain is frozen at genesis by construction; any update-record for such a DID is invalid.
- **Lost next keys = identity frozen at its current epoch, forever.** This is a feature (fail-closed), not an error state: the current keys keep working until holders choose to abandon the DID and mint a new one. There is no recovery path, by design — a recovery path is a backdoor.

### 7.2 Rollback and duplicity (stated honestly)

Two distinct adversarial behaviors, requiring two distinct defenses:

**Rollback** — serving an *older but genuinely valid* chain state (pre-rotation, pre-revocation). Bounded by:
- strictly increasing `epoch` + **persistent** client-side high-watermark (a client that has accepted epoch N never accepts < N — R6);
- short `expires-at` on update records (stale state eventually fails closed, up to local-clock error — §4.3);
- multi-provider agreement (prefer the max **duplicity-checked** valid epoch across N independent providers).

**Duplicity (fork)** — an attacker with stolen pre-committed keys, or an equivocating holder, publishing a *divergent* valid successor: a second, different record at an epoch the client has already accepted, or a higher-epoch record chaining through the attacker's fork rather than the client's accepted history. An epoch watermark alone does NOT catch this — the attacker's epoch N+1 exceeds the watermark N while silently replacing the honest key state; naive "accept max epoch" would actively select the fork. Defense:

> Clients persist `(epoch, H512(accepted update-record))` per DID — not just the epoch number. Every new chain segment must chain through the persisted digest (`prev-record-digest` linkage, §5.5). A record that is *cryptographically valid* but divergent from the persisted digest at ≤ watermark epoch is **duplicity: hard fail-closed for that DID plus an operator alarm** — the DID is provably equivocating or its rotation keys are stolen; both mean "stop trusting, investigate," never "pick the higher epoch."

This is KERI's duplicity-detection made first-class. Residual exposure: a **first-contact** client (no persisted history) meeting only the attacker fork cannot distinguish it from the honest chain — that window cannot be closed to zero without a witness layer. No amount of DHT cleverness fixes it; any design claiming otherwise is wrong.

### 7.3 Witnesses (the standalone boundary)

The capsule MAY list `witnesses` (PDS instances / discovery services we or the identity holder operate). A witness countersigns the latest epoch and, crucially, gives first-contact clients a second independent view of the chain (duplicity becomes visible when views diverge). Clients that can reach a witness get freshness and fork-evidence up to witness honesty; clients that cannot still get the standalone guarantee (integrity + authenticity + bounded rollback + duplicity detection after first contact). This is KERI's witness model; the atproto PDS (#392) is the natural implementation — it is already exactly a signed-sequenced record store (MST, signed commits, CAR proofs).

## 8. Adversarial analysis

Adversary classes considered: (i) fully malicious DHT (Sybil majority, eclipse-capable), (ii) malicious providers, (iii) network MITM, (iv) HNDL / future CRQC adversary, (v) compromised current subject key, (vi) compromised next (pre-committed) key, (vii) malicious/equivocating identity holder, (viii) malicious time source, (ix) malicious namespace above (out of scope but boundary-checked).

| # | Attack | Capability required | Effect under this design | Verdict |
|---|--------|--------------------|--------------------------|---------|
| A1 | **Forge capsule** for a known cid512 | BLAKE3-512 second preimage | None achievable (2⁵¹² preimage; ~2²⁵⁶ post-Grover, and Grover is sequential/non-parallelizing so not wall-clock-feasible) | Defeated (GATE 1) |
| A2 | **Serve junk / oversized data** | none (permissionless announce) | Genesis: fetch-whole + hard size cap bounds waste per provider (no manifest amplification — §5.3). Post-genesis shards: manifest sig verified against established keys + `total-len` policy cap + per-shard digests. Providers scored/blacklisted locally | Nuisance, bounded |
| A3 | **Rendezvous squatting / poisoning** — flood provider lists with attacker peers | none | Client dials junk, all fetches fail GATE 1; availability harm only. Mitigate: k derived rendezvous keys, provider scoring, parallel fetch | Availability-only |
| A4 | **SHA-1 collision on infohash** (two cids share a rendezvous) | ~2⁸⁰ birthday work | Provider lists mingle; each client filters by its own cid512 gate. No trust impact | Defeated (trust); cosmetic (liveness) |
| A5 | **Name substitution** — answer `name → wrong cid512` | control of DHT | Structurally impossible: the DHT is never asked names (§3.1). Residual risk is *implementation* drift — see R1 | Excluded by construction |
| A6 | **Rollback** — serve old valid chain state | control of provider set reachable by victim | Bounded by persistent watermark + expires-at + multi-provider max-valid-epoch + witnesses (§7.2–7.3). **Nonzero residual window** for witness-unreachable clients, stretched by clock error (A17) | **Accepted residual**, bounded and disclosed |
| A7 | **Duplicity / chain fork** — divergent valid successor via stolen pre-committed keys or equivocating holder | next-key theft (A11) or holder malice | Detected after first contact via persisted `(epoch, record-digest)` chaining (§7.2): hard fail-closed + alarm, never max-epoch selection. **First-contact clients** on an attacker-only view cannot detect the fork without a witness | Detected-and-fail-closed post-first-contact; first-contact residual closed only by witnesses |
| A8 | **Eclipse / censorship** — victim sees only attacker DHT nodes | Sybil positioning around victim | Total denial of discovery for that client; zero forgery. Fallback: endpoints cached from prior capsule, PDS/discovery resolver path, direct dial | Availability-only; out of scope to fix at this layer |
| A9 | **Lookup surveillance** — observe who resolves which DID | passive DHT participation | Real leak: rendezvous key ↔ cid512 linkable by anyone knowing the cid. Same class as the existing pkarr publication leak (#385); mitigations track #361 (oblivious relay) / resolving via PDS; BEP5 traffic blends with torrent noise. **Disclosed, unsolved at this layer** | Accepted residual (Non-goal 3) |
| A10 | **HNDL / future CRQC** | record now, break later | Nothing confidential crosses this layer (public authority material). Authenticity chain: BLAKE3-512 + ML-DSA-65 both PQ; the classical composite component is redundant, not load-bearing; locator-layer SHA-1/Ed25519 carry no trust (§4.2) | Defeated for authenticity; confidentiality N/A (#550) |
| A11 | **Compromised current subject key** | key theft | Attacker signs envelopes until rotation; cannot rotate (rotation authority = pre-committed next keys). Recovery = publish rotation record; propagation bounded by A6 window | Standard; recoverable |
| A12 | **Compromised next (pre-committed) keys** | cold-key theft | Attacker can rotate — but any use forks the chain and trips duplicity detection (A7) for every client with history; DID becomes loudly untrustworthy rather than silently owned. Inherent to pre-rotation (KERI-known). Mitigate: cold storage, future threshold commitments; identity abandoned, new DID minted | Contained to first-contact victims; disclosed |
| A13 | **Malicious witness** | witness key | Can *withhold* countersignature (freshness degrades to standalone bound) or countersign a fork it was shown (its signatures on both forks are then themselves duplicity evidence). Cannot mint fake state — epochs remain chain-validated. Witness set is holder-chosen | Bounded: witnesses add freshness/evidence, cannot subtract authenticity |
| A14 | **Mailbox-hint pruning** — bogus epoch hint in the public BEP44 mailbox causes client to skip real rotations | none (seed is public) | Excluded by the hint rule (§6.1): hints add lookup targets, never prune/replace/terminate the watermark-anchored scan. Without the rule this would be a rollback assist, not mere availability | Excluded by construction (rule is load-bearing — R7) |
| A15 | **Truncation misuse inside hyprstream** — 20-byte infohash used as identity/cache/dedup key | implementation bug | Would promote a 2⁸⁰-collision value onto the trust path | **Must be excluded by review rule R2** |
| A16 | **Canonicalization ambiguity** — two encodings of "the same" capsule | encoder divergence | GATE 1 preceded by strict canonical check: `reserialize(decode(bytes)) == bytes`, else reject (never normalize). The PDS deterministic DAG-CBOR codec supports this | Must-fix in implementation (R3) |
| A17 | **Clock manipulation** — attacker controls victim's time source | NTP/host control | Stretches every `expires-at` bound; A6's window grows with clock error. Disclosed trust dependency (§4.3), shared with JWT/TLS expiry across the stack | Accepted residual, disclosed |
| A18 | **Cross-context signature reuse** — replay a capsule-key signature from another context (RPC envelope, other record type) as a capsule/update signature or vice versa | valid signature from any context | Record-type context strings (`"at9p-capsule/1"`, `"at9p-update/1"`) bound in the COSE protected header and checked at GATE 2; RPC envelopes have their own AAD/context. No cross-context bytes verify | Defeated (given R8) |
| A19 | **Downgrade to classical sig** | strip ML-DSA component | GATE 2 pins Hybrid verification independent of node `CryptoPolicy` (§4.4); a `Classical`-configured node still rejects classical-only capsules. Signing side must use the policy-gated path (R4) | Defeated, by pinned verification |

### 8.1 Residual-risk register (the honest list)

1. **A6 rollback window** for witness-unreachable clients — bounded, nonzero, stretched by clock error (A17).
2. **A7 first-contact duplicity** — a fresh client meeting only an attacker fork; closed only by witnesses.
3. **A9 lookup privacy** — real, unsolved at this layer; tracks #385/#361.
4. **A12 next-key compromise** — detection (loud fork) not prevention; operational mitigation only.
5. **A17 local-clock trust** — all freshness bounds are modulo clock error.
6. **Availability under eclipse/censorship (A3/A8)** — classical-DHT-grade, by design; PDS/discovery path is the fallback.

### 8.2 Review rules for implementation (adversarially motivated invariants)

- **R1 — No names in the DHT, ever.** The DHT client API takes `cid512` and returns peer contacts. There is no code path accepting a string name. Any future "cache the NameRecord in the DHT" proposal re-opens A5 and must be rejected (or the record must be PQ-signed and verified against a pinned namespace authority — at which point the DHT is still only a locator for it).
- **R2 — The 20-byte infohash never escapes the DHT client.** All internal keying (cache, dedup, logs, AVC object identity, audit records) uses the full cid512. The truncated form exists only at the lookup call site.
- **R3 — Canonical bytes or reject.** Fetched capsules/manifests/updates must round-trip byte-exact through the deterministic DAG-CBOR codec before GATE 1; non-canonical encodings are rejected, never normalized.
- **R4 — Fail-closed hybrid, at the right layer.** Capsule/update/manifest signing goes through the **policy-gated composite path**; direct `sign_composite(_, None, ..)` (classical-only) is forbidden in at9p code — `sign_composite` itself does not enforce policy, its callers do. Verification pins Hybrid unconditionally (§4.4), independent of node `CryptoPolicy`.
- **R5 — Label hints are hints; assurance has one provenance.** Capsule `label-hints` feed the D1/D2 clamp path (`import_label` / `bind_time_label`, restrict-only), never the PDP directly. `PqHybrid` assurance is *created only* at the D2 admission boundary from verified capsule material and carried via `SecurityContext`/`SubjectCtx`/`VerifiedKeyMaterial` (→ MAC `SecurityLabel::assurance`). Consistent with the ratified MAC interface policy: caller/wire-supplied labels are forbidden as authoritative inputs — and, per the downstream MAC review, the mount-surface instances are named explicitly: **`#at/<did>`, `#task/at/<did>`, `aname`, mount metadata, PodSpec/CSI fields, and wire labels are selectors/hints only and MUST NOT assert assurance.**
- **R6 — Watermark = (epoch, record digest), persistent.** Clients persist the pair, not the bare epoch; regression below the epoch is a reject, and a valid-but-divergent record at ≤ watermark is duplicity → fail-closed + alarm (§7.2). A bare-epoch watermark is insufficient and must fail review.
- **R7 — DHT hints are add-only.** Mailbox/hint channels may only add lookup targets; they never prune, replace, or terminate the watermark-anchored update scan (A14).
- **R8 — Context strings are checked, not decorative.** Every at9p record type binds its context string in the COSE protected header; GATE 2 rejects a signature whose context does not match the record type being verified (A18).

## 9. Relationship to existing architecture

- **`cid.rs` (#705, dual-hash decision):** cid512 is a CIDv1 with a private-use `at9p-capsule` multicodec (xet LOCAL_CONVENTION pattern) over a **BLAKE3-512** multihash (`0x1e`, length 64). Same BLAKE3 family as the CAS default (BLAKE3-256, #812) — full width for the witness-path collision margin — and the same self-describing machinery as git SHA-256, XET xorbs. Adding it is relaxing `cid.rs`'s 32-byte assumption, not a new codec.
- **`cose_sign` (#550):** capsule and update signatures reuse the existing composite EdDSA + ML-DSA-65 envelope via the policy-gated path (§4.4/R4). This design adds no new crypto primitives — only pinned-Hybrid verification and per-record-type context strings.
- **`hyprstream-discovery`:** the DHT locator slots beside the existing `RecordResolver` as a capsule/record *source* under the Network resolver profile (#873) — it is not itself a `Resolver` impl (those resolve `(service, socket_kind) → endpoints`). PDS/discovery remain the trusted name path and the witness layer.
- **iroh pkarr (§6.4, #385/#361):** at9p supersedes classical pkarr trust for federated identity; pkarr may remain as an iroh-internal liveness mechanism, but no *authority* derives from it.
- **`gittorrent`:** its libp2p-Kademlia DHT is a different network; its domain-separated key-derivation pattern is reused, its stack is not.
- **9P uniform plane (arch law):** resolved capsules surface as files (a peer's capsule readable in the namespace); the DHT locator is a dial-time carrier, not an app-layer plane.
- **MAC (#547):** capsules may carry label hints (D1); resolver output feeds `SecurityContext` derivation only via verified key material, per the interface policy. Nothing here grants clearance.

## 10. Implementation sketch (if/when scheduled)

- **Phase 0 — spike hardening:** capsule DAG-CBOR schema + canonical round-trip tests; **codec placement decision** (move `hyprstream-pds/src/dag_cbor.rs` to a shared crate vs. dependency on hyprstream-pds); BLAKE3-512 (0x1e, len 64) support + `at9p-capsule` multicodec in `cid.rs`; grammar tests (this issue's EBNF as source of truth).
- **Phase 1 — locator:** mainline BEP5 client (announce/get_peers) behind the discovery record-source seam; k-derived rendezvous; provider scoring; GATE 1/2 verification path with pinned-Hybrid + context strings; R1/R2 enforced by API shape.
- **Phase 2 — rotation:** update-record chain validation with **duplicity detection** (persistent `(epoch, digest)` watermarks, R6), `at9p:update` rendezvous, add-only hint rule (R7).
- **Phase 3 — witnesses:** PDS countersignature record + client freshness/fork-evidence preference.
- **Not in scope:** PQ DHT. If/when one exists (or Saorsa-like PQ routing matures), it replaces the locator backend without touching the trust model — that substitution being free is the acceptance test for this design's layering.

## 11. Open questions

1. Genesis capsule hard size cap — what number? (Whole-fetch, no sharding, per §5.3; endpoints + delegations + witnesses growth pushes toward referencing large material by separate cid512.)
2. Threshold `next-key-commitments` format (strengthens A12) — v1 or deferred?
3. Should the BEP44 deterministic-mailbox channel (§6.1 opt 2) ship at all, or is BEP5-only cleaner for v1? (If it ships, R7 is mandatory.)
4. ~~Interaction with `did:web`/`did:key` peers during transition — `alsoKnownAs` aliasing and authoritative direction.~~ **RESOLVED** (see #896/D4 + #905 §2/§6): bidirectional links required to be believed; a verifier takes the **minimum** assurance of the leg it actually walked (at9p ⊃ plc — capsule→plc vouch is hybrid, plc→at9p vouch is classical). Public atproto infra allowlists only did:plc/did:web authorities, so public at:// records publish under the classical DID with `alsoKnownAs` to the at9p identity; our own PDS accepts did:at9p natively (#908).
5. Rendezvous context-string registry ownership (`at9p:provider:v1`, `at9p:update:v1`) and versioning discipline.
6. Duplicity alarm plumbing — where does the fail-closed signal surface (audit store per S7? operator notification path)?

## 12. Adversarial review record

This draft was adversarially reviewed before filing by two independent passes: a **crypto/protocol adversary** (instructed to refute, with specific probes on chain forks, signature contexts, BEP semantics, and fact-checks against FIPS 204 / BEP5 / BEP44 / multihash), and a **systems/integration adversary** (instructed to verify every repo-integration claim against actual code, not the doc).

**Findings incorporated (all):**
- *BLOCKING:* chain fork/duplicity was undetected — a bare epoch watermark lets a stolen-next-key fork sail past R6, and "accept max epoch" actively selects the attacker fork. Fixed: §7.2 duplicity detection, `(epoch, digest)` watermarks, `prev-record-digest` chaining, fail-closed + alarm semantics; A7/A12/A13 rewritten.
- Capsule verification now **pins Hybrid** independent of node `CryptoPolicy` (a Classical-configured node previously would have accepted Ed25519-only forgeries) — §4.4, A19.
- Genesis manifest signature was circular (signed by keys inside the object being fetched) and enabled unbounded bandwidth amplification — genesis is now whole-fetch + size-capped, sharding is post-genesis only — §5.3, A2.
- Local-clock trust added as a disclosed dependency — §4.3, A17, §8.1.
- Mailbox hints restricted to add-only (a pruning hint was a rollback assist, not mere availability) — §6.1, A14, R7.
- Per-record-type COSE context strings, checked at GATE 2 (cross-context signature reuse) — A18, R8.
- Empty/exhausted `next-key-commitments` specified (declared-immutable vs. frozen) — §7.1.
- `sign_composite` does **not** itself enforce fail-closed (verified against `cose_sign.rs` — it emits classical-only when `pq_sk` is `None`); R4 rewritten to target the policy-gated caller path.
- cid grammar aligned with the repo's actual CIDv1 convention (`cid.rs`: version ‖ multicodec ‖ multihash; BLAKE3-512 support (`0x1e`, len 64) must be *added*; `at9p-capsule` multicodec follows the xet private-use pattern) — §5.1. **Hash choice: BLAKE3-512** (Bernstein-lineage, CAS-aligned, 256-bit collision for the witness path) — SHA-2-512 rejected (NSA/Merkle–Damgård provenance + CAS-misaligned), SHA-3 rejected (slower, unaligned).
- The doc's "untapped DHT" pitch was wrong: gittorrent has a (libp2p, disjoint-network) DHT and **iroh pkarr already publishes classical-signed records to real mainline** (#385/#361) — §1, §6.4, A9 now positioned as an upgrade over the status quo.
- Minor: BEP5 token (~10 min) vs re-announce (~15–30 min) cadence corrected; rendezvous context strings unified with fixed-width components (§5.2).

**Fact-checks that survived:** ML-DSA-65 = 3,309 B / ML-DSA-44 = 2,420 B (FIPS 204); BEP44 `v` ≤ 1,000 B; multihash 0x1e = blake3 (len is a uvarint, so BLAKE3-512 = 0x1e/len 64, no new codec); BLAKE3-512 gives 256-bit collision + ~256-bit post-Grover preimage; KERI pre-rotation + duplicity semantics; PDS fitness as witness store (MST/signed commits/CAR proofs); `import_label`/`bind_time_label` restrict-only clamps exist as claimed.

**Residual risks are enumerated in §8.1** — the design's honesty criterion is that every "accepted residual" verdict in §8 appears there.

ewindisch:
Track C story #892 (resolves #879 Q3 — R1/R7, A5/A14) on the `hyprstream-p2p` mainline locator API surface. Depends on C2 (#890, merged).

## R1 — no name path compiles in

The locator already keyed its public API on `Cid512`; this makes the invariant **machine-checked** and documents the A5 rationale: a name→cid cache *inside the DHT client* re-opens threat A5 — it turns the DHT into a trusted name authority (whoever poisons the cache picks which capsule a name resolves to), the exact "toxic mold" failure #879 §3 excludes.

- `Cid512` gains no `FromStr` / `From<&str>` / `from_name` — the only way in is the full 64-byte digest.
- `r1_no_name_path_compiles_in` scans the module's production source (everything before `#[cfg(test)]`) and fails if any `pub fn` grows a `&str`/`String` parameter or the CID type gains a string constructor.

Names continue to resolve **above** this layer (a PQ-signed `NameRecord` from a trusted namespace authority — PDS/discovery/MoQ group), never inside it.

## R7 — add-only hints + mailbox decision

The BEP44 deterministic-mailbox decision (Q3) is recorded as:

```rust
pub const V1_MAILBOX_HINT_CHANNEL: MailboxHintDecision = MailboxHintDecision::Bep5Only;
```

**BEP5-only for v1** (per #879 §6.1, option 1 "chosen"); the deterministic mailbox is deferred. Whether or not it ships, R7's constraint is encoded **structurally**:

- `LookupHints` is an add-only accumulator (`add_rendezvous` / `add_peer`; no `remove`/`replace`/`clear`).
- `MainlineLocator::providers_with_hints` **always** performs the watermark-anchored BEP5 scan, then unions the hints in. A hint can enlarge the candidate set — never prune it or skip the scan. A14 (bogus epoch hint skips real rotations) is excluded by construction.

## Tests

R1 source-scan regression · content-addressed-only query proof · recorded v1 decision · add-only hint growth · scan-results-survive-hints · empty-hints-equal-plain-scan · misleading-hint-cannot-prune-scan.

## Acceptance (#892)

- [x] no name path compiles in (R1)
- [x] mailbox decision recorded (R7 / Q3)
- [x] if shipped, hint channel is provably add-only (R7) — deferred for v1, but the add-only mechanism (`LookupHints` + `providers_with_hints`) lands now so the rule binds the moment any channel feeds it

Build: `cargo build -p hyprstream-p2p` ✅ · `cargo test -p hyprstream-p2p locator` — 15/15 ✅ · `cargo clippy -p hyprstream-p2p --all-targets` clean ✅


ewindisch:
Pushed `94d515bba` addressing both review findings.

**BLOCKER — R1 now structural, not grep-based.** The line-by-line `include_str!` scan was replaced by `tests/locator_r1_invariants.rs`, which parses `src/locator.rs` with `syn` and walks **every** public item signature (free fns, inherent/trait methods, multi-line + generic forms). The detector recurses through references, boxes, slices, tuples, parens, generic args, and impl-trait/trait-object bounds, so all enumerated bypass forms resolve true: `&str`, `String`, `&String`, `Cow<'_, str>`, `Box<str>`, `Vec<String>`, `impl AsRef<str>`, `Into<String>`, etc. It also fails if `Cid512` (or any gated public type) gains a string-conversion trait impl (`FromStr`, `AsRef`, `Deref`, `Borrow`, or `From`/`TryFrom<str>`). A grep over single-line sigs is bypassable; an AST walk is not. A detector self-test pins the recognizer against literal type fragments, and I verified a negative case: temporarily adding a multi-line `pub fn resolve_name(cid, name: &str)` made the test fail with the expected diagnostic, then reverted.

**MEDIUM — hints are now bounded.** `LookupHints` caps rendezvous (`MAX_HINTED_RENDEZVOUS = 8`) and peers (`MAX_HINTED_PEERS = 64`), dedupes by value, and `add_rendezvous`/`add_peer` return `bool` so callers detect truncation. `providers_with_hints` dedups the final candidate list and caps it at `MAX_LOCATOR_RESULTS = 256`, so untrusted hint input (PDS record, relay bridge, mailbox) cannot amplify into unbounded DHT/network work. R7 add-only is unchanged: the canonical watermark scan always runs (existing `r7_*` tests still pass, plus new `r7_hints_are_bounded_and_deduped` and `r7_locator_dedupes_and_caps_results`).

`cargo build -p hyprstream-p2p` ✅ · `cargo test -p hyprstream-p2p` all green (18 locator + 2 invariant tests) ✅ · `cargo clippy -p hyprstream-p2p --all-targets` clean ✅

ewindisch:
Pushed `937a10aa2` — hardened the R1 detector against the re-review's holes. Only the detector changed; the DoS bounds and the legit `Cid512`/`RendezvousKey` path are untouched.

**1. generics + where-clauses (PRIMARY).** `check_sig` now reads `sig.generics` (type-param bounds) and the where-clause, not just `sig.inputs`. So `pub fn resolve<S: AsRef<str>>(name: S)` and `fn f(s: S) where S: AsRef<str>` are now flagged via the same bound predicate as `impl AsRef<str>` — the `<T: Trait>` desugaring can no longer smuggle a name type past the concrete-param check (`generics_admit_stringish`).

**2. type aliases.** `Item::Type` is now handled: `type X = String;` (any visibility) is denied, since a `String` alias used as a `pub fn` param type wouldn't otherwise look string-like.

**3. macros and `pub use`.** `Item::Macro` and public `Item::Use` are forbidden outright in this module (fail by name, requiring explicit human approval) rather than silently unresolvable — a macro could generate, or a re-export surface, a name-taking `pub fn` the AST walk can't see.

**4. generics self-test.** Added `check_sig_flags_generic_and_where_clause_string_bounds`, which drives `check_sig` directly against `<S: AsRef<str>>`, the where-clause form, `Into<String>`, `FromStr`, and a multi-line generic — the generics path was previously UNtested. Also added a type-alias walk self-test and put `Borrow<[u8]>` in the negative type set so byte-view bounds stay clean.

`bound_trait_is_stringish` now also flags `FromStr` as a bare bound; `AsRef`/`Into`/`TryInto`/`Borrow` stay arg-gated so `AsRef<[u8]>`/`Borrow<[u8]>` byte views are not over-flagged.

Verified end-to-end: temporarily injecting `pub fn resolve<S: AsRef<str>>(name: S)` + `pub type CapsuleName = String` into the real module made the test fail with the expected diagnostics, then reverted. Docstrings corrected to stop over-claiming (the earlier doc falsely advertised type-alias coverage) and now state the honest coverage matrix + known limitation.

`cargo nextest run -p hyprstream-p2p` → 78/78 ✅ · `cargo clippy -p hyprstream-p2p --all-targets` clean ✅

ewindisch:
Pushed `6d34e3499`. Per the meta-point, I stopped adding forbidden-form cases and flipped to a **closed-world positive allowlist** — the substantive structural fix feasible in this PR.

**The structural flip.** Instead of enumerating forbidden string forms (whack-a-mole), the detector now enumerates the *approved* public types and rejects **everything else by omission**: `Cid512`, `RendezvousKey`, `PeerContact`, `LookupHints`, `SocketAddrV4`, primitives, and byte views (recursing through `&`/`Option`/`Vec`/`Box`/slices/arrays). All three remaining bypasses close via one mechanism:
1. **struct-wrapped string** `pub fn locate(q: NameQuery)` → `NameQuery` not on the list; public struct/enum fields are also type-checked, so `pub struct NameQuery { pub name: String }` is caught at the definition.
2. **`impl ToString` / `impl Display`** (and `<S: ToString>`) → generics/impl-trait params fall through to not-approved; no per-trait enumeration needed.
3. **private external alias** `pub fn resolve(name: CapsuleName)` → `CapsuleName` not on the list regardless of what it aliases.

Adding a public parameter/field type is now a deliberate, reviewable allowlist edit (`APPROVED_LEAF_TYPES`/`APPROVED_WRAPPERS`) — the "un-vetted public param structs are forbidden" gate the review asked for. Trait-impl method params are no longer type-checked (their signatures are trait-mandated, e.g. `Debug::fmt`'s `Formatter`); that risk is captured by `forbidden_string_trait` at the impl level.

Verified end-to-end: injecting `locate(NameQuery)` + `render(impl std::fmt::Display)` + `resolve(CapsuleName)` into the real module failed the test with all three (plus the struct field and the alias def), then reverted. Self-tests pin the allowlist and drive `check_sig` against every round-1/2/3 bypass form.

`cargo nextest run -p hyprstream-p2p` → 79/79 ✅ · clippy clean ✅.

**On the compile-error ideal.** A true `compile_error!` enforcement (build-script / dylint / proc-macro) is preferred but too large for this PR. Filed **#1005** to track it; the allowlist is the closed-world interim guard and #1005 retires or supplements it.

PR #1021, opened 2026-07-11
ci: move Rust compile jobs to arm64 Graviton container

ewindisch:
Moves `clippy`, `wasm`, and the merge-gate `build` job off GitHub-hosted `ubuntu-latest` onto the self-hosted Graviton (arm64) runners, running inside `ghcr.io/hyprstream/rust-builder-arm64` (rust + capnproto + clang + libtorch). Builds on the merged arm64 builder (#1019/#1020).

## Why
Those jobs pull the sccache S3 bucket (us-east-1). From off-AWS runners that is billed as **S3 DataTransfer-Out — ~560 GB / ~$41 last window, the entire recent S3 bill**. Same-region self-hosted Graviton transfers from S3 for free; the metal-aws side also gets an S3 gateway VPC endpoint (cyberdione/ingest MR!1) to keep that traffic on the AWS backbone. Toolchain lives in the image, so the golden AMI stays toolchain-free.

## Changes per job
- `runs-on:` → `[self-hosted, linux, arm64, graviton, hyprstream-merge-gate]` + `container: ghcr.io/hyprstream/rust-builder-arm64:latest`
- Drop host `apt-get`/`dtolnay/rust-toolchain` steps (image provides them); `clippy`/`wasm` add only their component/target via the image's rustup.
- Replace `--features download-libtorch` (x86-only zip) with `LIBTORCH=/opt/libtorch` from the image (workflow-level env).
- `cargo-deny` stays on `ubuntu-latest` (metadata only — no compile, no S3).
- `container:` jobs get node injected by the self-hosted actions/runner, so `checkout`/`configure-aws-credentials`/`sccache-action` run normally.

## ⚠️ DRAFT — one validation gap before flipping the required gate
The earlier arm64 validation (#1020) covered `--no-default-features --features otel,gittorrent,xet` and `cargo build` only. This PR builds **default features** (systemd / kata-vm / oci-image) and runs the **full test suite** on arm64 — neither is validated yet. I'm running that on the `ewindisch/runner-smoke-test` branch (default-feature build + nextest in the container on Graviton). **Do not mark `build` required-green until that validation passes** — otherwise an arm64-specific default-feature or test failure would red the merge queue for everyone.

Ratified direction: merge-gate build is arm64-only on Graviton (operator decision).

PR #1022, opened 2026-07-11
ci: migrate clippy+wasm to Graviton (build stays x86 pending arm64 test validation)

ewindisch:
Supersedes closed #1021. Graviton runner validation established that job-level `container:` expects a Docker API socket unavailable on the rootless-Podman AMI; these jobs use the repository’s established host-step `podman run` pattern instead.

## Scope

- Inherits the focused PDS compile fix from #1041 as ancestry only; the #1022-specific diff remains `.github/workflows/rust.yml`.
- Moves the PR-triggered `clippy` and `wasm` checks to the trusted self-hosted ARM64 Graviton runner pool using `ghcr.io/hyprstream/rust-builder-arm64:latest`.
- Keeps the merge-queue `build` gate on `ubuntu-latest`: the default-feature arm64 nextest suite is not yet validated green.
- Keeps `cargo-deny` on GitHub-hosted Ubuntu.
- Uses bounded 30-minute timeouts and disables checkout credential persistence before mounting the workspace into Podman.

`download-libtorch` deliberately remains outside the ARM64 clippy job because its upstream archive is x86-only; the builder image supplies ARM64 libtorch.

Follow-up: move the merge gate only after the runner-smoke/default-feature suite is green.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -25,14 +25,15 @@ env:
   CARGO_INCREMENTAL: 0
   BUILDER_IMAGE: ghcr.io/hyprstream/rust-builder-arm64:latest
 
-# The Rust compile jobs run on the self-hosted Graviton (arm64) runners and drive
-# the build through `podman run` against the prebuilt rust-builder image (rust +
-# capnproto + clang + libtorch). We DELIBERATELY do NOT use a job-level
-# `container:` — the runner's native container support wants a Docker-API socket
-# at /var/run/docker.sock, which the rootless-podman golden AMI does not expose
-# (a `container:` job fails with `statfs /var/run/docker.sock: permission
-# denied`). Driving `podman run` from host steps sidesteps that entirely and is
-# the same pattern proven in graviton-build-validate.yml.
+# The fast lint jobs (clippy, wasm) run on the self-hosted Graviton (arm64)
+# runners and drive the build through `podman run` against the prebuilt
+# rust-builder image (rust + capnproto + clang + libtorch). We DELIBERATELY do
+# NOT use a job-level `container:` — the runner's native container support wants
+# a Docker-API socket at /var/run/docker.sock, which the rootless-podman golden
+# AMI does not expose (a `container:` job fails with `statfs
+# /var/run/docker.sock: permission denied`). Driving `podman run` from host
+# steps sidesteps that entirely and is the same pattern proven in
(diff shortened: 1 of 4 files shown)


ewindisch:
Prepared for review.

- Rebased onto current `main` and narrowed this PR to `.github/workflows/rust.yml`; the unrelated PDS serialization work is preserved separately in #1041.
- Kept the repository-established rootless-Podman/Graviton pattern and the validated runner labels. Added 30-minute bounds for the fast shared-runner jobs (last green run: Clippy ~6m, WASM ~2m) and `persist-credentials: false` before mounting the checkout into Podman.
- Review disposition: timeout and credential findings are addressed (threads auto-resolved). I did not add `--all-features`: it would lint `download-libtorch`, whose upstream archive is x86-only and is deliberately excluded from the ARM64 image-backed path. I also did not move Rust components into the builder image in this PR: doing so would require publishing a new image before these checks can rely on it.

Validation: local YAML parse and diff check; focused PDS split validated with rustfmt, clippy, and 203 tests. Hosted `cargo-deny` and CodeQL Actions/Go/JS analyses are green.

Remaining external blocker: the two repository Graviton runners (`hyprstream-graviton-i-02e98b3a357874d78`, `hyprstream-graviton-i-06afcfb6ae48ae4e4`) are both offline, so Clippy and WASM are queued without a runner. Claude review also failed before analysis with the action reporting `is_error:true` and zero findings: https://github.com/hyprstream/hyprstream/actions/runs/29354665068/job/87159182983. No merge performed.

ewindisch:
Final rebase update: #1022 is now based on the exact #1041 head and is pushed at `74ac9cb5d9d5dbf96676bb4e993d7ac1f331ef20`; the #1022-specific delta after #1041 is only `.github/workflows/rust.yml`.

Also addressed the final CodeRabbit reproducibility finding by pinning the verified ARM64 Linux builder manifest to `sha256:92d8b445f8020e38a4ab08a5868b98819630e3bbf3425f3e281b0e66f1e4b9d5`. Local YAML/diff validation, PDS clippy, 203 PDS tests, and wasm32 check pass. CodeRabbit and cargo-deny are green; CodeQL is running; Graviton Clippy/WASM are queued.

Claude review again failed before analysis (same action `is_error:true`, no findings): https://github.com/hyprstream/hyprstream/actions/runs/29358346254/job/87171630662. No merge performed.

PR #1033, opened 2026-07-14
fix(discovery): content-bind CAR blocks to CIDs at ingest + commit.did check (security)

ewindisch:
## Summary
Follow-up hardening for #980 (merged). `PlacementIndex::is_member` verified the commit **signature** but never verified that each CAR block's bytes actually hash to its declared CID — since `Commit::verify` only signs the commit's own bytes (which merely *name* the MST root CID), an attacker could replay a genuine signed commit with **substituted MST-node/record blocks** and forge group membership.

- Mirrors `hyprstream_pds::mst::Proof::verify` / `car.rs`: when a verifying key is present (verified ingest), recompute `Cid::from_dag_cbor(bytes)` for every consumed block (commit, each MST node, each record) and fail closed on any CID mismatch.
- `ensure!(commit.did == did)` — a validly-signed repo for DID A must not be replayable under DID B.
- Bound `walk_mst` recursion depth (`MST_MAX_DEPTH=1024`) to defend against an attacker-shaped CAR stack-overflow DoS.
- Corrects the trust-statement rustdoc (module, `ingest_did`, `is_member`) — it previously overclaimed signature-only verification when both commit-signature and per-block CID content-binding are enforced.
- Adds a test: a repo CAR with a genuine signed commit but a substituted record block (declared CID, replaced bytes) is rejected and produces no membership.

## Test plan
- [x] New test: substituted-record-block CAR is rejected, produces no membership
- [x] `cargo clippy` clean
- [x] Workspace release build green

### DIFF
--- a/crates/hyprstream-discovery/src/placement_index.rs
+++ b/crates/hyprstream-discovery/src/placement_index.rs
@@ -15,12 +15,21 @@
 //! handed it over. Ingest is **verified by construction**: when the resolver
 //! can provide the DID's `#atproto` verifying key
 //! ([`RecordResolver::resolve_verifying_key`]), [`PlacementIndex::ingest_did`]
-//! verifies the repo CAR's commit signature against it *before* any record
-//! enters the index. A record whose commit signature fails verification (or
-//! whose CAR is malformed) is refused — it cannot reach the derived indices, so
-//! it cannot produce membership or any other fact derived from
-//! [`PlacementIndex::is_member`]. This makes signature verification a property
-//! of the index itself, not a caller obligation.
+//! verifies the repo CAR's commit signature against it **and content-binds
+//! every consumed block to its declared CID** — the commit block, each MST
+//! node, and each record block have their CID recomputed from their bytes and
+//! checked against the CID the CAR / parent declared for them (mirroring
+//! `hyprstream_pds::mst::Proof::verify`) — *before* any record enters the
+//! index. The signature alone is not sufficient: `Commit::verify` only signs
+//! the commit's own bytes (which name the MST root CID), so without per-block
+//! CID binding an attacker could replay a genuine signed commit with
+//! substituted MST-node/record blocks and forge membership; the content-bind
+//! check closes that. A repo whose commit signature fails verification, whose
+//! `commit.did` is not the requested DID, or whose any block's recomputed CID
+//! ≠ its declared CID is refused — it cannot reach the derived indices, so it
+//! cannot produce membership or any other fact derived from
+//! [`PlacementIndex::is_member`]. This makes verification a property of the
+//! index itself, not a caller obligation.
 //!
 //! When the resolver returns `Ok(None)` (it cannot provide a key for this DID
 //! — e.g. a foreign DID whose DID-document `#atproto` method this crate can't
@@ -54,7 +63,7 @@
 
 use std::collections::{HashMap, HashSet};
 
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, ensure, Result};
 use p256::ecdsa::VerifyingKey;
 use parking_lot::RwLock;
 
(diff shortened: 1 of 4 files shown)


ewindisch:
Security disposition after rebase onto current `origin/main` (`aad84f325`):

- Preserved upstream's traversal-wide visited-CID set and total node-visit budget, and combined them with this PR's depth limit and per-block content binding. Shared/cyclic MST nodes now reject the whole ingest.
- CID integrity is mandatory for every CAR block even when the resolver has no verifying key; signature verification remains conditional on key availability. Unsupported/non-canonical CID/hash claims fail the recomputed-CID comparison.
- Ingest is fail-closed and replacement-transactional: resolver, key-resolution, CAR decode, CID, signature, DID-binding, MST, missing-reference, and placement-record validation failures remove the DID's prior snapshot and rebuild derived state.
- Added strict rejection for duplicate block CIDs, non-singleton CAR roots, missing MST-referenced record blocks, malformed placement records, and malformed MST keys.
- Added focused regressions for substituted content, DID-confused replay, stale state with the no-key fallback, duplicate/root/missing-block structure, and traversal-wide shared-node reuse.

Validation: targeted placement-index suite passes (11/11); targeted `clippy -p hyprstream-discovery --all-targets -- -D warnings` passes. Current `origin/main` independently has a compile defect in `hyprstream-pds/src/at9p_duplicity.rs` where `to_dag_cbor()` now returns `Result`; I used an uncommitted temporary compatibility adjustment only to validate this PR, then removed it. Fresh CI is running. No merge performed; ready for human security review once checks settle.

ewindisch:
Follow-up security review disposition (`5a9f35531`): both new CodeRabbit findings were valid and are fixed.

- Group-item authority binding: a repo may only publish claims for an `at://<commit.did>/ai.hyprstream.placement.group/<rkey>` URI. Third-party owner impersonation and wrong-collection URIs reject the whole snapshot; regression added.
- Atomic publication: raw snapshots and all derived maps now live under one `RwLock<PlacementState>`. Insert/clear plus rebuild occur under the same write guard, and readers take one generation-consistent read guard, preventing an older concurrent rebuild from restoring stale membership after a validation failure.

Focused placement tests: 12/12 pass. Targeted discovery clippy with `-D warnings`: pass (using the same temporary, uncommitted origin/main compatibility adjustment noted above, then removed). Fresh CI is running.

ewindisch:
Scope confirmation: `git diff --name-only origin/main...HEAD` contains only `crates/hyprstream-discovery/src/placement_index.rs`. The `hyprstream-pds/src/at9p_duplicity.rs` E0308 regression is not committed here and is tracked/fixed separately in #1041; #1033 validation used that fix only as a temporary local compatibility patch, removed before each commit.

ewindisch:
CI rebase update: #1033 is now based on the exact final pushed #1041 head `63faddfb2926746607ab8d9580c6c748dbbc63f4` while the PR target remains `main`.

- #1041 commits remain intact ancestors (`7472de35c`, `63faddfb2`); no PDS change was copied, squashed, amended, or independently edited in #1033.
- Rebased #1033 head: `f7b3334ea`.
- `rustfmt --check` for the discovery change: pass.
- `cargo test -p hyprstream-discovery --all-targets`: 47/47 pass.
- `cargo clippy -p hyprstream-discovery --all-targets -- -D warnings`: pass.
- Practical workspace clippy is locally blocked before project diagnostics by the host Perl installation missing `FindBin.pm` while vendored OpenSSL configures; hosted CI uses the project build image and is running now.

No merge performed.

ewindisch:
Latest CodeRabbit review is green. Its sole nitpick concerns a possible shared DAG-CBOR trait in `hyprstream-pds/src/at9p.rs`; that code is inherited unchanged from the exact #1041 ancestor, is explicitly low-value/non-security, and is outside #1033 discovery scope. No change is justified here; any future DRY refactor belongs to a separate PDS PR after #1041.

PR #876, opened 2026-07-07
feat(rpc): install explicit resolver profiles (#873)

ewindisch:
Closes #873.

Stacked on #875 / #874.

## Summary
- Adds explicit resolver profile names and concrete resolver types:
  - `LocalInprocResolver`
  - `IpcResolver`
  - `NetworkDiscoveryResolver`
- `registry::init()` now installs a named local resolver profile instead of a generic registry resolver.
- Local profile resolvers first use explicitly registered endpoints, then synthesize only their own local endpoint shape.
- `NetworkDiscoveryResolver` wraps a DiscoveryService/PDS-backed resolver and rejects local-only transports (`inproc`, `ipc`, `systemd-fd`) so network mode cannot silently fall back to local reach.
- Exports the profile types for startup wiring.

## Validation
- `CARGO_TARGET_DIR=/tmp/hyprstream-876-target CARGO_INCREMENTAL=0 cargo test -p hyprstream-rpc registry::tests --lib`
- `CARGO_TARGET_DIR=/tmp/hyprstream-876-target CARGO_INCREMENTAL=0 cargo test -p hyprstream-rpc resolver::tests --lib`
- `CARGO_TARGET_DIR=/tmp/hyprstream-876-target CARGO_INCREMENTAL=0 cargo clippy -p hyprstream-rpc -p hyprstream-rpc-derive -p hyprstream-discovery --all-targets -- -D warnings`

Issue #873, opened 2026-07-07
Service-to-service network resolver profile over Discovery/PDS + iroh


ewindisch:
## Context

The `#846` Helm discussion exposed a non-Kubernetes-specific architecture gap: hyprstream services still have implicit local endpoint fallback (`inproc` or same-host UDS/`ipc`) while the intended multi-host topology is identity-bound network reach over Discovery/PDS + iroh.

This should not live in the Helm chart. Kubernetes is only one consumer. The same contract is needed for any multi-host deployment: bare metal, systemd units across hosts, PDS/cell deployments, and later operator-managed environments.

## Goal

Make service-to-service resolution an explicit runtime profile, where generated `for_service(...)` clients ask the installed resolver for reach and never infer locality themselves.

Invariant:

```text
for_service() never decides locality.
The installed Resolver decides locality.
dial() only dials the resolved TransportConfig.
```

## Proposed resolver profiles

- `local-inproc`: single process / local tests only.
- `same-host-uds`: same-host systemd/dev IPC only.
- `network-discovery`: production multi-host profile; resolve service dependencies through DiscoveryService/PDS records and dial iroh-first.

## Scope

- Define explicit resolver-profile config and startup wiring.
- Install the correct global `Resolver` for each service startup mode.
- Make `DiscoveryClient` or a PDS-backed resolver usable as the production network resolver.
- Bootstrap DiscoveryService explicitly, then resolve all other services through Discovery/PDS.
- Advertise service reach as typed network metadata:
  - service name
  - service DID / verifying key
  - iroh node id
  - accepted ALPNs, e.g. `hyprstream-rpc/1`, `moql`
  - relay/direct reach
  - optional QUIC/WebTransport fallback
- Ensure `inproc` and UDS/`ipc` are never advertised as routable service reach.
- Preserve `EndpointRegistry` as the local registration table; do not replace the existing resolver architecture.

## Acceptance

- Two separate hosts can run different hyprstream services and satisfy a dependency chain such as `oai -> model -> registry/policy` over Discovery/PDS + iroh, with no `inproc` or UDS fallback.
- Single-process local mode still works through `local-inproc`.
- Same-host IPC mode still works through `same-host-uds`.
- Network mode fails closed when a dependency has no fresh network reach.
- `dial()` remains transport-execution code, not resolver policy code.

## Related

- `#282` iroh substrate + pkarr dial-by-node-id.
[...]

PR #1037, opened 2026-07-14
fix(cli): initialize RPC verification and recover TUI key

### LINKED ISSUE #1018: New nix build install compiles but unable to do anything
System: amd64, nixos, kernel 6.18.36, nvidia gpu, cuda 13.2, nvidia driver 595.71.05

Installed with `nix build github:hyprstream/hyprstream#hyprstream-cuda130` on a fresh nix system with nvidia nonfree drivers and cudatoolkit, git, and git-lfs

Post install ran:

```
hyprstream service install
hyprstream wizard -y —start
```

`hyprstream service status` shows services running. 

All quick commands give this warning 

`2026-07-11T00:52:03.324379Z  WARN hyprstream_rpc::envelope: crates/hyprstream-rp
c/src/envelope.rs:1064: response verify config not installed; defaulting to fail
-closed Hybrid policy. Production code MUST call install_response_verify_config(
) at startup. `

Downloading a model using quick times out (I assume due to the warning above?):

```
$ ./result/bin/hyprstream quick clone https://huggingface.co
/Qwen/Qwen3-0.6B --name q
📦 Cloning model from: https://huggingface.co/Qwen/Qwen3-0.6B
   Depth: 1 commits
2026-07-11T01:01:08.087208Z  WARN hyprstream_rpc::envelope: crates/hyprstream-rp
c/src/envelope.rs:1064: response verify config not installed; defaulting to fail
-closed Hybrid policy. Production code MUST call install_response_verify_config(
) at startup.
   Progress:    Fetching objects....
   Indexing.....................................................................
..................................................................
   Downloading LFS files.....Error: Failed to clone model: uds RPC timeout after
 30s

```

Starting a TUI causes a panic and crash (the tui the wizard uses works just fine though):
```
$ ./result/bin/hyprstream tui new
2026-07-11T01:06:39.661631Z  WARN hyprstream_rpc::envelope: crates/hyprstream-rp
c/src/envelope.rs:1064: response verify config not installed; defaulting to fail
-closed Hybrid policy. Production code MUST call install_response_verify_config(
) at startup.
🚨 PANIC occurred:
   Location: crates/hyprstream/src/cli/tui_handlers.rs:54
   Message: Failed to resolve TUI service key: service key 'tui' not registered
This was likely caused by a threading issue or memory corruption.
Please check for unsafe RefCell usage or race conditions.
```

Did I miss a part of the setup somewhere? Thanks!

Would love to try this project.

ewindisch:
## Summary

- install process-global request/response envelope verification before dispatching ordinary CLI commands
- remove the ambiguous global-trust-store fallback for the TUI key
- propagate TUI client construction failures as normal CLI errors instead of panicking

This intentionally fixes the response-verification startup half of #1018 while leaving functional TUI recovery fail-closed until identity and transport can be resolved atomically.

Fixes #1018 partially; the resolver migration is tracked under #873.

## Investigation and scope

`install_response_verify_config()` was introduced by `feef937a9` for response-side hybrid enforcement, but its startup call remained confined to service and standalone-worker branches. Ordinary `quick` and `tui` clients could therefore reach response verification without configuration.

Policy-based service-key resolution was introduced by `6de885e1a`. The original revision attempted to recover through `global_trust_store().resolve_one("tui")`, but review confirmed that this can select an arbitrary key within the service scope and still leaves transport selection independent from identity selection. Loading or selecting a bootstrap key more specifically would preserve that split-brain design rather than fix it.

The safe stopping point is therefore:

1. install envelope verification early;
2. return a clear error when PolicyService cannot resolve the TUI identity;
3. implement functional recovery through an identity-bound resolver result under #873, after the resolver-profile work from #876 is restored to main.

PR #1036/#550 is separately hardening HyKEM request envelopes and production network transports; this PR does not duplicate that work.

## Validation

- `git diff --check` passes
- targeted host checking with system PyTorch/OpenSSL reached the known unrelated `hyprstream-pds/src/at9p_duplicity.rs` type errors (`to_dag_cbor()` now returns `Result<Vec<u8>>`)
- the pre-commit workspace Clippy hook failed on the same existing PDS errors before it could complete; the follow-up commit was made with hook bypass only after recording that baseline failure
- prior CI passed CodeQL, cargo-deny, and the WASM browser build; the Claude review action failed internally with `is_error:true` and produced no findings

### DIFF
--- a/crates/hyprstream-pds/src/at9p.rs
+++ b/crates/hyprstream-pds/src/at9p.rs
@@ -708,7 +708,7 @@ impl UpdateRecord {
 
     pub fn to_dag_cbor(&self) -> Result<Vec<u8>> {
         self.validate()?;
-        Ok(self.to_value().encode())
+        Ok(self.encode_value())
     }
 
     /// Serialize without re-validating. For crate-internal digest/state
--- a/crates/hyprstream-pds/src/at9p.rs
+++ b/crates/hyprstream-pds/src/at9p.rs
@@ -637,7 +637,7 @@ impl Capsule {
 
     pub fn to_dag_cbor(&self) -> Result<Vec<u8>> {
         self.validate()?;
-        Ok(self.to_value().encode())
+        Ok(self.encode_value())
     }
 
     pub fn from_dag_cbor(bytes: &[u8]) -> Result<Self> {
@@ -646,6 +646,14 @@ impl Capsule {
 
     pub fn cid512(&self) -> Result<String> {
         at9p_capsule_cid512(&self.to_dag_cbor()?)
+    }
+
+    /// Serialize without re-validating. For crate-internal digest/state
+    /// projection over capsules already known to be valid — they reached here
+    /// via [`Self::from_dag_cbor`] (validated) or the signing path. External
+    /// callers must use [`Self::to_dag_cbor`].
+    pub(crate) fn encode_value(&self) -> Vec<u8> {
+        self.to_value().encode()
     }
 
     fn to_value(&self) -> DagCbor {
--- a/crates/hyprstream-pds/src/at9p_duplicity.rs
+++ b/crates/hyprstream-pds/src/at9p_duplicity.rs
@@ -217,8 +217,8 @@ impl PredecessorRecord<'_> {
     /// the watermark pins is verified to describe *these* bytes.
     pub fn record_digest(&self) -> [u8; H512_LEN] {
         match self {
-            PredecessorRecord::Genesis(cap) => h512(&cap.to_dag_cbor()),
-            PredecessorRecord::Update(rec) => h512(&rec.to_dag_cbor()),
+            PredecessorRecord::Genesis(cap) => h512(&cap.encode_value()),
+            PredecessorRecord::Update(rec) => h512(&rec.encode_value()),
         }
     }
 
@@ -2263,7 +2263,6 @@ mod tests {
             Some(DuplicityKind::HigherEpochFork)
         );
     }
-
 
     // Small accessors so the guard's sink can be inspected in-test.
     impl DuplicityGuard<InMemoryWatermarkStore, RecordingSink> {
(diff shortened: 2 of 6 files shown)


ewindisch:
I validated both CodeRabbit findings against the broader resolver architecture. The observation that the fallback was not reliably provisioned is correct, and `resolve_one("tui")` is indeed ambiguous when multiple attestations exist. I did **not** add a bootstrap-specific lookup because that would still select identity separately from `EndpointRegistry` transport selection and preserve the underlying race/ambiguity.

Commit `a26bd3e` instead removes the trust-store fallback entirely and changes TUI client construction to return normal contextual errors rather than panic. Early envelope verification remains in place. Functional TUI recovery is deliberately deferred to the #873 identity-bound resolver seam, where the selected transport, verifying key, service identity, and selection evidence can be returned atomically.

Validation is currently bounded by the unrelated main-branch PDS compile regression at `at9p_duplicity.rs:220-221`; both targeted checking and the pre-commit workspace Clippy hook reach the same `to_dag_cbor()` `Result<Vec<u8>>` mismatch.

ewindisch:
Addressed the valid cleanup-order finding in `385db9137`.

- `forward_process_output` now resolves both auxiliary clients (resize and conditional stdin polling) before entering raw mode, creating the cancellation pipe/thread, or spawning tasks.
- `handle_shell_tui` now resolves the SIGWINCH resize client immediately after the primary client, before terminal/raw-mode, VFS proxy, and background-task setup.

Consequently, an identity-resolution failure returns a normal contextual CLI error while the terminal and task set are still untouched. `git diff --check` passes and rustfmt successfully parses both changed files. Full workspace checking remains blocked before the affected application crate by the existing `hyprstream-pds/src/at9p_duplicity.rs:220-221` `E0308` regression already documented on this PR.

ewindisch:
CI repair update: rebased cleanly onto the exact final #1041 repair head `63faddfb2926746607ab8d9580c6c748dbbc63f4` while retaining `main` as the PR target. New head: `90344cd6a95c9bb1ff8f1abfbec88a9e84d15db6`.

Scope remains the three CLI files only; no independent PDS edits were made. `git diff --check` passes. Targeted application Clippy passed with the installed PyTorch/libtorch environment (`LIBTORCH_USE_PYTORCH=1`, version bypass for the host package). The PDS crate compiled cleanly, confirming the shared E0308 blocker is removed. Fresh hosted checks are now queued/running.

Issue #1011, opened 2026-07-10
ci: validate Linux arm64 CPU libtorch + hyprstream tch-rs fork


ewindisch:
Validate Linux arm64 CPU libtorch + hyprstream tch-rs fork for local ARM64/Graviton CI

### Problem

ARM64 is a preferred canary and required release target, but the `download-libtorch` path and the hyprstream `tch-rs` fork must be validated on Linux arm64 before relying on either the local LX2K or cloud Graviton for release/merge-gate coverage.

### Proposal

Run a focused Linux arm64 validation on the first Graviton host; repeat later on local LX2K:

- install normal build deps;
- use the pinned Rust toolchain;
- run the `download-libtorch` path;
- build the workspace or a reduced crate set that exercises `tch`;
- run a smoke test that proves CPU libtorch loads and executes;
- confirm CUDA/ROCm marker logic does not assume x86 for CPU CI.

### Acceptance Criteria

- `cargo build --features download-libtorch` succeeds on Linux arm64, or failures are reduced to explicit tracked patches.
- A CPU libtorch smoke test passes on Linux arm64.
- Cloud Graviton can serve as the Phase-1 ARM blocking merge-gate.
- The local LX2K can later run as the arm64 canary/release-builder candidate.
- Any arch-specific libtorch URL or env var differences are documented in workflow comments.

### Follow-Up

If validation passes, enable the Graviton merge-gate first. Local ARM64 is Phase 2 and should start non-blocking; promote it only after reliability/speed proof.

---

_Epic: #1009._

PR #1073, opened 2026-07-17
ci: automate BUILDER_IMAGE digest bumps via self-validating PR (#1011)

ewindisch:
## What

Automates the digest re-pin so determinism (digest-pinning) costs no manual toil. After `build-image.yml` publishes a new `rust-builder-arm64` image, it captures the pushed digest (`podman push --digestfile`), rewrites the `BUILDER_IMAGE` pin in `rust.yml` if it changed, and opens a **self-validating** digest-bump PR.

## Why digest (not `:latest`)

Digest pins keep CI reproducible + supply-chain-safe (exact reviewed bytes; a bad/compromised push can't silently become active). The bump PR runs the full arm64 merge gate against the new image, so **a broken image never merges** and the previous digest stays live; rollback is a one-line revert.

## Credential: GitHub App token (not a PAT, not OIDC)

The PR is authored with a **short-lived GitHub App installation token** minted per-run via `actions/create-github-app-token` — scoped to `contents` + `pull-requests` write on this repo, expires ~hourly, not tied to a human. This is required because PRs opened by the default `GITHUB_TOKEN` don't trigger workflows (the merge gate would never run). Actions **OIDC can't be used** for this — it federates to external providers (AWS/Azure/GCP), not GitHub's own API.

## Setup (one-time)
Create a GitHub App with `contents`+`pull-requests` write, install it on this repo, and set `vars.BUILDER_IMAGE_BUMP_APP_ID` + `secrets.BUILDER_IMAGE_BUMP_APP_PRIVATE_KEY`. Until then the bump step **skips with a warning** (image still publishes) — safe to merge now.

## Follow-ups
- metal AMI `prepull_image` → rolling `:latest` (cache-warm only, non-authoritative), so it never needs re-baking on a digest bump.
- Adopt **immutable OIDC subject claims** on the existing AWS trust policies (sccache-S3 role + metal deploy roles) — spoof-resistance hardening; auto-enforced for new repos/renames since 2026-07-15.

### DIFF
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -74,11 +74,16 @@ jobs:
       # PR self-validates: the merge gate runs against the new image, so a broken
       # image simply never merges and the old digest stays live.
       #
-      # The PR MUST be authored by a non-GITHUB_TOKEN identity (a fine-grained PAT
-      # or GitHub App installation token, secret BUILDER_IMAGE_BUMP_TOKEN) — PRs
-      # opened with the default GITHUB_TOKEN do not trigger workflows, so the merge
-      # gate would never run and the PR could not validate. If the secret is unset
-      # the step is skipped with a warning (image still published; bump is manual).
+      # The PR MUST be authored by a non-GITHUB_TOKEN identity — PRs opened with the
+      # default GITHUB_TOKEN do not trigger workflows, so the merge gate would never
+      # run and the PR could not validate. We use a GitHub App installation token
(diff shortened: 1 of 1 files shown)


ewindisch:
Addressed both review points: pinned `peter-evans/create-pull-request` to its v8.1.1 commit SHA (`5f6978fa…`) since the step runs with a write-capable token, and refined the credential note (a GITHUB_TOKEN-authored PR lands approval-gated and won't self-validate unattended). Switched to a fine-grained PAT (`BUILDER_IMAGE_BUMP_TOKEN`) per operator preference.

Issue #314, opened 2026-06-18
[P1] Finish production 2b: range forward, boundary copies, and true stage loading


ewindisch:
Part of #310. Plan: `docs/plans/2026-06-18-multi-gpu-multi-host-inference-spike.md`. (B2; rust M1/M3)

Per-layer `device_map`; thread device through `from_weights`/`build_layer`
(`runtime/architectures/qwen3_5.rs:1218`); `hidden.to_device(next)` at group boundaries in the
forward loop (`:1514`); new `forward_layers(hidden, a..b, start_pos)` trait method
(`torch_engine.rs:902` is whole-model today). Layer-subset loading: stage-aware `from_weights`
(`is_first`/`is_last`; middle stages have no embed/norm/lm_head), global↔local layer-index remap,
KV/SSM vec sizing per stage. Capacity-aware stage assignment (prep for asymmetric VRAM).

ewindisch:
## Spike: `forward_layers` stage-execution design

The trait surface is **mostly already there** — `ModelOperations` (`runtime/architectures/mod.rs`) exposes `embed_tokens`, `forward_from_embeddings`, `apply_final_norm`, `lm_head`, `num_layers`. The only missing piece is a layer-range runner.

### Add one method
```rust
fn forward_layers(
    &self,
    hidden: &Tensor,          // [batch, seq, hidden]; embeddings if first stage, else prior stage output
    range: Range<usize>,      // GLOBAL layer indices this stage owns [a..b)
    start_pos: usize,
    delta: Option<&TenantDelta>,
) -> Result<Tensor>;          // hidden after layer b-1, on the last owned layer's device
```
- **Stage boundary carries only `hidden` + `start_pos`.** `position_ids` is recomputed inside from `start_pos`+seq. **Stage-local, never transferred:** KV cache (per-layer) and qwen3_5 SSM `conv_states`/`rec_states`.
- **Orchestration (arch-agnostic):** stage-0 `embed_tokens` → `forward_layers(0..b)`; middle `forward_layers(a..b)`; last `forward_layers(a..N)` → `apply_final_norm` → `lm_head` (the prefill last-token narrow moves into the last stage). `is_first`/`is_last` are implicit in the range.
- **TTT-on-split (#316):** add a sibling `forward_layers_train` (no KV, full causal, `start_pos=0`) looping the existing training path — keeps inference/training paths separate as today.

### Global↔local index remap
[...]

Issue #311, opened 2026-06-18
[P1] Two-device lifecycle gate: prove one-thread multi-device on real hardware


ewindisch:
Part of #310. Plan: `docs/plans/2026-06-18-multi-gpu-multi-host-inference-spike.md`. (finding C-SEND)

Confirm one process can drive multiple CUDA/ROCm devices given tch `!Send` tensors and the
thread-local RoPE cache (`runtime/architectures/qwen3_5.rs:990`; engine-per-thread because
tensors are `!Send`, `services/inference.rs:191`). Determine: one thread driving N devices, or
thread-per-device with in-process hand-off channels. **Blocks Phase-1 2b.**

ewindisch:
## Spike verdict: VIABLE AS-IS (single-thread / multi-device) — `!Send` is a non-issue for 2b

One process **can** drive multiple GPUs for the 2b milestone; no thread-per-device needed.

- **One thread can own N devices.** A tch tensor carries its own device; libtorch dispatches each op via an internal `DeviceGuard` (no manual `cudaSetDevice`). Holding tensors on `Cuda(0)` and `Cuda(1)` in one thread and `.to_device()`-ing the activation between them is exactly PyTorch naive pipeline parallel. The codebase never touches the CUDA current-device global, so no contention.
- **`!Send` only blocks cross-*thread* tensor moves** (why engines are thread-pinned, `services/inference.rs:191`). It does not stop one thread owning multiple devices. **2b design: the existing dedicated inference thread owns all local GPUs** + a layer→device map + `hidden.to_device()` at group boundaries. Thread-per-device is only needed *later* for overlapped pipeline throughput, not for correctness.
- **`.to_device(Cuda j)`** = P2P copy (`cudaMemcpyPeer`/HIP `hipMemcpyPeer`) over NVLink / Infinity-Fabric xGMI when available, else stages through host (correct either way). The P2P-vs-host measurement is the #312 MI210 hop-cost item.
- **ROCm = byte-identical path** — HIP presents MI210s as `Device::Cuda(i)`; no CUDA-specific enumeration anywhere. Multi-MI210-in-one-process == multi-CUDA.

### Landmines (localized)
- Thread-local RoPE cache keys `(layer_idx, rope_theta)` but **not device** (`qwen3_5.rs:1016`) — safe by accident in a split; **defensively add `device` to the key**.
[...]

Issue #325, opened 2026-06-18
[P4] Executable inter-host pipeline planning, admission, and recovery


ewindisch:
Part of #310. Plan: `docs/plans/2026-06-18-multi-gpu-multi-host-inference-spike.md`. (M-FAIL/M-BUBBLE/M-2C-SCALE)

Multi-hop failure/recovery (stage death mid-request, partial KV, drain, HOL blocking). Confirm
continuous batching exists or build it; quantify bubble fraction `(S-1)/(m+S-1)` and required
microbatch depth. Capacity-aware stage planner from `config.json` (asymmetric VRAM). Envelope:
LAN-only, minimize boundaries; for multi-TB throughput serving, not single-stream latency.

ewindisch:
Code-level recon at `origin/main` / `69eda14916c0870c16f101eb4147f637e09f2043`:

The repository has the right **pipeline-shaped primitives**, but neither production 2b nor 2c is
complete.

* `DevicePool` explicitly chooses one thread driving tensors on multiple devices
  (`crates/hyprstream/src/runtime/device_pool.rs:13-28`), and inference is constructed inside a
  current-thread Tokio runtime/`LocalSet` on a dedicated OS thread
  (`crates/hyprstream-rpc/src/transport/iroh_rpc.rs:387-437`,
  `crates/hyprstream/src/services/inference.rs:2507-2568`). This answers #311's architecture
  choice.
* The factory constructs a multi-device Llama/Qwen model
  (`crates/hyprstream/src/runtime/model_factory.rs:790-814`), but production cached inference calls
  `model.forward_with_cache(...)` (`crates/hyprstream/src/runtime/torch_engine.rs:1091-1098`).
  Llama's corresponding loop has no device-boundary move
  (`crates/hyprstream/src/runtime/architectures/llama.rs:2296-2339`). The intended
  `hidden.to_device(target)` exists only in `forward_layers_inner` (`:2632-2644`); an
  `rg -n "forward_layers\\(" crates/hyprstream/src --glob '*.rs'` search found only architecture
  tests and the unused training helper, not a production inference caller.
* The loader materializes all shard tensors on the primary device before constructing the model
  (`crates/hyprstream/src/runtime/model_factory.rs:252-284`, `:425-499`), so it cannot yet load a
  model larger than one GPU. Qwen3.5 explicitly falls back to one device (`:702-711`).
* There is no 2a engine pool. `DevicePool` calls N engines a future use
[...]

Issue #324, opened 2026-06-18
[P4] Inter-host activation protocol: codec, credits, ordering, and secure stage link


ewindisch:
Part of #310. Plan: `docs/plans/2026-06-18-multi-gpu-multi-host-inference-spike.md`. (M-CHUNK/M-FLOW; M-WIRE)

Ship hidden states between stages over dial_stream/moq (vendor-neutral bytes via
`serialize_state_dict_to_bytes`; receiver chooses device). Chunk large prefill activations
(>64 MiB moq frame; one-frame-per-blob today). Bounded queues + explicit backpressure — a dropped
frame silently corrupts the hidden-state stream. Document the per-token CPU-bounce + RTT cost.

ewindisch:
**Rescoped (Epic B): the activation-link protocol.** Not merely "send a tensor over `dial_stream`." **Remove** the assertion that a prefill activation crosses a verified 64 MiB MoQ application cap — no such cap was found (the only 64 MiB constant is the legacy ZMTP frame cap; Iroh RPC control plane is 4 MiB). Chunking is still required for bounded memory/credits/retry/GPU-CPU staging, but the ceiling must be measured, not copied from ZMTP.

Add: full activation/chunk envelope (model OID, plan digest, request/session, epoch, microbatch/sequence, stage boundary, phase, `start_pos`, accepted-token index, dtype/shape/length, chunk coords, whole-activation digest, provenance/key epoch, terminal/error); gap-free reassembly before execution + stale-epoch rejection; byte-bounded producer/in-flight/reassembly/ready/executor queues; receiver credits/ACKs; fail-closed AEAD + host provenance, no downgrade; prefill-before-decode + per-sequence ordering; cancellation invalidates the epoch; measured raw-header vs safetensors codec. Needs #1236 (stage service) + #1237 (dial). — Epic B ([roadmap](https://claude.ai/code/artifact/8dad534b-c405-4d24-9637-f1781591bab4))

PR #1078, opened 2026-07-17
SPIKE: inter-host pipeline planner + recovery primitives (#325) — blocked on #324 transport

ewindisch:
> **STATUS: SPIKE — not for merge in its current form.** #1222 (v0 admission, #1215) has been merged
> into this branch so the planner contract and its first consumer evolve together. Further review
> happens here.

## What this is

A capacity-aware inter-host pipeline planner, plus microbatch sizing and job-recovery primitives, for
[#325](https://github.com/hyprstream/hyprstream/issues/325) (2c of the closed epic
[#310](https://github.com/hyprstream/hyprstream/issues/310), Multi-GPU & Multi-Host Inference).

- capacity-aware planner: reads Hugging Face model metadata, reserves runtime VRAM, enforces a single
  LAN, minimises stage boundaries, assigns contiguous layer ranges across asymmetric hosts
- exact microbatch bubble/depth calculations, including the four-stage 75% single-microbatch bubble
  and the 27-microbatch threshold for ≥90% utilisation
- per-job pipeline recovery: epoch fencing, stale-completion rejection, planned drain, failed-stage
  replacement, conservative full-prompt/KV replay
- (via #1222) `HostId` migration, `HostResolver`, `ReachAttestation`, the `admit()` step

## Why it is a SPIKE and not a deliverable

**It has no callers, and the execution path it plans for does not exist.**

`git grep InterhostPipelinePlanner` outside the module returns only the `pub use` in
`crates/hyprstream/src/runtime/mod.rs`. Nothing constructs a planner. More decisively:
`services/model.rs` and `services/inference.rs` contain **no** match for `pipeline`, `shard`,
`multi_host` or `device_map` — there is no path that loads or executes a model split across hosts.

**Intra-host staging does exist** and is not in question: `DevicePool`
(`crates/hyprstream/src/runtime/device_pool.rs`) is threaded through `model_factory.rs` at four call
sites, and `model_factory.rs:802` logs *"{} device(s) ({} stage boundary copies per forward)"*. So 2b
— per-layer `device_map` + stage-aware execution, [#314](https://github.com/hyprstream/hyprstream/issues/314) — landed.
**What is missing is the inter-host hop.**

Merging this would freeze a public API — `runtime::{InterhostPipelinePlanner, HostCapacity,
PlannerOptions, PipelineStageAssignment, InterhostPipelinePlan, MicrobatchSizing, PipelineJobState,
RestartDirective, CompletionDisposition, PipelineJobRecovery}` — **before the transport that defines
its interface exists.** A planner whose consumer is unbuilt is shaped by guesses about that consumer.
That is precisely what its independent review found: *"the public planner contract cannot represent
#1215's ratified identity/admission boundary."*

## Blockers

| Blocker | Why it gates this |
|---|---|
| **[#324](https://github.com/hyprstream/hyprstream/issues/324) — inter-host pipeline transport** (`dial_stream` activations + chunking + failover) | **The real missing rung.** This planner decides *which layers go on which hosts*; #324 moves activations *between* them. Without it no stage boundary crosses a machine, so the plan cannot be executed however correct it is. #324 also defines the interface this contract should be shaped to. |
| **[#311](https://github.com/hyprstream/hyprstream/issues/311) — single-process multi-device tch feasibility** | Open, and states *"Blocks Phase-1 2b"* — the `!Send` tch tensor / thread-per-device question. Either it was resolved when #314 landed and this ticket is stale, or 2b shipped over an unresolved feasibility question. Resolve or close before building 2c on top. |
| **[#1215](https://github.com/hyprstream/hyprstream/issues/1215) admission boundary** | #1222 is merged here, but its own review found the acceptance path model-only with no production path enforcing admission, and `ReachAttestation` carrying epoch and nonce with **no expected values for either**. |

## Open review findings on this branch

- **HIGH** — recovery control acknowledgements are **not epoch-fenced**; fencing is the feature this advertises.
- **HIGH** — an existing shard manifest can **silently lose its authoritative capacity check**.
- **MEDIUM** — same-epoch microbatch ID reuse **admits an ABA completion**.
- Checks currently red: Clippy.

## What would make this mergeable

Close or confirm #311 → land #324 → reshape this contract against what the transport actually needs →
validate on real asymmetric hardware ([#317](https://github.com/hyprstream/hyprstream/issues/317),
[#323](https://github.com/hyprstream/hyprstream/issues/323)).

**Hardware is no longer the constraint: a dual-MI210 machine is available**, which is exactly #317's
requirement ("Validate 2b (+2a replication) vs single-GPU ground truth on 2×MI210"), and cross-host
validation is available for 2c.

### DIFF
--- a/crates/hyprstream-rpc/src/envelope.rs
+++ b/crates/hyprstream-rpc/src/envelope.rs
@@ -1160,30 +1160,8 @@ pub fn global_pq_store() -> Option<std::sync::Arc<dyn PqTrustStore>> {
     VERIFY_CONFIG.get().and_then(|c| c.pq_store.clone())
 }
 
-#[cfg(not(target_arch = "wasm32"))]
-static BROWSER_CURRENTNESS_VERIFIER: std::sync::OnceLock<
-    std::sync::Arc<dyn crate::browser_provisioning::BrowserCurrentnessVerifier>,
-> = std::sync::OnceLock::new();
-
-/// Install the checkpoint-backed authority used at WebTransport dispatch.
-/// First write wins; there is deliberately no test/global fallback.
-#[cfg(not(target_arch = "wasm32"))]
-pub fn install_browser_currentness_verifier(
-    verifier: std::sync::Arc<dyn crate::browser_provisioning::BrowserCurrentnessVerifier>,
-) -> Result<()> {
-    BROWSER_CURRENTNESS_VERIFIER
-        .set(verifier)
-        .map_err(|_| anyhow!("browser currentness verifier already installed"))
-}
-
-#[cfg(not(target_arch = "wasm32"))]
-pub fn global_browser_currentness_verifier(
-) -> Option<std::sync::Arc<dyn crate::browser_provisioning::BrowserCurrentnessVerifier>> {
-    BROWSER_CURRENTNESS_VERIFIER.get().cloned()
-}
-
-/// Apply the mandatory process-global policy and trust store to an
-/// `UnwrapOptions`.
+/// Apply the process-global verify configuration to an `UnwrapOptions`,
+/// unless it has already been explicitly downgraded/overridden.
 #[cfg(not(target_arch = "wasm32"))]
 pub fn apply_global_verify_config<'a>(
     mut opts: UnwrapOptions<'a>,
--- a/crates/hyprstream-rpc/src/envelope.rs
+++ b/crates/hyprstream-rpc/src/envelope.rs
@@ -1160,8 +1160,30 @@ pub fn global_pq_store() -> Option<std::sync::Arc<dyn PqTrustStore>> {
     VERIFY_CONFIG.get().and_then(|c| c.pq_store.clone())
 }
 
-/// Apply the process-global verify configuration to an `UnwrapOptions`,
-/// unless it has already been explicitly downgraded/overridden.
+#[cfg(not(target_arch = "wasm32"))]
+static BROWSER_CURRENTNESS_VERIFIER: std::sync::OnceLock<
+    std::sync::Arc<dyn crate::browser_provisioning::BrowserCurrentnessVerifier>,
+> = std::sync::OnceLock::new();
+
+/// Install the checkpoint-backed authority used at WebTransport dispatch.
+/// First write wins; there is deliberately no test/global fallback.
+#[cfg(not(target_arch = "wasm32"))]
+pub fn install_browser_currentness_verifier(
+    verifier: std::sync::Arc<dyn crate::browser_provisioning::BrowserCurrentnessVerifier>,
+) -> Result<()> {
+    BROWSER_CURRENTNESS_VERIFIER
+        .set(verifier)
+        .map_err(|_| anyhow!("browser currentness verifier already installed"))
+}
+
+#[cfg(not(target_arch = "wasm32"))]
+pub fn global_browser_currentness_verifier(
+) -> Option<std::sync::Arc<dyn crate::browser_provisioning::BrowserCurrentnessVerifier>> {
+    BROWSER_CURRENTNESS_VERIFIER.get().cloned()
+}
+
+/// Apply the mandatory process-global policy and trust store to an
+/// `UnwrapOptions`.
 #[cfg(not(target_arch = "wasm32"))]
 pub fn apply_global_verify_config<'a>(
     mut opts: UnwrapOptions<'a>,
--- a/crates/hyprstream-rpc/src/lib.rs
+++ b/crates/hyprstream-rpc/src/lib.rs
@@ -253,14 +253,10 @@ pub use hyprstream_rpc_derive::{authorize, service_factory, FromCapnp, ToCapnp};
 
 #[cfg(not(target_arch = "wasm32"))]
 pub use resolver::{
-    NetworkDiscoveryResolver, Resolver, ResolverProfile, ServiceQuery,
-};
-
-#[cfg(all(not(target_arch = "wasm32"), test))]
-pub use resolver::{
     retry_validated_candidates, select_service_candidate, select_service_candidates,
     AcceptedStateEvidence, AnchoredKemRecipient, CandidateDecision, ResolutionEvidence,
-    SelectedService, ServiceCandidate,
+    NetworkDiscoveryResolver, ResolvedRpcClient, Resolver, ResolverProfile, SelectedService,
+    ServiceCandidate, ServiceQuery, ServiceResolver,
 };
 
 #[cfg(not(target_arch = "wasm32"))]
--- a/crates/hyprstream-rpc/src/lib.rs
+++ b/crates/hyprstream-rpc/src/lib.rs
@@ -253,10 +253,14 @@ pub use hyprstream_rpc_derive::{authorize, service_factory, FromCapnp, ToCapnp};
 
 #[cfg(not(target_arch = "wasm32"))]
 pub use resolver::{
+    NetworkDiscoveryResolver, Resolver, ResolverProfile, ServiceQuery,
+};
+
+#[cfg(all(not(target_arch = "wasm32"), test))]
+pub use resolver::{
     retry_validated_candidates, select_service_candidate, select_service_candidates,
     AcceptedStateEvidence, AnchoredKemRecipient, CandidateDecision, ResolutionEvidence,
-    NetworkDiscoveryResolver, ResolvedRpcClient, Resolver, ResolverProfile, SelectedService,
-    ServiceCandidate, ServiceQuery, ServiceResolver,
+    SelectedService, ServiceCandidate,
 };
 
 #[cfg(not(target_arch = "wasm32"))]
--- a/crates/hyprstream-rpc/src/resolver.rs
+++ b/crates/hyprstream-rpc/src/resolver.rs
@@ -427,28 +427,6 @@ impl RpcClient for ResolvedRpcClient {
         })
         .await
     }
-    async fn call_streaming_for_service_with_method(
-        &self,
-        service: &str,
-        method_discriminator: u16,
-        payload: Vec<u8>,
-        ephemeral: [u8; 32],
-    ) -> anyhow::Result<Vec<u8>> {
-        anyhow::ensure!(
-            service == self.service_name,
-            "generated service authority mismatch"
-        );
-        let service = service.to_owned();
-        self.attempt(|c| {
-            let p = payload.clone();
-            let s = service.clone();
-            async move {
-                c.call_streaming_for_service_with_method(&s, method_discriminator, p, ephemeral)
-                    .await
-            }
-        })
-        .await
-}
     async fn open_stream(&self, payload: Vec<u8>) -> anyhow::Result<Box<dyn StreamHandle>> {
         self.attempt(|c| {
             let p = payload.clone();
--- a/crates/hyprstream-rpc/src/resolver.rs
+++ b/crates/hyprstream-rpc/src/resolver.rs
@@ -30,15 +30,11 @@ use std::sync::Arc;
 use anyhow::anyhow;
 use parking_lot::RwLock;
 
-#[cfg(test)]
 use crate::identity::Did;
 use crate::registry::SocketKind;
-#[cfg(test)]
 use crate::rpc_client::{CallOptions, RpcClient};
-#[cfg(test)]
 use crate::stream_consumer::StreamHandle;
 use crate::transport::{EndpointType, TransportConfig};
-#[cfg(test)]
 use crate::VerifyingKey;
 
 /// Canonical service-resolution request.
@@ -82,7 +78,6 @@ impl ServiceQuery {
 /// It contains only owned public material. Its producer is responsible for
 /// obtaining it through the PDS checkpoint-verifying typed read.
 #[derive(Debug, Clone, PartialEq, Eq)]
-#[cfg(test)]
 pub struct AcceptedStateEvidence {
     pub service_did: Did,
     pub digest: [u8; 64],
@@ -94,7 +89,6 @@ pub struct AcceptedStateEvidence {
 
 /// An anchored, suite-complete request recipient and its current key id.
 #[derive(Debug, Clone, PartialEq, Eq)]
-#[cfg(test)]
 pub struct AnchoredKemRecipient {
     pub key_id: String,
     pub recipient: crate::crypto::hybrid_kem::RecipientPublic,
@@ -103,7 +97,6 @@ pub struct AnchoredKemRecipient {
 
 /// One owned candidate acquired from Discovery/PDS.
 #[derive(Debug, Clone, PartialEq, Eq)]
-#[cfg(test)]
 pub struct ServiceCandidate {
     pub service_name: String,
     pub service_did: Did,
@@ -120,7 +113,6 @@ pub struct ServiceCandidate {
 
 /// Non-secret deterministic audit record for filter/rank/select.
 #[derive(Debug, Clone, PartialEq, Eq)]
-#[cfg(test)]
 pub struct CandidateDecision {
     pub candidate_fingerprint: String,
     pub accepted: bool,
@@ -129,7 +121,6 @@ pub struct CandidateDecision {
 
 /// Evidence carried with the selected atomic result.
 #[derive(Debug, Clone, PartialEq, Eq)]
-#[cfg(test)]
 pub struct ResolutionEvidence {
     pub accepted_state_digest: [u8; 64],
     pub accepted_state_epoch: u64,
@@ -139,7 +130,6 @@ pub struct ResolutionEvidence {
 
 /// One identity-bound, owned service resolution result.
 #[derive(Debug, Clone)]
-#[cfg(test)]
 pub struct SelectedService {
     service_name: String,
     service_did: Did,
@@ -153,7 +143,6 @@ pub struct SelectedService {
     evidence: ResolutionEvidence,
 }
 
-#[cfg(test)]
 impl SelectedService {
     pub fn same_authority(&self, other: &Self) -> bool {
         self.service_name == other.service_name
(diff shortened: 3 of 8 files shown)


ewindisch:
Addressed all three findings at `5afd349b1`:

- hosts with zero post-reserve capacity are filtered as ineligible, so undersized hosts no longer poison otherwise feasible plans
- old completion epochs are discarded as stale, while future epochs now fail as protocol desynchronization without consuming in-flight work
- drain intent and aborted IDs survive stage replacement; only pending aborted IDs can replay while normal admission remains closed

Validation on this head: 13 focused tests passed; affected package tests passed (1,034 library tests, one existing ignore); affected and workspace strict Clippy passed; cargo-deny passed; both WASM guests built; nextest passed 3,159/3,159 with 10 repository-declared skips; release doctests passed.

@coderabbitai review

ewindisch:
Blocked on #1083 (design spike: general-purpose worker/kernel networking). The `host_id`/`lan_id` fields here are unenforced string tags with no binding to a real network identity or verified topology — holding this PR until the networking design lands. Not a code-quality issue with this PR specifically; it's a missing foundation underneath it and the rest of the inter-host family (#324, #325, #327).

ewindisch:
# PR #1078 frozen-API review

Reviewed head: `5afd349b1c1de16f8d2abf25f2153135a127c9ab`

VERDICT: BLOCK

This change has no callers and changes no runtime behavior today, but it freezes the wrong public contract for its first consumer, #1215. The arithmetic tests pass and data-completion epoch fencing is sound; those do not offset an API that contradicts the ratified #1090 §4.5 migration.

## Findings

### BLOCK — The public planner contract cannot represent #1215's ratified identity/admission boundary

`HostCapacity` and `PipelineStageAssignment` expose `host_id: String`, while `HostCapacity` and `InterhostPipelinePlan` expose `lan_id: String` (`interhost_pipeline.rs:260-301`). `InterhostPipelinePlanner::plan` then requires every candidate to carry the same non-empty LAN string (`:316-329`).

#1090 §2.0/§4/§4.5 says the opposite, explicitly:

- migrate `host_id: String` to cryptographic `HostId(Did)`;
- delete `lan_id`;
- prove every adjacent boundary with a fresh `ReachAttestation` in `admit(plan, NetworkView)`;
- keep affinity only as an unverified cost hint in `HostAttributes`;
- fail closed when host resolution or attestation is absent.

#1215 is the named first consumer and repeats those requirements. With this surface it must either (a) stringify and later re-parse trusted identities while fabricating a meaningless common `lan_id`, leaving the planner's advertised LAN invariant unbound to evidence, or (b) immediately break the newly frozen public structs and planner contract. Public struct fields make the required `HostId`/`HostAttributes` migration semver-breaking.

This is also semantically stale against main: since the branch cut, main has required hybrid identified setup for same-cell/federated streams (`75298e469`) and mandatory hybrid posture (`524fd96a3`). A plan keyed only by arbitrary strings has no represented binding to that accepted identity. There is no textual merge conflict, but the identity contracts conflict.

The same shape omits the static attributes #1090 assigns to the planner (affinity, declared bandwidth/latency, and later per-link mover facts), so RTT-bearing `ReachAttestation` data cannot influence placement or the claimed utilization model without another breaking API.

### HIGH — Recovery control acknowledgements are not epoch-fenced

Data completions are handled correctly: `complete(old_epoch, id)` returns `StaleEpochDiscarded` before touching the current in-flight set, and a future epoch errors without consuming work (`:641-655`).

The replacement control path has no equivalent fence. `replacement_ready(stage_count)` accepts no epoch, restart token, or plan digest (`:709-724`). A delayed readiness acknowledgement for replacement N can therefore be accepted while the state machine is `AwaitingReplacement` for failure N+1:

1. failure N advances the epoch; replacement N becomes ready;
2. failure N+1 advances it again;
3. a duplicate/delayed readiness event for N arrives and is accepted as N+1's replacement.

The public primitive claims epoch fencing but cannot distinguish those events. `stage_failed(failed_stage)` likewise has no source epoch, so a delayed failure report from an old chain can fence healthy current work. Pass the expected epoch/restart identity into both control transitions and reject mismatches.

### HIGH — An existing shard manifest can silently lose its authoritative capacity check

When `model.safetensors.index.json` exists, `from_model_dir` parses it but turns a missing or wrongly typed `/metadata/total_size` into `None`, then falls back to the analytic estimate (`:52-63`, `:188-196`). This is the repository's known “absent expected value disables the check” failure class.

A valid JSON index with incomplete metadata can therefore understate actual weights and produce a plan that OOMs while claiming to fit. Conversely, single-file or quantized checkpoints without the index can be materially overestimated and legitimate models refused. If the manifest exists, invalid/missing `total_size` must be an error or the actual checkpoint bytes must be derived by another checked path; absence must not silently select a weaker estimator.

### MEDIUM — Same-epoch microbatch ID reuse admits an ABA completion

`admit` rejects only IDs currently in `in_flight`; `complete` removes the ID, after which the same ID can be admitted again in the same epoch (`:627-654`). A delayed duplicate completion from the first incarnation then has the current epoch and matching ID, so it is accepted and consumes the second incarnation. The API documents no global/per-epoch uniqueness requirement and does not enforce one. Keep completed IDs for the epoch or add a per-attempt generation/token to completion identity.

## Arithmetic and claimed behavior

The advertised constants are correct for the module's ideal equal-stage, zero-network-cost formula:

- four stages, one microbatch: bubble `3/(1+3) = 75%`, utilization `25%`;
- `m/(m+3) >= 0.9` first holds at `m = 27`; `26/29 < 0.9`.

That is not an exact utilization prediction once stage times, activation size, bandwidth, or RTT differ. #1090 already identifies RTT as an input/open scoping question; the API should call this idealized sizing or gain the cost inputs before it is treated as a planning guarantee.

Planned drain, pending replay, full-KV invalidation, and old/future data-completion handling are otherwise internally consistent on inspection.

## What the change breaks

- Current runtime behavior: none. `rg` found no caller outside the new module's re-export.
- The #1215 consumer contract: typed host identity, deletion of `lan_id`, and evidence-driven admission require immediate API replacement/adapters.
- Legitimate candidate sets: any differently labeled candidate poisons the entire call even if a feasible selected subset is mutually reachable.
- Recovery correctness under delayed/duplicated control messages and reused same-epoch IDs.
- Capacity correctness when expected checkpoint-size metadata is absent.

## Evidence

- Read PR #1078, issue #325, issue #1215, and #1090 §2.0/§2.2/§4/§4.5; applied the no-caller scaffold precedent from PR #1108.
- Confirmed clean worktree and pinned head before and after review; `git merge-tree` found no textual conflict with current `origin/main` (`a7097a051d69308bc8d549ba0e02b3302780882f`).
- `env OPENSSL_NO_VENDOR=1 LIBTORCH=.../cuda130/libtorch LD_LIBRARY_PATH=.../lib LIBTORCH_BYPASS_VERSION_CHECK=1 CARGO_INCREMENTAL=0 buildq -- cargo test -p hyprstream interhost_pipeline --lib -- --nocapture`: **13 passed, 0 failed, 1,022 filtered**.
- The first queued attempt lacked `LIBTORCH` and failed in `torch-sys`; this was an environment-selection failure, not a source diagnostic. The corrected queued command above passed.
- Mutation verification is **unverified**. I temporarily reverted the future-epoch rejection and drain-preservation expressions and queued the same focused suite, but the run had not acquired a build slot when the deliver-now instruction arrived. I stopped it, restored both hunks, and verified `git diff --exit-code` plus a clean `git status`. Do not count the current passing tests as proof that those tests fail without the fixes.
- `cargo fmt --check` was not run, per the brief's clean-main #1140 exception.

ewindisch:
## BLOCK — the contract cannot represent what its own consumer needs

Final review (gpt-5.6-sol) at the current head. This answers the question you raised: **what depends on #1078, given it introduces no callers?**

> **BLOCK — The public planner contract cannot represent #1215's ratified identity/admission boundary.**

That is decisive. #1215 (v0 of the #1090 §4.5 ladder) is the consumer, now filed and being built — `HostId` migration, `HostResolver`, `ReachAttestation`, the `admit()` step. The planner API frozen here **cannot express that boundary**, so merging it commits a contract the only consumer must work around.

This is the same failure as https://github.com/hyprstream/hyprstream/pull/1108, found by three independent reviewers there: an inert no-caller scaffold whose contract could not express what its own epic needed. Merging changes no runtime behaviour — the *only* effect is the freeze, so the contract is the whole deliverable and it does not hold.

Three further findings, all real bugs rather than shape problems:

- **HIGH — recovery control acknowledgements are not epoch-fenced.** Fencing is the mechanism this PR advertises; an unfenced ack defeats it.
- **HIGH — an existing shard manifest can silently lose its authoritative capacity check.** Silent loss of a capacity check mis-plans every subsequent pipeline.
- **MEDIUM — same-epoch microbatch ID reuse admits an ABA completion.** A stale completion is accepted as current within one epoch.

**Recommendation: rework against #1215's boundary rather than merging and adapting later.** The consumer is under construction right now, so the two can be shaped together — which is the one moment when fixing a frozen contract is free. Holding this PR open for that; the alternative is closing it and folding the planner into #1215 directly.

Refs https://github.com/hyprstream/hyprstream/issues/1215, https://github.com/hyprstream/hyprstream/issues/1090, https://github.com/hyprstream/hyprstream/issues/325, https://github.com/hyprstream/hyprstream/issues/1083

ewindisch:
Code-level recon at `origin/main` / `69eda14916c0870c16f101eb4147f637e09f2043`:

The repository has the right **pipeline-shaped primitives**, but neither production 2b nor 2c is
complete.

* `DevicePool` explicitly chooses one thread driving tensors on multiple devices
  (`crates/hyprstream/src/runtime/device_pool.rs:13-28`), and inference is constructed inside a
  current-thread Tokio runtime/`LocalSet` on a dedicated OS thread
  (`crates/hyprstream-rpc/src/transport/iroh_rpc.rs:387-437`,
  `crates/hyprstream/src/services/inference.rs:2507-2568`). This answers #311's architecture
  choice.
* The factory constructs a multi-device Llama/Qwen model
  (`crates/hyprstream/src/runtime/model_factory.rs:790-814`), but production cached inference calls
  `model.forward_with_cache(...)` (`crates/hyprstream/src/runtime/torch_engine.rs:1091-1098`).
  Llama's corresponding loop has no device-boundary move
  (`crates/hyprstream/src/runtime/architectures/llama.rs:2296-2339`). The intended
  `hidden.to_device(target)` exists only in `forward_layers_inner` (`:2632-2644`); an
  `rg -n "forward_layers\\(" crates/hyprstream/src --glob '*.rs'` search found only architecture
  tests and the unused training helper, not a production inference caller.
* The loader materializes all shard tensors on the primary device before constructing the model
  (`crates/hyprstream/src/runtime/model_factory.rs:252-284`, `:425-499`), so it cannot yet load a
  model larger than one GPU. Qwen3.5 explicitly falls back to one device (`:702-711`).
* There is no 2a engine pool. `DevicePool` calls N engines a future use
  (`crates/hyprstream/src/runtime/device_pool.rs:8-11`, `:185-199`); a source search for a
  `Vec<TorchEngine>`/per-device engine construction found none.
* Continuous batching is built as a Llama/TorchEngine primitive
  (`crates/hyprstream/src/runtime/architectures/mod.rs:258-279`,
  `crates/hyprstream/src/runtime/torch_engine.rs:1151-1184`) but is not wired:
  `rg -n "forward_batched_step\\(" crates/hyprstream/src --glob '*.rs'` found only its definition.

Therefore #311 is **partially stale**: the one-thread/N-device design is settled, but its
real-device feasibility gate is not. The existing GPU E2E deliberately stays single-device
(`crates/hyprstream/tests/e2e_inference.rs:721-724`), while the two-GPU-shaped test only constructs
device enums (`:610-631`). It should not close as “proved” until a real two-device load, prefill,
cached decode, RoPE reuse, and teardown pass exists.

For #324, the local contract to replace is a contiguous stage range where stage 0 embeds, the final
stage performs norm/head, and the only boundary tensor is `hidden [B,Q,H]`; `start_pos` is scalar
metadata, position IDs are reconstructed, and KV/SSM state remains stage-local
(`crates/hyprstream/src/runtime/architectures/mod.rs:376-413`). Payload size is
`B*Q*H*sizeof(dtype)`: at BF16 and H=4096, B1/Q8192 prefill is 64 MiB, B1/Q1 decode is 8 KiB, and
B16/Q1 decode is 128 KiB per boundary.

#324 needs to own an executable stage link, not just tensor serialization: model/plan identity,
request/session/epoch/microbatch/sequence identity, stage/layer boundary, prefill-vs-decode and
`start_pos`, dtype/shape/length, chunk layout/digest, byte-based receiver credits, exact
reassembly/order, cancellation, and authenticated terminal state. The existing MoQ path has
AEAD/HMAC/provenance support (`crates/hyprstream-rpc/src/moq_stream.rs:593-623`, `:733-805`,
`:1885-1928`), but its producer constructs a whole encoded frame and does not await downstream
credits (`:652-658`, `:694-826`). No explicit MoQ application payload cap was found; the 64 MiB
constant is the legacy ZMTP cap (`crates/hyprstream-rpc/src/zmtp_framing.rs:26-27`), so the exact
chunk ceiling remains **UNVERIFIED** and should be measured.

Failure should fence a pipeline epoch, reject stale completions, rebuild the chain, and replay the
prompt plus committed generated-token history to reconstruct stage-local caches. The durable
coordinator record needs model/plan/stage identities, prompt and token commits, sampling/RNG state,
tenant/delta revision, per-microbatch admission/completion, and the last client-visible output
commit.

#1078/#1222 are a useful spike, not merge-shaped production code. Keep contiguous ranges,
capacity-aware asymmetric placement/minimal boundaries
(`origin/pr/1078:crates/hyprstream/src/runtime/interhost_pipeline.rs:311-368`), bubble arithmetic as
a metric (`origin/pr/1222:crates/hyprstream/src/runtime/interhost_pipeline.rs:912-963`), and #1222's
typed identity/admission plus epoch/replay invariants (`:733-769`, `:993-1183`). Reshape placement
around exact indexed tensor bytes, runtime/KV/activation/reassembly memory, backend and link
capabilities, and executable stage descriptors. The current assignment has only host/range/two
byte counts (`origin/pr/1078:crates/hyprstream/src/runtime/interhost_pipeline.rs:287-302`).
`git grep -n InterhostPipelinePlanner origin/pr/1222 --
':!crates/hyprstream/src/runtime/interhost_pipeline.rs'` found only the `pub use`, not a caller.

Dependency order:

1. Wire the production device-aware forward and split-training helper (no open ticket; closed
   #314/#316 are incomplete by code).
2. Add true stage/subset loading and either wire Qwen3.5 stages or narrow the model contract (no
   ticket).
3. Resolve #311 with a real two-device gate.
4. Build 2a's N-engine replica pool (no ticket), then run #317.
5. Finish #329's scheduler around the existing batched kernel.
6. Add a network-addressable stage lifecycle/service (no ticket), then real remote replica
   resolution/dialing (no open ticket in the requested set).
7. Implement #324's activation protocol and stage integration.
8. Reshape/integrate #1078/#1222 under #325.
9. Run #323 cross-host/cross-vendor validation.
10. Add #327 inter-host backward communication.
11. Add tensor/expert parallelism only if a single layer cannot fit one device/host (conditional,
    no ticket); no collectives were found by
    `git grep -n -E 'all_reduce|tensor_parallel|world_size|nccl' HEAD --
    crates/hyprstream/src/runtime`.

#317 is diagnostic-runnable now but not acceptance-runnable: the current path will likely expose
the missing production copy, while no numerical-parity harness or 2a pool exists. After the
forward/loading fixes it needs deterministic single-vs-split full-logit and cached-decode
comparison, explicit proof both GPUs executed, repeated teardown, peak HBM, prefill/TTFT/decode,
boundary-copy latency, utilization, and P2P measurements.

#323 is not end-to-end runnable because no stage service/link exists. This week it can still
measure generic Iroh/MoQ RTT/throughput at 8 KiB, 128 KiB, and chunked 64 MiB; ROCm/CUDA
GPU↔CPU+codec+crypto costs; independent same-model backend drift; and synthetic stream faults.
The external MI210 checkout/runtime/topology/model inventory is **UNVERIFIED**.

Focused validation, all through `buildq -- cargo ...`: `device_pool --lib` **14 passed**;
`pipeline --lib` **20 passed**; `--test e2e_inference e6_layer` **5 passed**. These validate
planning and CPU/same-device equivalence, not a real cross-device boundary.

Full report:
`/home/birdetta/projects/cyberdione/.fleet-coord/mhi/REPORT.md`

ewindisch:
**Independent review (frontline seat) — disposition assessment, bound to head 778429ab3.**

Still hard-blocked: #324 (inter-host activation transport) and #325 remain open — the transport this plans for does not exist, and the PR body concedes no callers. mergeStateStatus DIRTY, though the conflict is only runtime/mod.rs (small); the heavily-rewritten runtime files are untouched.

Two default-on, non-feature-gated behavior flips are buried here and must NOT ride in on a spike: oci_backend.rs network_mode "" → "none" with the is_empty() guard deleted (every OCI sandbox loses networking), and nspawn.rs network_veth true → false (+--private-network). Defensible hardening — as its own reviewed PR with a migration note. resolver.rs also promotes ~7 cfg(test) types to permanent public API with no production consumer.

Extractable now: crates/hyprstream/src/runtime/interhost_pipeline.rs — ~1240 lines of pure, transport-free planning/recovery logic + 16 tests, conflicts with nothing. Clean standalone PR.

Recommendation: (1) standalone PR for interhost_pipeline.rs (+ rebased mod.rs export); (2) separate PR for the sandbox network defaults; (3) drop the cfg(test) removals; (4) close this spike with a summary into #325. Converting to draft meanwhile so it can't merge with the network-default landmines aboard.

ewindisch:
Closing this draft spike after triage. The PR body and independent review confirm it has no production caller, is blocked on the inter-host transport contract, and bundles network-default changes that must not merge with planner work. Preserve the branch and extract: (1) transport-shaped planner contract after #324, (2) separate sandbox network-default change with migration evidence, and (3) cross-host validation under the Phase 2 plan.

PR #1118, opened 2026-07-19
fix(pds): bounded #atproto overlap verification (closes #918 Problem A)

### LINKED ISSUE #918: PDS commit verification breaks on #atproto key rotation — overlap (drain/lead slot) publication
## Scope (rescoped 2026-07-22)

**This issue is now Problem A only: rotation survivability.** Position-aware verification — proving a key was valid *at a commit's position*, so a retired or compromised key cannot verify forever — is a different problem and lives at #1169 (C3 of epic #1158). Do not merge the two; A fixes an outage, B fixes a forgery.

## The problem

`oauth::did_document` publishes only the **single active** ES256 key as the `#atproto` verification method. The ES256 store rotates (~14 days).

- A commit signed by ES256 key *K* is persisted verbatim (`commit\0{did}`) and served keyless on reads.
- When the active key rotates *K → K′*, the DID document advertises only *K′*.
- Every **already-persisted** commit still signed by *K* now **fails verification**, until the repo is re-published (which re-signs with *K′*).

Verification of historical commits silently breaks at each rotation boundary for any repo that has not re-published since.

## Fix: overlap publication (drain + lead slots)

Publish the drain (and lead) ES256 slots as additional verification methods so a verifier accepts any currently-published slot.

- **New commits are still signed by the active key only.** The overlap window is for *verification*, not for signing.
- **The window must be bounded and the drain slot must actually leave the document when it expires.** An unbounded overlap is a retired key that verifies forever — which is the forgery problem this explicitly does not solve.
- Alternative considered: re-sign-on-rotation (one commit per repo, keeps a single-VM document, but couples the rotation task to the PDS store).

## Why the earlier "reserved for did:at9p" ruling was wrong

An earlier comment on this issue ruled out multi-key publication on the atproto-compat key, reserving multi-key/hybrid rotation for `did:at9p` native identities, on the grounds that atproto's model is a signed rotation-log rather than multiple concurrent verification methods.

**That conflated what the spec prescribes with what implementations do — and our own code already depends on the permissive behavior.**

`build_did_document` (`crates/hyprstream/src/services/oauth/did_document.rs:327-375`) ships, in a single document alongside the `#atproto` P-256 Multikey:

- Ed25519 mesh/OAuth verification methods, in **both** multibase and JWK forms
- an ML-DSA-65 post-quantum mesh verification method (#157)
- `#mesh-kem` hybrid keyAgreement entries — X25519 + ML-KEM-768 (S1 / #552)

with the function's own doc comment stating these are **"ignored by atproto resolvers (which match by id/type)"**. If resolvers were strict about the verification-method array, our current documents would already be broken. The tolerate-what-you-do-not-understand property is not hypothetical here; it is load-bearing today.

## The PQ-hybrid mechanism this generalizes (operator input, 2026-07-22)

**Hybrid key rotation works by publishing multiple keys and letting verifiers ignore the ones they do not understand.** In JWKS this is the `kid` mechanism: a JWKS may carry a classical key and a post-quantum or composite key side by side; a verifier selects by `kid` and **skips entries whose `kid` or `alg` it does not recognize**, rather than failing the set. That is what makes a PQ rollout possible without a flag day — classical verifiers keep working against the classical entry while PQ-capable verifiers use the PQ entry, and neither needs to know about the other.

Erica verified this behavior against each mainstream JWT library, matching RFC guidance and PQUIP recommendations, and needed **one patch to ruby-jwt** where the behavior was errant. So the property is empirically established across the ecosystem, not merely specified.

**A DID document's `verificationMethod` array has the same shape and the same property**, which is exactly why our mesh Ed25519, ML-DSA-65 and ML-KEM entries can coexist with `#atproto` today. Overlap rotation and PQ-hybrid rollout are therefore *the same mechanism*: publish more than one key, let each verifier take the one it understands, and bound the window.

The design consequence: **hyprstream should treat "one published key at a time" as an anti-pattern wherever it appears**, not as a property to preserve on the atproto-compat surface. See https://github.com/hyprstream/hyprstream/issues/1183.

## Remaining open question (narrow, empirical)

Do atproto resolvers tolerate **two entries carrying the `#atproto` fragment**, or must the drain slot use a distinct id? Our code comment says atproto takes the **first** matching entry — if that is accurate, a strict resolver silently ignores a second entry, which decides whether the drain slot needs a distinct fragment id and verifier-side try-each.

This must be answered from **real implementation source** (`@atproto/identity`, indigo, others), not spec prose — the spec-versus-implementations distinction is the entire point above. Recon is dispatched on PR #1118.

## Scope note

Not required for #910a landing — #913 fixed only the frozen-key defect. See the KNOWN LIMITATION comment on `Es256KeyGetter` in `crates/hyprstream/src/services/discovery.rs`.

ewindisch:
## What

Closes **#918 Problem A only**: bounded overlap publication lets Hyprstream verify a commit signed by a just-rotated-out `#atproto` key while that key remains explicitly published.

## Design

- The active signing key remains the sole exact `#atproto` method and is the only key `PdsPublisher` uses for new commits.
- Drain and lead are separate, bounded DID verification methods: `#atproto_drain` and `#atproto_lead`, each with required `nbf`/`exp`. They are removed from the ES256 publication source at their expiry.
- Hyprstream parses at most active + drain + lead, rejects duplicate/malformed recognised methods, and verifies only slots live at verification time.
- Stock atproto consumers continue selecting only the active `#atproto` fragment. Recon and implementation-source citations: https://github.com/hyprstream/hyprstream/issues/918#issuecomment-5052404053

## Explicit non-goal: #918 Problem B remains blocked

This does **not** prove that a key was valid at a commit’s causal position and does not protect against a compromised retired key during its bounded overlap window. Position-aware historical verification remains #1169 (C3), downstream of #1167/#1168/#1170. The existing production DID-authority/history work is not replaced or worked around here.

## Tests

- old-key commit verifies during a bounded drain window;
- a key absent from the document or past its expiry does not verify;
- currently-published lead verifies;
- new commits are signed by active, never lead;
- drain removal occurs at its published expiry.


ewindisch:
## Review revision pushed (rebased onto latest main)

All four REVISE findings are addressed. Findings → commits:

**Finding 1 (BLOCKER, producer — DID document must publish the drain key).**
`AtprotoIdentity` now carries an optional bounded `DrainKey { vk, nbf, exp }`; `build_did_document` emits it as a second `#atproto-drain` Multikey with explicit `nbf`/`exp`; `root_did_document` sources it from the ES256 store's new `drain_slot()` accessor. The DID document now advertises both keys during a drain window.

**Finding 2 (BLOCKER, consumer — wire the real verification boundary).** `RecordResolver` gains `resolve_verifying_keys() -> Option<RotationKeySet>` (default `Ok(None)`). `PlacementIndex::ingest_did` prefers it (falling back to `resolve_verifying_key` as a one-slot unbounded set, then the trusted-resolver posture), and `decode_repo_car` now takes `Option<&RotationKeySet> + now` and calls `Commit::verify_against_keys` — the production ingest boundary no longer calls single-key `verify(vk)`. The production `PdsRecordResolver` implements `resolve_verifying_keys` by building the set from the shared ES256 rotation store via a DID-document encode/decode round-trip, wired in the discovery service factory.

**Finding 3 (HIGH — typed bounded key set, not bare keys).** `RotationKeySet` replaces the bare `IntoIterator<VerifyingKey>`. Each slot carries `[nbf, exp]`; `verify_against_keys` filters to live slots at the verification instant, so an expired drain key or a pre-`nbf` lead key is never consulted, and a cached/stale drain key cannot accept forged commits after its `exp`. The set is capped (`MAX_SLOTS`) and constructible only from a resolved DID document (`from_did_document`) or, for tests, `testing_from_slots` (+ an infallible `single_unbounded` for the fallback path). A `#atproto-drain` slot without explicit `nbf`/`exp` is rejected (an unbounded drain key is exactly the stale-key-forever failure this prevents); the active `#atproto` slot honors explicit bounds when published, defaulting to unbounded.

**Finding 4 (MEDIUM — real regression tests via the wired path, not hand-built arrays).** The discovery placement suite exercises `PlacementIndex::ingest_did → decode_repo_car → verify_against_keys` with a resolver returning a `RotationKeySet` built through a DID-document encode/decode round-trip: (a) drained-key commit verifies within the window; (b) fails after `exp`; (c) fails before `nbf`; (d) fails closed when no slot is live; (e) forged signature fails. pds unit tests add a `from_did_document` round-trip and the unbounded-drain rejection; did_document tests add a producer-side drain-publication round-trip.

### Verification
```
cargo check -p hyprstream-pds -p hyprstream-discovery   ✔
cargo test  -p hyprstream-pds -p hyprstream-discovery   ✔
  pds 216; discovery 82 lib + 21 doc; +5 wired rotation cases
cargo clippy -p hyprstream-pds -p hyprstream-discovery -p hyprstream \
  --all-targets --release -- -D warnings                  ✔ (pre-commit hook green)
```

### Drive-by (flagged, not #918)
`hyprstream-discovery` dev-dep now enables `hyprstream-rpc`'s `test-classical-policy` feature. A pre-existing announcement test (`ordinary_announcement_handler_populates_production_resolver`) uses the gated `SignedEnvelope::new_signed` helper, which is invisible cross-crate without that feature — the full-workspace build unified the feature and masked it; building `-p hyprstream-discovery` in isolation surfaced it. This is API drift, unrelated to #918.

Closes #918.

ewindisch:
## Round-2 revision pushed (rebased onto latest main)

Round-2 review: finding 1 (drain publication) FIXED in r1; finding 2's cross-process store gap + the additional blocker (frozen publisher) are split to **#1123** by the round-2 task scope (deployment architecture, not a login-MVP verification blocker). This push addresses the in-scope MUST-FIX items:

**A. Sealed the bounded-keyset API (r2 finding 3).** `RotationKeySlot` fields are private; `RotationKeySlot::new`/`unbounded` are `pub(crate)`; `RotationKeySet::testing_from_slots` is `#[cfg(test)]` (absent from non-test builds); `RotationKeySet::single_unbounded` is removed. The ONLY public constructor is `RotationKeySet::from_did_document`. The single-key resolver fallback now builds a minimal DID-document fragment via the new public `p256_public_key_multibase` encoder and constructs through `from_did_document` — there is no bare-key constructor anywhere.

**B. `from_did_document` validates authority, not just suffix (r2 finding 3).** New signature: `from_did_document(doc, expected_did, authoritative_active)`. It requires `doc.id == expected_did`; matches VMs by FULL id (`{did}#atproto` / `{did}#atproto-drain`); rejects duplicate methods (no silent first-match); checks each VM is `type: "Multikey"` + `controller: expected_did`; and cross-checks the `#atproto` key against the resolver-supplied authoritative current active key. **A stale pre-rotation document (active = rotated-out key) fails closed** — new test `from_did_document_rejects_stale_pre_rotation_document` proves it; plus `from_did_document_rejects_wrong_did` and `from_did_document_rejects_duplicate_atproto`.

**C. Producer frozen-key limitation made explicit in-code.** Doc comments at `factories.rs` (the `active_key()`-once-at-construction site) and the `PdsPublisher.signing_key` field both point to #1123. Live per-publish key resolution + cross-service store propagation left to #1123 by scope (operational, not verification).

PR description updated to state the split: this PR = rotation-survivable **verification** + sealed API; cross-service propagation + publisher live-key = #1123.

```
cargo test -p hyprstream-pds -p hyprstream-discovery   ✔  (pds 213; discovery 82 lib + 21 doc; +3 authority tests)
cargo clippy -p hyprstream-pds -p hyprstream-discovery -p hyprstream --all-targets --release -- -D warnings  ✔
```

ewindisch:
## Re-steer: atproto-spec-aligned re-sign-on-rotation (replaces drain-slot)

PR #1118 is re-steered per the #918 design decision. The earlier drain-slot approach (`#atproto-drain` with `nbf`/`exp`) was a hyprstream invention — a stock atproto relay/AppView only reads `#atproto`, so it only "worked" when hyprstream was on both sides, defeating interop. This push replaces it with the spec model.

### What changed (drain → re-sign-on-rotation)

**Producer:**
- DID document publishes ONLY the single current `#atproto` method. All `#atproto-drain` / `DrainKey` / `atproto_verification_methods` / nbf-exp-on-VM code removed from `did_document.rs`.
- `PdsPublisher` resolves the LIVE active signing key at sign time from the shared `Es256SigningKeyStore` (which it holds by `Arc`, not a frozen `SigningKey`). After K→K' the next commit is signed with K'. The ES256 store was switched `tokio::RwLock` → `parking_lot::RwLock` to make this sync.

**Verifier:**
- Reverted to single-key `Commit::verify(vk)`. Removed `RotationKeySet`, `RotationKeySlot`, `verify_against_keys`, `live_slots`, `verify_record_proof_against_keys`, and all the window-enforcing machinery.
- Added `atproto_verifying_key_from_did_document(doc, expected_did)` — authority-hygienic single-key resolver (validates doc.id, full-id match, duplicate rejection, controller/type).
- `PlacementIndex::ingest_did` → `decode_repo_car` → `Commit::verify(vk)`.

**did:at9p paths untouched** (multi-key/hybrid rotation stays per cross-repo contract §4.1).

### Tests (replaced the drain suite)
- Rotate K→K', re-sign head with K', verify against K' → passes.
- Head NOT re-signed (still K) fails against rotated doc (K') → both directions asserted.
- Authority hygiene: wrong DID / wrong controller / duplicate `#atproto` → rejected.
- Forged signature → fails.

### #1123 scope split
Cross-`--ipc` rotation-event delivery (ensuring the publisher process observes the rotation) is #1123 — the single-process re-sign works here; multi-process is the deployment follow-up. In-code comments at `factories.rs` + `PdsPublisher` point to #1123.

```
cargo test -p hyprstream-pds -p hyprstream-discovery   ✔ (pds 210; discovery 80 lib + 21 doc)
clippy -D warnings                                      ✔
```

ewindisch:
Round-2 re-signed findings are addressed in f1ecb09af.

- ES256 lead-to-active promotion now invokes an in-process publisher hook that re-signs the existing persisted head over identical canonical unsigned DAG-CBOR bytes. Normal publish holds the ES256 read guard across load/sign/persist, while promotion takes the write guard, closing the stale-key persist race. Hook installation also performs a startup reconciliation; cross-IPC event delivery remains deferred to #1123.
- Placement ingest now fails closed on missing verification material. Local verification materializes the same #atproto verification-method shape used by the published DID document and passes it through atproto_verifying_key_from_did_document; foreign DID resolution remains unsupported and therefore rejects rather than skipping verification.
- Removed stale drain/multi-key guidance and added the missing wrong-VM-type test.

Validation:
- cargo test -p hyprstream-pds -p hyprstream-discovery: 205 PDS unit + 6 PDS integration; 80 discovery unit + 21 doctests; all green.
- cargo test -p hyprstream services::discovery::pds_store_tests --lib: 19 passed, including the production promotion/re-sign test.
- cargo clippy -p hyprstream-pds -p hyprstream-discovery -p hyprstream --all-targets -- -D warnings: clean.
- Repository pre-commit Clippy hook: clean.

ewindisch:
Round 3 is finalized at a6296fcdb after rebasing onto current origin/main (the transactional implementation is 4169bb58f).

Atomic/recoverable promotion:
- K-prime remains invisible behind the ES256 write guard until active-slot persistence and persisted-head re-sign both succeed.
- A persistence or re-sign failure restores/retains K, leaves K-prime queued as lead, returns not-promoted, and retries next tick.
- The publisher hook signs with the explicit candidate key while the write guard is held; publication now follows the same key-store -> publisher lock order.
- New production-path coverage proves a failed re-sign leaves the DID-visible K matching the K-signed stored head, then the next tick re-signs and exposes K-prime. A separate test covers active-slot persistence failure without invoking the hook.

Finding 2 is split to #1124 as directed. The repo subject is an Ed25519 node did:key but the real served document is rooted at the OAuth issuer did:web authority; aligning and provisioning that identity is #1124 work. The verifier now carries an explicit TODO/comment and the PR body documents the split. did:at9p remains untouched.

Post-rebase validation:
- cargo test -p hyprstream-pds -p hyprstream-discovery: passed (314 package/integration tests plus 21 discovery doctests)
- cargo test -p hyprstream --lib services::discovery::pds_store_tests -- --nocapture: 20 passed
- strict affected-crate Clippy -D warnings: passed
- repository release-mode pre-commit Clippy: passed

The final mainline rebase also incorporated merged #1089; its new liveness-admission fixture was adapted to the signed-CAR helper and is covered by the package suite.

ewindisch:
Round 4 is pushed in 3d9068fb6.

Mandatory fail-closed fix:
- The local Ed25519 did:key no longer fabricates a document containing the unrelated P-256 key. resolve_verifying_key returns None, and PlacementIndex rejects the local repo until #1124 supplies a real authority.
- The acceptance test is replaced with an end-to-end rejection assertion.

Bounded rotation fixes:
- Local resolve_record/resolve_repo reads hold the same ES256 generation guard as DID/key reads, closing the K-prime-head/K-key reader window.
- A barrier test pauses after the K-prime head write and proves the repo reader cannot return until K-prime is live.
- Rotation with no re-sign hook stays pending and reports not-promoted.

Crash/restart durability is split to #1126 (durable pending marker + startup reconciliation/publication gating); #1123 remains cross-process delivery. did:at9p is untouched.

Validation:
- cargo test -p hyprstream-pds -p hyprstream-discovery: passed (314 package/integration tests + 21 doctests)
- focused PDS production suite: 21 passed
- focused no-hook test: passed
- strict affected-crate Clippy -D warnings: passed
- repository release-mode pre-commit Clippy: passed

ewindisch:
Round 5 revision complete in 21c6a6ec6.

- The production ES256 promotion hook now treats a failed Weak<PdsPublisher>::upgrade() as an error. A dropped publisher can no longer report a successful promotion without re-signing the head; rotation remains pending instead.
- Added dropped_publisher_hook_keeps_rotation_pending_and_head_consistent: publish a K-signed head, install the real weak hook, drop the publisher, attempt K→K′ rotation, and verify K remains active, K′ remains queued/unadvertised, and the stored head verifies under K but not K′.
- The existing fail-closed local resolver/verifier behavior is unchanged.

Validation:
- cargo test -p hyprstream --lib dropped_publisher_hook_keeps_rotation_pending_and_head_consistent -- --nocapture (1 passed)
- cargo test -p hyprstream --lib services::discovery::pds_store_tests -- --nocapture (22 passed)
- cargo test -p hyprstream-pds -p hyprstream-discovery (all unit, integration, and doctests passed)
- cargo clippy -p hyprstream -p hyprstream-pds -p hyprstream-discovery --all-targets -- -D warnings
- release-mode pre-commit Clippy hook passed

Branch is current with origin/main (0 behind).

ewindisch:
**fable final review: CLEAR.** Five rounds on trust-chain code, re-steered mid-flight from a non-spec drain-slot design to atproto-aligned re-sign-on-rotation — the harder, correct path. Verified: single `#atproto` key published (no drain method), repo head re-signed with the live active key on rotation over identical canonical DAG-CBOR bytes, the verifier fails closed on every path (no fabricated-document acceptance for the transitional did:key mismatch — confirmed #1124 owns the real repo-authority fix), the reader-guard barrier prevents any observable mismatched generation, and a dropped-publisher promotion hook now errors rather than silently reporting success. Bot findings checked directly: 0 unresolved GitHub review threads, CodeRabbit SUCCESS, CodeQL clean on the Rust delta (the neutral aggregate is a pre-existing Python-job config gap, not a finding in this PR). Scoped follow-ups cleanly filed and referenced: #1123 (cross-process rotation delivery), #1124 (real repo authority), #1126 (crash-atomic durable pending state). Enqueueing.

ewindisch:
# PR #1118 review — `fix(pds): survive #atproto key rotation in commit verification`

**VERDICT: BLOCK**

Reviewed detached head `21c6a6ec67179a07915f47c8fe23ed361183f747` against base `bbbef97c04c38319f8a5e8bd1159163791dd9a58`. The worktree remained clean and read-only.

## Findings

### 1. BLOCKER — verification is not keyed to the commit's position in rotation history

`PlacementIndex::ingest_did` resolves exactly one current key and passes it to `Commit::verify` (`crates/hyprstream-discovery/src/placement_index.rs:176-197,264-266`). `Commit::verify` verifies only against that supplied key; it does not consult `rev`, `prev`, a DID operation, or any authenticated rotation history (`crates/hyprstream-pds/src/commit.rs:222-247`). The comments explicitly defer historical verification (`commit.rs:231-236`).

Consequently the verifier cannot distinguish these two cases:

- a legitimate commit signed by K at a log position before K→K';
- a forged commit made with retired K after K→K'.

The PR chooses to reject both once the live document advertises K'. Its own `old_key_head_does_not_verify_against_rotated_doc` test proves the first required case is rejected (`commit.rs:505-523`). Re-signing the current head with K' makes current-head interop work, but it does not establish whether an old K signature was valid at its own position in the log.

This needs a verifiable key history (for did:plc, the authenticated operation/audit log; for did:web, an explicitly defined verifiable-history policy) and a verification API that selects the key valid at the commit's authenticated log position. A commit-supplied timestamp alone would not suffice because retired K can sign a backdated value.

### 2. BLOCKER — the production verifier never resolves the DID document or any authority history

`atproto_verifying_key_from_did_document` has good local authority checks, but every call site is a unit test or the DID-document producer test; there is no production consumer call. The production `PdsRecordResolver::resolve_verifying_key` unconditionally returns `Ok(None)` for local and foreign DIDs (`crates/hyprstream/src/services/discovery.rs:1433-1448`). `PlacementIndex` then clears prior state and rejects ingest (`placement_index.rs:181-194`).

That is fail-closed, not accept-on-error, but it also means this PR cannot successfully verify any production repo CAR. The tests inject a `VerifyingKey` directly and therefore do not exercise DID fetch, document freshness, rotation-history absence, or an unknown-key/unknown-`kid` condition at the real trust boundary. The local did:key/did:web authority mismatch is deferred to #1124, but until that dependency lands this PR does not close #918 operationally.

Positive fail-closed evidence: wrong DID, wrong controller/type, duplicate `#atproto`, missing key, resolver error, malformed signature, and forged signature are rejected. There is no accept-on-error fallback in the inspected ingest path.

### 3. BLOCKER — K/head promotion is not crash-atomic and startup reconciliation can fail open operationally

Promotion persists filesystem active K' first (`crates/hyprstream/src/auth/key_rotation.rs:1608-1626`), then invokes the RocksDB head re-sign hook, and only afterward changes the in-memory active slot. A crash or concurrent process startup between the first two durable writes leaves active K' on disk with a K-signed head. The source acknowledges this gap at `key_rotation.rs:1664-1668`.

On restart, `install_es256_promotion_hook` attempts one reconciliation (`crates/hyprstream/src/services/discovery.rs:995-998`), but the registry factory merely logs a warning if it fails and continues installing the publisher (`crates/hyprstream/src/services/factories.rs:775-779`). The DID document can therefore publish persisted K' while the served head remains signed by K. There is no durable pending marker or publication gate.

This is exactly #1126's invariant. #1118 is **not safe to merge ahead of #1126** as a completed rotation fix.

### 4. HIGH — the supported `--ipc` topology cannot execute the rotation fix

The ES256 store is a process-local `OnceLock` (`crates/hyprstream/src/auth/key_rotation.rs:1027-1038`). The rotation task is spawned by OAuth (`crates/hyprstream/src/services/oauth/mod.rs:612-622`), while the promotion hook is installed by the registry factory (`crates/hyprstream/src/services/factories.rs:768-779`). In split processes those are different stores: OAuth has no publisher hook, and `rotate_es256_keys` deliberately leaves promotion pending when no hook exists (`key_rotation.rs:1608-1613`).

The passing `es256_rotate_without_hook_stays_pending` test concretely demonstrates that behavior (`key_rotation.rs:3326-3356`). It avoids a mismatched head by stopping rotation entirely, but it does not deliver rotation survivability in the supported topology. #1118 must be sequenced behind #1123, or the cross-process event/reconciliation mechanism must be included here.

### 5. HIGH — required boundary tests are absent

The in-process re-sign, failure retry, stale weak-hook, and reader-lock tests are useful. However, there is no test in which an authenticated rotation history accepts an old-key commit before the transition and rejects an old-key commit after it. The existing negative test supplies only the live K' and therefore cannot represent the required boundary. There is also no crash-injection/restart test at the filesystem-key/RocksDB-head write boundaries and no split-process integration test. Those omissions correspond directly to findings 1, 3, and 4.

## Merge sequencing

Do **not** merge #1118 as the fix for #918 before #1126 and #1123. The safe order is:

1. define and implement authenticated, position-aware historical key resolution (and land #1124's real repo authority needed by the production resolver);
2. land #1126's durable pending marker, startup reconciliation, and DID-publication gating;
3. land #1123's cross-IPC promotion delivery/reconciliation;
4. integrate the re-sign/current-head path and the required positive/negative boundary tests, then merge #1118.

If the current in-process code is desired as preparatory plumbing, it should be split into an explicitly non-closing PR and kept unreachable from production rotation until those invariants are present.

## Validation run

- `cargo test -p hyprstream-pds -p hyprstream-discovery` — passed (including 82 discovery unit tests, 21 discovery doctests, 205 PDS unit tests, and integration suites).
- `cargo test -p hyprstream --lib services::discovery::pds_store_tests -- --nocapture` — 22 passed.
- `cargo test -p hyprstream --lib auth::key_rotation::tests::es256_ -- --nocapture` — 6 passed.
- `cargo build -p hyprstream-pds -p hyprstream-discovery -p hyprstream` — passed.
- `cargo clippy -p hyprstream-pds -p hyprstream-discovery -p hyprstream --all-targets -- -D warnings` — passed.

Only the existing `proc-macro-error2 v2.0.1` future-incompatibility warning was emitted.

ewindisch:
## Fix-round analysis (post-BLOCK): #1118 cannot close #918 without #1124/#1126/#1123 — stopping rather than shipping a local workaround

I picked up the BLOCK verdict (three blockers, two HIGHs), re-verified the reviewed head is current (`21c6a6ec6`), and read the #1124 decision (Erica, 2026-07-22: **host-form did:web**, did:plc federate-only) plus the recon at `.fleet-coord/t1124/recon.md` before touching anything. Conclusion up front: **every blocker in this verdict is owned by a dependency that has not landed, and the #1124 design explicitly forbids building the stopgap versions of them inside this PR.** No code was changed; here is the per-finding accounting.

### Finding 1 (BLOCKER — verification not keyed to rotation-history position) — NOT addressed here; owned by #1124's DidOp log

The #1124 decision comment (§3) defines the mechanism this finding demands, and it is not a #1118-local change: a per-account signed hash-chained operation log (`DidOp { version, prev, seq, did, doc_cid, rotation_keys, head_at_op, sig }`), stored canonically as records in the account's own PDS repo, projected at `/.well-known/did-log.json`, witnessed by the `hyprstream-ledger` checkpoint chain. The position-aware rule is already specified there:

> K was valid for commit C **iff** C is a descendant of `head_at_op` of the op that introduced K, and an ancestor-or-equal of `head_at_op` of the op that superseded K.

Any history mechanism I built inside #1118 to satisfy this finding would be exactly the "second, incompatible rotation-history mechanism" the fix scope prohibits. What #1118 already does remains correct as far as it goes: it fails closed — an old-key commit is *rejected* once the live document advertises K′, so nothing forged is ever accepted. The missing piece is only the *positive* case (accepting a legitimately pre-rotation commit), and that requires the DidOp log to exist.

### Finding 2 (BLOCKER — production resolver never resolves a DID document) — NOT addressed here; fail-closed `Ok(None)` is the correct interim state

`PdsRecordResolver::resolve_verifying_key` returning `Ok(None)` is deliberately fail-closed; the alternative (accept-on-error) would be worse. Making it real requires an authority source, and the authority shape was only decided today: did:web host-form, the DID document as a **sealed artifact** recorded in the op log, resolution WebPKI-rooted (intersecting #1135), did:plc resolve-only. Implementing a production resolver now — before #1124 lands the mint/serve/seal shape — means building against an authority that does not yet exist and likely ripping it out. And even a working current-document resolver would not close #918 without finding 1's history: it would still verify against the *current* key set, which is the exact failure mode of finding 1.

### Finding 3 (BLOCKER — promotion not crash-atomic, reconciliation can fail open) — NOT addressed here; #1124 says "#1126 and the DidOp are the same record — do not build two"

The tempting local fix is a durable pending marker + publication gate in `key_rotation.rs`. The #1124 design explicitly pre-empts that: "Never publish K′ until the head is durably re-signed by K′" becomes *structural* — the op that publishes K′ embeds the K′-signed head, so an op that can't be sealed can't publish K′. Order: re-sign head → durably write head → seal op → seal doc → publish. A separate pending-marker mechanism built here would be the second record the decision says not to build. This finding is #1126's invariant, implemented as part of the DidOp work.

### Finding 4 (HIGH — `--ipc` topology cannot execute rotation) — CONFIRMED: #1118 requires #1123 (or its DidOp-shaped resolution) first

Stating it plainly as asked: **yes, #1118 requires #1123 to land first for the supported split-process topology.** The ES256 store is process-local; OAuth rotates, the registry factory holds the promotion hook, and `rotate_es256_keys` deliberately stays pending with no hook (proven by `es256_rotate_without_hook_stays_pending`). Per the #1124 design, #1123 itself is expected to be closed by the DidOp shape — the sealed log head is the durable cross-process record of the active generation, so no process needs to observe OAuth's `OnceLock`. Faking cross-process delivery locally in #1118 would be a third mechanism to rip out.

### Finding 5 (HIGH — missing boundary tests) — NOT added; the tests require the mechanisms above

The required tests are: (a) authenticated rotation history accepts old-K commit before the transition and rejects it after (needs finding 1's DidOp log), (b) crash-injection/restart at the filesystem-key/RocksDB-head write boundaries (needs finding 3's durable record), (c) split-process integration (needs finding 4's delivery). Each must FAIL without its fix, and each fix lives outside this PR. Writing them against mocks of mechanisms that don't exist would produce tests that pass against a fiction — worse than absent.

### What this PR does and does not do, net

Does, at `21c6a6ec6`: spec-aligned single-`#atproto` publication; in-process head re-sign on promotion with fail-closed retry; reader-guard barrier against mixed-generation observation; fail-closed ingest with no accept-on-error path (independently confirmed in the BLOCK review's positive evidence).

Does not, and cannot without dependencies: historical/position-aware verification (#1124 DidOp), production DID resolution (#1124 + #1135), crash-atomic publication (#1126 as DidOp), cross-process rotation (#1123 as DidOp).

### Recommended sequencing (agrees with the verdict)

1. Land #1124's DidOp log (op log + sealed documents + `head_at_op`).
2. Land #1126 as part of that record (not separately).
3. Close #1123 with the sealed log head as the cross-process generation record.
4. Return to #1118: wire `resolve_verifying_key` to the real authority, add position-aware verification against the op log, add the three boundary tests — each then fails without its real fix.
5. Only then does "closes #918" become true.

If the in-process plumbing here is wanted as preparatory work, the honest move is to retitle this PR to drop "closes #918" and merge it explicitly as non-closing plumbing; I have not changed the title or merged anything, per scope. Otherwise this PR should sit until steps 1–3 land.

ewindisch:
## Affected by the #1124 decision — hyprstream-hosted accounts get host-form did:web

Decision approved 2026-07-22; design epic: https://github.com/hyprstream/hyprstream/issues/1158

C3 of the epic supplies what review finding 2 said was missing — the authority history this PR needs in order to answer 'was this key valid at the commit position'. Do **not** invent a second rotation-history mechanism here; C1 defines the log format and verifier library, and C3 wires it into this verifier. Review finding 4 (the `--ipc` topology cannot execute the fix) is answered by C4, which closes #1123.

ewindisch:
Tracking ticket filed for the missing authority history: **https://github.com/hyprstream/hyprstream/issues/1169 (C3)**, which depends on **https://github.com/hyprstream/hyprstream/issues/1167 (C1)** for the log format and verifier. Review finding 4 (`--ipc` topology) is answered by **https://github.com/hyprstream/hyprstream/issues/1170 (C4)**. Epic: https://github.com/hyprstream/hyprstream/issues/1158 — do not build a second rotation-history mechanism in this PR.

ewindisch:
## Reverification: blocked — no verifier workaround shipped

I independently re-read the BLOCK review, the final #1124 decision/recon, the current PR head (`21c6a6ec6`), and the production paths. The prior analysis remains correct; the new ticket mapping makes the required sequence explicit. **#1118 cannot truthfully close #918 yet, and I made no code changes.**

- `PdsRecordResolver::resolve_verifying_key` still returns `Ok(None)` for local and foreign DIDs, so production ingest fails closed rather than resolving any current authority—let alone authority at a commit position.
- The current verifier accepts one supplied/current key only. It cannot distinguish a valid K-signed commit before K→K′ from a forged K-signed commit after it. Host-form `did:web` now requires the authenticated DidOp history; C1/#1167 defines it and C3/#1169 wires the position-aware query into this verifier.
- Rotation is not crash-atomic: active-key persistence precedes the PDS-head re-sign, and publication is not durably gated. C2/#1168 replaces the proposed separate pending marker with the single sealed record (`re-sign head → durable head → seal op with `head_at_op` → seal doc → publish`).
- The supported `--ipc` topology still cannot deliver the in-process hook: OAuth and registry have distinct process-local stores. #1170 (C4, dependent on C2) closes #1123 by making the sealed op-log head the cross-process source of truth.

Therefore the prerequisites are C1/#1167 → C2/#1168 → C4/#1170 (closing #1123), plus C3/#1169 for historical validity. Implementing a second history, pending-marker, or IPC mechanism in this PR would directly conflict with the settled #1124 design. This PR should remain blocked (or be explicitly retitled as non-closing preparatory plumbing) until those land; it must not merge as the #918 fix.

ewindisch:
## Recommendation: SPLIT. Two different problems are merged in this PR.

The BLOCK stands as written, but re-reading https://github.com/hyprstream/hyprstream/issues/918 against the review makes clear they are asking for different things, and only one of them is genuinely blocked.

**Problem A — what #918 actually asks for: rotation survivability.** Historical commits signed by K stop verifying once the document advertises only K-prime. #918 own "possible fixes" are drain-slot publication or re-sign-on-rotation. Neither needs an operation log. This is small, testable, and **not blocked by anything in the epic.**

**Problem B — what the review demands: position-aware verification.** Prove key K was valid *at the commit position*, so a retired or compromised key cannot verify forever. This genuinely needs the authenticated `DidOp` history: #1167 (C1, format + verifier) → #1168 (C2, crash-atomic rotation) → #1170 (C4, cross-process head, closes #1123) → #1169 (C3, wires the causal-lineage query into this verifier). Plus `PdsRecordResolver::resolve_verifying_key` returning `Ok(None)` for local **and** foreign DIDs, so production ingest resolves no authority at all today.

A fixes the outage; B fixes the forgery. Conflating them is why this PR reads as blocked on a four-ticket chain when its stated goal is not.

### The A path is now open, and the reason it was closed was wrong

The prior design comment on #918 ruled out multi-key publication as "reserved for did:at9p native identities." **Our own `build_did_document` disproves the premise** — see https://github.com/hyprstream/hyprstream/issues/918#issuecomment-5052383967. We already ship Ed25519 mesh VMs, an ML-DSA-65 PQ VM, and hybrid `#mesh-kem` keyAgreement entries in the same document, with the code commenting that atproto resolvers ignore what they do not match. Operator input (Erica, 2026-07-22): this is the same tolerate-unknown-`kid` property that makes hybrid JWKS work, verified against mainstream JWT libraries per RFC and PQUIP guidance.

The one open question is narrow and empirical: do resolvers tolerate two `#atproto`-fragment entries, or does the drain slot need a distinct id? The code says atproto takes the **first** matching entry, which decides the shape.

### Concrete path for this PR

1. **Rescope #1118 to Problem A** — drain/lead-slot publication (or re-sign-on-rotation), gated on the empirical resolver check. It can then legitimately close #918.
2. **Do not claim position-aware verification here.** That is #1169 (C3), downstream of C1/C2/C4.
3. If you prefer not to rescope, retitle as non-closing preparatory plumbing and leave #918 open — but do not merge this as the #918 fix.

Recommend (1). It converts a four-ticket blocked chain into one shippable change plus a correctly-sequenced follow-on.

ewindisch:
Implemented and pushed `b77af361a` for #918 **Problem A**.

- DID documents keep the exact active `#atproto` key first and publish bounded, distinct `#atproto_drain` / `#atproto_lead` verification slots with `nbf`/`exp`.
- Commits are signed only by the active key; discovery verifies against only currently-live published slots, and unknown fragments are ignored.
- The source-level interoperability recon is recorded on #918: https://github.com/hyprstream/hyprstream/issues/918#issuecomment-5052404053.

This deliberately does **not** claim causal-position verification or protect against a compromised old key while it remains in the overlap window. Problem B remains #1169, downstream of #1167 / #1168 / #1170.

Validation: PDS and discovery unit suites, focused DID-document test, and the commit hook CI-mirroring clippy pass. The full `-p hyprstream` suite reached the network test environment and aborted in `noq-proto` on its GSO datagram-alignment assertion; it was not a test failure in this patch. Shared-target full build/clippy are queued behind other worktrees.

ewindisch:
# PR #1118 review — bounded `#atproto` overlap

VERDICT: BLOCK

Reviewed detached head `b77af361a4c4cba3481b46bb9730d0f54c9423cb` against base `bbbef97c04c38319f8a5e8bd1159163791dd9a58`. The PR worktree remained clean and read-only.

## Findings

### 1. BLOCKER — the bounded keys are not reachable from any production commit authority

The isolated parser/verifier works, but it is not wired to a deployable verification path.

`PdsPublisher` writes commits whose `did` is the node's Ed25519-derived `did:key` (`crates/hyprstream/src/services/factories.rs:744-748`). The new P-256 active/drain/lead methods are instead published on the OAuth issuer's root `did:web` document (`crates/hyprstream/src/services/oauth/did_document.rs:512-558`). A verifier resolving `commit.did` therefore cannot reach the document containing the overlap keys.

The production code acknowledges this mismatch: `PdsRecordResolver::resolve_verifying_key` returns `Ok(None)` for every DID (`crates/hyprstream/src/services/discovery.rs:1431-1448`), and it does not override `resolve_verifying_keys`. The trait default consequently maps that `None` to `None` (`crates/hyprstream-discovery/src/service.rs:568-576`), after which `PlacementIndex::ingest_did` rejects and clears the repo (`crates/hyprstream-discovery/src/placement_index.rs:172-193`).

`PublishedAtprotoKeys::from_did_document` has no production call site: `rg` found only its definition and tests. Thus the required end-to-end case—persist K-signed commit, rotate K→K', resolve the currently published document for that commit's DID, and verify under drain—does not exist and cannot pass in production. The PR body overclaims “Hyprstream parses ... and verifies”; only the isolated library/test path does.

### 2. BLOCKER — supported split-process deployment never creates the drain window

The ES256 store is a process-local `OnceLock` (`crates/hyprstream/src/auth/key_rotation.rs:1015-1026`). OAuth owns the rotation task (`crates/hyprstream/src/services/oauth/mod.rs:603-620`), while Registry installs the promotion/re-sign hook (`crates/hyprstream/src/services/factories.rs:766-780`). Under `--ipc`, these are different stores/processes. OAuth's store has no hook, and `rotate_es256_keys` intentionally leaves promotion pending when no hook is installed (`crates/hyprstream/src/auth/key_rotation.rs:1600-1606`).

The included `es256_rotate_without_hook_stays_pending` test asserts this exact behavior. In the supported topology K never promotes to K', old active never becomes drain, and Problem A is avoided only by stopping ES256 rotation. This does not solve rotation survivability.

The old re-sign-on-rotation machinery is unnecessary for the newly selected overlap design and is the reason promotion still requires an in-process hook. It also rewrites the persisted head's signature/CID while preserving `rev`, `data`, and `prev` (`crates/hyprstream/src/services/discovery.rs:1192-1240`), creating two commit CIDs for one revision even though normal publication requires revisions to strictly advance.

### 3. HIGH — expiry admission and publication do not share one boundary

`PublishedAtprotoKey::is_live_at` accepts a slot when `now <= exp` (`crates/hyprstream-pds/src/commit.rs:313-315`), while rotation deletes it when `now >= exp` (`crates/hyprstream/src/auth/key_rotation.rs:1670-1676`). At `now == exp`, a cached document still authorizes the key even though the publication source considers it expired and removes it. The new rejection test checks only `exp + 1`, so it misses the deadline required by the review brief.

Physical removal is also tick-driven, not expiry-driven. The rotation task skips its immediate tick and defaults to a six-hour interval (`crates/hyprstream/src/auth/key_rotation.rs:1321-1328`; `crates/hyprstream/src/config/mod.rs:1259-1261`). An expired drain loaded at startup can remain in the served document for up to six hours. An expired lead can remain indefinitely when promotion is blocked, because cleanup removes only drain. The unit test named `es256_drain_is_removed_at_its_published_expiry` manually invokes a rotation tick at the deadline and does not cover either runtime case.

### 4. HIGH — this PR breaks production placement/discovery ingest

The diff changes missing verification material from trusted-resolver ingest to unconditional rejection. Because the only production resolver returns `None`, local and foreign repo ingestion now always fails, stale facts are cleared, and cross-node placement/group membership cannot be populated (`crates/hyprstream-discovery/src/placement_index.rs:172-193`). This is a real availability regression in the same PR, not a hypothetical safer failure mode.

### 5. MEDIUM — the served DID document can mix rotation generations

`root_did_document` snapshots active, drain, and lead through three separate lock acquisitions (`active_slot()`, `drain_slot()`, `lead_slot()`) at `did_document.rs:537-539`. Promotion can occur between those calls. For example, the handler can capture old K as active, then capture K as the post-promotion drain while the persisted head has already been re-signed by K'. Stock atproto consumers still select old K from exact `#atproto`, producing a transient verification outage. Take one atomic slot snapshot under one read guard.

## Checks that do pass

- The explicit non-goal is stated correctly in the PR body and `Commit::verify_against_published_keys`: this does not prove key validity at a commit position and does not contain a compromised overlap key. Problem B remains #1169.
- The interoperability recon is real implementation-source recon. Pinned TypeScript code selects the first exact `#atproto`; pinned Indigo code maps by fragment, preserves the first duplicate, and later reads only `atproto`. Distinct `#atproto_drain` / `#atproto_lead` fragments match that conclusion and do not disturb stock selection.
- New commits use only `Es256SigningKeyStore.active`; `publisher_signs_new_commits_with_active_not_lead_key` directly checks active success and lead failure.
- The isolated parser rejects wrong DID/controller/type, duplicate full recognized methods, missing active, malformed bounds, absent keys, and signatures under no live slot.
- No “absent expected value disables the check” fallback was found at the modified placement trust boundary; `None` rejects. The problem is that this also rejects every production path.

## Validation

- `buildq -- cargo test -p hyprstream-pds --lib -- --nocapture` — **PASS**, 209 passed, 0 failed, 0 ignored. This includes drain admission, `exp + 1` rejection, lead admission, absent-key rejection, and DID-method authority/duplicate checks.
- `buildq -- cargo test -p hyprstream --lib auth::key_rotation::tests::es256_ -- --nocapture` — first attempt stopped in `torch-sys` because `LIBTORCH` was not selected; environment-only, not a source failure.
- The same command was requeued with the repository's cached CUDA 13.0 libtorch, `OPENSSL_NO_VENDOR=1`, and version bypass, then cancelled on operator instruction to deliver immediately. **Unverified.**
- A read-only temporary mutation/probe tree was prepared to prove the drain-admission test fails when overlap iteration is restricted to active and to test rejection exactly at `exp`, but the queued Cargo runs were not started before the delivery instruction. **Mutation evidence unverified; no claim is made that these tests were run.**

## Required before re-review

Wire a real authority-matching production resolver and an end-to-end K→K' test; make split-process rotation actually promote and publish drain without the in-process re-sign dependency; use one exclusive expiry boundary and prove physical startup/runtime removal; restore legitimate placement ingest; and snapshot all DID slots atomically.

ewindisch:
Addressed the re-review findings in afbda9814:

1. The published authority is now reachable: registry commits use the root `did:web` DID that owns the served `#atproto` methods, and the production PDS resolver supplies the same atomic bounded publication snapshot for that DID. The new end-to-end test publishes with K, rotates to K′, resolves the local authority, and verifies/ingests the K-signed commit through drain.

2. ES256 promotion no longer needs the in-process PDS re-sign hook. It persists K as drain before K′ becomes active; IPC registry/discovery workers refresh the durable slot source before signing/resolving, so supported split-process deployment produces and consumes the overlap window.

3. Bounded slots are half-open `[nbf, exp)`: verifier admission and cleanup both reject/remove at `now >= exp`. Added the exact-deadline test and startup cleanup; runtime rotation polls at most every minute. I temporarily restored the old inclusive comparison and confirmed that exact-deadline test fails, then restored the fix.

4. Placement ingest is restored for the local production authority: it resolves the did:web bounded set and verifies the persisted K-signed commit through drain before indexing. Foreign/unresolvable DIDs remain fail-closed.

5. `root_did_document` now takes one `Es256KeySlots` snapshot, eliminating mixed-generation active/drain/lead output.

Validation: `buildq -- cargo test -p hyprstream-pds --lib -- --nocapture` (209 passed); focused queued Hyprstream ES256/production-path tests; pre-commit clippy passed. `cargo fmt --check` remains known to fail on clean main (#1140), so I formatted only the touched files.

Problem B / position-aware verification remains out of scope (#1169).

ewindisch:
VERDICT: BLOCK

# PR #1118 re-review — bounded `#atproto` overlap

Reviewed head `afbda9814373fd805cf9756c6490c18054c81a25` against base
`bbbef97c04c38319f8a5e8bd1159163791dd9a58`. The source worktree remained
clean and read-only.

## Findings

### 1. BLOCKER — the production systemd/IPC processes still do not share the ES256 rotation source

The new disk refresh closes the standalone `--ipc` case only when every
process resolves the same writable secrets directory. The supported systemd
split does not have that property.

Installed systemd services set `HYPRSTREAM__SECRETS__PATH=%d`, where `%d` is
that unit's private systemd credential directory
(`crates/hyprstream-service/src/service/manager/units.rs:219-229`). The
credential lists imported into each unit contain RSA/TLS/bootstrap/service
credentials, but no `es256-signing-key.{active,drain,lead}` key or metadata
files (`units.rs:28-92`, `169-215`). The encryption/provisioning path likewise
does not manage any ES256 slot files
(`crates/hyprstream-service/src/service/manager/systemd.rs:470-566`).

Every OAuth, Registry, and Discovery process nevertheless calls
`resolve_secrets_dir()` and binds its process-local `Es256SigningKeyStore` to
that private `%d` (`factories.rs:710-712`, `1822-1824`;
`auth/key_rotation.rs:1018-1029`). On a missing active slot,
`load_or_init_es256_key_store` generates a random process-local key; when
persistence into the read-only credential ramfs fails, it logs but still
retains that random key in memory (`auth/key_rotation.rs:1588-1597`).
Consequently OAuth can publish key A, Registry can sign commits with key B,
and Discovery can resolve key C. OAuth also cannot persist a promotion/drain
generation for the other processes to refresh.

The new tests do not exercise this topology. Despite its name,
`es256_rotate_without_hook_publishes_drain_for_ipc_workers` uses one store and
directly reloads one ordinary temp directory
(`auth/key_rotation.rs:3377-3404`). The production-path test also rotates,
publishes, resolves, and ingests through the same in-memory store
(`services/discovery.rs:1931-1985`). There is no second process/store, private
credential directory, or systemd unit boundary.

This leaves the prior split-process blocker open in a production deployment.
Mutable rotation state needs a shared writable state location/protocol
separate from per-unit secret credentials, and the test must instantiate the
actual split topology.

### 2. HIGH — an expired drain still remains in the served DID document until the next polling tick

Verifier admission is now correctly half-open (`nbf <= now < exp`), and
cleanup uses the same `now >= exp` comparison. But
`root_did_document` publishes every slot returned by `slots_snapshot` without
filtering against the current time
(`services/oauth/did_document.rs:535-562`). Runtime deletion remains
poll-driven (`auth/key_rotation.rs:1325-1335`, `1675-1689`). Capping the poll
interval at 60 seconds reduces the old six-hour lag but does not satisfy the
requirement that the drain actually leave the document at expiry: it remains
served for up to another minute.

Filter bounded slots at the document-publication boundary (and the equivalent
local publication snapshot) or schedule expiry exactly. The startup cleanup
test does not cover a live handler between `exp` and its next rotation tick.

### 3. HIGH — changing the durable repo identity from `did:key` to root `did:web` has no migration

The new factory changes the publisher/resolver repo DID from the node's stable
Ed25519 `did:key` to the OAuth root `did:web`
(`services/factories.rs:695-751`, `1819-1834`). `PdsRecordStore` is explicitly
keyed by repo DID, and both record and commit RocksDB keys include that DID
(`services/discovery.rs:159-180`, `265-284`). There is no migration, alias, or
continuity check.

After upgrade, every existing `did:key` repo and its `at://did:key/...` record
URIs become unreachable through the production resolver; the next publish
starts a new `did:web` repo with no `prev` link to the old head. Existing
placement facts and references using the node DID also no longer address the
new repo. This is the concrete thing the authority fix breaks and must be
migrated explicitly (or gated as a knowingly destructive identity reset).

## What the new head does fix

- The local writer and resolver are now configured for the same root
  `did:web` authority in the ordinary shared-directory topology.
- Promotion no longer requires the obsolete head re-sign hook, and the old
  active key becomes a bounded drain before the new active is persisted.
- Active/drain/lead publication uses one in-memory snapshot.
- Verification and cleanup use the same exclusive expiry comparison.
- Missing active authority still fails closed; no
  “absent expected value disables the check” fallback was found.
- New commits continue to use active only, never lead or drain.
- Problem B / position-aware verification remains correctly out of scope
  under #1169.

## Validation

- `buildq -- cargo test -p hyprstream-pds --lib overlap_slot_expires_and_is_not_consulted -- --nocapture`
  — **PASS** (1 passed).
- `buildq -- cargo test -p hyprstream --lib es256_ -- --nocapture`
  — **PASS** (12 passed).
- `buildq -- cargo test -p hyprstream --lib local_did_web_authority_admits_pre_rotation_commit_through_drain -- --nocapture`
  — **PASS** (1 passed).
- `buildq -- cargo test -p hyprstream --lib publisher_signs_new_commits_with_active_not_lead_key -- --nocapture`
  — **PASS** (1 passed).
- Negative control in an isolated archive of `afbda9814`: changed
  `PublishedAtprotoKey::is_live_at` back from `now < exp` to `now <= exp`.
  `overlap_slot_expires_and_is_not_consulted` then **FAILED** at the exact
  expiry assertion. The hunk was restored and the archive matched the reviewed
  source again.
- GitHub checks at review time: Clippy, CodeQL, browser WASM, and cargo-deny
  passed. As instructed, the known clean-main `cargo fmt --check` failure was
  not used as a finding.

## Required before re-review

Provide one shared writable, crash-safe ES256 rotation source for the actual
systemd/IPC process layout and test it across independent stores/processes;
remove bounded slots from the served document at their exact expiry; and
migrate or explicitly preserve the existing `did:key` repo identity and
records when moving production publication to root `did:web`.

ewindisch:
VERDICT: BLOCK

# PR #1118 re-review — bounded `#atproto` overlap

Reviewed head `afbda9814373fd805cf9756c6490c18054c81a25` against base
`bbbef97c04c38319f8a5e8bd1159163791dd9a58`. The source worktree remained
clean and read-only.

## Findings

### 1. BLOCKER — the production systemd/IPC processes still do not share the ES256 rotation source

The new disk refresh closes the standalone `--ipc` case only when every
process resolves the same writable secrets directory. The supported systemd
split does not have that property.

Installed systemd services set `HYPRSTREAM__SECRETS__PATH=%d`, where `%d` is
that unit's private systemd credential directory
(`crates/hyprstream-service/src/service/manager/units.rs:219-229`). The
credential lists imported into each unit contain RSA/TLS/bootstrap/service
credentials, but no `es256-signing-key.{active,drain,lead}` key or metadata
files (`units.rs:28-92`, `169-215`). The encryption/provisioning path likewise
does not manage any ES256 slot files
(`crates/hyprstream-service/src/service/manager/systemd.rs:470-566`).

Every OAuth, Registry, and Discovery process nevertheless calls
`resolve_secrets_dir()` and binds its process-local `Es256SigningKeyStore` to
that private `%d` (`factories.rs:710-712`, `1822-1824`;
`auth/key_rotation.rs:1018-1029`). On a missing active slot,
`load_or_init_es256_key_store` generates a random process-local key; when
persistence into the read-only credential ramfs fails, it logs but still
retains that random key in memory (`auth/key_rotation.rs:1588-1597`).
Consequently OAuth can publish key A, Registry can sign commits with key B,
and Discovery can resolve key C. OAuth also cannot persist a promotion/drain
generation for the other processes to refresh.

The new tests do not exercise this topology. Despite its name,
`es256_rotate_without_hook_publishes_drain_for_ipc_workers` uses one store and
directly reloads one ordinary temp directory
(`auth/key_rotation.rs:3377-3404`). The production-path test also rotates,
publishes, resolves, and ingests through the same in-memory store
(`services/discovery.rs:1931-1985`). There is no second process/store, private
credential directory, or systemd unit boundary. The real systemd deployment
was not run end-to-end and is therefore **UNVERIFIED**; the static production
path above is sufficient to show that its stores cannot share the new files.

This leaves the prior split-process blocker open in a production deployment.
Mutable rotation state needs a shared writable state location/protocol
separate from per-unit secret credentials, and the test must instantiate the
actual split topology.

### 2. HIGH — an expired drain still remains in the served DID document until the next polling tick

Verifier admission is now correctly half-open (`nbf <= now < exp`), and
cleanup uses the same `now >= exp` comparison. But
`root_did_document` publishes every slot returned by `slots_snapshot` without
filtering against the current time
(`services/oauth/did_document.rs:535-562`). Runtime deletion remains
poll-driven (`auth/key_rotation.rs:1325-1335`, `1675-1689`). Capping the poll
interval at 60 seconds reduces the old six-hour lag but does not satisfy the
requirement that the drain actually leave the document at expiry: it remains
served for up to another minute.

Filter bounded slots at the document-publication boundary (and the equivalent
local publication snapshot) or schedule expiry exactly. The startup cleanup
test does not cover a live handler between `exp` and its next rotation tick;
that live-handler timing case is **UNVERIFIED by test** and established by the
handler/interval control flow.

### 3. HIGH — changing the durable repo identity from `did:key` to root `did:web` has no migration

The new factory changes the publisher/resolver repo DID from the node's stable
Ed25519 `did:key` to the OAuth root `did:web`
(`services/factories.rs:695-751`, `1819-1834`). `PdsRecordStore` is explicitly
keyed by repo DID, and both record and commit RocksDB keys include that DID
(`services/discovery.rs:159-180`, `265-284`). There is no migration, alias, or
continuity check.

After upgrade, every existing `did:key` repo and its `at://did:key/...` record
URIs become unreachable through the production resolver; the next publish
starts a new `did:web` repo with no `prev` link to the old head. Existing
placement facts and references using the node DID also no longer address the
new repo. This is the concrete thing the authority fix breaks and must be
migrated explicitly (or gated as a knowingly destructive identity reset).
An upgrade against a populated production database was **UNVERIFIED by test**;
the break follows directly from the changed DID and DID-keyed storage schema.

## What the new head does fix

- The local writer and resolver are configured for the same root `did:web`
  authority in the ordinary shared-directory topology.
- Promotion no longer requires the obsolete head re-sign hook, and the old
  active key becomes a bounded drain before the new active is persisted.
- Active/drain/lead publication uses one in-memory snapshot.
- Verification and cleanup use the same exclusive expiry comparison.
- Missing active authority still fails closed; no
  “absent expected value disables the check” fallback was found.
- New commits continue to use active only, never lead or drain.
- Problem B / position-aware verification remains correctly out of scope
  under #1169.

## Validation

- `buildq -- cargo test -p hyprstream-pds --lib overlap_slot_expires_and_is_not_consulted -- --nocapture`
  — **PASS** (1 passed).
- `buildq -- cargo test -p hyprstream --lib es256_ -- --nocapture`
  — **PASS** (12 passed).
- `buildq -- cargo test -p hyprstream --lib local_did_web_authority_admits_pre_rotation_commit_through_drain -- --nocapture`
  — **PASS** (1 passed).
- `buildq -- cargo test -p hyprstream --lib publisher_signs_new_commits_with_active_not_lead_key -- --nocapture`
  — **PASS** (1 passed).
- Negative control in an isolated archive of `afbda9814`: changed
  `PublishedAtprotoKey::is_live_at` back from `now < exp` to `now <= exp`.
  `overlap_slot_expires_and_is_not_consulted` then **FAILED** at the exact
  expiry assertion. The hunk was restored and the archive matched the reviewed
  source again.
- GitHub checks at review time: Clippy, CodeQL, browser WASM, and cargo-deny
  passed. As instructed, the known clean-main `cargo fmt --check` failure was
  not used as a finding.

## Required before re-review

Provide one shared writable, crash-safe ES256 rotation source for the actual
systemd/IPC process layout and test it across independent stores/processes;
remove bounded slots from the served document at their exact expiry; and
migrate or explicitly preserve the existing `did:key` repo identity and
records when moving production publication to root `did:web`.

ewindisch:
Rebased `fix/plc-rotation-verification` onto `origin/main` at `89befb09c` and force-pushed with lease (`1332c0370`).

Conflict resolutions:
- `crates/hyprstream/src/services/oauth/did_document.rs` (four replay conflicts): retained #1121’s issuer-host validation and per-user DID/PDS behavior, while retaining this PR’s bounded `#atproto_drain`/`#atproto_lead` publication. The final IPC hunk takes this PR’s atomic `slots_snapshot()` (the production reachability fix) plus main’s validated handle derivation. I also applied that snapshot-based bounded publication to #1121’s newly added per-user DID path, since it is a production commit authority.
- `crates/hyprstream/src/auth/key_rotation.rs`: took main’s OAuth-conformance additions (`issuer` and `client_assertion_jkt`) over the PR’s formatting-only older initializer; retained the PR’s non-conflicting overlap/IPC rotation changes.
- `crates/hyprstream/src/services/oauth/xrpc.rs`: retained this PR’s new `lead: None` alongside the existing drain field, so hosted snapshot DID construction matches the expanded identity shape.

Re-verified:
- `buildq -- cargo fmt --check` passed.
- `buildq -- cargo test -p hyprstream-pds pre_rotation_commit_verifies_during_bounded_drain_overlap` passed.
- `services::oauth::did_document::tests::build_did_doc_publishes_bounded_overlap_slots` passed from the buildq-produced test binary.
- `auth::key_rotation::tests::es256_rotate_without_hook_publishes_drain_for_ipc_workers` passed.
- `services::discovery::pds_store_tests::local_did_web_authority_admits_pre_rotation_commit_through_drain` passed.

Negative regression proof: temporarily bypassing `PublishedAtprotoKeys::live_keys` bounds caused `overlap_slot_expires_and_is_not_consulted` to fail at the exclusive drain expiry; restored the bound check immediately. I also temporarily removed overlap publication before restoring it; its producer regression test requires the missing drain/lead VMs.

Not merged.

ewindisch:
Blocking on #1190 (option a). The ES256 active/drain/lead rotation source has the same cross-process requirement as the #atproto generation state: OAuth, Registry, and Discovery must derive one crash-safe authoritative view even when their systemd credential directories are distinct. The seam is the sealed op-log head introduced by #1190; I will depend on that mechanism rather than create a second mutable rotation store that could diverge.

#1190 is still OPEN and is not merged into this branch (`e7aa49415` is not an ancestor of `1332c0370`). Per the task direction, I am stopping here rather than locally faking a cross-process verifier view. There is no source commit or push for this pass.

After #1190 lands, the remaining findings will be addressed on that shared source: (1) exercise independent systemd/IPC stores against it; (2) filter bounded drain/lead slots at the DID-document and local-publication boundary using the same half-open expiry rule; and (3) preserve/migrate existing DID-keyed repos before changing durable publication identity. Problem B remains out of scope (#1169).

ewindisch:
VERDICT: BLOCK

# PR #1118 re-review — bounded `#atproto` overlap

Reviewed head `1332c0370cfd477215b4384aee0faba58800ac8d` against merge base
`89befb09cfe8bb77eb521036fb4c68ba791b7c91`. The PR worktree remained clean
and read-only.

## Findings

### 1. BLOCKER — the supported systemd/IPC deployment still has no shared ES256 authority

The new `refresh_from_disk` mechanism only refreshes from the directory already
bound inside that process (`auth/key_rotation.rs:1439-1461`). It does not make
OAuth, Registry, and Discovery share a directory or an authority.

The installed systemd units still set `HYPRSTREAM__SECRETS__PATH=%d`, the
private credential directory for each unit
(`hyprstream-service/src/service/manager/units.rs:162-228`). None of the
node-, service-, OAuth-, or policy-credential lists includes
`es256-signing-key.{active,drain,lead}` or their metadata
(`units.rs:28-92`). This PR does not modify the systemd/service-manager path.
Registry and Discovery each call `resolve_secrets_dir()` and initialize their
process-local global ES256 store from that directory
(`services/factories.rs:710-712,1839-1842`), just as OAuth does.

When a unit's private directory has no active ES256 slot,
`load_or_init_es256_key_store` generates a random active key and retains it in
memory even when persistence fails (`auth/key_rotation.rs:1588-1597`).
Therefore OAuth can publish/rotate A, Registry can sign with B, and Discovery
can verify against C. Refreshing each cache from its own `%d` cannot converge
them. The new "IPC" test uses one ordinary temp directory and reloads it; the
production-path test uses one in-memory store. Neither instantiates independent
unit credential directories.

The author-selected solution is explicitly #1190's sealed op-log seam. PR
#1190 remains OPEN at `e7aa49415` and is not an ancestor of this head. The
author's latest #1118 comment likewise says this branch is blocked on #1190 and
stopped without another source commit. Thus the prior production IPC blocker
is genuinely still open, and the overlap is not reachable under the supported
deployment.

### 2. HIGH — expired drain/lead keys remain in served DID documents until a poll

Verifier admission is correctly half-open (`nbf <= now < exp`) at
`hyprstream-pds/src/commit.rs:308-318`, and rotation cleanup deletes at
`now >= exp`. Publication does not apply that boundary.

Both root and per-user DID handlers call `slots_snapshot()` and publish every
returned drain/lead without filtering against the request time
(`services/oauth/did_document.rs:564-582,735-755`).
`slots_snapshot()` only reloads files; it does not remove expired slots
(`auth/key_rotation.rs:1482-1487`). Physical removal is performed by the
background rotation task, whose first tick is skipped and whose later ticks
are merely capped at 60 seconds (`auth/key_rotation.rs:1324-1353`).
Consequently an expired retired key remains in the live root and user DID
documents for up to the next tick (or indefinitely if that task stops while
the HTTP service remains live).

`es256_drain_is_removed_at_its_published_expiry` manually calls the rotation
function at the deadline; it does not request a DID document between expiry
and the next poll. The producer test constructs a document from hand-supplied
slots and likewise has no current-time publication assertion. The PR-body
claim that bounded slots "are removed ... at their expiry" is therefore
stronger than the implementation.

The overlap is bounded at the Hyprstream verifier, but the drain slot does not
actually leave the served document at the promised boundary.

### 3. HIGH — legitimate foreign placement/discovery ingest remains broken

The local resolver returns bounded keys only when `did == local_did`; every
foreign DID returns `Ok(None)` (`services/discovery.rs:1440-1449`).
`PlacementIndex::ingest_did` treats that result as an error and clears the
DID's contribution (`hyprstream-discovery/src/placement_index.rs:175-192`).

This is on the production heartbeat path: an admitted node liveness report
lazily calls `ingest_did` for the reported candidate DID
(`hyprstream-discovery/src/service.rs:4870-4875,4919-4937`). Liveness remains
recorded, but the foreign node's placement resources, labels, and group facts
never enter the directory. The code and test explicitly acknowledge that
foreign resolution is unsupported.

The prior outage is only partially closed for a fresh local, same-store
fixture. The positive test named
`local_did_web_authority_admits_pre_rotation_commit_through_drain` uses the
suite-wide `DID = "did:key:zTestNode"` and one in-memory store
(`services/discovery.rs:1500,1932-1985`); it does not exercise a real did:web
document, an independent worker, or the foreign production path. Per the
review rule, changing missing-key ingest to fail closed is not a complete fix
when it denies this legitimate cross-node path.

### 4. MEDIUM — legacy single-key compatibility paths bypass the bounded type

`PublishedAtprotoKeys::single` wraps a bare key as an unbounded active slot
(`hyprstream-pds/src/commit.rs:329-337`). The `RecordResolver` default keeps
old implementations source-compatible by converting
`resolve_verifying_key()` through that constructor
(`hyprstream-discovery/src/service.rs:550-575`). The public
`atproto_verifying_key_from_did_document` is explicitly a compatibility helper
that discards bounded overlap slots (`hyprstream-pds/src/commit.rs:495-507`).

No production call at this head uses those helpers, so this is not the IPC
blocker. It is still the exact greenfield compatibility escape the repository
rule disallows: an old/cached single-key resolver compiles unchanged and can
represent retired K as live forever instead of being forced to supply the
bounded publication set.

## What is fixed / not a finding

- The bounded parser requires one active method, explicit non-empty bounds on
  drain/lead, full DID/controller/type authority, and rejects duplicates.
- Hyprstream verifier admission and cleanup now use the same exclusive expiry
  comparison.
- New commits select only the active key; the active-not-lead test passes.
- `Ok(None)` no longer skips placement authentication. The remaining problem
  is the legitimate path it denies.
- The PR correctly states that position-aware validity and compromise
  containment are out of scope under #1169.
- Classical P-256 is confined to the atproto interoperability surface; no
  internal PQ-hybrid downgrade was found.
- I am not carrying forward the prior did:key data-migration request: the
  current operator ruling says this repository is greenfield and unpublished,
  so legacy migration is not required. That does not cure the fresh systemd
  topology failure above.

## What the change breaks

- In installed split-process deployments, the DID publisher, commit writer,
  and placement verifier can use three different ES256 authorities.
- Foreign admitted nodes keep liveness but lose placement/resources/groups
  because their repo ingest always fails key resolution.
- Root and hosted-user DID documents can advertise a retired overlap key after
  its stated expiry.
- Old single-key resolver implementations silently retain no-overlap,
  potentially unbounded behavior through compatibility defaults.

## Validation

All Cargo invocations ran through `buildq -- cargo ...` with the repository's
cached CUDA 13.0 libtorch environment.

- `cargo test -p hyprstream-pds --lib -- --nocapture` — **PASS**:
  209 passed, 0 failed, 0 ignored.
- `cargo test -p hyprstream --lib es256_ -- --nocapture` — **PASS**:
  12 passed, including promotion, startup cleanup, exact-expiry cleanup, and
  the one-directory IPC fixture.
- Focused `hyprstream` tests — **PASS**:
  `local_did_web_authority_admits_pre_rotation_commit_through_drain`,
  `publisher_signs_new_commits_with_active_not_lead_key`, and
  `build_did_doc_publishes_bounded_overlap_slots`.
- Negative control in an isolated archive of this exact head: changed
  `PublishedAtprotoKey::is_live_at` from `now < exp` back to `now <= exp`.
  `commit::tests::overlap_slot_expires_and_is_not_consulted` then **FAILED**
  at the exact-expiry assertion. The hunk was restored, and the archive's
  `commit.rs` SHA-256 again matched the reviewed source.
- `git diff --check origin/main...HEAD` — **PASS**.

## Required before re-review

Land and consume the single cross-process authoritative rotation source chosen
in #1190 (or keep this PR honestly blocked on it), then test independent
systemd/IPC workers. Filter bounded slots at both DID publication handlers
using the same half-open boundary. Restore verified foreign placement ingest
instead of treating its outage as a safer fallback. Remove the greenfield
single-key compatibility escapes so bounded authority is required by the
trait/API.

ewindisch:
Fixed the BROKEN re-sign-on-rotation path in 352efb004.

- Restored the ES256 promotion notification and wired rotate_es256_keys() to invoke the installed hook with K′ while the store write guard still hides the candidate active key.
- Hook failures now roll the durable active/drain generation and repo head back, leaving the lead queued for retry.
- Reconciled the contradictory test: an installed failing hook must be observed and keep promotion pending; the explicit no-hook IPC overlap test remains.
- rotation_resigns_stored_head_with_new_key now passes and proves the stored signature changes and verifies under K′.

Validation:
- buildq -- cargo test -p hyprstream rotation --lib -- --nocapture: 44 passed, 0 failed
- pre-commit buildq workspace release Clippy with -D warnings: passed

ewindisch:
VERDICT: PASS (352efb004 re-sign fix) — rotate_es256_keys() now invokes the promotion hook before publishing K′, the stored head is re-signed under K′, the contradictory hook test is reconciled as rollback/retry coverage, and the buildq rotation suite passes 44/44.

Issue #698, opened 2026-07-03
feat(mac): subject clearance provenance — an issuer path that sets Claims.clearance (MAC activation prereq)


ewindisch:
**Epic #547 · activation prerequisite (greenfield).** Surfaced by the independent enforcement-activation gap survey (2026-07-03).

## The gap
The MAC subject-context machinery is complete and dormant: `Claims.clearance: Option<SecurityLabel>` exists (`hyprstream-rpc/src/auth/claims.rs`), `SecurityContext::from_clearance` derives the two-input context (clearance × verified key material, assurance clamped down), and `ClaimsSubjectContextResolver` turns a `Claims` into a `SecurityContext`. But:

- **No production issuer ever calls `Claims::with_clearance`** — every call is in tests (`token_exchange.rs`, `svc.rs` test modules). No mint path (authorization_code / refresh_token / device_code / jwt-bearer / token-exchange in `services/oauth/token.rs`) attaches a clearance.
- The production dispatch wires `DenyUnlabeledResolver` (deny-all), not `ClaimsSubjectContextResolver` (`services/oauth/token.rs:143`).

Net: even if the resolver were wired, it resolves to `None` (deny) because nothing sets `clearance` upstream. This is one of the four things blocking live MAC enforcement.

## Why it needs a decision, not just wiring
**Deciding which DID gets which clearance is authority-owned security policy** (epic #547 principle #5: "the grant decision is authority-owned; no principal authors the function that grants its own request"). A clearance-issuance mechanism must NOT be invented ad-hoc. Options to choose among:
- **Enrollment-time assignment** — an admin/UCAN-authored mapping (DID → clearance) applied at token-issuance, re-attested on a TTL (matches the ceiling model).
- **Casbin/policy-derived** — clearance as a function of the subject's roles/attributes in the existing policy store.
- **Static node/service labels** — services get schema-annotated clearances (S3 baseline); users get enrollment-assigned.

## Scope
1. Choose the clearance-assignment authority + mechanism (the decision above).
2. Wire the chosen issuer path to call `Claims::with_clearance` at mint time (for the paths that should carry MAC clearance).
3. Swap `DenyUnlabeledResolver` → `ClaimsSubjectContextResolver` at the production dispatch (`token.rs:143`) once (2) exists — never before (fail-closed until real clearances flow).
4. Assurance stays derived from verified key material (already enforced, #548) — this ticket sets only the level+compartments half.

## Acceptance
[...]

PR #1139, opened 2026-07-22
fix(xet): close reference-face gaps

### LINKED ISSUE #1134: xet: close the gaps between XetService and the OAI reference face
## Problem

`XetService` (epic #654) is the reference "thin HTTP face" alongside `OAIService`, but it currently sits a rung below OAI on several axes. Each item below is independently shippable; together they make Xet a correct template for future faces (an out-of-tree execution-DAG service is being modelled on this pattern, so the divergences propagate).

## Items

**1. Registers an RPC endpoint it never serves.**
`xet.rs:394-400` registers a `SocketKind::Rep` transport into the `EndpointRegistry`, but `run()` only calls `serve_app` — nothing ever listens on it. `OAIService` has the same defect (`oai.rs:107-113`). `OAuthService` does it correctly, spawning `serve_bridged` on the Rep transport alongside the HTTP server (`oauth/mod.rs:864-874`). Either serve it or stop claiming it — a registry entry nothing answers is a liveness lie.

**2. Hand-rolled auth middleware that has already drifted from OAI's.**
`xet_auth_middleware` (`xet.rs:151-176`) takes only `jwt_verifying_key` + `audience`. It therefore lacks everything `server/middleware.rs:52` provides: DPoP sender-binding (`cnf.jkt`) and proof/jti replay checks, JTI revocation via the shared blocklist, federated-issuer resolution, rotation-slot and composite ML-DSA-65 keys, per-subject rate limiting. This gap widens every time the OAI path gains a feature. Either reuse `auth_middleware`/`ServerState`, or extract the shared core so there is one implementation.

**3. Authenticated floor, not an authorization decision.**
`xet.rs:207` passes `AllowAllCasAuthorizer` — a deliberate downgrade from `CasMount::new`'s `DenyAllCasAuthorizer` default (`storage/cas/mount.rs:41-55`). By contrast OAI makes a real AVC call per request: `policy_client.check(PolicyCheck{subject, resource: "model:X", operation: "infer"})` → 403 (`routes/openai.rs:307-330`). Xet cannot reach the equivalent until object provenance labels (#699) and subject clearances (#698) land — this issue tracks making the ratchet explicit and closing it when they do (plane #1 of #1091 R4b).

**4. No discovery surface for the HTTP face.**
OAI serves `/.well-known/oauth-protected-resource` (RFC 9728, `server/mod.rs:51`); Xet serves no well-known route, so a client cannot discover its issuer or audience. Note there is no `SocketKind::Http` (`hyprstream-rpc/src/registry/mod.rs:59` — `Req | Rep | Quic` only), so HTTP faces are not announceable through DiscoveryService at all; RFC 9728 metadata is the substitute and Xet should serve it.

**5. `XetConfig.enabled` is dead config.**
The factory never reads it (`factories.rs:1182-1216`); the service runs iff `"xet"` appears in `[services] startup`, which the default list omits (`config/mod.rs:1629`). Two contradictory enable switches — remove one.

**6. Audience derivation is a silent-failure trap.**
`resource_url()` (`config/mod.rs:670-687`) derives `http` vs `https` from the global `[tls] enabled` flag and rewrites `0.0.0.0`→`localhost`. Any mismatch between the issuer's `aud` and this string is a total auth failure with no useful diagnostic. Worth a startup-time assertion or a logged effective-audience line.

## Not in scope

The 501 write routes (`/v1/xorbs`, `/v1/shards`, `/v1/reconstructions`) are deliberately unimplemented pending the xorb-hash → `grantRepo` mapping — "the hash is not a capability" (`xet.rs:34-44`). That is correct behaviour and should stay 501 until entitlement can be enforced.

ewindisch:
Fixes #1134.

## Implemented

- I1: removed the unserved REP registration. Xet has no RequestService or RPC control handler, so advertising a REP endpoint was a liveness lie. The same strategy generalizes to OAI, which retains the identical defect outside this PR.
- I2: extracted ResourceAuthState and made OAI plus Xet use one auth/rate-limit middleware implementation. Xet now inherits DPoP sender binding and replay checks, federated issuer resolution, Ed25519 rotation slots, composite ML-DSA-65 keys, the shared JTI revocation blocklist, strict audience validation, and per-subject rate limiting.
- I4: added the public RFC 9728 /.well-known/oauth-protected-resource metadata route with Xet resource and issuer values.
- I5: removed XetConfig.enabled; services.startup is now the single enablement switch.
- I6: logs the effective audience and issuer once at startup and warns when the served transport scheme disagrees with the configured audience scheme.

## Deferred by design

- I3 remains blocked on provenance labels (#699) and subject clearances (#698). The existing BootstrapCasAuthorizer ratchet documentation is preserved; no fake AVC was added.
- The write routes remain 501 until entitlement can be enforced; a hash is not treated as a capability.
- OAI REP registration was not changed, per issue scope.

## Tests and validation

- cargo build -p hyprstream: pass
- cargo test -p hyprstream --lib services::xet: pass (16 tests)
- cargo test -p hyprstream --lib server::middleware: pass (22 tests)
- cargo clippy -p hyprstream -- -D warnings: pass
- foreground pre-commit release Clippy hook: pass
- cargo fmt --check: baseline failure under Rust 1.97 across untouched workspace files, beginning in crates/cas-serve/src/store.rs and crates/chat-core/src/capnp.rs; git diff --check passes.

### DIFF
--- a/crates/hyprstream/src/config/mod.rs
+++ b/crates/hyprstream/src/config/mod.rs
@@ -620,19 +620,14 @@ fn default_oai_host() -> String { "0.0.0.0".to_owned() }
 fn default_oai_port() -> u16 { 6789 }
 fn default_oai_timeout() -> u64 { 300 }
 
-/// XetService configuration — dual-stack HuggingFace-XET CAS HTTP face.
+/// XetService configuration — HuggingFace-XET CAS HTTP face.
 ///
-/// When enabled, exposes the HF-XET CAS wire routes (`/get_xorb/{hash}/`,
+/// When listed in `[services] startup`, exposes the HF-XET CAS wire routes (`/get_xorb/{hash}/`,
 /// `/v1/reconstructions/{hash}`, `/v1/chunks/{key}`, `/v1/xorbs/{key}`,
 /// `/v1/shards`) so a standard xet-enabled git repo can point its CAS endpoint
 /// at hyprstream as an alternative XET backend to HuggingFace. See epic #654.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct XetConfig {
-    /// Whether the XetService is served. Disabled by default: this is a
-    /// foundation surface (most routes 501 pending interop verification).
-    #[serde(default)]
-    pub enabled: bool,
-
     /// Host address for the HTTP server.
     #[serde(default = "default_xet_host")]
     pub host: String,
@@ -658,7 +653,6 @@ pub struct XetConfig {
 impl Default for XetConfig {
     fn default() -> Self {
         Self {
-            enabled: false,
             host: default_xet_host(),
             port: default_xet_port(),
             external_url: None,
--- a/crates/hyprstream/src/server/middleware.rs
+++ b/crates/hyprstream/src/server/middleware.rs
@@ -1,7 +1,7 @@
 //! Middleware for authentication, logging, and request processing
 
 use crate::auth::jwt;
-use crate::server::state::ServerState;
+use crate::server::state::{ResourceAuthState, ServerState};
 use axum::{
     extract::{Request, State},
     http::{header, HeaderValue, StatusCode},
@@ -51,7 +51,7 @@ impl std::fmt::Debug for AuthenticatedUser {
 ///
 /// On success, inserts `AuthenticatedUser` into request extensions.
 pub async fn auth_middleware(
-    State(state): State<ServerState>,
+    State(state): State<ResourceAuthState>,
     mut request: Request,
     next: Next,
 ) -> Response {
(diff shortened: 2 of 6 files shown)


ewindisch:
**Independent review (kimi K3 high): PASS.**

> "No blocking or revision-worthy findings. The PR does what it claims, holds the auth bar, and stays in scope."

Verified against the #1134 items:
- **I1** — the unserved `SocketKind::Rep` registration is removed and test-asserted; OAI deliberately untouched (it has the same defect but was out of scope).
- **I3** — `BootstrapCasAuthorizer` default-deny retained; **no fake AVC decision** was introduced. This was the item I most wanted checked, since faking it would have looked like progress while removing the ratchet.
- **I4/I5/I6** correct; the 501 write routes remain 501 ("the hash is not a capability").
- Validation reproduced independently: build pass, xet tests 16/16, middleware tests 22/22, `clippy -D warnings` pass.

Three non-blocking nits recorded, notably **OAI's silent private-blocklist fallback** — pre-existing, not introduced here, and a good follow-up candidate.

Full verdict on disk at `.fleet-coord/rev1139/verdict-kimi.md`. Not merged — human approval only.

Issue #1140, opened 2026-07-22
chore: cargo fmt --check fails on a clean main checkout (pre-existing baseline drift under Rust 1.97)


ewindisch:
## Problem

`cargo fmt --check` fails on `main` across **untouched** workspace files under Rust 1.97, starting in `crates/cas-serve/src/store.rs` and `crates/chat-core/src/capnp.rs`. `git diff --check` passes, so this is pre-existing formatting drift, not anything a current branch introduced.

## Why it matters more than it looks

Every agent and contributor is told to run `cargo fmt --check` as a gate. Because it fails on a clean checkout, everyone hits it, has to determine it's a baseline failure rather than their own, and then reports "fmt fails but not because of me." That is happening repeatedly right now across parallel work — it costs a verification step per branch and trains people to ignore a real gate.

It also means **a genuine formatting regression would be invisible**: nobody can distinguish new drift from the existing baseline.

## Fix

One mechanical `cargo fmt` pass over the workspace, landed on its own so it never mixes with substantive review. Then keep it green.

Worth confirming while doing it: whether the drift came from a toolchain bump (1.97 changing formatting of existing code) or from commits landing without `fmt`. If the former, pin the formatting toolchain (`rust-toolchain.toml`) so it can't silently re-drift on the next bump.

## Acceptance

`cargo fmt --check` exits 0 on a clean `main` checkout.

Issue #1136, opened 2026-07-22
design: DID-anchored trust bootstrap — replace the /etc+/run pin with did:at9p + did:web (unblocks k8s, remote workers, and the metal stack)


ewindisch:
> **Design proposal against #1135.** The spike on #1135 concludes we should support out-of-tree services (v0 pattern → v1 credentials → v2 SDK). This issue proposes the **mechanism for the credential/trust half** — and the same mechanism unblocks the k8s epic and auto-registering workers. It is mostly *wiring components that already exist and are already ratified*.

## Problem: the current trust root is undeployable

Production discovery pins its trust root to two hardcoded paths (`hyprstream-discovery/src/service.rs:1497-1499`):

```
/etc/hyprstream/trust/deployment-ca.ed25519
/run/hyprstream/credentials/registry-service.jwt
```

`read_os_owned_file` (`service.rs:1869-1899`) requires both — **and every ancestor directory up to `/`** — to be a regular file (not a symlink), `uid == 0`, and not group/world-writable. Anything else fails closed.

Consequences, each verified:

- **The metal Quadlet stack cannot boot.** It runs `service start <svc> --foreground --ipc` as user `rocky` and provisions *neither* file (grep across `stacks/hyprstream/` → zero hits), while `install_process_production_resolver` runs unconditionally before subcommand dispatch (`bin/main.rs:1893-1902`; only `wizard`, `mac genesis`, `pds join` dispatch earlier). Tracked separately.
- **Kubernetes projected Secret volumes fail outright** — they are symlink farms (`..data/<key>` → `..2024_xx/<key>`), and the check uses `symlink_metadata` + requires `is_file()`. The ancestor-ownership walk is also hostile to CSI-mounted paths.
- **This directly conflicts with #790**, which insists *"the pod's identity attaches — never a node-scoped or plugin-scoped ambient credential."* A fixed `/etc` + `/run` file pair **is** a node-scoped ambient credential.
- **systemd `LoadCredential` is explicitly excluded** by design: *"The executable does not consult `CREDENTIALS_DIRECTORY` … ambient credential-directory fallback is not supported"* (`docs/deployment-registry-trust.md`). So the obvious workaround is foreclosed.
- **Auto-registering workers are impossible by construction** — the trust root must be provisioned out of band, which an ephemeral node has no channel for.
[...]

Issue #790, opened 2026-07-04
K3b: CSI node plugin csi.hyprstream.io — pods declaratively mount MAC-enforced hyprstream 9P volumes


ewindisch:
Part of #778 (Phase 1). The filesystem axis made K8s-native: **any stock pod declares a hyprstream VFS volume and inherits the security spine underneath (WIT → ticket → Tattach → MAC floor → PQC channel) without knowing it.**

**K3b — a CSI driver (`csi.hyprstream.io`) whose Node service mounts a tenant VFS subtree into the pod.**

### Deliverables
- CSI Node plugin (DaemonSet, chart-installed via K1): `NodePublishVolume` performs the mount; `volumeAttributes` carry `{namespacePath, tenant, plane}`; ephemeral inline volumes first (the natural fit — per-pod, per-identity), PV/PVC static provisioning second.
- **Mount mechanism — reuse, don't invent:** FUSE→9P client machinery from #766/#751 (the kata-guest POSIX mount path) as the default; kernel v9fs (`trans=tcp`/fd, #762) as the alternative where the node kernel supports it. The node plugin is the mount *executor*; the 9P planes (#764/#765) are the wire.
- **Credential flow:** pod's projected SA token (CSI `podInfoOnMount` + token requests) → WIT (K2a) → mount ticket (K3a) → `Tattach`. The *pod's* identity attaches — never a node-scoped or plugin-scoped ambient credential (that would bypass the per-workload MAC clearance).
- Liveness: reconnect/backoff on plane disruption; a broken session surfaces as EIO, never silently reattaches as a different subject.
- MAC posture: fail-closed today (deny-floor until #568/#699); the CSI path *is* the "K8s workloads benefit from our secure designs" delivery vehicle once those activate.

### Acceptance
- On kind: a vanilla pod with an ephemeral `csi.hyprstream.io` volume reads/writes its tenant subtree; a pod without a valid TenantBinding-scoped identity mounts nothing (fail-closed); node plugin restart doesn't leak stale attaches.

### Depends on / consumes
- K3a (#786) tickets; K2a (#780) exchange; #764/#771 + #765/#773 (planes); #766/#751 (FUSE→9P client); #762 (kernel v9fs option); K1 (#779) for chart install.
- Consumed by: K4b (backend delivery), any tenant workload.

ewindisch:
## Design gate for implementation

Security spine PR #862 now defines the ticket API consumed here:

- `POST /oauth/mount-ticket`
[...]

Issue #780, opened 2026-07-04
K2a: SPIFFE/WIT identity bridge (consume) — cluster/SPIRE JWKS as federated issuer, SA-token→WIT exchange


ewindisch:
Part of #778 (Phase 0). The **consume** half of SPIFFE compatibility — the entry point of the epic's security spine.

**K2a — accept Kubernetes-native workload identities and re-encode them as WITs.**

A pod presents its projected ServiceAccount token (or a SPIRE JWT-SVID); hyprstream validates it against the cluster's OIDC discovery/JWKS (kube-apiserver is an OIDC issuer; SPIRE serves a bundle endpoint) and exchanges it for a short-lived **WIT** (`wit+jwt`) — the credential every downstream surface (9P mount tickets K3a, RPC envelopes, MAC clearance derivation) already understands.

### Deliverables
- Register a cluster/SPIRE issuer as a **federated trust domain**, gated by the existing `federation:register` Casbin policy (one trust decision, multiple protocols — per `docs/spiffe.md`).
- Token exchange endpoint: SA-token/JWT-SVID → WIT, built on the existing OAuth machinery (`services/oauth/jwt_bearer.rs` JWT-bearer grant / RFC 8693-style exchange; sender-binding via the S6 `mac::exchange` DPoP model where the caller can prove a key).
- Claims mapping: `sub` (`system:serviceaccount:<ns>:<sa>` or `spiffe://<domain>/<path>`) → hyprstream identity; namespace→tenant mapping is **not** decided here (that's `TenantBinding`, K5a) — until then, exchanged WITs carry issuer+subject only, and MAC clearance stays floor/deny (fail-closed, consistent with #568/#699 not yet active).
- Audit every exchange (S7 model).

### Acceptance
- A pod in an allowed cluster can exchange its projected SA token for a WIT; a pod from an unregistered issuer is rejected at the `federation:register` gate; tests cover expiry/audience/issuer mismatch fail-closed.

### Depends on / consumes
- Consumes: `federation:register` (shipped), `jwt_bearer.rs`, S6 token-exchange (#572, landed), S8 PQ posture (#574).
[...]

PR #1142, opened 2026-07-22
fix(auth): make service secret profiles explicit

### LINKED ISSUE #802: systemd unit generation: missing per-service signing key silently falls back to ImportCredential of the CA root key

### LINKED ISSUE #805: Replace resolve_service_signing_key's systemd_mode bool with an explicit SecretsProfile (deployment-profile enum) before K1a consumes it
## Summary

`resolve_service_signing_key(secrets_dir, service_name, systemd_mode)` (PR #774) conflates three orthogonal decisions into one `bool`, and the bool itself is detected by an overloadable heuristic. Before K1a (#783) consumes this model for the Helm chart — as it explicitly plans to — the bool should become an explicit deployment profile.

## The three conflated decisions

1. **Layout** — flat `{secrets_dir}/signing-key` vs per-service `{secrets_dir}/{service}/signing-key`.
2. **Provisioning posture** — may-generate-on-missing (standalone, writable shared dir) vs must-exist-fail-closed (systemd read-only `%d`, pre-populated by bootstrap; k8s read-only Secret/CSI mount, pre-populated by a provisioning Job).
3. **The `policy == root key` identity rule** — a topology fact baked in as a string compare.

## Why the bool breaks

- **Detection is spoofable-by-accident**: `bin/main.rs:2024` infers `systemd_mode` from `HYPRSTREAM__SECRETS__PATH` being set — but that env var is *also* the general "relocate secrets dir" config knob. An operator (or the k8s chart, which would naturally set exactly this var) pointing it at a **shared** dir flips every service into the flat branch → every service loads the **root/CA private key** as its own. systemd's own `CREDENTIALS_DIRECTORY` export (or explicit config) is the honest signal.
- **K8s is a third profile, not a variant of either**: per-pod read-only mount, flat filename (systemd-like layout) but chart/Job provisioning semantics and rotation = re-mount-not-rewrite. The SPIFFE tickets (#780/#785) add a fourth: pods with *no* mounted signing key at all (projected SA token → WIT exchange), where the resolver's question isn't "which file" but "file or exchange".

## Proposed shape

```rust
enum SecretsProfile {
    Standalone,          // shared writable dir, per-service subdirs, generate-on-missing OK
    SystemdCredentials,  // per-unit %d, flat name, read-only, bootstrap-prepopulated, fail-closed
    ProvisionedReadOnly, // k8s Secret/CSI mount: flat name, read-only, MUST exist, rotation=re-mount
    TokenExchange,       // no key file; identity via SA-token→WIT exchange (#780)
}
```

selected by explicit config (unit generator / chart sets it), with the heuristic kept only as a deprecated fallback. The `policy`-is-root-key rule moves into the profile/bootstrap contract rather than a string compare in a shared resolver. Pair with the startup consistency check from the PR #774 review (`own_key.verifying_key() == bootstrap_pubkeys[name]` → hard error).

## Related

PR #774 (introduces the bool), #759, #778 (k8s epic), #783 (K1a — blocks-ish: K1a says "consume the #759/PR #774 resolution model rather than forking it", which risks freezing the bool in), #780/#785 (TokenExchange profile), #802 (install-time twin of the shared-dir hazard), #804 (resolver unification).

ewindisch:
## Summary

- replace the systemd-mode boolean with `SecretsProfile::{SharedDirectory, PerServiceScoped}` so credential layout reflects provider scoping and can cover projected Kubernetes Secret/CSI volumes
- preserve the policy service root/CA-key special case and all existing standalone/scoped resolution behavior
- fail systemd unit generation closed when a non-policy service signing key is missing, preventing fallback to the flat node/CA root credential
- add regression coverage for both profile layouts and the missing-key privilege-escalation path

## Validation

- `cargo build -p hyprstream`
- `cargo test -p hyprstream --lib auth::` (332 passed)
- `cargo test -p hyprstream-service missing_service_signing_key_never_falls_back_to_node_key --lib`
- `cargo clippy -p hyprstream -- -D warnings`
- `cargo clippy -p hyprstream-service --all-targets -- -D warnings`
- foreground pre-commit Clippy hook (release profile)
- `cargo fmt --check` remains red repository-wide on pre-existing formatting drift beginning in untouched `crates/cas-serve/src/store.rs`; all changed hunks match rustfmt and `git diff --check` passes

Closes #805
Closes #802

### DIFF
--- a/crates/hyprstream-service/src/service/manager/units.rs
+++ b/crates/hyprstream-service/src/service/manager/units.rs
@@ -91,6 +91,56 @@ pub const ALL_CREDENTIAL_NAMES: &[&str] = &[
     "service-jwt",
 ];
 
+/// Build credential directives for a user service unit.
+///
+/// A non-policy service must have its own plaintext signing key available for
+/// `LoadCredential=`. Falling through to `ImportCredential=signing-key` would
+/// import the node/CA root key from the flat credstore into that service.
+fn credential_directives(
+    service: &str,
+    names: &[&str],
+    credstore: &std::path::Path,
+    plain_creds: &std::path::Path,
+) -> Result<String> {
+    let mut import_lines = String::new();
+
+    for name in names {
+        // Per-service credentials (signing-key, service-jwt): each service
+        // has its own independent Ed25519 key stored in a subdirectory.
+        //
+        // ImportCredential requires the credstore file name to match the
+        // runtime name, but per-service creds are encrypted with prefixed
+        // names (e.g. model-signing-key). Use LoadCredential to load from
+        // the plaintext file directly, making it available as the unprefixed
+        // name the runtime expects.
+        //
+        // Policy is special: its "signing-key" IS the CA key, stored flat
+        // in the encrypted credstore, so ImportCredential works for it.
+        if SERVICE_CREDENTIAL_NAMES.contains(name) && service != "policy" {
+            let plain_path = plain_creds.join(service).join(name);
+            if plain_path.is_file() {
+                import_lines.push_str(&format!(
+                    "LoadCredential={name}:{}\n",
+                    plain_path.display()
+                ));
+                continue;
+            }
+            if *name == "signing-key" {
+                anyhow::bail!(
+                    "per-service signing key '{}' is missing; refusing to import the node/CA root key as service '{}'. Re-run 'hyprstream service install' after provisioning service credentials",
+                    plain_path.display(),
+                    service
+                );
+            }
+        }
+        if credstore.join(name).exists() {
+            import_lines.push_str(&format!("ImportCredential={name}\n"));
+        }
+    }
+
+    Ok(import_lines)
+}
+
 /// Generate a systemd service unit for a service.
 ///
 /// # Arguments
@@ -184,37 +234,7 @@ pub fn service_unit(service: &str, use_systemd_creds: bool, depends_on: &[&str])
             .map(|d| d.join("hyprstream").join("credentials"))
             .unwrap_or_default();
 
-        let mut import_lines = String::new();
-
-        for name in &names {
-            // Per-service credentials (signing-key, service-jwt): each service
-            // has its own independent Ed25519 key stored in a subdirectory.
-            //
-            // ImportCredential requires the credstore file name to match the
-            // runtime name, but per-service creds are encrypted with prefixed
-            // names (e.g. model-signing-key).  Use LoadCredential to load from
-            // the plaintext file directly, making it available as the unprefixed
-            // name the runtime expects.
-            //
-            // Policy is special: its "signing-key" IS the CA key, stored flat
-            // in the encrypted credstore, so ImportCredential works for it.
-            if SERVICE_CREDENTIAL_NAMES.contains(name) && service != "policy" {
-                let plain_path = plain_creds.join(service).join(name);
-                if plain_path.exists() {
-                    import_lines.push_str(&format!(
-                        "LoadCredential={name}:{}\n",
-                        plain_path.display()
-                    ));
-                    continue;
-                }
-                // Fall through: no per-service plaintext file, try encrypted credstore
-            }
-            if credstore.join(name).exists() {
-                import_lines.push_str(&format!("ImportCredential={name}\n"));
-            }
-        }
-
-        let all_cred_lines = import_lines;
+        let all_cred_lines = credential_directives(service, &names, &credstore, &plain_creds)?;
 
         if all_cred_lines.is_empty() {
             String::new()
@@ -373,4 +393,36 @@ WantedBy=multi-user.target
 "#,
         exec = exec.display(),
     ))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn missing_service_signing_key_never_falls_back_to_node_key() -> Result<()> {
+        let temp = tempfile::TempDir::new()?;
+        let credstore = temp.path().join("credstore.encrypted");
+        let plain_creds = temp.path().join("credentials");
+        std::fs::create_dir_all(&credstore)?;
+        std::fs::create_dir_all(plain_creds.join("model"))?;
+
+        // This is the dangerous fallback credential: the node/CA root key is
+        // present under the runtime name expected by the service.
+        std::fs::write(credstore.join("signing-key"), b"encrypted-root-key")?;
+
+        let error = match credential_directives(
+            "model",
+            SERVICE_CREDENTIAL_NAMES,
+            &credstore,
+            &plain_creds,
+        ) {
+            Err(error) => error,
+            Ok(lines) => anyhow::bail!("dangerous signing-key fallback returned: {lines}"),
+        };
+        let message = error.to_string();
+        assert!(message.contains("per-service signing key"));
+        assert!(message.contains("refusing to import the node/CA root key"));
+        Ok(())
+    }
 }
--- a/crates/hyprstream-service/src/service/manager/units.rs
+++ b/crates/hyprstream-service/src/service/manager/units.rs
@@ -91,11 +91,14 @@ pub const ALL_CREDENTIAL_NAMES: &[&str] = &[
     "service-jwt",
 ];
 
-/// Build credential directives for a user service unit.
+/// Build credential directives for a service unit.
 ///
-/// A non-policy service must have its own plaintext signing key available for
-/// `LoadCredential=`. Falling through to `ImportCredential=signing-key` would
-/// import the node/CA root key from the flat credstore into that service.
+/// This is the single, path-independent fail-closed guard shared by both the
+/// user-unit (`service_unit`) and system-unit (`system_service_unit`) emission
+/// paths. A non-policy service must have its own plaintext signing key
+/// available for `LoadCredential=`. Falling through to
+/// `ImportCredential=signing-key` would import the node/CA root key from the
+/// flat credstore into that service.
 fn credential_directives(
     service: &str,
     names: &[&str],
@@ -139,6 +142,40 @@ fn credential_directives(
     }
 
     Ok(import_lines)
+}
+
+/// Build the credential section for a system-wide service unit.
+///
+/// Kept separate from `system_service_unit` so the system path can be tested
+/// with injectable paths while still routing every credential through the
+/// shared [`credential_directives`] fail-closed guard.
+fn system_creds_section(
+    service: &str,
+    is_appimage: bool,
+    credstore: &std::path::Path,
+    plain_creds: &std::path::Path,
+) -> Result<String> {
+    let mut names: Vec<&'static str> = NODE_CREDENTIAL_NAMES.to_vec();
+    names.extend(SERVICE_CREDENTIAL_NAMES.iter().copied());
+    if service == "oauth" {
+        names.extend(OAUTH_CREDENTIAL_NAMES.iter().copied());
+    }
+    if service == "policy" {
+        names.extend(POLICY_CREDENTIAL_NAMES.iter().copied());
+    }
+
+    let import_lines = credential_directives(service, &names, credstore, plain_creds)?;
+
+    Ok(if import_lines.is_empty() {
+        String::new()
+    } else {
+        let private_mounts = if is_appimage {
+            ""
+        } else {
+            "\nPrivateMounts=yes"
+        };
+        format!("\n{import_lines}Environment=HYPRSTREAM__SECRETS__PATH=%d{private_mounts}")
+    })
 }
 
 /// Generate a systemd service unit for a service.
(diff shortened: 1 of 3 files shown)


ewindisch:
## Independent review — REVISE (kiro)

Full verdict: `.fleet-coord/rev1142/verdict-kiro.md`

**What passes:** the #802 fail-closed guard is real and correct — no path reaches the root key as a non-policy fallback, and the regression test fails if the fallthrough returns. The #805 refactor is behaviour-preserving and exact; all four (profile x service) combinations map identically to the old bool, and the policy -> root/CA special case survives verbatim as the first match arm. Local: build OK, 332 auth tests OK, clippy clean on both crates.

**Why REVISE:**

1. **[SECURITY, cross-PR] #1141 reintroduces #802.** https://github.com/hyprstream/hyprstream/pull/1141 adds a new credential loop to `system_service_unit` using the old vulnerable pattern (`ImportCredential=signing-key` with no `bail!`). #1142 only guards the *user-unit* path. On merge the system-unit path regains the CA-key escalation — and the two PRs collide textually on `units.rs`. **The fail-closed guard must not be path-specific.**
2. **[SECURITY] #805s actual complaint is unaddressed.** The spoofable `HYPRSTREAM__SECRETS__PATH` heuristic that #805 wanted replaced by `CREDENTIALS_DIRECTORY` is unchanged at `main.rs:2120`. Pointing that env at a shared dir still flips non-policy services to the flat root/CA key at runtime — the install-time fix does not cover standalone/manual/k8s.
3. The recommended `own_key == bootstrap_pubkeys[name]` startup interlock is absent — no runtime backstop.
4. The enum under-models #805s named futures: two layout variants serve k8s Secret/CSI (#790) but cannot express TokenExchange (#780) or DidAnchored (#1136), and the provisioning-posture axis is not modeled. Additive, so it does unblock K1a, but it will need another touch.

ewindisch:
Follow-up: the #805 residue found during this review — the spoofable `HYPRSTREAM__SECRETS__PATH` profile-selection heuristic (`main.rs:~2120`), the missing `own_key == bootstrap_pubkeys[name]` startup interlock, and `SecretsProfile` under-modeling the #780/#790/#1136 providers — is tracked in #1148 (kept out of scope of this PR and the #1141 fix). The system-unit twin of the #802 fallback (introduced by #1141's second credential loop) is fixed in #1141, which now routes both unit-emission paths through this PR's shared `credential_directives()` guard.

ewindisch:
Blocking-review residue cleared — opened #1189 (stacked on this branch as the base) to land the three out-of-scope findings as a peer:

- **F2 (the real #805):** the spoofable `HYPRSTREAM__SECRETS__PATH` heuristic is retired. `resolve_runtime_secrets_profile()` now selects the profile from non-spoofable signals only — systemd `CREDENTIALS_DIRECTORY`, then an explicit `HYPRSTREAM__SECRETS__PROFILE` / `[secrets] profile` declaration, else fail-closed. `HYPRSTREAM__SECRETS__PATH` is no longer consulted for the profile, so an operator/k8s-chart pointing it at a shared dir can no longer flip a non-policy service onto the flat node/CA root key at runtime (the case this PR's install-time guard does not see for standalone/manual/k8s).
- **F3:** runtime key-consistency interlock (`assert_key_matches_bootstrap`) — a service that boots holding a key not registered for its own name, or that clashes with a key registered for a *different* name (the root/CA-key escalation signature), refuses to start. Wired into both `--ipc` and single-process startup.
- **F4:** `SecretsProfile` split into `CredentialLayout` × `ProvisioningPosture` plus `TokenExchange` / `DidAnchored` (fail-closed until #780/#1136 resolvers land). Additive / exhaustive-match; behavior this PR established is preserved.

#1189 does not touch `units.rs` and extends (not rewrites) the enum. Regression tests are mutation-verified (each fails without its fix). The two can land together; #1189 is rebased on top of this branch so it fast-forwards cleanly once this merges.

Posting for awareness per the review — not requesting action here.

ewindisch:
# Independent Review — PR #1142

VERDICT: BLOCK

Reviewed detached head `f46ef92f32213363289ed466da5aca75db19bb6d` on 2026-07-23. The local #802 fail-closed fix is sound and mutation-tested, and the bool-to-enum refactor preserves all four existing profile/service behaviors. This PR must nevertheless remain blocked: operator-required peer PR #1189 has **not** landed, so merging #1142 alone would bank only the systemd install-time half while leaving the exploitable #805 runtime selector live.

## Findings — severest first

### 1. [BLOCKING / SECURITY / merge sequencing] #1189 has not landed

As rechecked immediately before this verdict, PR #1189 is `OPEN`, `mergedAt: null`, with no merge commit. It is stacked on `fix/805-secrets-profile` at `070ec543190688e68c191f296ce444b95e7d0cf6`.

#1142 still selects `SecretsProfile::PerServiceScoped` from the mere presence of `HYPRSTREAM__SECRETS__PATH` (`bin/main.rs:2120-2128`). That variable is also the general secrets-directory relocation knob. Setting it to a shared directory therefore selects the flat-key branch for a non-policy service and can hand that service the node/CA root signing key. This is the exact “absent/present expected value disables the intended distinction” class called out in #805: the value is not validated; any present value means scoped, while absence means shared.

#1189 owns the required peer half: retire that presence heuristic, require an honest/explicit profile signal, add the runtime key/bootstrap consistency interlock, and model provisioning posture. Per the operator’s ruling, #1142 and #1189 land as peers. Until #1189 lands, #1142 is correctly **BLOCKED**, not passed into a merge it must not take.

### 2. [SECURITY / cross-path] The #802 guard is path-independent in the current composed stack

At #1142’s own head, only the user-unit path emits credentials; `service_unit()` routes its credential list through the single `credential_directives()` guard, while `system_service_unit()` does not yet emit credentials.

I also inspected the latest #1141 head, `000556598b344fdde6a2086b72365bdc25730220`, which is a descendant of #1142. #1141’s revised system-wide emission path calls `system_creds_section()`, which calls the same `credential_directives()` helper used by the user-unit path. It also adds negative and positive system-path tests. Thus the original per-path regression is fixed in the intended composition: one guard serves both emission paths. #1141 is still open, so this routing must be preserved through its eventual rebase/merge.

### 3. [SECURITY] #802’s non-policy root-key fallback is closed at #1142

For every non-policy `signing-key` request:

- A regular per-service key emits `LoadCredential=signing-key:<service path>` and continues.
- A missing or non-file per-service key bails before the flat credstore lookup.
- The dangerous `ImportCredential=signing-key` fallback is therefore unreachable for non-policy services.
- Policy skips the non-policy branch and may still import flat `signing-key`, correctly preserving policy’s root/CA identity.

The new test is causal, not merely green: after temporarily removing only the `if *name == "signing-key" { bail!(...) }` hunk, it failed with:

`dangerous signing-key fallback returned: ImportCredential=signing-key`

After restoring the hunk, the same test passed and the source diff returned clean.

### 4. [CORRECTNESS] The #805 refactor preserves all four existing behaviors

| Profile | Service | Old bool behavior | New enum behavior |
|---|---|---|---|
| `PerServiceScoped` | policy | flat root/CA key | flat root/CA key |
| `PerServiceScoped` | non-policy | flat scoped key | flat scoped key |
| `SharedDirectory` | policy | flat root/CA key | flat root/CA key |
| `SharedDirectory` | non-policy | independent `{service}/signing-key` | independent `{service}/signing-key` |

The policy special case remains the first match alternative: `(_, "policy") | (SecretsProfile::PerServiceScoped, _)`. All five focused resolver tests passed, covering both policy profiles, both non-policy layouts, and repeat stability.

### 5. [AUDIT] “Absent expected value disables the check” scan

The new `SecretsProfile` type itself contains no `Option`, default, or missing-value bypass. The retained call-site selector does contain the risky form: `std::env::var("HYPRSTREAM__SECRETS__PATH").is_ok()`. It treats any present value—including one naming a shared directory—as proof of per-service scoping. This is not newly introduced behavior, but the refactor does not abolish it; it is the blocking residue assigned to #1189.

The #802 guard does not have this flaw: absence of the required non-policy signing key is an error, not permission to fall through. Missing `service-jwt` remains omitted as before, but cannot collide with or import the flat root signing key.

## What this change breaks

- **Intentional install-time break:** non-policy user-unit generation with `use_systemd_creds=true` now errors if `{credentials}/{service}/signing-key` is absent. Deployments that previously generated a unit before provisioning the service key must provision first or rerun `hyprstream service install`.
- **Intentional malformed-input break:** a directory or other non-regular object at the signing-key path no longer passes the old `.exists()` test; `.is_file()` rejects it.
- **Rust API source break:** downstream callers of public `resolve_service_signing_key(..., bool)` must migrate to `SecretsProfile`. Both in-tree call sites were migrated.
- **No legitimate credential-path outage found:** a review-only happy-path test confirmed a present non-policy key emits `LoadCredential` without root import, and policy still emits `ImportCredential=signing-key`. The temporary test was removed after it passed.
- **No behavior break in the four existing profile × service combinations:** focused tests confirm the matrix above.

## Validation

All Cargo invocations used `buildq -- cargo ...` as required.

- `buildq -- cargo test -p hyprstream-service missing_service_signing_key_never_falls_back_to_node_key --lib -- --nocapture`
  - PR head: **1 passed**
  - Guard hunk temporarily reverted: **failed as expected**, exposing `ImportCredential=signing-key`
  - Guard restored: **1 passed**
- Review-only happy-path unit test, run through `buildq`: **1 passed**; then removed.
- With cached CUDA 13.0 libtorch selected:
  - `buildq -- cargo test -p hyprstream --lib resolve_service_signing_key -- --nocapture`
  - **5 passed, 0 failed, 1096 filtered**
- With cached CUDA 13.0 libtorch selected:
  - `buildq -- cargo build -p hyprstream`
  - **passed**
- The first resolver-test attempt lacked `LIBTORCH` and stopped in `torch-sys`; this was environment-only. Rerunning with `/home/birdetta/projects/cyberdione/hyprstream/appimage/libtorch-cache/cuda130/libtorch` passed.
- `cargo fmt --check` was not used as a gate because clean main is known red under #1140.
- Final source status before writing this verdict: clean detached checkout; no source edits, push, or merge retained/performed.

## Disposition

Do not merge #1142 alone. Keep it blocked until #1189 lands as the operator-required peer. The #802 guard and the behavior-preserving enum refactor are acceptable components of that paired landing.

ewindisch:
Rebased #1142 onto current `origin/main` and removed the merged #1189 commits from this PR's rewritten history.

Fixed the BLOCK: `system_service_unit()` now emits its `/etc`-backed credentials through the same path-independent `credential_directives()` fail-closed guard used by user units. Added negative and positive system-unit regressions proving a missing per-service signing key cannot fall through to `ImportCredential=signing-key`, while a present key emits `LoadCredential=signing-key` without the node/CA-root fallback.

Validation:
- `buildq -- cargo test -p hyprstream-service --lib --tests --no-fail-fast` — passed (31 unit tests passed, 1 systemd-only test ignored; integration target 1/1 passed)
- pre-commit workspace Clippy gate — passed

Force-pushed with lease. Head: `129f0c23d335978f4fcd6f8c50aa0f65e73cabfc`

ewindisch:
VERDICT: APPROVE — at 129f0c23d, user and system service credentials both route through the single credential_directives() fail-closed guard; buildq hyprstream-service tests pass.

Issue #1146, opened 2026-07-22
OAuth AS conformance follow-ups from the #1121 dual review — tiered work chain (T2.1 gates #1119)


ewindisch:
Tracking issue for the fixes surfaced by the dual independent review of #1121 (gpt-5.6-sol high → BLOCK, kimi K3 high → REVISE, reconciled — see #1121 comments). Grouped by execution order, not by reviewer.

Sizes are the reviewers' estimates. "Blocks #1119" is the reconciled gate rule both reviewers now agree on.

---

## Tier 1 — must land before clients depend on the behavior (ship with or immediately after #1121)

**T1.1 — Client-metadata scope ceiling is silently dropped.** *(sol Blocker 2 = kimi finding 3 — the one thing both reviewers independently flagged.)*
`ClientIdMetadataDocument` doesn't deserialize `scope` (or `dpop_bound_access_tokens`) (`registration.rs:251`); fetched metadata converts with `scope: None` (`registration.rs:481`); PAR passes no ceiling (`par.rs:182`); scope validation reads that as permission to use the full server-supported set (`state.rs:1679`). **A stock CIMD document declaring only `atproto` can request `transition:generic`.** Also harden `resolve_client_declared_scopes` (`authorize.rs:668`) to re-resolve on cache miss rather than silently drop the ceiling. The conformance test only covers the dynamic-registration path.
Size **S–M**. Ceilings must be tight from day one — tightening later breaks deployed clients.

**T1.2 — Client-assertion audience is wrong for atproto.** *(kimi R1, overlaps sol High 4.)*
Expected assertion audience is the token endpoint URL (`token.rs:273`); the atproto OAuth profile mandates the **AS issuer**. RFC 7523 permits both, atproto does not. This currently **rejects the reference confidential client**. Fix in `client_auth.rs::check_claims` / the `expected_audience` construction in `token.rs::enforce_client_authentication`.
Size **S** (~20 lines + a handler test variant; the wrong-aud/canonical-aud harness already exists).

**T1.3 — Reject `atproto` scope at the device-authorization endpoint.** *(kimi R2.)*
One guard in `device.rs::device_authorize`. Currently mints malformed `token_type: "DPoP"` tokens with `cnf` bound to the user's login key.
Size **XS**.

## Tier 2 — blocks #1119 (the authority bridge)

**T2.1 — Bind minted authority to granted scopes.** *(sol Blocker 1 = kimi R5, added on reconciliation.)*
[...]

Issue #1145, opened 2026-07-22
audit: 'absent expected value disables the check' — three instances found today; make the unsafe spelling impossible


ewindisch:
## Pattern

Three separate places today, found by three different reviewers, all the same shape: **an expected value that is `None`/unset causes the check to be skipped entirely rather than to fail closed.**

1. **www-cyberdione-ai !48** — `verify_request` guards the verifying key but not the audience. With `EXECUTION_DAG_JWT_AUDIENCE` unset, `cfg.audience` is `None`, and `decode` skips the audience branch (`if let Some(expected) = expected_aud`). Any token minted by the same cluster CA — or with no `aud` at all — is accepted. The module's own `from_env` logs *"will fail closed (503) until an audience is configured"*; it does not.
2. **#1121** (`services/oauth/auth.rs:74`, `introspection.rs:42`) — both call `decode(..., None)`, disabling expected-audience checking on privileged OAuth routes, and neither enforces the canonical issuer.
3. **www-cyberdione-ai !47** (fixed) — `create_session`'s "fallback" re-read the same env var it compared against, so the comparison always passed and control fell through **without returning**.

`server/middleware.rs:401` already demonstrates the correct pattern: supply an exact expected resource and issuer, and fail closed when absent.

## Why this is a class, not three incidents

The unsafe shape is idiomatic and reads as correct:

```rust
if let Some(expected) = expected_aud {
    // verify
}
// ...fall through as if verified
```

`Option<T>` for an expected value silently encodes "no constraint" as "unconstrained," and Rust's ergonomics make that the natural spelling. Every instance was written by a different author and passed review at the time. Two of the three were caught only because a reviewer was explicitly told to hunt for this shape.

## Proposed work

1. **Audit** every verification call site that takes an optional expected value — `decode`, audience, issuer, `cnf`/`jkt`, nonce, `kid`, and their equivalents — and classify each as intentionally-optional or accidentally-skippable.
2. **Make the unsafe spelling hard to write.** Options to weigh: require an explicit `Audience::Any` variant so "no constraint" must be typed deliberately; or split the API into `decode_strict(expected)` and an explicitly-named `decode_unverified_audience()` that is greppable in review.
3. **Fail closed on absent configuration** where the code documents that it will (both !48 and #1121 promise this in comments/logs while doing the opposite).
4. **A lint or test** that fails when a verification helper is called with a `None` expected value outside an allowlist.

## Related
[...]

Issue #1119, opened 2026-07-19
feat(identity): atproto-account→UCAN exchange endpoint — server-side did:plc resolution, ai.hyprstream.identity.exchangeUcan lexicon, jti one-time-use


ewindisch:
## Context

The cross-repo DID contract (PR #1116, §10) specifies an atproto-account→UCAN exchange: a www-cyberdione-ai user signed in via atproto OAuth (did:plc/did:web) presents a com.atproto.server.getServiceAuth JWT to a hyprstream host, which verifies it and mints a UCAN grant (Assurance::Classical floor). Round-2 contract review (PR #1116) found the implementation work is referenced but owned by no issue. This issue owns it.

## Scope

1. **Server-side did:plc resolution** for THIS exchange: a did:plc arm in `MethodDispatchResolver` (currently fails closed), configurable PLC directory base (default `https://plc.directory`), used to resolve the caller's DID and verify the service-auth JWT signing key. Fail closed on resolution failure.
2. **Lexicon endpoint**: define `ai.hyprstream.identity.exchangeUcan` (XRPC, under the #908 `ai.hyprstream.*` lexicon namespace) as the exchange endpoint, so clients can request `getServiceAuth` with `lxm: ai.hyprstream.identity.exchangeUcan`. (The consumer's earlier Cap'n Proto shape cannot carry an NSID for lxm — superseded.)
3. **Token verification**: verify iss/aud/exp/lxm/jti per atproto service-auth semantics; `aud` MUST be the host's atproto-compatible did:web service DID (NOT did:at9p — upstream PDSes reject non-atproto audiences); document how that did:web is discovered/bound to the host.
4. **Replay**: server-side jti one-time-use cache scoped to the token lifetime.
5. **Honest security posture**: no sender-binding claim — service-auth JWTs carry no cnf/DPoP binding; mitigations are short exp, aud+lxm restriction, jti one-time-use.

## Related
- PR #1116 (contract, §10) — normative source
- #1113 (atproto AS conformance — separate concern: hyprstream AS-side)
- #1112 (XRPC surface — shares HTTP/XRPC routing infrastructure)
- #1114 (did:at9p verification — explicitly NOT this issue; at9p fails closed here)
- #908 (ai.hyprstream.* lexicons)
- cyberdione/www-cyberdione-ai GitLab epic #23 (consumer)

ewindisch:
**Hard gate identified by the #1121 dual review (see #1146 T2.1).**

[...]

PR #1147, opened 2026-07-22
fix(oauth): bind and validate minted authority (#1146 Lane B)

ewindisch:
## Summary

- choose T2.1 shape (a): embed the granted OAuth scope set as an optional signed `scope` claim, with fail-closed exact-scope helpers for downstream consumers such as #1119
- route OAuth bearer authentication and JWT introspection by the protected JWT algorithm so PolicyService composite tokens verify against their exact published pair
- require the OAuth-self audience and an exact profile-valid local issuer on both acceptance paths
- treat both a configured path-bearing issuer and its canonical origin as local key-routing aliases

Shape (a) is intentional: #1119 needs to attenuate authority from the grant presented by the caller without a subject-wide lookup. A signed optional claim gives the bridge that verifiable ceiling directly, while remaining backward-compatible with non-OAuth service tokens.

## Regression coverage

- minted atproto, refresh, generic, and device tokens carry their granted scope
- a composite token minted by the production PolicyService path authenticates to a protected OAuth route and can be introspected
- a token for another audience is rejected by bearer middleware
- a token from another issuer is inactive under JWT introspection
- path-bearing configured issuers route their canonical-origin atproto tokens as local, while sibling paths remain rejected

## Validation

- `cargo test -p hyprstream-rpc --lib --no-fail-fast` — 774 passed, 3 ignored
- `cargo build -p hyprstream` — passed
- `cargo test -p hyprstream --lib --no-fail-fast` — 1,135 passed, 1 ignored
- `cargo clippy -p hyprstream --all-targets -- -D warnings` — passed
- `cargo clippy -p hyprstream-rpc --all-targets -- -D warnings` — passed
- repository pre-commit release workspace Clippy — passed
- `git diff --check` — passed

`cargo fmt --check` was not run because the clean-base failure is tracked in #1140.

Refs #1146 (Lane B), #1121, #1119, #1145.

### DIFF
--- a/crates/hyprstream-rpc/schema/common.capnp
+++ b/crates/hyprstream-rpc/schema/common.capnp
@@ -164,6 +164,7 @@ struct Claims {
   token @6 :Text;        # Original JWT for e2e verification by downstream services
   iss @7 :Text;          # Issuer URL (RFC 7519); hyprstream node that minted this token
   pubKey @8 :Text;       # Ed25519 public key (base64url) for service identity tokens
+  oauthScope @9 :Text;   # Signed OAuth grant ceiling (space-delimited scope claim)
 }
 
 # UTC timestamp with nanosecond precision
--- a/crates/hyprstream-rpc/src/auth/claims.rs
+++ b/crates/hyprstream-rpc/src/auth/claims.rs
@@ -2,7 +2,8 @@
 //!
 //! Two claim types:
 //! - [`Claims`] — access token claims used throughout the RPC layer.
-//!   Authorization is enforced server-side via Casbin policies, not JWT scopes.
+//!   OAuth grants carry their granted scope set in the signed JWT so consumers
+//!   can attenuate authority without falling back to subject-wide policy.
 //! - [`IdTokenClaims`] — OIDC ID Token claims (Section 2 of OpenID Connect
 //!   Core 1.0). Only used by the OAuth token endpoint when `scope=openid`.
 
@@ -109,8 +110,9 @@ pub struct ActClaim {
 
 /// JWT claims for authentication.
 ///
-/// Note: Authorization is enforced via Casbin policies server-side.
-/// Scopes are NOT embedded in JWTs.
+/// Casbin remains the server-side policy authority. OAuth access tokens also
+/// carry the grant's scope ceiling so a verifier can enforce the intersection
+/// of subject policy and the authority delegated by that specific grant.
 #[derive(Clone, Serialize, Deserialize)]
 pub struct Claims {
     /// Issuer URL — the hyprstream node that minted this token.
@@ -128,6 +130,13 @@ pub struct Claims {
     /// RFC 8707 audience claim for resource indicator binding.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub aud: Option<String>,
+    /// OAuth 2.0 granted scope set, serialized as a space-delimited string.
+    ///
+    /// This is optional because non-OAuth service tokens do not originate from
+    /// an OAuth grant. Consumers that derive authority from OAuth scopes MUST
+    /// use [`Self::has_scope`] and fail closed when this claim is absent.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub scope: Option<String>,
     /// RFC 8705 / WIMSE proof-of-possession confirmation claim.
     ///
     /// Binds the Ed25519 signing key to the JWT-attested identity via a standard
@@ -202,6 +211,7 @@ impl std::fmt::Debug for Claims {
             .field("iat", &self.iat)
             .field("jti", &self.jti)
             .field("aud", &self.aud)
+            .field("scope", &self.scope)
             .field("cnf", &self.cnf)
             .field("token", &self.token.as_ref().map(|_| "[REDACTED]"))
             .field("clearance", &self.clearance)
@@ -285,6 +295,9 @@ impl FromCapnp for Claims {
             iat: reader.get_iat(),
             jti: None,
             aud,
+            // OAuth scopes are JWT-carried authority and deliberately are not
+            // reconstructed from the legacy Cap'n Proto scope list.
+            scope: None,
             cnf,
             token,
             // MAC clearance (S8/#574) is not carried on the Cap'n Proto envelope
@@ -311,6 +324,7 @@ impl Claims {
             iat,
             jti: None,
             aud: None,
+            scope: None,
             cnf: None,
             token: None,
             clearance: None,
@@ -340,6 +354,25 @@ impl Claims {
     pub fn with_audience(mut self, audience: Option<String>) -> Self {
         self.aud = audience;
         self
+    }
+
+    /// Set the OAuth 2.0 granted scope claim.
+    pub fn with_scope(mut self, scope: Option<String>) -> Self {
+        self.scope = scope.filter(|value| !value.trim().is_empty());
+        self
+    }
+
+    /// Iterate over the exact scopes granted to this token.
+    ///
+    /// An absent claim yields an empty iterator, keeping scope-based consumers
+    /// fail closed for legacy and non-OAuth tokens.
+    pub fn granted_scopes(&self) -> impl Iterator<Item = &str> {
+        self.scope.as_deref().into_iter().flat_map(str::split_whitespace)
+    }
+
+    /// Return whether this token's signed grant contains `required` exactly.
+    pub fn has_scope(&self, required: &str) -> bool {
+        self.granted_scopes().any(|granted| granted == required)
     }
 
     /// Attach original JWT token for e2e verification by downstream services.
@@ -510,6 +543,22 @@ impl crate::auth::mac::SubjectContextClaims for Claims {
 #[allow(clippy::unwrap_used, clippy::expect_used)]
 mod tests {
     use super::*;
+
+    #[test]
+    fn oauth_scope_claim_is_exact_and_absence_fails_closed() {
+        let plain = Claims::new("alice".to_owned(), 1000, 2000);
+        assert!(!plain.has_scope("atproto"));
+        assert!(!serde_json::to_string(&plain).unwrap().contains("\"scope\""));
+
+        let scoped = plain.with_scope(Some("atproto transition:generic".to_owned()));
+        assert!(scoped.has_scope("atproto"));
+        assert!(scoped.has_scope("transition:generic"));
+        assert!(!scoped.has_scope("transition"));
+        assert_eq!(
+            scoped.granted_scopes().collect::<Vec<_>>(),
+            vec!["atproto", "transition:generic"]
+        );
+    }
 
     #[test]
     fn test_act_claim_serde_roundtrip_and_default_none() {
--- a/crates/hyprstream-rpc/src/auth/claims.rs
+++ b/crates/hyprstream-rpc/src/auth/claims.rs
@@ -243,6 +243,9 @@ impl ToCapnp for Claims {
                 builder.set_pub_key(&jwk.x);
             }
         }
+        if let Some(ref scope) = self.scope {
+            builder.set_oauth_scope(scope);
+        }
         // Write empty scopes list for wire compatibility
         builder.reborrow().init_scopes(0);
     }
@@ -252,9 +255,15 @@ impl FromCapnp for Claims {
     type Reader<'a> = common_capnp::claims::Reader<'a>;
 
     fn read_from(reader: Self::Reader<'_>) -> Result<Self> {
-        // Scopes on the wire are ignored - authorization is via Casbin
         let aud = reader
             .get_aud()
+            .ok()
+            .and_then(|s| s.to_str().ok())
+            .map(std::borrow::ToOwned::to_owned)
+            .filter(|s| !s.is_empty());
+
+        let scope = reader
+            .get_oauth_scope()
             .ok()
             .and_then(|s| s.to_str().ok())
             .map(std::borrow::ToOwned::to_owned)
@@ -295,9 +304,9 @@ impl FromCapnp for Claims {
             iat: reader.get_iat(),
             jti: None,
             aud,
-            // OAuth scopes are JWT-carried authority and deliberately are not
-            // reconstructed from the legacy Cap'n Proto scope list.
-            scope: None,
+            // The dedicated OAuth field preserves the signed grant ceiling;
+            // the legacy structured scope list remains unused by Claims.
+            scope,
             cnf,
             token,
             // MAC clearance (S8/#574) is not carried on the Cap'n Proto envelope
@@ -558,6 +567,21 @@ mod tests {
             scoped.granted_scopes().collect::<Vec<_>>(),
             vec!["atproto", "transition:generic"]
         );
+    }
+
+    #[test]
+    fn oauth_scope_survives_capnp_envelope_roundtrip() -> anyhow::Result<()> {
+        let claims = Claims::new("alice".to_owned(), 1000, 2000)
+            .with_scope(Some("atproto transition:generic".to_owned()));
+        let mut message = capnp::message::Builder::new_default();
+        let mut builder = message.init_root::<common_capnp::claims::Builder>();
+        claims.write_to(&mut builder);
+
+        let decoded = Claims::read_from(builder.into_reader())?;
+        assert_eq!(decoded.scope, claims.scope);
+        assert!(decoded.has_scope("atproto"));
+        assert!(decoded.has_scope("transition:generic"));
+        Ok(())
     }
 
     #[test]
--- a/crates/hyprstream-rpc/src/auth/key_source.rs
+++ b/crates/hyprstream-rpc/src/auth/key_source.rs
@@ -117,10 +117,27 @@ pub trait JwtKeySource: Send + Sync + 'static {
 #[derive(Clone)]
 pub struct ClusterKeySource {
     ca_verifying_key: VerifyingKey,
-    local_issuer_url: String,
     local_issuers_vec: Vec<String>,
     ml_dsa_vks: Arc<std::sync::RwLock<Vec<crate::crypto::pq::MlDsaVerifyingKey>>>,
     composite_keys: Arc<super::CompositeKeySet>,
+}
+
+fn local_issuer_aliases(local_issuer_url: &str) -> Vec<String> {
+    if local_issuer_url.is_empty() {
+        return Vec::new();
+    }
+    let mut issuers = vec![local_issuer_url.to_owned()];
+    // Keep this byte-for-byte aligned with OAuthState::canonical_issuer_origin:
+    // the atproto issuer is the configured scheme + authority with its path
+    // removed, without URL-library normalization of ports or casing.
+    if let Some((scheme, rest)) = local_issuer_url.split_once("://") {
+        let authority = rest.split('/').next().unwrap_or_default();
+        let origin = format!("{scheme}://{authority}");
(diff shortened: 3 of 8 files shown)


ewindisch:
# Independent Security Review — PR #1147 "fix(oauth): bind and validate minted authority (#1146 Lane B)"

Reviewer: kimi (independent adversarial pass; graded against `laneB.txt`, not the PR's self-description)
Head reviewed: `bf260eaac` (diff base `85265b045`, +249/−48 across 6 files)

Validation run (shared warmed target, tree untouched at end):
- `cargo build -p hyprstream` — pass
- `cargo test -p hyprstream --lib --no-fail-fast` — 1,135 passed, 0 failed, 1 ignored
- `cargo test -p hyprstream-rpc --lib --no-fail-fast` — 774 passed, 0 failed, 3 ignored
- `cargo clippy -p hyprstream --all-targets -- -D warnings` — clean
- `cargo clippy -p hyprstream-rpc --all-targets -- -D warnings` — clean
- Revert-verify (temporary hunk reverts, each restored): see "Tests are real" below.

## VERDICT: REVISE

T3.1 is done properly — both halves landed, the tests are real, and the #1145 instance in
`auth.rs`/`introspection.rs` is closed. T2.1's core (signed scope claim on the OAuth grant paths)
is correct and the fail-closed helpers are honest. But one mint path feeds **unvalidated
caller input** into the new signed authority claim (F1), which is exactly the confused-deputy
shape #1119 is gated against. Narrow fix. Until F1 lands, #1119 is **not** genuinely unblocked.

## Findings (severest first)

### F1 — BLOCKER: token-exchange mints arbitrary caller-supplied scopes into the signed claim
`token_exchange.rs:91-92` builds `requested_scopes` by splitting the caller's RFC 8693 `scope`
parameter on whitespace — no `validate_requested_scopes`, no intersection with the subject
token's authority, not even `normalize_scope_token` — and `policy.rs:367-373` now signs that
verbatim into the JWT `scope` claim. Before this PR the field was ignored at mint; the PR makes
it a signed authority claim while leaving this path unvalidated. A holder of any valid subject
token (e.g. a service WIT via `verify_jwt`) can POST `grant_type=token-exchange` with
`scope=transition:generic` (or any invented string) and receive a token whose **signed** claim
asserts that scope. The moment #1119's bridge attenuates from `claims.scope`, this is a live
privilege-escalation primitive — worse than subject-wide fallback, because the fabricated
ceiling is attacker-chosen. The OAuth grant paths (PAR/authorize/device) all validate through
`validate_requested_scopes`; this one never did, and T2.1 made that matter. Fix either by
validating/intersecting in `token_exchange.rs` (subset of the subject token's scopes, or the
server-supported set) or by passing `None` there until a real grant exists. Needs a test that
mints through token-exchange with an out-of-grant scope and asserts the claim is rejected or absent.

### F2 — MODERATE: the AS now rejects its own default-audience tokens when `external_url` carries a path
`validate_oauth_access_token` (auth.rs:211) expects exactly `state.atproto_issuer_url()` (canonical
origin) as audience. But `PolicyService`'s `default_audience` is the **raw configured issuer**
(`factories.rs:627` ← `config.oauth.issuer_url()`), so any token minted without an explicit
`resource` indicator gets `aud = <configured issuer, path included>`. In a path-bearing
deployment (the configuration this whole chain exists for) those tokens are now 401 at every
protected OAuth route — the AS mints default tokens it refuses to accept. `key_source.rs` was
given exactly this alias treatment for issuers; the audience side needs the same (accept both
aliases) or the default audience must be canonicalized. Previously these callers passed
(`decode(..., None)`), so this is a behavior change either way — decide deliberately and test it.

### F3 — MODERATE: `verify_access_token` (token-exchange subject verification) still routes everything through the fixed EdDSA key with `decode(..., None)`
`token_exchange.rs:183-188` is the third adjacent acceptance path with the pre-fix shape: composite
access tokens **cannot** be used as `subject_token` in an exchange (signature never verifies against
the Ed25519-only key), and audience is unchecked for the EdDSA tokens that do verify. It wasn't one
of the two sites #1145 item 2 names, so the #1145 instance "here" is still closed (see below), but
the same class survives one hop away and should share `validate_oauth_access_token` rather than
re-implement verification. Note the lane's own T2.1 shape-(b) alternative
(`PendingAuthCode.scopes` → `RefreshTokenEntry.scopes`) is unaffected.

### F4 — NOTE: envelope-reconstructed `Claims` drop `scope`
`claims.rs` `FromCapnp` sets `scope: None` (deliberate, documented, consistent with
`clearance`/`act`/`cap`). #1119's bridge must read the grant ceiling from the **verified JWT**
(`Claims::with_token` carries the raw token), not from envelope-reconstructed Claims — otherwise it
will see "no claim" on every RPC hop and fail closed into denying everything. Worth one sentence in
the #1119 handoff so the consumer doesn't trip on it.

### F5 — NOTE: who now fails that used to pass (intended, enumerated)
- Tokens with **no `aud`** at protected OAuth routes: rejected (both composite and EdDSA decode
  paths are strict — verified at `jwt.rs:459-466` and `jwt.rs:625-630`). Production mints always set
  `aud` (explicit resource or `default_audience`), so legitimate impact is confined to hand-rolled
  callers like the old test's `introspection_caller` (which the PR itself rewrote to the real token).
- Tokens with `aud` ≠ canonical origin (e.g. minted for an MCP/other resource): rejected at OAuth
  routes. Correct RFC 8707 discipline — this was the fail-open the fix exists to close.
- Foreign `iss`: rejected (exact-match against both local profile issuers — good).
- `wit+jwt` and other non-`at+jwt` types: rejected at HTTP OAuth routes. WITs are a ZMQ-envelope
  credential; no legitimate HTTP consumer exists. Fine.
- Composite tokens whose `kid` is not in the current published snapshot: rejected. Previously they
  failed too (EdDSA-only key), so no regression.

## T3.1 — both halves confirmed

- **Algorithm routing**: `validate_oauth_access_token` dispatches on the protected-header `alg`
  (`ML-DSA-65-Ed25519` → exact published pair by `kid` from the global snapshot; `EdDSA` →
  `jwt_bearer_verifying_key()` with rotation-slot preference). A composite token minted by the real
  in-process PolicyService now authenticates to a protected OAuth route (handler test) — the exact
  original defect.
- **Audience + issuer**: both decode wrappers are strict (missing or wrong `aud` ⇒ reject), and the
  issuer must exactly match the canonical origin or the configured issuer. The #1145 instance in
  `auth.rs:74`/`introspection.rs:42` is **closed** (both `decode(..., None)` sites gone). Issuer
  routing for path-bearing `external_url` is fixed via `local_issuer_aliases` in
  `ClusterKeySource`/`JwksKeySource` — byte-aligned with `canonical_issuer_origin`, sibling paths
  still denied.
- Bonus: `jti` blocklist check added on the acceptance path.

## T2.1 — shape (a), graded

The two tokens are now distinguishable: `atproto` vs `transition:generic` grants produce different
signed `scope` claims (conformance assertions on the minted JWT for atproto/refresh/generic/device
flows). `has_scope`/`granted_scopes` fail closed on absence, `with_scope` filters empties, and the
claim is optional so non-OAuth service tokens are untouched. Enforcement on the resource side is
#1119's job by design — what this PR owed was a **trustworthy** claim, which F1 breaks on the
token-exchange path. Fix F1 and the gate is honestly met.

## Tests are real (revert-verify, each restored after)

1. Removed `.with_scope(granted_scope)` in `policy.rs` →
   `oauth_handler_atproto_and_legacy_conformance` fails (`claims["scope"]`: `Null` vs `"atproto"`). ✔
2. Composite-branch audience `Some(&expected_audience)` → `None` → same test fails: wrong-audience
   bearer token gets `200` where `401` was asserted. ✔
3. Disabled the composite routing arm → same test fails: the real minted composite token gets
   `401 invalid_token` at the protected route where `200` was asserted. ✔
4. Removed the canonical-origin alias in `local_issuer_aliases` →
   `local_key_sources_trust_canonical_origin_for_path_issuer` fails (`is_trusted(origin)` false). ✔

## Verdict conditions

- F1 must be fixed (validate/intersect or drop the claim on the token-exchange path, plus a test)
  before #1119 is treated as unblocked. Everything else is landable as-is.
- F2 should be decided deliberately in this chain (it bites exactly the path-bearing deployments the
  atproto profile targets); F3 is the natural follow-up site for the shared validator.
- #1145: the `auth.rs`/`introspection.rs` instance is closed here; say so on #1145. The
  `token_exchange.rs` site should be added to the audit list rather than silently left.

## Confidence and what I could not verify

High on F1–F3 (traced to code, F1/F2 mechanics confirmed by reading the mint and acceptance paths;
revert-verifies run locally). Not verified: a live deployment with path-bearing `external_url`
(mechanics traced, not executed); Valkey-backed refresh store (feature-gated); CodeQL not rerun.

ewindisch:
Addressed every finding in `e3b075693`:

- **F1 — scope inflation:** token exchange now derives authority only from a verified local OAuth access token. A requested scope must be a normalized, non-empty subset of that verified grant; the minted token carries the intersection in subject-grant order. ID tokens and generic JWTs cannot authorize OAuth scopes. Out-of-grant requests return `400 invalid_scope` before replay consumption. Counterfactual test: reverting this check changed the inflation case from 400 to 200.
- **F2 — path-bearing default audience:** access-token validation accepts exactly either the canonical origin or the configured local issuer URL as the audience, then performs strict signature/audience decoding against the selected exact value. This preserves path-bearing deployments without wildcard audience acceptance. Counterfactual test: restoring canonical-only audience handling changed the protected route from 200 to 401.
- **F3 — subject-token verification:** token exchange now reuses the shared OAuth access-token validator, including composite/EdDSA verification, exact issuer/audience checks, and JTI revocation. Counterfactual test: restoring fixed EdDSA plus `decode(None)` rejected the production composite subject with 401 before scope evaluation.
- **F4 — envelope scope loss:** added wire-compatible `oauthScope @9` to the Capn Proto claims schema and mapped it in both directions. Counterfactual test: removing the write mapped the roundtrip scope from `Some("atproto transition:generic")` to `None`.

Validation:
- `cargo build -p hyprstream`
- `cargo test -p hyprstream --lib` — 1135 passed, 0 failed, 1 ignored
- `cargo clippy -p hyprstream --all-targets -- -D warnings`
- focused assembled-router OAuth conformance test
- focused Capn Proto envelope roundtrip test
- foreground pre-commit release workspace Clippy

Per repository guidance, `cargo fmt --check` was not used as a gate because clean main is already tracked as failing in #1140.

Leaving PR #1147 open for re-review; not merging.

ewindisch:
Coordination for #1194 review: I repaired the combined service-WIT registration contract by minting a dedicated exact registration audience and requiring it, plus required cnf.jwk equality in Policy registration. I also migrated the CrossApp OIDC ID-token path to an explicit reason-bearing audience exemption because that boundary does not receive the external OIDC client ID. These touch policy.rs/token_exchange.rs to avoid the cross-PR availability regressions called out in the independent review; please account for these adjacent hunks when rebasing.

ewindisch:
# Re-Review — PR #1147 "fix(oauth): bind and validate minted authority (#1146 Lane B)"

Reviewer: fresh independent adversarial pass (kiro-cli), graded against `rr-1147.txt` and the prior
BLOCKER verdict `verdict-1147-kimi.md`. The author's fix account was treated as a claim to test, not
evidence.

Head reviewed: `e3b075693` ("fix(oauth): bind token exchange to verified authority"), diff base
`bf260eaac` (the prior-reviewed head). Fix delta: +239/−20 across 5 files
(`common.capnp`, `claims.rs`, `auth.rs`, `oauth/mod.rs` tests, `token_exchange.rs`).

Validation (all via the shared `buildq` pool; tree untouched at end — `git diff --stat` clean):
- `cargo test -p hyprstream-rpc --lib --no-fail-fast` — 775 passed, 0 failed, 3 ignored
- `cargo test -p hyprstream --lib --no-fail-fast` — 1,135 passed, 0 failed, 1 ignored
- `cargo clippy -p hyprstream-rpc --all-targets -- -D warnings` — clean
- `cargo clippy -p hyprstream --all-targets -- -D warnings` — clean
- Revert-verify of all four new assertions (each restored after; final tree clean) — see below.

## VERDICT: PASS

The prior BLOCKER (F1) and both MODERATE findings (F2, F3) and the NOTE (F4) are genuinely closed.
The fix does **not** introduce a new escalation primitive — I tried to inflate scope through every
path into the claim and could not (attempt below). Lane B is done.

Per-finding severity: **F1 BLOCKER → CLOSED · F2 MODERATE → CLOSED · F3 MODERATE → CLOSED · F4 NOTE → CLOSED.**

## F1 — scope inflation via token-exchange — CLOSED (was BLOCKER)

The mint itself is unchanged and still naive: `policy.rs:handle_issue_token` signs
`requested_scopes` verbatim into the `scope` claim (`.with_scope(granted_scope)`), with no validation
of its own. So the whole security of the claim rests on **every** HTTP path pre-attenuating before it
reaches the mint. I enumerated every caller that feeds `requested_scopes`:

- `authorize.rs:433` and `par.rs:184` — route through `validate_requested_scopes` (server-supported ∩
  client-declared) before minting. Unchanged, still correct.
- `token.rs:885` — scopes come from the already-validated stored grant snapshot.
- `jwt_bearer.rs:106` — passes `None` (service defaults) ⇒ `granted_scope = None` ⇒ no scope claim
  (fail-closed).
- `token_exchange.rs` — **was the hole.** Now fixed.

The token-exchange path now:
1. Derives `granted_scopes` **only** from a subject token that passed the full
   `validate_oauth_access_token` (algorithm routing + exact aud + local iss + jti), reading the
   subject token's own *signed* `scope` claim (`token_exchange.rs:203-210`).
2. `attenuate_exchange_scopes` (`token_exchange.rs:262-308`) enforces requested ⊆ granted after
   `normalize_scope_token`, returning `invalid_scope` on any out-of-grant or malformed token, and
   emits the intersection in grant order — never caller-supplied text.
3. Subject types with no server-authorized OAuth grant (`verify_id_token`, `verify_jwt` — the WIT /
   federated paths) set `granted_scopes: None`, so any requested scope is rejected and no scope claim
   can be minted. **This closes the exact attack the prior review named** ("a service WIT via
   `verify_jwt`").

**My inflation attempt (traced against the code):** present a valid subject access token carrying
`scope=read:*:*`, POST `grant_type=token-exchange&scope=transition:generic`. `attenuate_exchange_scopes`
finds `transition:generic ∉ {read:*:*}` → `Err("requested scope exceeds the subject token grant")` →
`400 invalid_scope`, before replay consumption. The signed ceiling is the *server-signed subject
token's* grant, and to obtain that subject token in the first place you must go through
authorize/PAR/token, all of which intersect against server-supported ∩ client-declared. There is no
path where the request writes its own ceiling. Attempt fails — as it should.

Revert-verify: replacing `attenuate_exchange_scopes` with the pre-fix verbatim
`scope.split_whitespace()` flips the inflation case from `400` to `200` — the new test
(`oauth_handler_atproto_and_legacy_conformance`, mod.rs:1934) fails. Restored.

## F2 — AS rejecting its own default-audience tokens on path-bearing issuers — CLOSED

`validate_oauth_access_token` now accepts as audience **either** the canonical origin
(`atproto_issuer_url()`) **or** the configured issuer URL (`state.issuer_url`), then decodes/verifies
against the *exact selected value*. This is not a wildcard: it is exactly the two identity forms this
AS mints under, and the chosen value is still cryptographically enforced against the token's signed
`aud` inside `decode_composite`/`decode` (the pre-verify `decode_unverified` only *selects* which of
the two allowed values to enforce — it cannot inject a third). Issuer routing keeps the same two-value
exact match. The symptom is not "loosened away"; the check is narrowed to the AS's own two aliases.

Revert-verify: forcing canonical-only audience makes the path-bearing default-audience token get
`401` where `200` is asserted (mod.rs:1891, "the AS must accept its path-bearing default audience
alias"). Restored.

## F3 — token-exchange subject verification still on fixed EdDSA `decode(..., None)` — CLOSED

`verify_access_token` no longer re-implements verification against `state.verifying_key_bytes` with
`decode(token, &vk, None)`. It now delegates to the shared `validate_oauth_access_token` — the same
algorithm routing, exact audience, exact local issuer, and jti-revocation the protected OAuth routes
use. Composite (`ML-DSA-65-Ed25519`) subject tokens now verify; audience is checked on the EdDSA path
too. **F1 and F3 no longer compound**: the exchange path both validates the subject token's audience
*and* attenuates scope, so the "audience-unchecked precisely where scope is minted" pairing is gone.

Revert-verify: restoring the fixed-EdDSA + `decode(None)` body makes the production composite subject
token fail signature verification and return `401` before scope evaluation, where `400` is asserted
(mod.rs:1934). Restored.

## F4 — `scope` survives envelope reconstruction — CLOSED (stronger than the prior recommendation)

The prior review recommended #1119 read the ceiling from the verified JWT because `FromCapnp` dropped
`scope`. The author went further: added wire-compatible `oauthScope @9` to the Cap'n Proto `Claims`
schema and mapped it in **both** directions, so the signed grant ceiling now survives the envelope
hop. The value written is the server's own post-verification `Claims.scope` (not attacker-controlled
on that path), so this is safe, and it removes the "present at mint, absent at check" fail-open the
task flagged as the worst case. #1119's bridge can now read the ceiling from reconstructed Claims *or*
the raw JWT.

Revert-verify: reverting the read-side mapping to `scope: None` makes
`oauth_scope_survives_capnp_envelope_roundtrip` fail (`None` vs `Some("atproto transition:generic")`).
Restored.

## T2.1 acceptance (implementation-independent) — MET

- **Distinguishability + fail-closed:** an `atproto`-only token and a `transition:generic` token for
  the same subject now carry different signed `scope` claims (conformance assertions on the minted
  JWTs for atproto/refresh/generic flows). A verifier that cannot find the claim fails closed:
  `granted_scopes()` yields empty and `has_scope()` returns `false` on an absent claim — exact-match,
  no substring/prefix leniency.
- **Bridge attenuates from the GRANTED set, not subject-wide authority:** the token-exchange
  attenuation is itself the #1119-correct shape applied one layer early — it clamps to the subject
  token's *granted* scope set, never to subject-wide Casbin authority. This is the confused-deputy
  (#1119) outcome, not the #1119 bug.

## Tests are real (revert-verify; each restored, final tree clean)

| Finding | Revert | Observed failure |
|---|---|---|
| F1 | `attenuate_exchange_scopes` → verbatim `split_whitespace` | inflation exchange `200` vs asserted `400` |
| F2 | audience → canonical-origin only | default-audience token `401` vs asserted `200` |
| F3 | `verify_access_token` → fixed EdDSA + `decode(None)` | composite subject `401` vs asserted `400` |
| F4 | `FromCapnp` scope → `None` | roundtrip `None` vs `Some("atproto transition:generic")` |

## Downstream gating

- **#1119 is genuinely unblocked.** The signed ceiling is now trustworthy on every mint path
  (including token-exchange), fail-closed on absence, and survives the Cap'n Proto envelope. One
  handoff sentence for #1119: read the ceiling from `Claims.scope` (now populated on both the verified
  JWT via `with_token` and the reconstructed envelope via `oauthScope @9`); attenuate from that set,
  never from subject-wide Casbin authority.
- **#1121 may come out of draft.**
- **#1145:** the `token_exchange.rs` `decode(..., None)` site the prior review asked to be added to
  the audit list is now *fixed* (shares `validate_oauth_access_token`) rather than merely tracked. The
  `auth.rs`/`introspection.rs` instances remain closed.

## Confidence and what I could not verify

High: F1–F4 traced to source and each confirmed by an executed revert-verify against a real assertion.
Not verified: a live deployment with path-bearing `external_url` (mechanics traced, not deployed);
Valkey-backed refresh store (feature-gated, not exercised); CodeQL not rerun. `cargo fmt --check` not
used as a gate per repo guidance (#1140 tracks clean-base fmt failure).

Issue #1144, opened 2026-07-22
security: drop allow_credentials on the product API and allow any origin (bearer/DPoP already carries all authority)


ewindisch:
## Problem

The product API router (`/oai/v1`, `/models`, `/.well-known/export9p`, `/.well-known/planes`) is CORS-gated by an **enumerated origin allow-list with `allow_credentials: true`** (`default_cors_credentials()` returns `true`, `crates/hyprstream/src/config/server.rs`). That combination is the wrong shape for this service, and it does not scale.

**The credentials flag buys nothing here.** `allow_credentials` exists so browsers may attach *ambient* authority — cookies, HTTP auth. hyprstream attaches none:

- No cookie usage anywhere in `crates/hyprstream/src/server/` (`grep -i "cookie|set-cookie|samesite"` → zero hits).
- Auth is purely header-based: `server/middleware.rs` reads `header::AUTHORIZATION`, accepting `Bearer` and `DPoP`, and **rejects plain Bearer when the token carries `cnf.jkt`** (RFC 9449 sender-binding).

**But it costs a lot.** `Access-Control-Allow-Origin: *` is illegal alongside credentials, so the product API is permanently stuck enumerating origins. Every new first-party domain is a config change and a deploy; arbitrary third-party origins (a user's own atproto app, a third-party client authenticating against this instance) are impossible by construction.

## Proposal

Set `allow_credentials: false` on the credentialed router and allow `*` (or reflect-with-validation). Every request already carries an explicit `Bearer`/`DPoP` token, so there is no ambient authority for a hostile origin to ride — and DPoP sender-binding is strictly stronger than anything cookies offered. This is the standard public-API shape.

This is a **security-posture change** and deserves its own review rather than riding along with an origin addition — hence this issue.

## Why now

It is the actual unlock for third-party/atproto clients calling the product surface. Widening the allow-list one domain at a time grows the problem; this removes it.

Note the OAuth surface is **already** wildcard-public and unaffected: `CorsConfig::public()` is `*` with `allow_credentials: false` (OAuth metadata/token/registration and the XRPC routes), and `CorsConfig::did_document()` is `*` with permissive headers. Login from a third-party origin already works — this issue is only about the product API.

## Must be established by the work

[...]

PR #1150, opened 2026-07-22
fix(cors): honor every configured origin, not just the last (pre-existing)

ewindisch:
## Why

This splits the genuine pre-existing runtime bug out of #1132, following the implementation sequence in #1144.

`CorsLayer::allow_origin` replaces its previous policy on every call. The old per-origin loop therefore retained only the last configured origin, causing browsers to reject every earlier configured origin.

## What

- pass the complete `Vec<HeaderValue>` in one `allow_origin` call so tower-http builds `AllowOrigin::list(...)`
- retain the assembled `Router + CorsLayer` runtime harness from #1132
- verify default, explicit multi-origin, exact-match, credential, and wildcard behavior through response headers
- leave the default origin list and config assertions unchanged; #1144 owns the any-origin posture change

## Regression proof

The focused multi-origin runtime test passes with this fix. I then temporarily restored the old per-origin loop without changing the test; it failed on the first entry because `https://a.example.com` received `Access-Control-Allow-Origin: https://c.example.com`, the final configured origin. Restoring the list-based call made the same test pass again.

## Verification

- `cargo build -p hyprstream`
- `cargo test -p hyprstream --lib cors` (5 passed)
- `cargo clippy -p hyprstream --all-targets -- -D warnings`
- foreground pre-commit CI-mirroring Clippy

`cargo fmt --check` is not used as a gate because clean main has the known #1140 failure; this one-file diff passes `git diff --check`.

Related: #1132, #1144.

### DIFF
--- a/crates/hyprstream/src/server/middleware.rs
+++ b/crates/hyprstream/src/server/middleware.rs
@@ -604,30 +604,37 @@ pub fn cors_layer(config: &crate::server::state::CorsConfig) -> tower_http::cors
         cors = cors.allow_origin(Any);
         // Never allow credentials with wildcard
         cors = cors.allow_credentials(false);
-    } else if config.allowed_origins.is_empty() {
-        // If no origins specified, use default localhost origins
-        let default_origins = vec![
-            "http://localhost:3000",
-            "http://localhost:3001",
-            "http://127.0.0.1:3000",
-            "http://127.0.0.1:3001",
-        ];
-        for origin in default_origins {
-            if let Ok(header_value) = origin.parse::<HeaderValue>() {
-                cors = cors.allow_origin(header_value);
-            }
-        }
-        // Configure credentials based on config
-        if config.allow_credentials {
-            cors = cors.allow_credentials(true);
-        }
     } else {
-        // Allow specific origins
-        for origin in &config.allowed_origins {
-            if let Ok(header_value) = origin.parse::<HeaderValue>() {
-                cors = cors.allow_origin(header_value);
-            }
-        }
+        // Allow specific origins.
+        //
+        // The whole set MUST be passed in a single `allow_origin` call:
+        // `CorsLayer::allow_origin` is documented to *override* on every call
+        // (`self.allow_origin = origin.into()`), and a single `HeaderValue`
+        // becomes `AllowOrigin::exact(...)` — a constant `Access-Control-Allow-Origin`
+        // emitted unconditionally, not a membership check. Iterating and calling
+        // it once per origin would therefore honor only the LAST origin in the
+        // list, silently dropping every earlier origin. Passing the
+        // `Vec<HeaderValue>` makes tower-http build `AllowOrigin::list(...)`,
+        // which performs exact-equality membership matching against the full set.
+        let origins: Vec<HeaderValue> = if config.allowed_origins.is_empty() {
+            // If no origins specified, use default localhost origins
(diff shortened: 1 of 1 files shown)


ewindisch:
**Adversarial review — verdict: PASS** (no blocking or revise-level findings; two informational notes). Full writeup in the linked gist-equivalent below; summary here.

## Verification performed (not just prose)

**1. The fix is on the assembled router's path.** `cors_layer()` is the *only* `CorsLayer` construction site in the tree; the product API (`server/mod.rs:112`), OAuth (`oauth/mod.rs:224-225`), and MCP (`factories.rs:1610`) routers all call it. Confirmed against **tower-http 0.6.11** source:
- `From<HeaderValue>` → `OriginInner::Const` → emits the origin *unconditionally* on every request (the bug — last origin wins and is echoed regardless of request `Origin`).
- `From<Vec<HeaderValue>>` → `OriginInner::List` → `origin.filter(|o| l.contains(o))` — exact membership match (the fix).

**2. The harness is not theatre — verified by revert.** Replaced `cors.allow_origin(origins)` with the old `for hv in origins { cors.allow_origin(hv); }` loop and reran `cors_layer_tests`: **3 of 5 failed**, reproducing the bug exactly — `https://a.example.com` got ACAO `https://c.example.com` (last entry), and a near-miss origin got a constant `https://www.hyprstream.com` echo instead of `None` (exposing a worse latent posture: constant-ACAO emission). The 2 that still passed take the `["*"]` branch or don't assert the ACAO value — they pass for the right reason. Restored; all 5 pass; tree clean.

**3. Split is clean.** PR touches only `middleware.rs`. `config/server.rs` untouched; default lists still localhost-only. `hyprstream.com` appears only as a *test fixture* (the configured set + hijack strings in the near-miss test), never as a default-list member. No config-vec assertions carried over — the new tests feed the builder then assert runtime headers, which is the whole point.

## Security / widening / credentials
- **No widening beyond configuration.** `List` is strict membership; `Const(last)` was an unconditional constant. The change narrows.
- **Credentials preserved.** `allow_credentials` stays config-gated; the wildcard branch still forces credentials off; `cors_layer_wildcard_never_pairs_with_credentials` verifies this even under adversarial `["*"]` + creds-true.

## Informational (not blocking)

- **F-1 (MCP behavior change, safe):** the MCP router rides the same fix (shares `CorsConfig::default()`), going from "only last localhost origin honored" to "all four honored." All loopback, all already in the configured set — no new authority. Flagged only because the surface isn't named in the PR description.
- **F-2 (pre-existing, fail-closed):** `filter_map(|o| o.parse::<HeaderValue>().ok())` silently drops malformed configured origins. Safe but a debug footgun; unchanged by this PR. A `warn!(...)` on the dropped `None` arm would be a nice out-of-scope follow-up.

## What now works / breaks (per ask)
- **Now works (intended):** every configured origin is honored; multi-origin `HYPRSTREAM_CORS_ORIGINS` operators get all entries.
- **Nothing legitimate breaks** — no previously-allowed origin is now rejected; the only "regression" is for callers relying on the constant-ACAO misemission, which was never supported.

Repro: `cargo test -p hyprstream --lib cors_layer_tests` → 5 pass; revert the single `allow_origin(origins)` call to the per-origin loop → 3 fail (incl. the exact PR regression proof). `cargo fmt --check` not used as a gate (#1140 pre-existing on clean main).

Issue #1176, opened 2026-07-22
E4: migrate existing local accounts to host-form DIDs


ewindisch:
Phase **E4** of https://github.com/hyprstream/hyprstream/issues/1158. Dep: B1, https://github.com/hyprstream/hyprstream/issues/1159 (A1).

First-login remint to host-form per the plan's §5. Deprecation response for path-form documents. Label-never-reused enforcement.

**Scope depends on A1's audit result.** If A1 finds zero path-form DIDs anchored in any `RepoSnapshot`/repo authority, this is a small remint path. If it finds any, this becomes a data migration over permanent identifiers that other people's data may reference — a materially different ticket. Do not size this before A1 reports.

ewindisch:
## Scope correction: the #1181 audit does NOT establish zero deployed path-form anchors

PR https://github.com/hyprstream/hyprstream/pull/1181 (A1) reports **0** path-form repo anchors and concludes this ticket needs no permanent-identifier data migration. **Both independent reviewers found that claim narrower than stated**, and gpt-5.6-sol rated it a BLOCKER.

What the audit actually establishes: **zero regex matches in selected source at one audited SHA.** What it does not establish: the absence of path-form DIDs in **deployed state** — a regex over in-tree source cannot see identifiers assembled at runtime, nor anything already persisted in a live database.

Concretely, sol found a live path by which path-form subjects persist and propagate *today*: `RefreshTokenEntry.username` is the persisted JWT subject, RocksDB serializes it unchanged, the default refresh lifetime is ~30 days, and refresh passes `claimed.username` straight through to issuance — minting a new access token **and** a newly rotated refresh token carrying the same path-form subject, with no check at either boundary. Valkey-backed user profiles may also retain path-form usernames.

**E4 must therefore plan for deployed-state migration rather than assume a clean slate**, and must decide explicitly whether startup revokes, scans, or migrates stored refresh and user records. Unmigrated permanent identifiers are unrecoverable later, which is why this correction is being recorded before #1181 merges rather than after.

Refs https://github.com/hyprstream/hyprstream/issues/1159, https://github.com/hyprstream/hyprstream/issues/1158

Issue #1163, opened 2026-07-22
B1: account record + genesis DidOp + mint path (decides the permanent genesis format)


ewindisch:
Phase **B1** of https://github.com/hyprstream/hyprstream/issues/1158. Dep: https://github.com/hyprstream/hyprstream/issues/1159 (A1), https://github.com/hyprstream/hyprstream/issues/1160 (A2), https://github.com/hyprstream/hyprstream/issues/1162 (A3). **BLOCKED on Q2.**

Allocate a label, generate the `#atproto` ES256 key, establish `rotation_keys` (ordered), seal the genesis op, write the account record.

**This ticket decides the permanent genesis format.** It carries `version` from op #1 — a log whose format changes incompatibly is retroactively unverifiable, so `version` must exist from the first operation ever sealed, not be added later.

**Q2 blocks this and is a one-way door per account:** host-only rotation keys with user enrollment offered later (better signup UX, every account starts fully captured), or a user-held top-priority key required at mint (real portability from op #1, worse UX, key-loss support burden). The recommendation is the hybrid — user key offered at signup, host key always present — but the *default* is permanent per account and cannot be retrofitted without the user cooperation that is exactly the property in question.

ewindisch:
## UNBLOCKED — Q2 answered (Erica, 2026-07-23): option B

Full rationale: https://github.com/hyprstream/hyprstream/issues/1158#issuecomment-5059754387

**The genesis `DidOp` seals `rotation_keys` in this shape:**

| Priority | Key | Required |
|---|---|---|
| 0 | **user-held** | **YES — no account is minted without it** |
| 1 | user-held **recovery** key (different device) | offered at mint, strongly encouraged |
| 2 | host key | **optional, and ONLY below the user's** |

**A host key at priority 0 is escrow and is forbidden.** Ordering is the property that decides this, not presence: a higher-priority key can override operations signed by a lower one inside a fork window, so a host key above the user's means the host can seize the account unilaterally. The ordering must be **explicit in the sealed operation**, not an artifact of insertion order — write it deliberately and test it.

**Do not conflate rotation keys with operational keys.** The account's `#atproto` ES256 signing key is held and rotated by hyprstream (`auth/key_rotation.rs:1028-1035`) so the PDS can sign repo commits without the user present; that is unchanged by this decision and is NOT a rotation key. Agent authority is UCAN delegation with attenuation, not a rotation key. **This decision does not require the user to be online for anything except changing identity control.**

[...]

Issue #1158, opened 2026-07-22
Epic: hyprstream-hosted account identity — host-form did:web, sealed DID documents, operation log


ewindisch:
Approved by Erica 2026-07-22. Supersedes the design phase of https://github.com/hyprstream/hyprstream/issues/1124 (closed as decided).
Full design: `.fleet-coord/t1124/plan.md` · recon behind the decision: `.fleet-coord/t1124/recon.md`

**Decision.** hyprstream-hosted accounts get **host-form did:web**. did:plc is rejected for its dependency on Bluesky-operated infrastructure. We **accept federation from did:plc identities** — the rule governs what we mint, not what we trust.

Recon confirmed it: self-hosted PLC is permissively licensed and runnable, but its DIDs **would not resolve anywhere else** (verified against four independent resolver defaults), and there is no DHT-based atproto directory mature enough to adopt.

**The cost we accepted.** did:plc supplies key-rotation history and a verifiable operation log for free; did:web supplies neither. **Phase C is that bill.** Mitigating: ~75-90% of Phase C is owed regardless, because *someone* must answer "was this key valid when this commit was signed" — see https://github.com/hyprstream/hyprstream/issues/1118.

---

### Phase A — stop the bleeding

| | Ticket | Depends on | Status |
|---|---|---|---|
| A1 | https://github.com/hyprstream/hyprstream/issues/1159 — freeze path-form minting, audit anchors | — | **READY · start here** |
| A2 | https://github.com/hyprstream/hyprstream/issues/1160 — label grammar + allocation policy | — | **READY** |
| A3 | https://github.com/hyprstream/hyprstream/issues/1162 — deployment-configured zone, wildcard DNS, DNS-01 TLS | A2 | **READY** (Q1 answered) |

### Phase B — mint and serve

| | Ticket | Depends on | Status |
|---|---|---|---|
| B1 | https://github.com/hyprstream/hyprstream/issues/1163 — account record + genesis DidOp + mint path | A1, A2, A3 | **blocked on Q2** |
| B2 | https://github.com/hyprstream/hyprstream/issues/1164 — canonical sealed DID-document artifact | B1 | blocked |
| B3 | https://github.com/hyprstream/hyprstream/issues/1165 — account HTTP face, credential-free origin | B2 | blocked · see Q6 |
| B4 | https://github.com/hyprstream/hyprstream/issues/1166 — handle resolution, bidirectional verification | B2 | blocked · see Q7 |

[...]

PR #1181, opened 2026-07-22
fix(auth): freeze path-form did:web account minting

### LINKED ISSUE #1159: A1: freeze path-form did:web minting (spec-invalid) and audit existing anchors
Phase A1 of https://github.com/hyprstream/hyprstream/issues/1158. **Start here — this is the only ticket with no dependency and a cost that grows every day.**

Path-form `did:web:{authority}:users:{name}` is **spec-invalid** and is being emitted today. Every one minted is permanent and wrong.

1. Make path-form a **hard error** in the account-DID constructors — not a warning, not a deprecation. Stop emitting it as a token `sub` or from any mint path.
2. **Audit** every `RepoSnapshot` / repo authority for a path-form did:web anchor and **report the count**. The migration shape in the epic (E4) depends on that answer being zero; if it is not zero, say so loudly, because it converts a freeze into a data migration.
3. Tests: minting path-form must fail; the audit must be reproducible.

Do not design the replacement here — host-form minting is B1 and is blocked on Q2. This ticket only stops the bleeding.

ewindisch:
Closes #1159.

Part of epic #1158. This intentionally does not design or enable host-form account minting; that remains #1163.

## Summary

- keep `UserMappingStrategy::DidWeb` as a hard configuration and runtime error, so new OIDC callbacks cannot create path-form account subjects;
- return `410 Gone` from the account DID route before any key or profile lookup, with a router-level regression test;
- reject path-form `did:web` at both the in-memory snapshot and durable PDS publisher write boundaries;
- reject a path-form local-account subject at the central token issuance boundary. A pre-upgrade refresh token is consumed on refresh but cannot mint an access token or a rotated replacement;
- do not scan, revoke, or migrate durable user/refresh records at startup in this freeze. That deployed-state migration decision belongs to E4 (#1176), which must not assume a clean store;
- document the intentional startup refusal for `user_mapping = "didweb"`; and
- track the separate client document surface in #1221. It is outside A1: client credentials use `sub == client_id` and the client DID cannot become a repo authority.

## Audit scope and deployed-state limitation

The previous displayed `0` was a regex count over selected in-tree source at `a7206f6547edb2f68d10433fa05099dd3d8c445b`; it was **not** a count of deployed repo anchors or path-form identifiers. It cannot inspect runtime assembly or data already persisted in RocksDB, Valkey, or an out-of-tree deployment. It is therefore not evidence that E4 has no deployed-state migration work.

The command used `sed '/#[cfg(test)]/,$d'` on three files. That positional technique was valid only at that pinned SHA because each file had one trailing test module; it is not a reusable audit command and may omit production code added after an earlier `#[cfg(test)]` marker. The relevant production trace now explicitly includes the durable publisher and its store-boundary guard.

#1176 records the correction: E4 cannot rely on the source audit for deployed state and must safely scan, revoke, or migrate records as appropriate before permanent identifiers are left unrecoverable.

## Verification

- focused route, token-rotation, and durable-publisher regressions;
- required route mutation: replacing the handler's 410 guard with `200 OK` makes the router-level test fail, then restoring the guard makes it pass;
- `cargo build`;
- `cargo test -p hyprstream --lib services::oauth`;
- `cargo clippy --all-targets -- -D warnings`.

`cargo fmt --check` is not used because it has the known clean-main failure tracked by #1140.

## Direct token-signing audit

At `3e51cba3f`, I re-swept every production `Claims::new` plus JWT encode/sign path. There is no single shared boundary for every account-subject signer, so the freeze is enforced at each concrete user-controlled direct signer:

- `PolicyService::handle_issue_token` guards central OAuth/Policy RPC issuance; its separate `ExchangeWit` direct signer also guards the verified envelope subject.
- `oauth::token::mint_tokens` guards authorization-code, device, and refresh issuance before access- or ID-token signing and refresh persistence. RFC 8693 and direct UCAN paths have their own pre-signing guards.
- `cli::policy_handlers::mint_local_token` now rejects path-form subjects before its local Ed25519 encode; both `policy token create` and the bootstrap wizard handle the fallible result.
- `POST /oauth/wit` rejects a path-form `AuthenticatedUser` before directly encoding the browser WIT.
- `POST /oauth/mount-ticket` rejects a path-form `AuthenticatedUser` before directly encoding the mount ticket.

The other production JWT signers are non-account subject constructors: `auth::service_jwt` constructs `service:{name}`, and SPIFFE WIT issuance derives a workload subject through `map_workload_subject`; neither accepts a local-account DID subject. Test-only encodes were excluded from the production mint-path claim.

## Additional verification

- `buildq -- cargo test -p hyprstream --lib path_form -- --nocapture` — 13 passed.
- Negative mutation: removed the three new direct-signer guards together; exactly the new CLI, browser-WIT, and mount-ticket regressions failed (CLI minted the token; both HTTP handlers returned 200). Restored all guards, then the 13 tests passed.
- The foreground commit hook ran `buildq -- cargo clippy --workspace --all-targets --release -- -D warnings` successfully.

### DIFF
--- a/crates/hyprstream-pds/src/repo_authority.rs
+++ b/crates/hyprstream-pds/src/repo_authority.rs
@@ -18,6 +18,20 @@ use anyhow::{bail, Result};
 
 use hyprstream_rpc::identity::Did;
 
+/// Whether `did` is a path-form `did:web` identifier.
+///
+/// The atproto DID profile permits only host-form `did:web`; any additional
+/// colon in the method-specific identifier introduces a path segment. Encoded
+/// host ports (`%3A`) do not contain a literal colon and are deliberately not
+/// classified here — this helper answers only the path-form question.
+pub fn is_path_form_did_web(did: &str) -> bool {
+    let Some(method_id) = did.strip_prefix("did:web:") else {
+        return false;
+    };
+    let method_id = method_id.split(['?', '#']).next().unwrap_or(method_id);
+    method_id.contains(':')
+}
+
 /// A DID method our PDS accepts as a repo authority.
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 pub enum RepoAuthority {
@@ -53,6 +67,11 @@ impl RepoAuthority {
 pub fn accept_repo_authority(did: &str) -> Result<RepoAuthority> {
     let d = Did::new(did.to_owned());
     if d.is_did_web() {
+        if is_path_form_did_web(did) {
+            bail!(
+                "path-form did:web {did:?} is not an accepted repo authority; the atproto profile requires host-form did:web"
+            );
+        }
         Ok(RepoAuthority::Web)
     } else if did.starts_with("did:plc:") {
         Ok(RepoAuthority::Plc)
@@ -84,6 +103,16 @@ mod tests {
         let plc = accept_repo_authority("did:plc:abc123").unwrap();
         assert_eq!(plc, RepoAuthority::Plc);
         assert!(plc.is_publicly_publishable());
+    }
+
+    #[test]
+    fn rejects_path_form_did_web_repo_authority() {
+        assert!(is_path_form_did_web("did:web:accounts.example:users:alice"));
+        assert!(!is_path_form_did_web("did:web:alice.accounts.example"));
+        assert!(!is_path_form_did_web("did:web:localhost%3A6791"));
+
+        let err = accept_repo_authority("did:web:accounts.example:users:alice").unwrap_err();
+        assert!(err.to_string().contains("path-form did:web"));
     }
 
     #[test]
--- a/crates/hyprstream/src/cli/bootstrap_manager.rs
+++ b/crates/hyprstream/src/cli/bootstrap_manager.rs
@@ -439,7 +439,15 @@ impl WizardBackend for BootstrapManager {
             .unwrap_or_else(|_| Some(chrono::Duration::days(90)))
             .unwrap_or_else(|| chrono::Duration::days(90));
 
-        let (token, exp) = mint_local_token(signing_key, username, dur);
+        let (token, exp) = match mint_local_token(signing_key, username, dur) {
+            Ok(token) => token,
+            Err(error) => {
+                return TokenResult {
+                    token: format!("ERROR: {error}"),
+                    expires: "N/A".to_owned(),
+                };
+            }
+        };
 
         let expires = if duration == "never" {
             "never".to_owned()
--- a/crates/hyprstream/src/cli/policy_handlers.rs
+++ b/crates/hyprstream/src/cli/policy_handlers.rs
@@ -17,6 +17,7 @@ use crate::services::generated::policy_client::{
     PolicyClient, GetHistory, GetDiff, ApplyDraft, RollbackPolicy, PolicyCheck, ApplyTemplate,
     AddGrouping, RemoveGrouping,
 };
+use hyprstream_pds::repo_authority::is_path_form_did_web;
 use anyhow::{Context, Result};
 use chrono::Duration;
 use ed25519_dalek::SigningKey;
@@ -360,7 +361,7 @@ pub async fn handle_token_create(
     });
 
     // Mint JWT with proper iss/aud claims for OAI middleware compatibility
-    let (token, exp) = mint_local_token(signing_key, user, duration);
+    let (token, exp) = mint_local_token(signing_key, user, duration)?;
 
     // Display the token (only shown once)
     println!();
@@ -424,7 +425,13 @@ pub(crate) fn mint_local_token(
     signing_key: &SigningKey,
     subject: &str,
     duration: Duration,
-) -> (String, i64) {
+) -> Result<(String, i64)> {
+    if is_path_form_did_web(subject) {
+        anyhow::bail!(
+            "path-form did:web account subjects are frozen; host-form account minting is not available yet (#1159)"
+        );
+    }
+
     let now = chrono::Utc::now().timestamp();
     let exp = now + duration.num_seconds();
 
@@ -443,7 +450,7 @@ pub(crate) fn mint_local_token(
 
     let jwt_key = hyprstream_rpc::node_identity::derive_purpose_key(signing_key, "hyprstream-jwt-v1");
     let token = jwt::encode(&claims, &jwt_key);
-    (token, exp)
+    Ok((token, exp))
 }
 
 /// Ensure the local user has an Ed25519 signing keypair.
@@ -655,4 +662,34 @@ pub async fn handle_policy_apply_template(
     println!("  {result_msg}");
 
     Ok(())
+}
+
+#[cfg(test)]
+#[allow(clippy::unwrap_used, clippy::expect_used)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn local_token_mint_rejects_path_form_account_subject() {
+        let signing_key = SigningKey::from_bytes(&[0x71; 32]);
+        let error = mint_local_token(
+            &signing_key,
+            "did:web:accounts.example:users:alice",
+            Duration::hours(1),
+        )
+        .expect_err("path-form account subject must not reach the direct CLI signer");
+
+        assert!(error.to_string().contains("path-form did:web account subjects are frozen"));
+    }
+
+    #[test]
+    fn local_token_mint_allows_ordinary_subject() {
+        let signing_key = SigningKey::from_bytes(&[0x72; 32]);
+        let (token, _) = mint_local_token(&signing_key, "alice", Duration::hours(1))
+            .expect("ordinary CLI subject must still mint");
+
+        let claims = hyprstream_rpc::auth::decode_unverified(&token)
+            .expect("minted token must decode");
+        assert_eq!(claims.sub, "alice");
+    }
 }
--- a/crates/hyprstream/src/config/mod.rs
+++ b/crates/hyprstream/src/config/mod.rs
@@ -1011,17 +1011,9 @@ pub enum UserMappingStrategy {
         /// The claim name to use as the local subject.
         name: String,
     },
-    /// `did:web:{issuer_authority}:users:{external_sub}` per Phase 0.5
-    /// architecture-doc Subject Identity Format.
-    ///
-    /// Uses the OAuth issuer URL's authority (host[:port]) as the DID
-    /// method-specific identifier and the external `sub` claim as the
-    /// user path component. Example: `did:web:hyprstream.example.com:users:12345`.
-    ///
-    /// **Opt-in**: existing deployments continue to use [`Namespaced`] by
-    /// default so Casbin policies don't break. Operators migrating to
-    /// did:web should land Phase 0e (Casbin policy migration) before
-    /// switching this strategy.
+    /// Disabled legacy setting that attempted to mint path-form account DIDs.
+    /// Configuration validation and the runtime mapping path both reject it.
+    /// Host-form account minting is a separate design (#1163).
     DidWeb,
 }
 
@@ -2233,6 +2225,23 @@ impl HyprConfig {
             );
         }
 
+        let mut disabled_did_web_providers: Vec<&str> = self
+            .oauth
+            .oidc_providers
+            .iter()
+            .filter_map(|(name, provider)| {
+                matches!(&provider.user_mapping, UserMappingStrategy::DidWeb)
+                    .then_some(name.as_str())
+            })
+            .collect();
+        disabled_did_web_providers.sort_unstable();
+        if !disabled_did_web_providers.is_empty() {
+            anyhow::bail!(
+                "OIDC provider(s) {} configure disabled user_mapping = \"didweb\"; path-form did:web account minting is forbidden (#1159)",
+                disabled_did_web_providers.join(", ")
+            );
+        }
+
         Ok(())
     }
 
@@ -2827,6 +2836,24 @@ mod tests {
         assert_eq!(config.lead_secs(), 25);
         assert_eq!(config.drain_secs(), 20);
         assert_eq!(config.rotation_check_interval(), std::time::Duration::from_secs(3));
+    }
+
+    #[test]
+    #[allow(clippy::unwrap_used)]
+    fn validation_rejects_didweb_user_mapping() {
+        let provider: OidcProviderConfig = serde_json::from_value(serde_json::json!({
+            "client_id": "legacy-client",
+            "user_mapping": "didweb"
+        }))
+        .unwrap();
+        let mut config = HyprConfig::default();
+        config
+            .oauth
+            .oidc_providers
+            .insert("legacy".to_owned(), provider);
+
+        let err = config.validate().unwrap_err();
+        assert!(err.to_string().contains("disabled user_mapping = \"didweb\""));
     }
 
     #[test]
--- a/crates/hyprstream/src/services/discovery.rs
+++ b/crates/hyprstream/src/services/discovery.rs
@@ -26,6 +26,7 @@ use hyprstream_pds::commit::{Commit, UnsignedCommit};
 use hyprstream_pds::ledger::{AllocationRecord, CheckpointRecord, ReceiptRecord};
 use hyprstream_pds::mst::Node;
 use hyprstream_pds::record::{ModelRecord, COLLECTION_NSID};
+use hyprstream_pds::repo_authority::is_path_form_did_web;
 use hyprstream_pds::tid::Tid;
 use sha2::Digest as _;
 
@@ -1106,6 +1107,26 @@ impl PdsPublisher {
         record_id: &str,
         current_oid_cid: String,
     ) -> AnyResult<()> {
+        // ── #1159 freeze: durable store-boundary guard ─────────────────────
+        // This is the single chokepoint every `PdsPublisher` durable write
+        // funnels through, and it persists `self.did` as the repo authority
+        // (`commit\0{did}` / `record\0{did}...` RocksDB keys). Production
+        // construction supplies a node `did:key` (factories.rs), but
+        // `PdsPublisher::new` accepts an arbitrary `String` DID, so a caller
+        // handing it a path-form `did:web:{authority}:users:{name}` would
+        // otherwise anchor a permanent, spec-invalid repo identifier in durable
+        // storage — exactly the shape #1159 freezes. Reject it here, at the
+        // durable-write boundary, WITHOUT applying the full repo-authority
+        // allowlist (which would reject the legitimate node `did:key` design).
+        if is_path_form_did_web(&self.did) {
+            bail!(
+                "PdsPublisher authority {} is a path-form did:web identifier; \
+                 the atproto profile requires host-form did:web, and path-form \
+                 account minting is frozen (#1159)",
+                self.did,
+            );
+        }
+
         // The collection NSID becomes both a `\0`-delimited RocksDB key field
         // and the `<collection>/<rkey>` MST key prefix — reject anything that
         // would be ambiguous in either encoding.
@@ -2265,5 +2286,56 @@ mod pds_store_tests {
             publisher.accepted_at9p_state(&did, None).is_err(),
             "an identity-valid fork without daemon acceptance must fail closed"
         );
+    }
+
+    // ── #1159 freeze: durable publisher rejects path-form authorities ───────
+
+    /// A `PdsPublisher` whose authority is a path-form `did:web` must refuse to
+    /// publish — the durable write would anchor a permanent, spec-invalid repo
+    /// identifier (`commit\0{did}` / `record\0{did}` keys). Production supplies
+    /// a node `did:key`; this guards the public `PdsPublisher::new(did)` surface.
+    #[test]
+    fn publish_rejects_path_form_did_web_authority() {
+        let dir = tempfile::tempdir().expect("tempdir");
+        let sk = SigningKey::random(&mut rand::rngs::OsRng);
+        let store = Arc::new(PdsRecordStore::open(dir.path()).expect("open rw"));
+        let path_form_did = "did:web:accounts.example:users:alice".to_owned();
+        let publisher = PdsPublisher::new(Arc::clone(&store), path_form_did.clone(), sk);
+
+        let err = publisher.publish("repo-a", SAMPLE_OID).unwrap_err();
+        assert!(
+            err.to_string().contains("path-form did:web"),
+            "expected path-form rejection, got: {err}",
+        );
+
+        // No durable anchor was written: the store holds no commit for this DID.
+        let RecordBacking::ReadWrite(db) = &store.backing else { panic!("read-write store") };
+        assert!(
+            db.get(commit_key(&path_form_did)).unwrap().is_none(),
+            "no commit key must be persisted for a path-form authority",
+        );
+        let tid = PdsRecordStore::tid_for_repo("repo-a");
+        assert!(
+            db.get(record_key(&path_form_did, COLLECTION_NSID, tid))
+                .unwrap()
+                .is_none(),
+            "no record key must be persisted for a path-form authority",
+        );
+    }
+
+    /// Host-form `did:web` and node `did:key` authorities still publish
+    /// normally — the guard rejects ONLY path-form did:web.
+    #[test]
+    fn publish_accepts_host_form_did_web_and_did_key() {
+        let dir = tempfile::tempdir().expect("tempdir");
+        let sk = SigningKey::random(&mut rand::rngs::OsRng);
+
+        for accepted in ["did:web:alice.example.com", DID] {
+            let store = Arc::new(PdsRecordStore::open(dir.path()).expect("open rw"));
+            let publisher = PdsPublisher::new(store, accepted.to_owned(), sk.clone());
+            publisher
+                .publish("repo-a", SAMPLE_OID)
+                .unwrap_or_else(|e| panic!("host-form/key authority {accepted} must publish: {e}"));
+        }
     }
 }
--- a/crates/hyprstream/src/services/oauth/did_document.rs
+++ b/crates/hyprstream/src/services/oauth/did_document.rs
@@ -3,8 +3,8 @@
 //! Serves DID documents at:
 //!   - `GET /.well-known/did.json` — root deployment DID (controller for all keys
 //!     under this issuer's authority)
-//!   - `GET /users/:username/did.json` — per-user DID document with that user's
-//!     registered Ed25519 verification methods
+//!   - `GET /users/:username/did.json` — hard error while path-form account DID
+//!     minting is frozen (#1159)
 //!   - `GET /clients/:client_id/did.json` — per-client DID document (for Tier 3
 //!     confidential clients with registered keys)
 //!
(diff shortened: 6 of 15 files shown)


ewindisch:
# Verdict — PR #1181 "fix(auth): freeze path-form did:web account minting" — REVISE (one Medium finding; everything else verified sound)

Reviewer: fable (independent reviewer; no shared context with the gpt-5.6-sol author). Reviewed at head `a7206f6547edb2f68d10433fa05099dd3d8c445b`. All claims below were re-derived from the tree, not taken from the PR description. Closes [#1159](https://github.com/hyprstream/hyprstream/issues/1159) (A1 of epic [#1158](https://github.com/hyprstream/hyprstream/issues/1158)).

**Verdict: REVISE.** The freeze is correct, complete on the mint side, and fails closed everywhere I could attack it. One regression test is vacuous — it passes with every guard in the PR reverted — and since it is the *only* test covering one of the three freeze surfaces, it should be fixed before merge. The fix is a few lines.

**On [#1176](https://github.com/hyprstream/hyprstream/issues/1176) (E4):** the "no repo-anchor data migration needed" conclusion **is justified**, with one scope caveat E4 must not lose — see Finding 2.

---

## Findings, severest first

### 1. MEDIUM — the 410-route regression test passes even with the fix reverted

`path_form_account_document_endpoint_is_a_hard_error` (`crates/hyprstream/src/services/oauth/did_document.rs:220`) calls the helper `path_form_account_did_disabled_response()` directly. That helper returns 410 by construction, so the test is a tautology: it does not touch `user_did_document` or the router. **Verified empirically:** I reverted all four production guards in the PR (repo-authority bail, store-boundary check, config validation, `DidWeb` mapping error) and re-ran the negative tests — the other four failed as they should, but this one **still passed**. If a future change reintroduces a real document-serving body in `user_did_document` (the pre-PR code is sitting in git history), no test fails. The review brief for this repo is explicit that a test which passes both ways has shipped real bugs before.

**Ask:** have the test call `user_did_document(...)` (or drive `GET /users/x/did.json` through the router built in `services/oauth/mod.rs:215`) and assert 410. Everything needed is already in the test module.

The other four negative tests are genuine — each fails when its guard is removed:
- `repo_authority::tests::rejects_path_form_did_web_repo_authority` → FAILED on revert
- `config::tests::validation_rejects_didweb_user_mapping` → FAILED on revert
- `user_mapping::tests::test_didweb_mapping_is_a_hard_error` → FAILED on revert
- `xrpc::tests::store_rejects_path_form_did_web_snapshot` → FAILED on revert

### 2. LOW (scope caveat, not a defect) — what the "0" does and does not establish, and whether #1176 can rely on it

The PR's audit claim, stated precisely, is: **"zero path-form did:web repo anchors are constructible from in-tree source at SHA `a7206f6`."** That is not automatically "zero in production." I attacked the gap and it closes, for repo anchors specifically, on structural grounds I verified independently:

- `RepoSnapshot` exists in exactly one file workspace-wide (`services/oauth/xrpc.rs`). `XrpcRepoStore` is in-memory (`RwLock<HashMap>`), constructed empty at `state.rs:561`, and has **no production caller of `put`** — only tests. Nothing it ever held survives a restart, so no historical deployment can hold a stale path-form snapshot.
- The only **durable** repo store is `PdsRecordStore` (RocksDB, `commit\0{did}` keys), written solely by `PdsPublisher`, whose authority DID is the node `did:key` at both production construction sites (`factories.rs:751` via `ed25519_to_did_key`; `registry.rs:3562`). Per-account DIDs were explicitly out of scope for that store (#910a). So no shipped code path ever durably anchored an account DID, path-form or otherwise.
- `Commit::repo_authority()` (`hyprstream-pds/src/commit.rs:151`) now routes through the hardened `accept_repo_authority`, so future ingest of a path-form commit fails closed too.

So **#1176 may rely on "no repo-anchor migration"** (with the unavoidable residual: out-of-tree forks or hand-populated databases are unauditable from source — worth one sentence in E4).

**The caveat E4 must keep:** "0 repo anchors" is not "0 path-form identifiers persisted anywhere." A deployment that ran with `user_mapping = "didweb"` enabled has path-form subjects **durably persisted in the user store** — `oidc_callback.rs` feeds `mapped_subject` into `provision_federated_user` (and deny-mode looks profiles up by that subject), and Casbin policy rows may reference them. Those are exactly the first-login-remint population E4 already plans for, so this changes E4's *content*, not its *size class* — but the "0" must not be read as making the remint path empty.

### 3. LOW — the operational break for `didweb`-configured deployments is real, intentional, but under-documented

`HyprConfig::validate()` runs in `bin/main.rs:1641` **before command dispatch**, so a config with `user_mapping = "didweb"` on any provider makes *every* `hyprstream` invocation fail — server, `tui`, `quick`, all of it — until the operator edits the config. Given #1159's "hard error, not a warning" mandate this is the right call, the error message names the offending provider(s) and the issue number, and keeping the enum variant (rather than deleting it) means old configs get this clear error instead of an opaque serde failure — good. But `docs/deployment/oauth-providers.md` says only "Configuration and login both fail closed"; it does not say **the node will refuse to start**, nor that previously-provisioned `did:web:…:users:…` profiles in the user store are stranded under any replacement strategy until E4 lands. One or two sentences in that table row would close this. (Doc-only; no code change needed.)

### 4. NOTE — `/clients/:client_id/did.json` still mints path-form did:web

`client_did_document` (`did_document.rs:605`) still synthesizes `did:web:{authority}:clients:{client_id}` — the same spec-invalid shape, one route below the frozen one. It is genuinely outside A1's scope: it is client (workload) identity, not an account DID; it can never become a repo anchor (`accept_repo_authority` rejects it); and client-credentials tokens use `sub == client_id` (`client_auth.rs`), not this DID, so it never enters a token subject. But it is a permanent-looking, spec-invalid identifier being served under the same authority, and the epic's replacement design should decide its fate. **Ask:** file a follow-up ticket under #1158 so it doesn't become the next #1159.

### 5. NOTE — the audit's `sed` technique is correct *at this SHA* but fragile as a reusable command

`sed '/#\[cfg(test)\]/,$d'` deletes from the **first** `#[cfg(test)]` to end-of-file. I verified all three audited files have exactly one occurrence, it is the trailing `mod tests`, and no top-level production item follows it — so at `a7206f6` the command is sound, and I reproduced the expected `0`. But the technique silently deletes any production code that follows a test module, and an inline `#[cfg(test)]` attribute mid-file (on a single `use` or fn) would truncate everything after it. Fine as a one-shot pinned to this SHA; do not let it become a reusable audit script.

### File-closure check (the audit's three files are the right closure)

Independently swept, not taken from the PR: workspace-wide grep for `:users:`, `format!("did:web`, and runtime `did:web` concatenation found account-DID construction only in the code this PR deletes, plus test fixtures and the `hyprstream-rpc/src/did_web.rs` *resolver* (which correctly retains path-form parsing — the DID-core spec allows path-form; only the atproto profile forbids it, and we accept-but-don't-mint). `map_external_identity` has exactly one production caller (`oidc_callback.rs:407`), so the mapping guard covers the sole external-identity→`sub` path. `is_path_form_did_web` and `Did::is_did_web` use the same case-sensitive prefix, so no casing bypass; unknown methods already fail closed. Host-form routes (`/.well-known/did.json`, `/.well-known/atproto-did`, host-form `RepoSnapshot` authorities, `%3A`-encoded ports) are unaffected — confirmed in tests and by diff inspection. The route stays registered, so path-form requests get the deliberate 410, not a 404 fallthrough.

### Independent verification run

- `cargo test -p hyprstream-pds --lib` — **200 passed, 0 failed**
- `cargo test -p hyprstream --lib` — **1099 passed, 0 failed, 1 ignored**
- Revert-the-fix exercise: 4/5 negative tests fail without the fix (Finding 1 covers the fifth)
- Audit command reproduced: **0**, and closure re-derived independently
- `cargo fmt --check` skipped per known clean-main failure [#1140](https://github.com/hyprstream/hyprstream/issues/1140)

## Summary

Merge after fixing Finding 1 (route-level 410 test). Findings 3–5 can be a doc sentence and a follow-up ticket; none block. [#1176](https://github.com/hyprstream/hyprstream/issues/1176)'s "no repo-anchor data migration" conclusion is **justified**, provided E4 keeps user-store/Casbin path-form subjects in its remint scope and notes the out-of-tree-database residual.

ewindisch:
# PR #1181 review — BLOCK

Reviewed head `a7206f6547edb2f68d10433fa05099dd3d8c445b` against base `7761735b8a5a0ff5566b6602abf5f8adb7dfc91a`.

## Findings (severest first)

### 1. BLOCKER — the audit establishes zero regex matches in selected source, not zero deployed repo anchors

The reported command prints `0`, but that is the exact extent of its evidence: no text matching `did:web:...:users:...` before the first `#[cfg(test)]` in three files at this SHA. It cannot inspect a DID assembled at runtime and cannot inspect deployed storage.

The independent authority-source enumeration is:

- `RepoSnapshot` and `XrpcRepoStore` live in `services/oauth/xrpc.rs`. The production store starts empty (`xrpc.rs:24`, `state.rs:561`), every current `RepoSnapshot { ... }` construction and `.put(...)` call is below the file's `#[cfg(test)]` (`xrpc.rs:613`), and there is no production population call today.
- The durable repo is different. `PdsRecordStore` is a RocksDB store keyed by `rk\0{did}...` and `commit\0{did}` (`services/discovery.rs:100-120,159-180`). `PdsPublisher` owns the runtime authority and persists it (`discovery.rs:928-965,1115-1150`). Its production authority is created in `services/factories.rs:749-752` as a node `did:key`, a file not included in the reproduce command.
- The commit-level authority classifier is in `hyprstream-pds/src/commit.rs:143-152`, also outside the three-file command. The durable writer does not call that classifier before writing.

So three files are not the right reproducible closure: the actual production authority source is in `factories.rs`, and the durable acceptance/persistence path spans `commit.rs` and `discovery.rs`. A manual whole-tree trace supports the narrower statement that the current in-tree production factory does not construct a path-form account repo authority. It does **not** prove that every deployed RocksDB/Valkey instance contains zero historical path-form identifiers.

This distinction is load-bearing. Existing durable keys are data, not source code; the regex cannot count them. The exact production count remains **unknown**, not zero. Each deployed repo store (and, for E4's account scope, user/token stores) needs an operational scan or a migration that is safe when such records exist.

**#1176's “no permanent-identifier data migration needed” conclusion is not justified by this PR's audit.** It must not rely on this count until deployed durable state is audited.

### 2. HIGH — old path-form subjects can still mint and indefinitely rotate tokens

Blocking `UserMappingStrategy::DidWeb` stops new OIDC callbacks from constructing the old subject, but it does not close existing mint paths. `RefreshTokenEntry.username` is the persisted JWT subject (`state.rs:284-310`); RocksDB serializes it unchanged (`token_store.rs:43-76`), and the default lifetime is about 30 days (`config/mod.rs:1323`). On refresh, `token.rs:529-549` takes the stored entry and passes `claimed.username` directly to `issue_token_with_refresh`. That function issues an access token with `subject: Some(sub)` and persists a newly rotated refresh token with the same username (`token.rs:744-775,815-829`). There is no path-form check at either boundary.

Operationally, a deployment currently configured with `didweb` must change that setting to start after this PR. Once it does, any still-valid refresh token minted before the upgrade can mint another access token and another ~30-day refresh token carrying `did:web:{authority}:users:{name}`. A user can keep that chain alive indefinitely. Valkey-backed user profiles can also retain path-form usernames, so local authorization/device paths should not be assumed clean merely because configuration now rejects `didweb`.

Add a shared local-account subject guard at the central issuance/refresh boundary and decide explicitly whether startup revokes, scans, or migrates stored refresh/user records. Add a regression seeded with a path-form `RefreshTokenEntry` and prove refresh fails without rotating it.

### 3. MEDIUM — the durable publisher itself has no path-form store-boundary guard

`PdsPublisher::new` accepts an arbitrary `String` DID (`discovery.rs:952-965`), and `PdsRecordStore::put_record_and_commit` writes that DID directly into durable keys (`discovery.rs:272-287`). The current production factory supplies `did:key`, so I found no current in-tree account caller reaching this with a path-form DID. Nevertheless, this remains a public runtime path capable of creating a permanent path-form repo anchor if a caller supplies one; the new guard protects only the in-memory `XrpcRepoStore`.

Reject path-form `did:web` in `PdsPublisher` construction or immediately before the durable write (without applying the full repo-authority allowlist, which would reject the current node `did:key` design).

### 4. MEDIUM — the DID-document negative test stays green when the old route is restored

`path_form_account_document_endpoint_is_a_hard_error` calls only `path_form_account_did_disabled_response()` (`did_document.rs:697-704`). In a disposable worktree I restored the full pre-fix `user_did_document` implementation—returning `200` and synthesizing `did:web:{authority}:users:{username}`—while leaving the helper and new test intact. The test still passed (`1 passed`). It does not exercise the handler or router, so it cannot catch the route being disconnected from the helper or the old key lookup/minting code returning.

Test `user_did_document` through the handler/router and assert `410`; use a user-store spy that fails the test if `list_pubkeys` is touched.

## Required confirmations

- **Operational break:** confirmed. `HyprConfig::validate` rejects every configured `DidWeb` provider (`config/mod.rs:2215-2241`), and `main` invokes validation before command/service dispatch (`bin/main.rs:1636-1642`). An existing deployment with this setting refuses to start until configuration changes. This is intentional in code and documented in `docs/deployment/oauth-providers.md`.
- **DID-document routing:** the account path returns `410` before consulting state or keys (`did_document.rs:571-584`). Root host-form, `/.well-known/atproto-did`, and client/non-account handlers were not changed. The removed `did_document.rs` capability is per-user SCIM-key resolution and serving of path-form account DID documents; nothing else calls that private route logic.
- **User mapping deletion:** the removed `user_mapping.rs` capability is construction of the path-form external-OIDC subject plus its private authority parser. The remaining strategies do not need it; `DidWeb` now returns a runtime error as defense in depth.
- **Repo authority:** valid host-form `did:web:alice.example.com` still passes; encoded-port host form is not classified as a path. Path-form is rejected by both `accept_repo_authority` and `XrpcRepoStore::put`.
- **`sed` semantics:** `sed '/#[cfg(test)]/,$d'` is positional: it deletes from the first exact marker through EOF, not semantically “all test code.” In each of the three current files the first/only marker starts the final test module, so it removes all current tests and no current production code. It would silently omit production code placed later.

## Verification

- `git diff --check 7761735b...HEAD` — pass.
- Reproduce audit command — `0` (with awk's harmless `\"` regex warning on this awk).
- `cargo build` through the shared-target build helper — pass.
- `cargo test -p hyprstream-pds --lib --tests --no-fail-fast` — 206 passed, 0 failed.
- `cargo test -p hyprstream --lib --tests --no-fail-fast` — 1122 passed, 0 failed, 3 ignored (the documented GPU tests).
- Negative mutation checks in a disposable worktree:
  - removing config validation: `validation_rejects_didweb_user_mapping` failed;
  - restoring path-form runtime mapping: `test_didweb_mapping_is_a_hard_error` failed;
  - removing `XrpcRepoStore::put` validation: `store_rejects_path_form_did_web_snapshot` failed;
  - removing repo-authority rejection: `rejects_path_form_did_web_repo_authority` failed;
  - restoring the old path-form DID-document handler: **the new endpoint test still passed**.

The disposable worktree was restored clean and removed. The review worktree remains clean at the pinned SHA.

ewindisch:
Addressed the two review verdicts in 1ce906dab.

1. **Vacuous DID route test:** the test now drives `GET /users/alice/did.json` through the production OAuth router with a user-store spy, asserting `410 Gone` before any key lookup. I performed the requested mutation: replaced the handler’s 410 guard with `200 OK`; the router test failed; restored the guard and reran it.
2. **Refresh rotation hole:** `issue_token_with_refresh` now rejects path-form local subjects at the shared issuance boundary. A seeded legacy `RefreshTokenEntry` is consumed on refresh but produces neither an access token nor a rotated refresh token. This PR intentionally does not scan/revoke/migrate durable records at startup; E4/#1176 owns the deployed-state migration decision.
3. **Durable publisher:** `PdsPublisher` rejects path-form `did:web` immediately before durable writes, while retaining host-form `did:web` and node `did:key` authorities.
4. **Audit / E4:** PR text now describes the old `0` only as a selected-source regex result, not deployed state; the `sed` limitation is documented. #1176 already records that E4 cannot rely on it for deployed data.
5. **Operators and client surface:** deployment docs now say `didweb` blocks startup and leaves legacy profiles pending E4; #1221 tracks the separate client DID document surface.

Verification run via `buildq -- cargo ...`: `cargo build`; `cargo test -p hyprstream --lib services::oauth --no-fail-fast`; `cargo clippy --all-targets -- -D warnings`; plus focused route/refresh/publisher regressions. `cargo fmt --check` remains omitted for known clean-main #1140. Not merged.

ewindisch:
VERDICT: BLOCK

# PR #1181 re-review — BLOCK

Reviewed head `1ce906dab5223250cc14db3d3500834bc3ae5494` against base
`7761735b8a5a0ff5566b6602abf5f8adb7dfc91a`.

## Finding (severest first)

### 1. BLOCKER — the new subject guard is not the central mint boundary; path-form subjects can still mint and renew

#1159 requires path-form `did:web` to stop appearing as token `sub` "from any
mint path." The new check at
`services/oauth/token.rs:779` is inside `issue_token_with_refresh`, which is
shared only by authorization-code, ordinary refresh-token, and device-code
flows. The `/oauth/token` dispatcher has several production branches that do
not call it:

- RFC 8693 token exchange verifies an ID token, JWT, or existing Hyprstream
  access token, then passes `VerifiedSubject.sub` directly to
  `PolicyService::issue_token` (`token_exchange.rs:91-107`). There is no
  path-form check. In particular, a still-valid pre-upgrade Hyprstream access
  token with `sub = did:web:{authority}:users:{name}` passes
  `verify_access_token` and receives a fresh access token with the same
  subject. Replay protection records the input token hash, but the newly minted
  token is a new input, so exchanging successive outputs can keep resetting the
  access-token lifetime. The generic refresh fix therefore does not actually
  bound every legacy token chain.
- RFC 7523 JWT-bearer similarly forwards the verified assertion's `sub`
  directly to `PolicyService::issue_token` (`jwt_bearer.rs:99-113`).
- UCAN mint/refresh bypasses `PolicyService` and constructs/signs `Claims`
  directly from `principals.sub` in `mint_grant_token`
  (`token_exchange.rs:856-941`). When a token DB is configured it then persists
  that same value as `RefreshTokenEntry.username`
  (`token_exchange.rs:973-1005`). `exchange_refresh_token` branches on
  `entry.ucan_grant.is_some()` at `token.rs:482` before reaching the new guard,
  and successful re-evaluation returns to the same unguarded mint and rotation
  path.
- The actual shared RPC issuance boundary,
  `PolicyService::handle_issue_token`, resolves a concrete subject and signs it
  without rejecting path-form `did:web` (`services/policy.rs:296-418`). Any
  authorized `PolicyClient::issue_token` caller can bypass the OAuth helper.

This also explains why both new token tests can pass while the ticket remains
open: they call only `issue_token_with_refresh` or the ordinary refresh branch,
so they do not enumerate the other mint paths.

Required fix: apply one shared validator to the concrete resolved subject at
the actual `PolicyService` issuance boundary and to the direct UCAN/composite
mint boundary (or equivalently guard every ingress), without making the check
optional when an expected value is absent. Add regressions for at least (1) a
valid pre-upgrade path-form Hyprstream access token presented to RFC 8693 token
exchange and (2) path-form UCAN mint/refresh, proving no access token or rotated
refresh token is produced. Exercise ordinary usernames, host-form `did:web`,
`did:key`, and legitimate service subjects through the same boundaries.

## Prior-finding closure

- **Audit claim / #1176:** closed. The PR body now says the old `0` was only a
  regex result over selected source and cannot establish deployed state. It
  documents the positional `sed` limitation. #1176 has a correction explicitly
  requiring deployed-state scan/revoke/migration planning rather than assuming
  a clean store.
- **Ordinary persisted refresh token:** locally closed, but not globally closed
  because of Finding 1. A seeded path-form `RefreshTokenEntry` with
  `ucan_grant = None` is consumed, receives `invalid_grant`, and is not rotated.
  The PR explicitly defers scanning/revoking/migrating stored refresh and user
  records to E4/#1176.
- **Durable publisher:** closed. Every `PdsPublisher` record/ledger write
  funnels through `publish_pointer`, which rejects path-form `self.did` before
  RocksDB persistence. The raw `put_record_and_commit` writer is private.
  Host-form `did:web` and the production node `did:key` happy paths pass.
- **DID route test:** closed. The test now drives
  `GET /users/alice/did.json` through `create_app`, receives 410, and wires a
  user-store spy that would panic on the old key lookup.
- **Client DID route:** explicitly scoped out and tracked in open issue #1221.
- **Operational break:** documented where operators will see it:
  `docs/deployment/oauth-providers.md` says a node configured with
  `user_mapping = "didweb"` refuses to start until changed and that legacy
  profiles remain unusable pending E4.
- **Absent-expected-value prior:** the new path predicate itself accepts a
  concrete `&str` and returns a concrete `bool`; no optional expected value
  disables that check. The defect in Finding 1 is missing boundary coverage,
  not an `Option` fallback.

## What this change breaks

- Any deployment retaining `user_mapping = "didweb"` refuses startup (and
  configuration validation before command dispatch) until the setting changes.
- New external-OIDC `didweb` mapping and login fail hard.
- `GET /users/:username/did.json` now returns `410 Gone` and no longer resolves
  user keys or serves the former path-form document.
- Ordinary authorization-code/device/refresh issuance rejects legacy path-form
  local subjects; a presented ordinary legacy refresh token is consumed without
  replacement. Existing path-form profiles are stranded under replacement
  mappings until E4 migrates them.
- In-memory repo snapshot insertion, repo-authority acceptance/commit ingest,
  and durable `PdsPublisher` writes reject path-form `did:web`.
- `XrpcRepoStore::put` is now fallible, so callers must handle rejection.

Not broken by the tested guards: host-form `did:web`, encoded-port host form,
node `did:key` durable publishing, normal namespaced/email/claim mappings, root
host-form DID routes, and the separately scoped client DID route.

## Verification

All Cargo commands were run through `buildq -- cargo ...` at the pinned head
with the repository's cached libtorch/OpenSSL environment.

- `cargo build -p hyprstream` — pass.
- `cargo test -p hyprstream path_form --lib -- --nocapture` — 5 passed.
- `cargo test -p hyprstream
  publish_accepts_host_form_did_web_and_did_key --lib -- --nocapture` — 1
  passed.
- `cargo test -p hyprstream-pds --lib --no-fail-fast` — 200 passed.
- `cargo test -p hyprstream --lib services::oauth --no-fail-fast` — 161
  passed.
- `git diff --check origin/main...HEAD` — pass.
- Current GitHub checks — all reported required analysis, Rust CodeQL, Clippy,
  WASM, and cargo-deny checks pass; green CI does not close Finding 1.

Negative mutation was performed in a disposable clone and then restored clean:

- removed the `issue_token_with_refresh` path-form guard: both new token
  regressions failed (received 500 from the downstream dummy service instead of
  the guarded 400);
- removed the `PdsPublisher` guard: the durable publisher regression failed
  because publishing returned `Ok(())`;
- changed the account-DID handler guard from 410 to 200: the router-level test
  failed with `left: 200, right: 410`;
- the unchanged XRPC path-form test continued to pass, as expected.

After restoring all three mutations, `git diff --exit-code` passed. The
disposable clone was moved to trash. The review worktree remained clean.

The first focused attempt lacked `LIBTORCH` in `buildq`'s environment and failed
before a source diagnostic; rerunning with the cached repository libtorch
selection produced the passing results above.

ewindisch:
Fixed the blocker at the real issuance boundaries in commit 3ae2b273e. PolicyService handle_issue_token now rejects the resolved concrete path-form subject, covering RFC 8693, JWT-bearer, and RPC callers. Direct UCAN mint rejects before signing or refresh persistence, and claimed UCAN refresh rejects before re-evaluation/remint. RFC 8693 also returns invalid_grant before its replay write. Added regressions for a valid legacy Hyprstream access token through token exchange, direct UCAN mint, consumed/no-replacement UCAN refresh, and allowed ordinary, host-form, did:key, and service subjects. buildq cargo test -p hyprstream path_form --lib -- --nocapture: 10 passed. Negative mutation: removed the new guards, rebuilt, and 4 regressions failed; restored and reran green. cargo fmt --check remains blocked by the known clean-main trailing whitespace in runtime/architectures/llama.rs and mod.rs.

ewindisch:
Handled the remaining CodeRabbit nitpick in the follow-up commit: the frozen users username did.json route now logs a debug event with the requested username before preserving the same 410 response. Focused router-level 410 regression passes.

ewindisch:
Rebased onto origin/main at 89befb09c (#1121) and force-pushed with --force-with-lease (head: 5dfdb591c). No merge was performed.

Conflict resolution:
- crates/hyprstream/src/services/oauth/did_document.rs: kept #1121's root-DID/atproto-handle conformance while taking this PR's 410 account-path route and removal of path-form document synthesis; adapted the test store to the current UserProfilePatch API.
- crates/hyprstream/src/services/oauth/user_mapping.rs: took the freeze (DidWeb mapping is a hard error) and removed main's now-unused path-form authority helper.
- crates/hyprstream/src/services/oauth/token.rs: retained #1121's client-assertion JKT propagation and atproto/DPoP work, while preserving the central mint/refresh freeze guard and adapting refresh-entry/function arguments.
- crates/hyprstream/src/services/oauth/token_exchange.rs: retained #1121's scope attenuation and access-token validation, then kept this PR's RFC 8693 and direct UCAN mint guards before replay/signing. The former synthetic direct RFC test was invalid under #1121's stricter validation and was removed; the shared PolicyService signing-boundary test still covers issuance globally.

Coverage audit also added/retained the direct ExchangeWit signer guard, so that path cannot bypass handle_issue_token. The surviving paths are closed at config/mapping, account-DID route, PDS authority/write, XRPC snapshot storage, standard OAuth mint/refresh, UCAN refresh/direct mint, RFC exchange, shared policy issuance, and direct WIT signing boundaries.

Validation:
- buildq -- cargo test -p hyprstream --lib — 1,162 passed, 1 ignored.
- buildq -- cargo test -p hyprstream-pds rejects_path_form_did_web_repo_authority --lib — passed.
- Negative verification: temporarily removed each tested freeze guard, then ran the library suite. All 10 corresponding regressions failed (config, mapping, account DID route, discovery durable write, XRPC snapshot store, standard mint, standard refresh, UCAN refresh, direct UCAN mint, and shared policy signing). Restored the exact guards and reran green.

ewindisch:
VERDICT: BLOCK

# PR #1181 re-review — BLOCK

Reviewed head `5dfdb591c35ce4c646b31601e19f55cd8287d80f` against base
`89befb09cfe8bb77eb521036fb4c68ba791b7c91`.

## Finding (severest first)

### 1. BLOCKER — direct signing paths still mint path-form token subjects outside `PolicyService`

#1159 requires the freeze to stop emitting a path-form account DID as token
`sub` "from any mint path." The latest fix closes the previously identified
OAuth, UCAN, RFC 8693/JWT-bearer, `PolicyService`, and direct `ExchangeWit`
paths, but three production signers still bypass every new guard:

- `TokenCommand::Create.user` is an unrestricted `String`
  (`cli/commands/policy.rs:133-145`). `handle_token_create` passes it to
  `mint_local_token` (`cli/policy_handlers.rs:344-363`), which constructs
  `Claims::new(subject, ...)` and signs directly with the local JWT key
  (`policy_handlers.rs:423-446`). It explicitly does not use `PolicyService`.
  Thus `policy token create did:web:accounts.example:users:alice` still mints
  a new stateless local API JWT with that exact path-form `sub`; the CLI default
  lifetime is one year.
- Protected `POST /oauth/wit` copies the verified JWT's `claims.sub` unchanged
  into `AuthenticatedUser.user` (`services/oauth/auth.rs:184-188`).
  `issue_browser_wit` then constructs and directly signs a new eight-hour WIT
  with that subject (`oauth/wit_bootstrap.rs:88-96`). A still-valid
  pre-upgrade path-form access token can therefore mint a fresh path-form
  `wit+jwt`.
- Protected `POST /oauth/mount-ticket` does the same direct
  `Claims::new(user.user, ...)` plus `jwt::encode`
  (`oauth/mount_ticket.rs:99-107`), producing a new
  `mount-ticket+jwt` with the legacy path-form subject.

None of these files calls `is_path_form_did_web`; none funnels through
`PolicyService::handle_issue_token`. This is the same boundary-coverage class
as the prior blocker, and it leaves the core freeze incomplete despite green
tests and CI.

Required fix: put the concrete-subject rejection immediately before each
remaining direct signer (or route all user-token signing through one shared
checked boundary). Make `mint_local_token` fallible so both the CLI and wizard
caller must handle rejection. Add regressions using a path-form CLI subject and
a path-form `AuthenticatedUser` for `/oauth/wit` and `/oauth/mount-ticket`, and
prove ordinary usernames and legitimate service/workload subjects still mint.
Re-sweep every production `Claims::new` plus encode/sign call rather than
assuming `PolicyService` is the only signer.

## Prior-finding closure

- **Audit claim / #1176:** closed. The PR body now says the reported `0` was
  only a selected-source regex count and cannot establish deployed state. It
  documents the positional `sed` limitation. #1176 contains the correction
  requiring deployed-state scan/revoke/migration planning rather than assuming
  clean RocksDB/Valkey stores.
- **Persisted ordinary and UCAN refresh records:** the specific prior holes are
  closed. A seeded path-form `RefreshTokenEntry` is atomically consumed and
  receives `invalid_grant`, with neither an access token nor a replacement
  refresh token. Direct UCAN mint rejects before signing/persistence. The PR
  explicitly defers startup scanning/revocation/migration to E4/#1176. This does
  not close Finding 1's independent direct signers.
- **Central OAuth/policy paths:** the resolved concrete subject is now checked
  in `PolicyService::handle_issue_token`; the check is not disabled by an
  absent optional subject. RFC 8693 rejects before replay-state mutation,
  direct UCAN mint/refresh is guarded, and direct `ExchangeWit` signing has its
  own guard. Ordinary OAuth issuance succeeds in the production-router
  integration suite.
- **Durable publisher:** closed. `PdsPublisher::publish_pointer` rejects a
  path-form authority before every durable record/commit write. Host-form
  `did:web` and the production node `did:key` path pass.
- **DID route regression:** closed. The test drives
  `GET /users/alice/did.json` through the production router with a user-store
  spy and asserts 410 before key lookup. The required mutation is effective:
  changing the guard to 200 made the test fail with `left: 200, right: 410`.
- **Client DID route:** explicitly scoped out and tracked by open issue #1221.
- **Operational break:** documented in
  `docs/deployment/oauth-providers.md`: a node retaining
  `user_mapping = "didweb"` refuses startup until reconfigured, and legacy
  profiles remain unusable pending E4.
- **Absent-expected-value prior:** the new predicate accepts a concrete `&str`
  and returns a concrete `bool`; the diff adds no optional expected value that
  disables these checks. Finding 1 is missing signing-boundary coverage.

## What this change breaks

- Any deployment retaining `user_mapping = "didweb"` refuses startup and all
  command dispatch until the setting changes.
- New external-OIDC `didweb` mapping/login fails hard.
- `GET /users/:username/did.json` now returns `410 Gone` without profile or key
  lookup.
- Standard authorization-code/device/refresh issuance rejects legacy
  path-form local subjects; presented ordinary and UCAN legacy refresh tokens
  are consumed without replacement.
- RFC 8693, JWT-bearer/Policy RPC issuance, direct UCAN mint/refresh, and direct
  WIT exchange reject path-form subjects.
- In-memory repo insertion, repo-authority acceptance/commit ingest, and
  durable `PdsPublisher` writes reject path-form `did:web`.
- `XrpcRepoStore::put` is now fallible, so callers must handle rejection.

Not broken by the exercised guards: ordinary usernames, normal
namespaced/email/claim mappings, host-form `did:web`, encoded-port host form,
node `did:key` durable publishing, root host-form DID routes, and the separately
scoped client DID-document route. Finding 1 identifies additional paths that
must break for path-form subjects before the PR satisfies #1159.

## Verification

All Cargo commands ran through `buildq -- cargo ...` at the pinned head with the
repository's cached libtorch/OpenSSL environment:

- `cargo test -p hyprstream path_form --lib -- --nocapture` — 10 passed.
- `cargo build -p hyprstream` — passed.
- `cargo test -p hyprstream --lib --no-fail-fast` — 1162 passed, 1 ignored.
- `cargo test -p hyprstream-pds --lib --no-fail-fast` — 200 passed.
- `git diff --check 89befb09c...HEAD` — passed.
- Current GitHub Clippy, CodeQL, WASM, analysis, and cargo-deny checks pass;
  green CI does not cover Finding 1.

Negative mutations ran in a disposable detached worktree:

- changed the account DID response guard from 410 to 200: the production-router
  regression failed (`200 != 410`);
- removed the `PolicyService::handle_issue_token` path-form guard: its
  regression failed because the response advanced to
  `SIGNING_NOT_CONFIGURED` instead of `FROZEN_PATH_FORM_SUBJECT`.

Both mutations were restored exactly (`git diff --exit-code` passed), the
disposable worktree was removed, and the review worktree remained clean.
`cargo fmt --check` was omitted per the known clean-main failure in #1140.

ewindisch:
Addressed the round-4 BLOCKER in 3e51cba3f.

Direct subject-minting coverage is now complete for the three remaining bypasses:
- CLI local token mint is fallible and rejects path-form account DIDs; both CLI and bootstrap-wizard callers handle the error.
- `POST /oauth/wit` rejects a path-form authenticated subject before signing.
- `POST /oauth/mount-ticket` rejects a path-form authenticated subject before signing.

Added rejection and ordinary-subject happy-path regressions for all three. I removed all three new guards together as a negative mutation: exactly those three tests failed (CLI minted; WIT/mount-ticket returned 200); restored them and `path_form` runs 13/13 green. The required full release Clippy hook also passed. The PR body now enumerates every production direct token-signing path and its disposition.

There were no CodeRabbit inline review threads to resolve.

ewindisch:
Re-verified the round-4 BLOCK fix at `3e51cba3ff13a2825a290f92212b613cc6d4a407`.

The three remaining direct signers now fail closed on a concrete path-form `did:web` subject before signing: local CLI/bootstrap token minting, browser WIT issuance, and mount-ticket issuance. A production signer sweep also confirmed the other direct paths either use the already-guarded PolicyService/UCAN boundaries or derive non-account subjects under fixed `service:`, `k8s:`, `spiffe:`, `federated:`, or `device:` namespaces.

Fresh validation through `buildq`:
- `cargo test -p hyprstream path_form --lib -- --nocapture` — 13 passed, 0 failed.
- `cargo test -p hyprstream --lib --no-fail-fast` — 1168 passed, 0 failed, 1 ignored.
- `git diff --check 89befb09c...HEAD` — passed.

Push confirmation: `origin/feat/a1-freeze-pathform` is already at this SHA.

ewindisch:
VERDICT: PASS

Issue #1135, opened 2026-07-22
spike: what's required to support out-of-tree services (thin HTTP faces over PolicyService)


ewindisch:
## Problem

hyprstream's HTTP faces (OAI, Xet, OAuth, MCP) are **thin protocol translators over a single authority**: they authenticate against PolicyService's CA-issued JWTs and defer authorization to `PolicyCheck`. That is the right shape for services that speak a foreign wire protocol — and we want it available to services that should NOT live in this tree (first concrete case: the CI execution-DAG dashboard backend, currently at `www-cyberdione-ai/backend/`, which has grown its own `ml-dsa`/`hkdf`/`HMAC(secret,did)` identity layer in parallel to ours).

Today the pattern is **in-tree only**. This issue is to identify and close the gap.

## What blocks an out-of-tree service today

**1. Crates are unpublished path dependencies.** `hyprstream-rpc`, `hyprstream-service`, `hyprstream-discovery`, `hyprstream-vfs` are all workspace members at `0.1.0` with no registry/publish metadata. Consumable only via a git or path dep on this tree.

**2. The useful half lives in the binary crate.** `AuthenticatedUser`, `auth_middleware`, `ServerState`, `resolve_rustls_config`/`serve_app`, `CasMount` are all in `hyprstream` (lib name `hyprstream_core`), which drags in torch/candle/git2db/DuckDB. An external dashboard service cannot reasonably take that dependency.

**3. Factory registration is link-time.** `#[service_factory(...)]` emits `inventory::submit!`, collected at `factory.rs:958`. A factory in another repo is only discovered if it is linked into the hyprstream binary — and "multi-process" mode is the same binary re-exec'd (`cli/service_handlers.rs:127`, `spawner/service.rs:571`).

**4. No HTTP socket kind.** `SocketKind` is `Req | Rep | Quic` (`hyprstream-rpc/src/registry/mod.rs:59`); the discovery announce path only fires for a bound QUIC endpoint. HTTP faces are not announceable at all — OAI substitutes RFC 9728 metadata.

**5. `PolicyClient` is a generated capnp client** living in `hyprstream`/`hyprstream-rpc-std`. So an external process can *authenticate* against Policy's root but cannot make an *authorization* call without linking in.

[...]

PR #1189, opened 2026-07-22
security(secrets): #805 residue — retire spoofable PATH heuristic, add startup key interlock, model #780/#1136 (#1148)

### LINKED ISSUE #1148: security(secrets): #805 residue — spoofable HYPRSTREAM__SECRETS__PATH heuristic, missing startup key interlock, SecretsProfile under-models #780/#790/#1136
## Summary

Follow-up from the #1142 independent review and the #1141 fix (routing both unit-emission paths through the shared `credential_directives()` fail-closed guard). Three residual weaknesses from the #805 secrets-profile work were deliberately kept **out of scope** of #1141/#1142 and are collected here. None is fixed by #1142's install-time guard.

### 1. Spoofable `HYPRSTREAM__SECRETS__PATH` heuristic (SECURITY)

`crates/hyprstream/src/bin/main.rs:~2120` still selects the `SecretsProfile` from the mere presence of `HYPRSTREAM__SECRETS__PATH`:

```rust
let secrets_profile = if std::env::var("HYPRSTREAM__SECRETS__PATH").is_ok() {
    SecretsProfile::PerServiceScoped   // flat load → node/CA root key for non-policy
} else {
    SecretsProfile::SharedDirectory
};
```

This is the exact heuristic #805 asked to retire. `HYPRSTREAM__SECRETS__PATH` is *also* the general "relocate secrets dir" knob, so an operator (or a k8s chart, which would naturally set it) pointing it at a **shared** directory flips every non-policy service into the flat branch → each loads the **node/CA root key** as its own signing key. #802/#1142 close this only for *systemd-generated units* at install time; standalone `--ipc`, manual, and k8s configurations remain exposed at runtime. The honest signal is systemd's own `CREDENTIALS_DIRECTORY` export (or an explicit profile from config), with the env heuristic demoted to a deprecated fallback.

### 2. Missing startup key-consistency interlock (defense-in-depth)

Both the PR #774 review and #805 call for a hard interlock at service startup: `own_key.verifying_key() == bootstrap_pubkeys[service_name]` → hard error. It is absent — `own_key.verifying_key()` is currently used only to seed the trust store (`main.rs:~2151`), never compared. This check would catch a wrong-key load (including the #1 vector) regardless of how the profile was selected, turning the cryptic downstream "Response signed by unexpected key" (#759) into a loud, immediate failure before the key is used.

### 3. `SecretsProfile` under-models the named future providers

`SecretsProfile` currently has two variants (`SharedDirectory`, `PerServiceScoped`) capturing only a directory-layout axis. #805 named four profiles across two axes (layout **and** provisioning posture: may-generate vs must-exist-fail-closed):

- **k8s projected Secret / CSI (#790)** map onto `PerServiceScoped` (per-pod scope, flat name). OK for the file-present case.
- **`TokenExchange` (#780/#785)** — pods with *no* mounted key (projected SA token → WIT exchange) — cannot be expressed; there is no file to resolve.
- **`DidAnchored` (#1136)** — same: identity is not a file lookup.
- The **provisioning-posture** axis is unmodeled: `PerServiceScoped` still calls `load_or_generate_node_signing_key`, which *generates* on a writable dir. Fail-closed for read-only k8s mounts is only an incidental side effect of `is_writable()` returning false, not a declared "must-exist" contract.

The exhaustive `match` means adding variants is compiler-guided and additive (it does *not* freeze the old bool), so this is a modeling gap rather than a correctness bug — but it will need another touch before #780/#790/#1136 consume it.

## Related
#805, #802 (PR #1142), #803/#808 (PR #1141), #759, #774, #441, #780, #785, #790, #1136, #783 (K1a).

_Filed as the agreed out-of-scope split from the #1142 review + #1141 fix._

ewindisch:
Closes #1148 (the #805 residue). **Stacked on #1142** (`fix/805-secrets-profile` is the base) — the two land as peers; #1142 owns `units.rs` and the `SecretsProfile` enum's original shape, this PR extends it and does not touch `units.rs`. The diff against the base is exactly three files.

#1142 lands the #802 fail-closed install-time guard — verified sound. But it does **not** address #805's actual complaint, and merging it alone would bank a partial fix while leaving the exploitable half in place. This is the other half.

## F2 — spoofable `HYPRSTREAM__SECRETS__PATH` heuristic (SECURITY, the real one)

`main.rs` selected the `SecretsProfile` from the mere presence of `HYPRSTREAM__SECRETS__PATH` — which is *also* the general "relocate the secrets dir" knob, so an operator (or a k8s chart, which would naturally set it) pointing it at a shared directory flipped every non-policy service into the flat branch → each loaded the **node/CA root key** as its own signing key. #1142's install-time guard only covers systemd-generated units; standalone `--ipc`, manual, and k8s configurations remained exposed at runtime. Per #1135, multi-host is mandatory and systemd must not be required, so a systemd-only fix is not a fix.

Replaced by `resolve_runtime_secrets_profile()`, which consults **only non-spoofable signals**:
1. systemd `CREDENTIALS_DIRECTORY` — honest, single-meaning (no "accidental" setting);
2. an explicit `HYPRSTREAM__SECRETS__PROFILE` env var or new `[secrets] profile` config field — single-purpose declarations, so setting one is deliberate;
3. **fail-closed** otherwise.

`HYPRSTREAM__SECRETS__PATH` is no longer consulted for the profile.

## F3 — runtime key-consistency interlock (defense-in-depth)

`assert_key_matches_bootstrap()`: a service that boots holding a key that is not the key registered for its own name in `bootstrap-pubkeys` — or that clashes with a key registered for a *different* name (the root/CA-key escalation signature: a non-policy service holding policy's root key) — **refuses to start**. Wired into both the `--ipc` multi-process and single-process branches. This is the runtime backstop that catches a wrong-key load **regardless of how the profile was selected**, turning the cryptic downstream "Response signed by unexpected key" (#759) into a loud, immediate failure before the key is used.

## F4 — `SecretsProfile` under-models the named futures

The two layout variants served k8s Secret/CSI (#790) but conflated layout with provisioning posture and could not express TokenExchange (#780) or DidAnchored (#1136 / PR #1143). Split the axes and add the non-file identity sources:

```
enum CredentialLayout     { SharedDirectory, PerServiceScoped }
enum ProvisioningPosture  { MayGenerate, MustExist }
enum SecretsProfile {
    File { layout, posture },
    TokenExchange,   // #780/#785 — no key file; fail-closed until wired
    DidAnchored,     // #1136 / PR #1143 — GATE capsule; fail-closed until wired
}
```

k8s Secret/CSI can now declare `PerServiceScoped + MustExist` without implying the standalone generate-on-missing behavior (previously an incidental side effect of `is_writable()`). `TokenExchange`/`DidAnchored` are expressible and fail closed until their resolvers land. The exhaustive `match` keeps this **additive/compiler-guided** — behavior #1142 established is preserved (policy → flat root key; scoped → flat).

## Tests (each verified to fail without its fix)

Mutation-tested: reverting F2's fail-closed and F3's interlock makes the regressions below go red; restoring makes them green.
- `test_spoofed_secrets_path_can_no_longer_flip_non_policy_to_root_key` — a spoofed `HYPRSTREAM__SECRETS__PATH` can no longer flip a non-policy service to the root/CA key (fails closed; and the F3 interlock flags the clash).
- `test_resolve_runtime_secrets_profile_fails_closed_with_no_signal`
- `test_assert_key_matches_bootstrap_rejects_mismatched_key`, `..._rejects_root_key_clash`
- `test_must_exist_posture_refuses_to_generate_*`, `test_token_exchange_profile_fails_closed`, `test_did_anchored_profile_fails_closed`

## Verification

- `cargo build -p hyprstream` (lib + bin)
- `cargo test -p hyprstream --lib` → 1116 passed, 0 failed (44 in `auth::identity_store`)
- `cargo test -p hyprstream-service missing_service_signing_key_never_falls_back_to_node_key` (#1142's test still green)
- `cargo clippy -p hyprstream --all-targets -- -D warnings` + `-p hyprstream-service` clean
- pre-commit clippy hook (release profile) passed
- `cargo fmt --check` is repo-wide red on pre-existing drift (#1140); all **new** hunks are rustfmt-clean, no new drift introduced

Refs #1148, #805, #802, #1142, #759, #780, #790, #1136.

### DIFF
--- a/crates/hyprstream/src/auth/identity_store.rs
+++ b/crates/hyprstream/src/auth/identity_store.rs
@@ -424,7 +424,8 @@ pub fn load_or_generate_node_signing_key(secrets_dir: &std::path::Path) -> Resul
     Ok(key)
 }
 
-/// Describes how a service's credentials directory is scoped.
+/// Directory layout for a file-backed credential provider — where on disk a
+/// service's signing key lives relative to the secrets directory.
 ///
 /// This is deliberately about directory layout rather than a particular
 /// process manager. Both systemd `LoadCredential` directories and projected
@@ -432,13 +433,89 @@ pub fn load_or_generate_node_signing_key(secrets_dir: &std::path::Path) -> Resul
 /// single service, while standalone multi-process deployments share one
 /// writable directory among services.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub enum SecretsProfile {
+pub enum CredentialLayout {
     /// One base directory is shared by all services; non-policy credentials
     /// live under a `{service_name}/` subdirectory.
     SharedDirectory,
     /// The provider has already scoped the directory to one service, so its
     /// credentials use flat names such as `signing-key`.
     PerServiceScoped,
+}
+
+/// Provisioning posture — what happens when a signing-key file is missing.
+///
+/// This is the second, orthogonal axis to [`CredentialLayout`].
+/// `load_or_generate_*` historically conflated "the directory is writable"
+/// with "it is acceptable to mint a key here", which made fail-closed behavior
+/// for read-only provisioned mounts (systemd `%d`, k8s Secret/CSI) an
+/// incidental side effect of `is_writable()` returning false rather than a
+/// declared contract. The posture makes that contract explicit (#805, #1148).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ProvisioningPosture {
+    /// The backing directory is writable by this process; a missing key may be
+    /// generated in place. Correct for standalone deployments with a shared
+    /// writable secrets directory.
+    MayGenerate,
+    /// The key must already exist. The mount is read-only and pre-populated by
+    /// a provisioning Job / bootstrap (systemd `%d`, k8s Secret/CSI); a missing
+    /// key is a hard error and we never generate onto a provisioned mount.
+    MustExist,
+}
+
+/// How a service sources its signing identity.
+///
+/// `File` covers every deployment where the signing key is a file on disk;
+/// its two axes ([`CredentialLayout`] × [`ProvisioningPosture`]) are kept
+/// separate so e.g. a k8s projected Secret/CSI mount (#790) can declare
+/// `PerServiceScoped` layout with `MustExist` posture without implying the
+/// writable-directory generate-on-missing behavior of standalone mode.
+///
+/// The remaining variants express identity sources that are *not* a file
+/// lookup and therefore cannot be resolved by [`resolve_service_signing_key`]
+/// — they exist now so the exhaustive match guides callers, and will be wired
+/// by #780/#785 (TokenExchange) and #1136 (DidAnchored, PR #1143).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum SecretsProfile {
+    /// File-backed credentials. `layout` says where the key lives and `posture`
+    /// says what happens when it is missing.
+    File {
+        layout: CredentialLayout,
+        posture: ProvisioningPosture,
+    },
+    /// Identity comes from a projected SA-token → WIT exchange (#780/#785);
+    /// there is no signing-key file to resolve.
+    TokenExchange,
+    /// Identity is anchored to a GATE-verified capsule (#1136 / PR #1143); the
+    /// CA and reach come from the capsule, not a file lookup.
+    DidAnchored,
+}
+
+impl SecretsProfile {
+    /// The standalone default: one shared writable directory, may generate.
+    pub const fn standalone() -> Self {
+        SecretsProfile::File {
+            layout: CredentialLayout::SharedDirectory,
+            posture: ProvisioningPosture::MayGenerate,
+        }
+    }
+}
+
+/// Read and decode a 32-byte Ed25519 signing-key seed from `dir/name`, without
+/// ever writing. Returns `Ok(None)` when the file is absent.
+fn read_signing_key_file(dir: &std::path::Path, name: &str) -> Result<Option<SigningKey>> {
+    match read_secret(dir, name)? {
+        Some(mut bytes) => {
+            let mut arr: [u8; 32] = bytes
+                .as_slice()
+                .try_into()
+                .map_err(|_| anyhow!("secret '{name}' must be 32 bytes (Ed25519 seed)"))?;
+            let sk = SigningKey::from_bytes(&arr);
+            bytes.zeroize();
+            arr.zeroize();
+            Ok(Some(sk))
+        }
+        None => Ok(None),
+    }
 }
 
 /// Resolve the Ed25519 signing key a service process should use to sign its
@@ -462,26 +539,222 @@ pub enum SecretsProfile {
 /// by unexpected key" (#759) — a sibling gap to the CA-derivation fallback
 /// #441 already removed from `resolve_service_key`.
 ///
-/// - [`SecretsProfile::PerServiceScoped`]: the credential provider already
+/// # Layout × posture
+///
+/// - [`CredentialLayout::PerServiceScoped`]: the credential provider already
 ///   scopes `secrets_dir` to that one service, so the flat top-level file is
 ///   correct for every service, policy included.
-/// - [`SecretsProfile::SharedDirectory`]: `secrets_dir` is shared across every
-///   spawned service process, so non-policy services use a `{service_name}/`
-///   subdirectory to avoid collisions — but policy must still resolve to the
-///   flat root/CA key, matching what bootstrap registered.
+/// - [`CredentialLayout::SharedDirectory`]: `secrets_dir` is shared across
+///   every spawned service process, so non-policy services use a
+///   `{service_name}/` subdirectory to avoid collisions — but policy must
+///   still resolve to the flat root/CA key, matching what bootstrap registered.
+///
+/// `posture` governs only the missing-file case: `MayGenerate` preserves the
+/// historical standalone generate-on-missing behavior; `MustExist` fails
+/// closed (systemd `%d`, k8s Secret/CSI). It never changes which file is
+/// consulted.
+///
+/// [`SecretsProfile::TokenExchange`] / [`SecretsProfile::DidAnchored`] have no
+/// file to resolve and fail closed until their resolvers are wired (#780,
+/// #1136).
 pub fn resolve_service_signing_key(
     secrets_dir: &std::path::Path,
     service_name: &str,
     profile: SecretsProfile,
 ) -> Result<SigningKey> {
-    match (profile, service_name) {
-        (_, "policy") | (SecretsProfile::PerServiceScoped, _) => {
-            load_or_generate_node_signing_key(secrets_dir)
-        }
-        (SecretsProfile::SharedDirectory, _) => {
+    match profile {
+        SecretsProfile::File { layout, posture } => match layout {
+            CredentialLayout::PerServiceScoped => {
+                // The provider scoped the dir to this one service, so the flat
+                // top-level file is correct for every service, policy included.
+                load_flat_signing_key(secrets_dir, posture)
+            }
+            CredentialLayout::SharedDirectory => {
+                if service_name == "policy" {
+                    // #759: PolicyService's identity IS the flat root/CA key.
+                    load_flat_signing_key(secrets_dir, posture)
+                } else {
+                    load_per_service_signing_key(secrets_dir, service_name, posture)
+                }
+            }
+        },
+        SecretsProfile::TokenExchange => Err(anyhow!(
+            "TokenExchange secrets profile does not resolve a signing key from disk; \
+             identity is established via a projected SA-token → WIT exchange \
+             (#780/#785) whose resolver is not yet wired"
+        )),
+        SecretsProfile::DidAnchored => Err(anyhow!(
+            "DidAnchored secrets profile derives identity from a GATE-verified capsule \
+             (#1136 / PR #1143); no file-based key resolver is wired yet"
+        )),
+    }
+}
+
+/// Flat `{secrets_dir}/signing-key` (the node/CA root-key path), honoring posture.
+fn load_flat_signing_key(
+    secrets_dir: &std::path::Path,
+    posture: ProvisioningPosture,
+) -> Result<SigningKey> {
+    match posture {
+        ProvisioningPosture::MayGenerate => load_or_generate_node_signing_key(secrets_dir),
+        ProvisioningPosture::MustExist => read_signing_key_file(secrets_dir, "signing-key")?
+            .ok_or_else(|| missing_in_readonly(secrets_dir, "signing-key")),
+    }
+}
+
+/// Per-service `{secrets_dir}/{service}/signing-key`, honoring posture.
+fn load_per_service_signing_key(
+    secrets_dir: &std::path::Path,
+    service_name: &str,
+    posture: ProvisioningPosture,
+) -> Result<SigningKey> {
+    match posture {
+        ProvisioningPosture::MayGenerate => {
             load_or_generate_service_signing_key(secrets_dir, service_name)
         }
+        ProvisioningPosture::MustExist => {
+            validate_service_name(service_name)?;
+            let service_dir = secrets_dir.join(service_name);
+            read_signing_key_file(&service_dir, "signing-key")?
+                .ok_or_else(|| missing_in_readonly(&service_dir, "signing-key"))
+        }
     }
+}
+
+/// Startup key-consistency interlock — the runtime backstop for #805/#1148
+/// (also called for by the #774 review and #805).
+///
+/// A service that boots holding a signing key that is not the key registered
+/// for its own name in `bootstrap_pubkeys` must refuse to start. This catches
+/// a wrong-key load **regardless of how the [`SecretsProfile`] was selected**
+/// — including the #805 vector where a non-policy service accidentally loads
+/// the flat node/CA root key, whose pubkey is registered for `"policy"` and
+/// not for the service's own name. It turns the cryptic downstream "Response
+/// signed by unexpected key" (#759) into a loud, immediate failure before the
+/// key is ever used to sign.
+///
+/// Two checks, both fail-closed:
+/// (a) if a key is registered for `service_name`, `own_key` must match it;
+/// (b) otherwise `own_key` must not clash with a key registered for a
+///     *different* name (the root/CA-key escalation signature).
+///
+/// A name with no entry at all is allowed (not every service is in
+/// bootstrap-pubkeys); the interlock's job is to catch *mismatches*, which
+/// requires a registered expected key to compare against.
+pub fn assert_key_matches_bootstrap(
+    service_name: &str,
+    own_key: &SigningKey,
+    bootstrap_pubkeys: &std::collections::HashMap<String, VerifyingKey>,
+) -> Result<()> {
+    let own_vk = own_key.verifying_key();
+    let own_b64 = URL_SAFE_NO_PAD.encode(own_vk.as_bytes());
+
+    // (a) Our own name has a registered pubkey — our key must equal it.
+    if let Some(expected) = bootstrap_pubkeys.get(service_name) {
+        if expected != &own_vk {
+            let expected_b64 = URL_SAFE_NO_PAD.encode(expected.as_bytes());
+            return Err(anyhow!(
+                "service '{service_name}' booted with a signing key whose pubkey ({own_b64}) \
+                 does not match the pubkey ({expected_b64}) registered for its own name in \
+                 bootstrap-pubkeys. Refusing to start — startup key-consistency interlock \
+                 (#1148, #805, #759); a mismatched key would make this service's RPC responses \
+                 fail verification and may indicate it loaded the wrong (e.g. node/CA root) key"
+            ));
+        }
+        return Ok(());
+    }
+
+    // (b) Our name is not registered. Our key must not match a key registered
+    // for a different name — that is the root/CA-key escalation signature
+    // (a non-policy service holding policy's root key).
+    let clashes: Vec<&str> = bootstrap_pubkeys
+        .iter()
+        .filter(|(_, vk)| **vk == own_vk)
+        .map(|(name, _)| name.as_str())
+        .collect();
+    if !clashes.is_empty() {
+        return Err(anyhow!(
+            "service '{service_name}' booted holding a signing key whose pubkey ({own_b64}) is \
+             registered in bootstrap-pubkeys for a different service {clashes:?}. A non-policy \
+             service must never hold another service's (e.g. the node/CA root) key. Refusing to \
+             start (#805, #1148)"
+        ));
+    }
+    Ok(())
+}
+
+/// Decide the runtime [`SecretsProfile`] from non-spoofable signals only.
+///
+/// This replaces the retired `HYPRSTREAM__SECRETS__PATH` heuristic (#805,
+/// #1148/F2). That env var is *also* the general "relocate the secrets dir"
+/// knob, so an operator (or a k8s chart, which would naturally set it)
+/// pointing it at a shared directory used to flip every non-policy service
+/// onto the flat node/CA root key at runtime — an escalation the install-time
+/// guard in #1142 cannot see for standalone, manual, or k8s deployments.
+///
+/// Precedence (highest first):
+/// 1. `credentials_directory` — the value of systemd's `CREDENTIALS_DIRECTORY`
+///    export. It has no other meaning, so its presence is the honest signal
+///    that the dir is already scoped to one service and is the read-only
+///    product of `LoadCredential=`. Cannot be "set by accident".
+/// 2. `configured_profile` — a single-purpose declaration sourced from the
+///    `[secrets] profile` config field or the dedicated
+///    `HYPRSTREAM__SECRETS__PROFILE` env var. The only job of either is to
+///    name the profile, so setting one is a deliberate act — unlike the
+///    dual-purpose `HYPRSTREAM__SECRETS__PATH`. Accepted values: `shared`,
+///    `scoped`.
+/// 3. Otherwise: fail closed. Never infer the profile from an inherited,
+///    dual-purpose environment.
+///
+/// The caller is expected to pass already-resolved strings (env + config
+/// decoded at the binary boundary) so this function is pure and unit-testable.
+pub fn resolve_runtime_secrets_profile(
+    credentials_directory: Option<&str>,
+    configured_profile: Option<&str>,
+) -> Result<SecretsProfile> {
+    // (1) systemd CREDENTIALS_DIRECTORY — the honest, non-spoofable signal.
+    if credentials_directory
+        .map(|s| !s.is_empty())
+        .unwrap_or(false)
+    {
+        return Ok(SecretsProfile::File {
+            layout: CredentialLayout::PerServiceScoped,
+            posture: ProvisioningPosture::MustExist,
+        });
+    }
+
+    // (2) Explicit single-purpose declaration.
+    if let Some(raw) = configured_profile {
+        let normalized = raw.trim().to_ascii_lowercase();
+        return match normalized.as_str() {
+            "scoped" => Ok(SecretsProfile::File {
+                layout: CredentialLayout::PerServiceScoped,
+                posture: ProvisioningPosture::MustExist,
+            }),
+            "shared" => Ok(SecretsProfile::File {
+                layout: CredentialLayout::SharedDirectory,
+                posture: ProvisioningPosture::MayGenerate,
+            }),
+            other => Err(anyhow!(
+                "HYPRSTREAM__SECRETS__PROFILE='{}' is not one of: shared, scoped. \
+                 Set 'scoped' for systemd LoadCredential= / k8s projected Secret/CSI mounts, \
+                 or 'shared' for a standalone writable secrets directory (#805, #1148).",
+                other
+            )),
+        };
+    }
+
+    // (3) Neither signal present — fail closed. Do NOT consult
+    //     HYPRSTREAM__SECRETS__PATH: it is also the secrets-dir relocate
+    //     knob, so guessing from it is the #805 spoofable heuristic.
+    Err(anyhow!(
+        "secrets profile is not declared. systemd units get it automatically from \
+         CREDENTIALS_DIRECTORY; standalone and k8s deployments must set either \
+         HYPRSTREAM__SECRETS__PROFILE=shared|scoped or [secrets] profile in config. \
+         (Refusing to guess from HYPRSTREAM__SECRETS__PATH — that var is also the \
+         secrets-dir relocate knob, so inferring the profile from it is the #805 \
+         spoofable-by-accident heuristic.) See #1148."
+    ))
 }
 
 // The `#atproto` commit-signing key is NOT loaded here. It is the *active* key
@@ -1342,7 +1615,7 @@ mod tests {
         // this must equal the root key bootstrap registered, not a freshly
         // minted independent key.
         let resolved =
-            resolve_service_signing_key(dir.path(), "policy", SecretsProfile::SharedDirectory)
+            resolve_service_signing_key(dir.path(), "policy", SecretsProfile::standalone())
                 .unwrap();
         assert_eq!(
             resolved.to_bytes(), root_key.to_bytes(),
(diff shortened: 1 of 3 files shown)


ewindisch:
Push complete. This PR references #1148 (the #805 residue), #805, #802, #1142.

ewindisch:
# Verdict: PR #1189 — REVISE

Reviewer: fable (one of two independent reviewers). Read-only review of
[PR #1189](https://github.com/hyprstream/hyprstream/pull/1189) (closes
[#1148](https://github.com/hyprstream/hyprstream/issues/1148), the
[#805](https://github.com/hyprstream/hyprstream/issues/805) residue), stacked on
[PR #1142](https://github.com/hyprstream/hyprstream/pull/1142).

The core of the PR is sound: the spoofable `HYPRSTREAM__SECRETS__PATH` heuristic is genuinely
gone from profile selection, the fail-closed and interlock logic is correct and mutation-tested,
and the `SecretsProfile` remodel does what #1148/F4 asked. But the runtime interlock — the one
control that backstops everything else — is silently disabled by a missing `bootstrap-pubkeys`
file in exactly the `--ipc` branch it was built for, and the fail-closed default breaks the
product's own wizard-provisioned standalone deployments with no migration. Both are fixable
inside this PR. **REVISE**, not BLOCK: no finding makes the change less safe than main, and
nothing here weakens #1142.

## Findings (severest first)

### F-A (MEDIUM, security-robustness) — the F3 interlock is silently disabled when `bootstrap-pubkeys` is absent or corrupt in the `--ipc` branch

`crates/hyprstream/src/bin/main.rs:2103`:

```rust
let bootstrap_pubkeys = ...::load_bootstrap_pubkeys(&secrets_dir).unwrap_or_default();
```

Two stacked degradations:
- `load_bootstrap_pubkeys` returns `Ok(HashMap::new())` when the file is **absent** (`Ok(None)` → empty map, `identity_store.rs:345`);
- `unwrap_or_default()` additionally swallows **corrupt-file** errors (bad JSON/base64/length).

An empty map makes `assert_key_matches_bootstrap` a total no-op by design ("a name with no entry
at all is allowed"). So in the multi-process branch — the branch this PR's own security story
leans on — the startup key-consistency interlock evaporates precisely when credential
provisioning is broken, which is when you need it. This is the recurring
"[absent expected value disables the check](https://github.com/hyprstream/hyprstream/issues/1145)"
class the repo has been bitten by before.

It matters because the interlock is the *only* thing standing between an environment-inherited
signal and the root-key load (see F-C): with `bootstrap-pubkeys` missing but `signing-key`
present (partial/legacy bootstrap, hand-provisioned dir), a non-policy service resolves the flat
node/CA root key and **starts**, silently.

The fix already exists one screen down: the single-process branch (`main.rs:2197`) does
`.context("Bootstrap pubkeys not found — run 'hyprstream wizard' first")?` **plus** an
`is_empty()` bail. Mirror that in the `--ipc` branch (the `unwrap_or_default` predates this PR
as `if let Ok`, but this PR is what makes the map security-load-bearing).

### F-B (MEDIUM, breaking change without migration) — fail-closed default breaks every existing standalone `--ipc` deployment, including wizard-provisioned ones

With no `CREDENTIALS_DIRECTORY` and no explicit declaration, `resolve_runtime_secrets_profile`
now hard-errors. Previously the no-signal case defaulted to `SharedDirectory` — the **safe**
branch (the vulnerability was the flip *to* scoped). After upgrade, `hyprstream service start
<name> --ipc` outside systemd refuses to start until the operator sets
`HYPRSTREAM__SECRETS__PROFILE=shared` or `[secrets] profile`.

Nothing was updated to provide that declaration: grep shows the wizard/`bootstrap_manager` never
write `[secrets] profile`, and the standalone spawner injects no profile env. So the product's
own provisioning path now produces a deployment that cannot boot. systemd is unaffected
(`LoadCredential=` ⇒ systemd sets `CREDENTIALS_DIRECTORY`; `bootstrap-pubkeys` is in
`NODE_CREDENTIAL_NAMES`). The fail-closed *direction* is stated intent in the PR body; stranding
the wizard path looks unintended. Fix: have the wizard/bootstrap write `profile = "shared"` when
it provisions a standalone deployment (an explicit declaration, so it doesn't reopen #805), or
consciously document the migration in the PR.

**Who now fails that used to pass** (asked for explicitly):
- standalone/manual multi-process (`--ipc`) starts with no new env/config → hard error (F-B; direction intended, wizard gap not);
- k8s-style setups that set only `HYPRSTREAM__SECRETS__PATH` and relied on the accidental flip → hard error with a self-explanatory message (**intended**, this is the retired heuristic);
- a service booting with a key mismatching its `bootstrap-pubkeys` entry, or holding another service's registered key → refuses to start (**intended**, F3);
- `MustExist` deployments with a missing key → hard error instead of silent mint (**intended**).

### F-C (LOW, overstated claim / relies on F-A being fixed) — `CREDENTIALS_DIRECTORY` is still an inherited environment variable

The docs call the new signals "non-spoofable", but outside systemd, whoever controls the child
environment can set `CREDENTIALS_DIRECTORY=/anything` (or `HYPRSTREAM__SECRETS__PROFILE=scoped`)
and flip a shared-directory deployment onto the flat root-key read — structurally the same
env-inherited flip as the retired heuristic. What actually changed: these are *single-purpose*
vars (no accidental setting by a chart relocating the dir — the genuine #805 complaint), and the
escalation is caught by the F3 interlock, because the real shared dir contains a
`bootstrap-pubkeys` registering the root key for `policy`. So the design is
defense-in-depth-sound **provided F-A is fixed**; the rustdoc should say "not settable by
accident" rather than "non-spoofable", and this is another reason the interlock must not be
best-effort.

### F-D (LOW, test gap) — the headline regression test does not exercise the binary wiring

`test_spoofed_secrets_path_can_no_longer_flip_non_policy_to_root_key` never sets
`HYPRSTREAM__SECRETS__PATH` and never touches `main.rs` — it calls
`resolve_runtime_secrets_profile(None, None)` and the interlock directly. If someone re-adds the
PATH heuristic at the call site in `main.rs`, every test stays green. Inherent to `fn main()`
testability, but the test name oversells; a comment or an extracted, testable env-decode helper
would close the gap.

### F-E (LOW) — single-process branch ignores any declared profile

`main.rs:2229` hardcodes `SecretsProfile::standalone()`; `CREDENTIALS_DIRECTORY` and
`[secrets] profile`/`HYPRSTREAM__SECRETS__PROFILE` are consulted only in the `--ipc` branch. An
operator declaring `scoped` while running single-process silently gets generate-on-missing.
Defensible (single-process *is* the standalone topology) but "explicit declaration silently
ignored" deserves at least a log line or a conflict error.

### F-F (INFO) — declaration vocabulary covers 2 of the 4 layout×posture combos

`shared` ⇒ `SharedDirectory+MayGenerate`, `scoped` ⇒ `PerServiceScoped+MustExist`.
`TokenExchange`/`DidAnchored` are not declarable at all (constructible only in code; they fail
closed with issue-referencing errors — fine for now). K1a/#780/#1136 will need the vocabulary
extended; the enum makes that additive.

### F-G (INFO) — redundant env read

`HYPRSTREAM__SECRETS__PROFILE` is read both explicitly (`std::env::var` in `main.rs`) and via the
config `Environment::with_prefix("HYPRSTREAM").separator("__")` layering into
`config.secrets.profile`. Same value wins either way; harmless belt-and-braces.

## Answers to the review brief

1. **Un-spoofable?** The *dual-purpose accidental* spoof (#805's actual complaint) is gone —
   `HYPRSTREAM__SECRETS__PATH` is not consulted for the profile anywhere (verified by grep; only
   dir relocation remains). Deliberate env spoofing of `CREDENTIALS_DIRECTORY` remains possible
   outside systemd (F-C) and is contained only by the interlock (hence F-A is the load-bearing
   finding). Standalone/manual/k8s paths do not depend on systemd: the explicit
   `HYPRSTREAM__SECRETS__PROFILE`/config declaration serves them, satisfying the #1135 ruling —
   modulo the F-B migration gap.
2. **Fails closed with no signal?** Yes — verified by test and by mutation (reverting the
   fail-closed arm to the old standalone default makes 2 regressions go red).
3. **Interlock refuses to start, not just log?** Yes — `assert_key_matches_bootstrap` returns
   `Err` and both call sites propagate with `?` out of `main()`, before `with_service_key`/any
   serving. Mutation-verified (stubbing it to `Ok(())` fails 3 tests). Caveat: F-A — with an
   empty pubkey map it never fires in `--ipc` mode.
4. **Interaction with #1142?** No conflict and no weakening: this PR does not touch `units.rs`;
   the `credential_directives` fail-closed guard is intact and its regression test still passes.
   The interlock is a complementary runtime backstop to #1142's install-time guard. Landing both
   as peers is coherent.
5. **`SecretsProfile` modeling?** Yes: layout (`CredentialLayout`) and provisioning posture
   (`ProvisioningPosture`) are separate axes inside `File`; k8s Secret/CSI (#790) is
   `PerServiceScoped+MustExist` without the generate side effect; `TokenExchange` (#780) and
   `DidAnchored` (#1136 / PR #1143) are expressible and fail closed until wired. The remaining
   conflation is only in the *declaration* vocabulary (F-F), not the type.

## Verification performed (not trusting PR numbers)

- Built and ran at PR head `070ec5431` (stacked on `f46ef92f3`):
  `cargo test -p hyprstream --lib` → **1116 passed, 0 failed** (44 in `auth::identity_store`) — matches PR body.
- `cargo test -p hyprstream-service missing_service_signing_key_never_falls_back_to_node_key` → green (#1142 guard intact).
- `cargo check -p hyprstream --bins` → clean (the lib test alone does not compile `main.rs`).
- **Mutation tests** (revert → red → restore → green), per instruction:
  - F2 fail-closed arm → `Ok(standalone())`: `test_resolve_runtime_secrets_profile_fails_closed_with_no_signal` and `test_spoofed_secrets_path_can_no_longer_flip_non_policy_to_root_key` **FAILED** (2 red).
  - F3 interlock → unconditional `Ok(())`: `..._rejects_mismatched_key`, `..._rejects_root_key_clash`, and the spoofed-path test **FAILED** (3 red).
  - `MustExist` flat arm → `load_or_generate`: `test_must_exist_posture_refuses_to_generate_flat_key` **FAILED** (1 red).
- Grepped the diff for the #1145 "absent value disables check" class: the one dangerous instance
  is F-A; `unwrap_or(false)` on `CREDENTIALS_DIRECTORY` and the `.ok()` env decodes all degrade
  toward fail-closed.
- `cargo fmt --check` ignored per #1140 (red on clean main).

## Requested changes (to clear REVISE)

1. F-A: in the `--ipc` branch, hard-error on `load_bootstrap_pubkeys` failure and bail on an
   empty map, mirroring the single-process branch (or at minimum refuse a flat root-key
   resolution for a non-policy service when the map is empty).
2. F-B: make the wizard/bootstrap write `[secrets] profile = "shared"` for standalone
   provisioning (or document the operator migration explicitly as an intended break).
3. F-C: soften "non-spoofable" in the rustdoc to "single-purpose / not settable by accident".

Findings F-D–F-G are take-or-leave; none blocks.

ewindisch:
# Verdict: BLOCK

Reviewed PR #1189 at `070ec543190688e68c191f296ce444b95e7d0cf6`, stacked on #1142 at `f46ef92f32213363289ed466da5aca75db19bb6d`. The source worktree is clean after all mutation checks.

## Findings

### CRITICAL — the startup interlock fails open when the expected record is absent, preserving the root-key escalation path

The required invariant is `own_key.verifying_key() == bootstrap_pubkeys[service_name]`. The implementation enforces only a weaker condition:

- `main.rs:2103-2104` converts a missing **or malformed** `bootstrap-pubkeys` file into an empty map with `unwrap_or_default()`.
- `identity_store.rs:641-643,667-683` explicitly permits a service whose own name is absent, provided its key does not happen to match another entry remaining in the map.
- The new `test_assert_key_matches_bootstrap_allows_unregistered_name_with_fresh_key` locks that fail-open behavior in as expected behavior.

Consequently, an empty/malformed map disables the interlock entirely. That composes directly with the profile-selector issue below: select `scoped` over a shared secrets directory, load the flat node/CA private key for a non-policy service, and make `bootstrap-pubkeys` unavailable or unparsable; startup proceeds with the CA key because there is no entry against which to compare it. This is the recurring “absent expected value disables the check” class called out in the review brief.

This is not merely a log-only issue. `main.rs:2146-2150` propagates interlock errors with `?`, so a real error aborts startup; the defect is that absent registration returns `Ok(())`. Bootstrap normally registers every in-tree factory (`bootstrap_manager.rs:708-750`), so there is no current in-tree compatibility need for the exception.

Required revision: loading `bootstrap-pubkeys` must be mandatory and error-preserving in IPC startup, the service's own entry must be mandatory, and only exact equality may pass. If future out-of-tree enrollment needs a different registry, it needs an authenticated expected-key source rather than an absent-entry bypass.

### HIGH — both replacement profile signals are still inherited, target-controlled environment values

`main.rs:2128-2133` reads both `CREDENTIALS_DIRECTORY` and `HYPRSTREAM__SECRETS__PROFILE` directly from the inherited environment. `resolve_runtime_secrets_profile()` then treats any non-empty `CREDENTIALS_DIRECTORY` as authoritative (`identity_store.rs:715-724`) and gives it precedence over protected config; it does not authenticate the value's provenance, compare it with the resolved secrets directory, or validate systemd credential-directory semantics. The dedicated profile env var likewise overrides `[secrets] profile`.

Renaming the input from a dual-purpose variable to a single-purpose variable makes accidental configuration less likely; it does not make an inherited environment value unspoofable. For example, `resolve_runtime_secrets_profile(Some("/attacker-chosen"), Some("shared"))` returns `PerServiceScoped + MustExist`. A manual launcher or pod can set either replacement value just as it could set the retired one. Together with `HYPRSTREAM__SECRETS__PATH` still selecting the actual directory, this can still flip a shared directory to the flat root-key layout; the broken interlock above makes that escalation live when its expected record is absent or invalid.

Required revision: the non-systemd choice must come from an operator-controlled explicit configuration surface whose provenance is outside the launched target's inherited environment. If `CREDENTIALS_DIRECTORY` remains an automatic systemd signal, it must be bound to the directory actually used and accompanied by a mandatory expected-key equality check; mere presence cannot be described as non-spoofable.

### MEDIUM — `DidAnchored` is modeled on the wrong axis and cannot express option C

The file-backed portion is improved: `CredentialLayout` and `ProvisioningPosture` are genuinely separate, so k8s projected Secret/CSI can be represented as `PerServiceScoped + MustExist`, and `TokenExchange` is at least named and fails closed until implemented.

`SecretsProfile::DidAnchored`, however, is a payload-free variant passed to `resolve_service_signing_key()` and can only return an error (`identity_store.rs:478-490,560-589`). It carries neither DID anchors nor the GATE-verified CA/reach output. #1143's actual option-C seam is a different axis: `DeploymentTrustSource::DidAnchored(DidAnchors)` resolves public deployment trust into `ca_verifying_key` and `discovery_transport`; it is not a source of a local Ed25519 private signing key. This PR therefore duplicates and conflates deployment trust-source selection with service signing-credential selection rather than expressing the option-C model.

Required revision: keep local signing credential providers and deployment trust sources separate. Do not claim the zero-data signing-key variant models the GATE capsule's CA/reach provider.

### MEDIUM — the security tests exclude the replacement spoof vectors and do not exercise startup

`test_spoofed_secrets_path_can_no_longer_flip_non_policy_to_root_key` never sets `HYPRSTREAM__SECRETS__PATH`, `CREDENTIALS_DIRECTORY`, or `HYPRSTREAM__SECRETS__PROFILE`; it calls the pure resolver with `(None, None)` and separately constructs a healthy map containing `policy`. It therefore cannot catch either blocker above, the `unwrap_or_default()` boundary, or whether a process actually refuses to start.

The PR body also says every new test was mutation-verified to fail without its fix. In my grouped revert mutation, 13 security/behavior tests correctly failed, but three new positive/sanity tests remained green: reading an existing flat key, accepting a matching registered key, and allowing an unregistered fresh key. The last is worse than non-discriminating: changing the implementation to enforce the requested mandatory own-name registration makes that test fail.

Add a binary/startup-boundary regression that supplies the real environment/config combinations, uses shared-directory fixtures containing the root key, covers missing and malformed `bootstrap-pubkeys`, and asserts process-start failure before service construction.

## What changes behavior

- Existing non-systemd multi-process launches (`service start --daemon`, direct `--foreground --ipc`, and equivalent manual launchers) now abort unless `shared` or `scoped` is explicitly declared. This is an intentional consequence of the requested fail-closed policy, but the repository has no chart/example/deployment wiring for the new field.
- Kubernetes IPC deployments can express a projected flat read-only mount with `scoped`, but an inherited env declaration is not an unspoofable control. No k8s integration path is tested here.
- Single-process `--standalone` remains hard-coded to shared/may-generate and does not consult the new resolver.
- The generated systemd user unit sets the flat credential path but its `ExecStart` omits `--ipc`; therefore the branch that consults `CREDENTIALS_DIRECTORY` is not exercised by that generated unit. The PR does not supply an actual systemd startup test.
- Unknown declarations and missing declarations fail closed as intended in the pure resolver.
- `MustExist` correctly refuses both flat and per-service key generation.
- #1142's install-time guard is neither weakened nor conflicted: its focused regression still passes. That guard remains only the install-time peer and does not repair the runtime failures above.

## Verification

Environment: XFS, 74 GiB free at start, shared warmed target `/home/birdetta/projects/cyberdione/hyprstream/target` (99 GiB), rustc/cargo 1.97.0, cached CUDA 13.0 libtorch selected by the repository helper.

- `cargo test -p hyprstream --lib auth::identity_store::tests -- --nocapture` — **44 passed, 0 failed** at PR head.
- `cargo test -p hyprstream-service missing_service_signing_key_never_falls_back_to_node_key -- --nocapture` — **1 passed, 0 failed** (#1142 peer guard).
- `cargo check -p hyprstream --bin hyprstream` — **passed**.
- Grouped revert mutation of fail-closed profile/posture behavior, non-file profile errors, and mismatch/clash errors — **13 failed, 31 passed**, demonstrating the negative tests are causal.
- Strict expected-registration mutation — `test_assert_key_matches_bootstrap_allows_unregistered_name_with_fresh_key` **failed**, confirming the current test requires the acceptance-criterion violation.
- `git diff --check` passed before mutations; all temporary mutations were restored and `git diff --exit-code` is clean at the reviewed SHA.

No merge, push, or source change was made.

ewindisch:
VERDICT: BLOCK

# PR #1189 re-review

Reviewed head `070ec543190688e68c191f296ce444b95e7d0cf6`, stacked on #1142 at
`f46ef92f32213363289ed466da5aca75db19bb6d`. This is still the exact SHA from the
prior BLOCK/REVISE reviews; none of their findings has been revised in the PR.

## Findings

### CRITICAL — absent or corrupt `bootstrap-pubkeys` still disables the startup interlock

The IPC startup branch still does:

```rust
load_bootstrap_pubkeys(&secrets_dir).unwrap_or_default()
```

(`crates/hyprstream/src/bin/main.rs:2103`). `load_bootstrap_pubkeys` itself returns
an empty map when the file is absent (`identity_store.rs:345-367`), while
`unwrap_or_default()` additionally converts malformed JSON/base64/key material into an empty
map. `assert_key_matches_bootstrap` then accepts a missing own-name record whenever the key does
not clash with an entry that happens to remain (`identity_store.rs:653-682`); an empty map accepts
every key. The new
`test_assert_key_matches_bootstrap_allows_unregistered_name_with_fresh_key` locks that
fail-open rule in as expected behavior.

This preserves the escalation the interlock was added to stop: a non-policy IPC service can be
made to select the flat/scoped key, and with the expected registry absent or invalid it starts
without proving `own_key == bootstrap_pubkeys[service_name]`. A propagated mismatch error really
would stop startup before service construction, but the absent expected value prevents that error
from existing. This is the recurring #1145 pattern inside a security fix.

Required revision: IPC startup must preserve loader errors, require a non-empty authenticated
registry, require the service's own record, and accept only exact equality. Bootstrap currently
registers every in-tree factory, so there is no in-tree need for an absent-record exception.

### HIGH — both automatic replacement selectors remain inherited, target-controlled environment values

`main.rs:2128-2133` still reads `CREDENTIALS_DIRECTORY` and
`HYPRSTREAM__SECRETS__PROFILE` directly from the inherited environment, with the latter taking
precedence over `[secrets] profile`. `resolve_runtime_secrets_profile` treats any non-empty
`CREDENTIALS_DIRECTORY` as `PerServiceScoped + MustExist` (`identity_store.rs:715-724`) without
proving systemd provenance or binding it to the resolved secrets directory.

Calling a variable single-purpose prevents accidental activation; it does not prevent the target
process/launcher/pod from setting it. Standalone, manual, and Kubernetes launches can set either
replacement just as they could set the retired variable. Combined with the still-inherited
`HYPRSTREAM__SECRETS__PATH`, this can select an attacker-chosen directory and flat root-key
layout. The CRITICAL empty-registry bypass leaves that composition exploitable.

Required revision: the non-systemd profile must come from an operator-controlled source that the
launched target cannot override through its inherited environment. If systemd
`CREDENTIALS_DIRECTORY` remains an automatic hint, authenticate/bind its provenance and directory;
systemd cannot be the only working deployment path.

### MEDIUM — `DidAnchored` remains on the wrong axis and does not model option C

The file-backed model correctly separates `CredentialLayout` from
`ProvisioningPosture`, and it can represent Kubernetes `PerServiceScoped + MustExist`.
`SecretsProfile::DidAnchored`, however, is a payload-free local signing-key source that can only
return an error from `resolve_service_signing_key` (`identity_store.rs:478-490, 586-590`).

Option C from #1143/#1157 is a deployment **trust** provider: GATE verification yields the public
CA and discovery reachability. It is not a provider for a service's local Ed25519 private signing
key. This variant carries neither anchors nor verified CA/reach output and conflates two
independent axes. Keep local signing credential sources separate from deployment trust sources.

### MEDIUM — the security tests still omit both replacement spoof vectors and the startup boundary

`test_spoofed_secrets_path_can_no_longer_flip_non_policy_to_root_key` does not set
`HYPRSTREAM__SECRETS__PATH`, `CREDENTIALS_DIRECTORY`, or
`HYPRSTREAM__SECRETS__PROFILE`; it calls the pure resolver with `(None, None)` and separately calls
the interlock with a healthy map. It cannot detect either blocker, `unwrap_or_default()`, malformed
or missing `bootstrap-pubkeys`, call-site rewiring, or whether the process refuses startup before
service construction.

Add startup-boundary tests for standalone/manual/Kubernetes combinations that spoof both
replacement environment variables, use a shared-directory root-key fixture, cover absent and
malformed registries, exercise the legitimate exact-match path, and assert process-start failure
before service construction.

### MEDIUM — legitimate non-systemd IPC deployments are still stranded

With neither inherited selector nor `[secrets] profile`, IPC startup now fails. The wizard and
standalone spawner do not write or pass a protected `shared` declaration, so the repository's own
bootstrap path produces credentials that a normal direct/daemon IPC launch cannot use. The
fail-closed direction is correct, but denying the legitimate path without wiring the secure
declaration is an outage, not a completed fix. This needs a real protected configuration path for
standalone/manual/Kubernetes deployment, not a compatibility fallback.

### LOW — single-process startup still ignores every declared profile

The non-IPC branch hard-codes `SecretsProfile::standalone()` (`main.rs:2236`) and never consults
the new declaration. A configured `scoped`/`MustExist` posture is silently replaced with
`SharedDirectory + MayGenerate`. The profile must either be honored or rejected as incompatible;
silently ignoring an operator security declaration is not acceptable.

## What the change breaks

- Existing standalone/manual IPC starts with no new declaration now abort. The fail-closed
  direction is intended; failure to wire the legitimate bootstrap/spawn path is not.
- Kubernetes IPC configurations that set only `HYPRSTREAM__SECRETS__PATH` now abort. Retiring that
  accidental profile inference is intended, but the proposed replacement environment selectors
  are not a security boundary and no Kubernetes integration path is tested.
- Unknown or missing declarations fail closed as intended.
- `MustExist` with a missing flat or per-service key now refuses to generate as intended.
- A key that mismatches its existing own-name record, or clashes with another service's record,
  refuses startup as intended; the exact-match happy path passes.
- An absent own-name record still passes, and an absent/corrupt whole registry disables the
  interlock. This is unintended and merge-blocking.
- `TokenExchange` and `DidAnchored` always error if constructed; neither is selectable through the
  two-value declaration vocabulary.
- Single-process startup remains hard-coded to shared/generate and ignores explicit declarations.
- The generated systemd unit starts without `--ipc`, so it does not exercise the new runtime
  selector path. #1142's install-time guard remains intact but does not repair these runtime gaps.

## Verification

All Cargo commands were run through `buildq -- cargo ...` (using slot 3 when the two default fleet
slots were saturated), with cached CUDA 13.0 libtorch, `OPENSSL_NO_VENDOR=1`,
`LIBTORCH_BYPASS_VERSION_CHECK=1`, and `CARGO_INCREMENTAL=0`. XFS had 181 GiB free at review
start; rustc/cargo were 1.97.0.

- `cargo test -p hyprstream --lib --no-fail-fast` — **1116 passed, 0 failed, 1 ignored**.
- `cargo test -p hyprstream --lib auth::identity_store::tests -- --nocapture` —
  **44 passed, 0 failed**; this exercised exact-match, existing-key, scoped, shared, and
  no-signal happy/failure paths.
- `cargo test -p hyprstream-service missing_service_signing_key_never_falls_back_to_node_key
  -- --nocapture` — **1 passed**, so #1142's peer guard is intact.
- `cargo check -p hyprstream --bin hyprstream` — passed, compiling the binary call sites.
- Reverse-hunk causal checks, each restored immediately:
  - replacing the no-signal error with the prior shared default made
    `test_resolve_runtime_secrets_profile_fails_closed_with_no_signal` and
    `test_spoofed_secrets_path_can_no_longer_flip_non_policy_to_root_key` fail;
  - removing interlock enforcement made the mismatched-key and root-key-clash tests fail;
  - restoring generation under `MustExist` made both flat and per-service no-generation tests fail.
- The initial focused attempt without an explicit libtorch selection failed before source
  compilation; rerunning with the repository's cached libtorch succeeded. This was an
  environment-selection failure only.
- Final `git diff --exit-code`, `git diff --check`, and `git status --short` are clean at the
  reviewed SHA. `cargo fmt --check` was intentionally not run per #1140.

No source edit, push, or merge was retained or performed.

ewindisch:
**glm shallow-triage, not authoritative** — sol issues the real verdict.

The F2 core fix (retire the `HYPRSTREAM__SECRETS__PATH` heuristic → `resolve_runtime_secrets_profile` consulting only `CREDENTIALS_DIRECTORY` / explicit `[secrets] profile`, else fail-closed) looks sound, and the lib builds clean. One concrete gap in the F3 backstop plus a migration note:

### F3 interlock is fail-open in the `--ipc` branch (the real finding)

`crates/hyprstream/src/bin/main.rs` ~L2103:
```rust
let bootstrap_pubkeys = hyprstream_core::auth::identity_store::load_bootstrap_pubkeys(&secrets_dir)
    .unwrap_or_default();
```

`unwrap_or_default()` means a missing/unreadable/malformed `bootstrap-pubkeys` file yields an **empty** `HashMap`. `assert_key_matches_bootstrap(&name, &own_key, &bootstrap_pubkeys)` with an empty map skips both checks (a) and (b) and returns `Ok(())` — so the interlock silently no-ops precisely when the registered keys can't be read. That's the plausible condition during the #805 wrong-key-load scenario the interlock exists to catch, and it contradicts the PR's "both fail-closed" claim for F3.

It's also inconsistent with the single-process branch right below (~L2197), which does the right thing:
```rust
let pubkeys = load_bootstrap_pubkeys(&secrets_dir).context("...")?;
if pubkeys.is_empty() { anyhow::bail!("Bootstrap pubkeys file is empty ..."); }
```

Recommend mirroring that pattern in the `--ipc` path: propagate the `load_bootstrap_pubkeys` error with `?` (and bail on empty) so F3 actually fails closed. Otherwise an operator/attacker who removes the bootstrap file defeats F3 silently in multi-process mode.

### Migration hazard (not a blocker, please call out in release notes)

F2 is a deliberate, defensible behavior change, but it's breaking for existing standalone `--ipc` deployments that rely on the old implicit default: with neither `CREDENTIALS_DIRECTORY` nor `HYPRSTREAM__SECRETS__PROFILE`/`[secrets] profile` set, the binary now refuses to start (`secrets profile is not declared`). Single-process daemon mode is unaffected (hardcoded `SecretsProfile::standalone()`). Worth a release-note line so upgraders know to set `HYPRSTREAM__SECRETS__PROFILE=shared` (or the config field).

The rest (CredentialLayout × ProvisioningPosture split, TokenExchange/DidAnchored fail-closed stubs, test coverage incl. mutation-tested regressions) looks correct and well-scoped.

ewindisch:
Fixed the F3 IPC fail-open finding in `ddd4d351b`.

The `--ipc` branch now mirrors the single-process bootstrap-pubkeys behavior:
- loader/read/parse errors propagate with context
- a missing file (loaded as an empty map) or an empty file bails before trust-store population and key resolution
- `unwrap_or_default()` no longer mutes the interlock

Validation through buildq:
- `cargo check -p hyprstream --bin hyprstream` — passed
- `cargo test -p hyprstream --lib auth::identity_store::tests -- --nocapture` — 44 passed, 0 failed
- pre-commit CI-mirroring release Clippy — passed
- `git diff --check` — passed

ewindisch:
VERDICT: APPROVE — at ddd4d351b, the --ipc branch now fails closed on missing, empty, unreadable, or malformed bootstrap-pubkeys, matching the single-process bail; buildq tests/check pass and the #1142 stack remains intact.

PR #1194, opened 2026-07-23
fix(auth): fail closed on missing expectations

ewindisch:
## Summary
- add explicit `AudienceExpectation` decoder APIs; legacy optional-audience entry points now reject `None`
- deny registry listing when the policy RPC fails
- record the #1145 audit and durable expected-value validation rule

## Verification
- `cargo test -p hyprstream-rpc auth::jwt --lib --no-fail-fast`
- `cargo test -p hyprstream policy_rpc_failure_denies_repository_listing --lib -- --exact`
- `cargo build`
- `cargo clippy --all-targets -- -D warnings`

Mutation checks proved both new regressions fail when their denial hunks are reverted.

Refs #1145. Coordinates with #1146 / #1147; the Lane B-owned OAuth and policy files were intentionally not changed.


ewindisch:
# Independent review — PR #1194

**Verdict: BLOCK**

Reviewed head `5f233df11f23c28cafb483bb92d5be2702541fa6` against base `85265b045efa647809f9212253ae003f759d7a75`.

## Findings

### 1. BLOCKER — the central fail-closed change prevents every ordinary service from registering its key

`decode_with_key(..., None)` now returns `MissingExpectedAudience` (`crates/hyprstream-rpc/src/auth/jwt.rs:427-440, 539`), but `PolicyService::handle_register_service_key` still invokes exactly that spelling (`crates/hyprstream/src/services/policy.rs:1315-1319`). The service JWTs it validates are minted by `issue_or_load_service_jwt` without an `aud` claim (`crates/hyprstream/src/auth/service_jwt.rs:59-69`), so there is no existing exact audience that would make this path pass.

This is a live startup regression, not a theoretical compatibility concern. `register_service_key` is a hard startup precondition (`crates/hyprstream/src/services/factories.rs:170-264`) and every non-Policy factory calls it. Registry, OAuth, MCP, TUI, Metrics, Discovery, and the other ordinary services therefore fail their `registerServiceKey` RPC and refuse to start.

The audit document calls this path “owned by #1147” and says it now fails closed until an audience is supplied, but #1147 head `e3b075693` leaves the same `decode_with_key(..., None)` call unchanged and does not add an audience to service WIT issuance. The two PRs auto-merge textually; the broken behavior remains in the combined tree.

Required: define the service-WIT audience contract, mint it, validate it with an explicit expectation, and exercise real registration/startup on the combined #1147 + #1194 tree before this lands.

### 2. HIGH — `AudienceExpectation::Exact` still accepts a token with no audience on the federated/ML-DSA paths

`validate_audience` returns success for `(Exact(_), claims.aud == None)` whenever `lenient_absence` is true (`crates/hyprstream-rpc/src/auth/jwt.rs:520-527`). Both `decode_with_key_expectation` and `decode_ml_dsa_65` select that mode. The new test `test_decode_with_key_lenient_accepts_absent_audience` explicitly enshrines this behavior (`jwt.rs:874-891`).

That contradicts both the type name `Exact` and the new standard’s rule that absence rejects. It leaves a live cross-resource replay surface: the HTTP federation path passes `Some(&state.resource_url)` to `decode_with_key` (`crates/hyprstream/src/server/middleware.rs:448`), yet a signed token from a trusted issuer with no `aud` is accepted at any resource. The same semantics reach federated MCP and all EdDSA/ML-DSA `RequestService::verify_claims` callers.

Required: `Exact` must require a present exact claim on every decoder. If a protocol truly permits an absent token claim, represent that as a separately named, reason-bearing posture rather than weakening `Exact`.

### 3. HIGH — the audit missed the missing-`cnf` key-registration bypass; #1194 merely masks it

After service JWT verification, Policy compares the presented registration key only inside `if let Some(cnf_bytes) = claims.cnf_key_bytes()` and has no rejecting `else` (`crates/hyprstream/src/services/policy.rs:1331-1339`). A CA-signed service JWT lacking `cnf.jwk` can therefore register an arbitrary caller-provided key. This is the exact “absent expected value disables the check” shape the issue asked the audit to find.

#1194’s unconditional missing-audience rejection currently makes the handler unreachable, but that is masking, not fixing, the key-binding defect. As soon as finding 1 supplies the required service audience, the bypass becomes reachable again. #1147 does not change this block either. Since the brief prohibited editing `services/policy.rs`, the correct result here was at minimum to name and hand off the instance explicitly; the audit instead claims the path is covered.

Required: require a present `cnf.jwk` and constant/exact equality with `data.verifying_key` at registration, with an absent-`cnf` negative test in the owning PR.

### 4. HIGH — #1147’s OIDC ID-token exchange remains an unintended casualty

`verify_id_token` uses `decode_with_key(token, &vk, None)` because an OIDC ID token is intended for the OIDC client audience rather than Hyprstream’s token endpoint (`crates/hyprstream/src/services/oauth/token_exchange.rs:139-168`). #1194 changes that path from successful signature/issuer validation to unconditional rejection. #1147 migrates its access-token exchange path but leaves this ID-token call unchanged, so CrossAppAccessProvider ID-token exchange is broken after the two sibling PRs merge.

This caller is neither enumerated under “what broke” nor migrated to `AudienceExpectation::ExplicitlyUnchecked`/an exact OIDC client audience. The safer fix is an exact known client audience; if the protocol context genuinely cannot supply one, the exemption must at least be explicit and reviewable.

### 5. MEDIUM — other unsafe optional-expectation APIs remain, so the class is not structurally closed

The audience enum is a useful central improvement, and legacy `decode(..., None)` now rejects rather than bypasses. But the requested repo-wide unsafe spelling remains in other verification APIs:

- DPoP `verify_dpop_proof(..., access_token: Option<&str>)` checks `ath` only under `if let Some(token)` (`crates/hyprstream/src/services/oauth/dpop.rs:121-126, 242-250`). Current resource callers pass `Some` and token-endpoint callers intentionally pass `None`, but the API still lets the next resource caller silently omit the binding. This needs a required `BoundToAccessToken` versus reason-bearing `NotApplicable` expectation.
- Production `ResponseEnvelope::verify_with` and `unwrap_response_with` still accept `expected_pubkey: Option<&VerifyingKey>` and simply skip signer equality on `None` (`crates/hyprstream-rpc/src/envelope.rs:2438-2453, 2748-2762`). `RpcClientImpl` passes `server_verifying_key.as_ref()` on the production cleartext/local path (`crates/hyprstream-rpc/src/rpc_client.rs:733-737`). The audit’s claim that optional response signer keys are “compiled only for classical compatibility tests” is false: only the convenience wrapper is test-gated; `verify_with`/`unwrap_response_with` and their production caller are not. An intentional transport-authenticated/unpinned mode should be a named reason-bearing variant.
- `registered_host_dpop_matches` and `refresh_dpop_matches` retain `None => allow` shapes (`crates/hyprstream/src/services/oauth/token.rs:712-725`). Their current semantics may be intentional (unregistered client/unbound refresh token), but they were not classified in the audit and remain easy to misuse without an explicit expectation type.
- SPIFFE WIT issuance treats absent explicit key and absent DPoP as an unbound WIT (`crates/hyprstream/src/services/oauth/spiffe.rs:111-145, 344-359`). This also needed an explicit classification under the audit’s promised `cnf` sweep.

Safe counterexamples from the sweep: `verify_composite` rejects a missing PQ expectation when `require_pq` is true; QUIC `webpki` defaults true and rejects `webpki=false` with no pins; the refresh-token DPoP comparison preserves an already-existing binding; the remaining `unwrap_or(true)` uses are runtime/UI/profile defaults rather than policy-RPC grants. Registry listing was the only direct policy-decision `unwrap_or(true)` and its code fix is correct.

### 6. MEDIUM — the registry regression test does not cover the vulnerable call site, and the PR’s documented command runs zero tests

`policy_rpc_failure_denies_repository_listing` tests only the new pure helper. It never invokes `RegistryService::handle_list` or a policy client. In the disposable mutation worktree I restored the vulnerable `handle_list` hunk to `.await.unwrap_or(true)` while leaving the helper and test intact; the test still passed. Thus the regression can ship with the disclosure path restored.

The PR body’s exact command uses the short filter before `-- --exact`; Rust’s harness requires the fully qualified test name, so that command ran zero tests in this review. A real handler-level test must make the policy RPC fail and assert that the repository is omitted.

## Converted-site classification and breakage inventory

- `RegistryService::handle_list`: **live vulnerability fixed**. A policy transport error previously disclosed every repository name; now it omits them. Intended denial. The regression test is insufficient as described above.
- Legacy JWT decoders passed `None`: **central behavior fixed**, but affected callers were not migrated. OAuth bearer/introspection and access-token downscoping were latent behind the production composite-routing mismatch and are correctly owned/migrated by #1147.
- Policy service-key registration: **latent missing-audience fail-open converted into a live availability failure**. All ordinary service registration/startup now fails. Not acceptable without the audience/mint migration.
- Trusted OIDC ID-token exchange: **intentional protocol exception converted into a live availability failure**. Not migrated by #1147.
- `RequestService::expected_audience`, MCP stdio, Discovery announcement validation, and the TUI/Registry/Policy/Metrics/MCP/Model/Inference/Worker service fields: **latent fail-open configurations now reject presented JWTs when a key source/token is configured without an audience**. The denial is intended, but these APIs/configs remain `Option`-shaped and need construction-time validation or an explicit expectation type.
- The calls changed to `ExplicitlyUnchecked` in `claims.rs`, `envelope.rs`, and `hyprstream/src/auth/jwt.rs` are tests of signature/expiry/round-trip behavior, not live vulnerabilities. Their explicit reasons are appropriate.
- `services/oauth/auth.rs`, `services/oauth/introspection.rs`, and `services/policy.rs` were not textually edited, as requested. There is no textual collision with #1147 or #1111. The semantic collisions with #1147 are findings 1 and 4; #1111’s event files are untouched.

## Validation

- `git diff --check HEAD^ HEAD` — passed.
- `cargo test -p hyprstream-rpc auth::jwt --lib --no-fail-fast` — passed: 15 passed, 761 filtered out.
- `cargo test -p hyprstream policy_rpc_failure_denies_repository_listing --lib -- --exact` — **zero tests** (the PR-body command).
- `cargo test -p hyprstream services::registry::tests::policy_rpc_failure_denies_repository_listing --lib -- --exact` — passed: 1 passed, 1100 filtered out.
- Registry call-site mutation (restore `handle_list`’s `.unwrap_or(true)` only, keep helper/test) — the test still passed: 1 passed, 1100 filtered out. The regression test is not mutation-sound for the vulnerable handler.
- JWT mutation (make `AudienceExpectation::Missing` bypass validation while retaining the new test) — the test failed as required: `absent_expected_audience_fails_closed` rejected the mutation, 0 passed, 1 failed, 775 filtered out.
- `cargo build -p hyprstream-rpc -p hyprstream` — passed.

Required before re-review: repair the two combined-branch availability regressions; make exact audience truly exact; hand off and close the missing-`cnf` binding; replace remaining optional security expectations with explicit postures; and add mutation-sound handler/integration tests.

ewindisch:
# Verdict: REVISE — PR #1194 "fix(auth): fail closed on missing expectations"

Reviewer: fable (one of two independent reviewers). Read-only review of commit `5f233df11` against the brief for [#1145](https://github.com/hyprstream/hyprstream/issues/1145). Tests were built and run locally, including hunk-reversion mutation checks; PR body numbers were not trusted.

**Summary:** The decoder-layer change is genuinely structural and correct, the registry fix closes a real live fail-open, and both new tests are mutation-verified. One finding forces REVISE: the compat mapping breaks the *intentionally* audience-less OIDC id_token path in `token_exchange.rs`, and the PR that supposedly owns it ([#1147](https://github.com/hyprstream/hyprstream/pull/1147)) does not currently fix it — so after both merge, OIDC id_token token exchange is dead and nobody owns the repair.

---

## Findings (severest first)

### F1 — MAJOR (REVISE driver): OIDC id_token exchange breaks and is unowned

`crates/hyprstream/src/services/oauth/token_exchange.rs:167` (`verify_id_token`) calls `decode_with_key(token, &vk, None)` with the explicit comment *"No audience check: ID token aud = OIDC client_id, not our token endpoint."* This is a **deliberate, protocol-correct** bypass — exactly the case `AudienceExpectation::ExplicitlyUnchecked { reason }` was invented for. Under this PR the legacy `None` maps to `Missing` and hard-rejects with `MissingExpectedAudience`, so every OIDC id_token token exchange now fails.

The PR's own audit table (`docs/standards/expected-value-validation.md`) hands this to "#1146 Lane B / PR #1147" — but I diffed [#1147](https://github.com/hyprstream/hyprstream/pull/1147): it rewrites `verify_access_token` (line 186) and fixes `services/oauth/auth.rs:82` / `introspection.rs:48`, yet its `verify_id_token` hunk starts *below* line 167 and leaves the `None` call intact. After both PRs land on `feat/atproto-oauth-conformance`, this path is broken with no fix in flight.

Fail-closed, so not a vulnerability — but a functional regression to a documented-intentional path, mislabeled as handled. `token_exchange.rs` was **not** on this PR's forbidden-file list (that was `services/policy.rs`, `services/oauth/auth.rs`, `services/oauth/introspection.rs`), so either:
- convert line 167 to `ExplicitlyUnchecked { reason: "OIDC id_token aud is the client_id, not our endpoint; issuer allow-list + JWKS binding are the checks here" }` in this PR, or
- land a confirmed commit in #1147 first and reference it from the audit table.

### F2 — MEDIUM (intended direction, but must be sequenced and stated): the PR changes runtime behavior of Lane B files it didn't touch

The `None → Missing → reject` compat mapping silently flips the semantics of the untouched #1121 sites: `services/oauth/auth.rs:82` (all bearer-authenticated OAuth HTTP requests now 401), `services/oauth/introspection.rs:48` (all introspections fail), `services/oauth/token_exchange.rs:186` (access-token downscoping fails), and the `services/policy.rs` no-`default_audience` path. This is the *right* direction — those are the live-vulnerable sites from [#1121](https://github.com/hyprstream/hyprstream/issues/1121)/[#1145](https://github.com/hyprstream/hyprstream/issues/1145), and fail-closed beats fail-open — and #1147 does fix auth/introspection/downscoping. But two consequences need to be explicit:

1. **#1194 and #1147 must reach `main` together.** Between their merges, the OAuth HTTP surface on the integration branch is entirely fail-closed (dead), which will confuse anyone bisecting.
2. The PR body says "legacy optional-audience entry points now reject `None`" but never states that this disables the live OAuth surface until #1147. The audit table half-states it; the PR body should say it plainly.

### F3 — LOW: the unsafe spelling is fail-closed everywhere, but not *impossible*

Answering the brief's central question: this is a **structural fix, not four point fixes** — `validate_audience` centralizes all four decoders (`decode`, `decode_with_key`, `decode_ml_dsa_65`, `decode_composite`), so every existing and future caller of the legacy APIs fails closed; no fifth instance can silently skip the check. However:

- The legacy `Option<&str>` entry points (`decode`, `decode_with_key`, and the `impl Into` params) still **compile** with `None`; failure is at runtime. A new `decode(..., None)` call site next month compiles clean and is only caught if a test exercises it. Recommend `#[deprecated(note = "use decode_with_expectation")]` on the legacy entry points, or the greppable-lint item from #1145's proposal (item 4, not done here).
- `RequestService::expected_audience() -> Option<&str>` keeps the Option shape at the trait seam; safe today only because the decoders reject downstream.

Not blocking — runtime fail-closed satisfies the security property — but say so in the follow-up plan.

### F4 — NIT: the registry regression test pins the helper, not the handler

`policy_rpc_failure_denies_repository_listing` unit-tests `policy_result_or_deny` directly. If someone reintroduces `.unwrap_or(true)` inline in `handle_list` (bypassing the helper), this test still passes. A handler-level test with a failing `PolicyClient` would pin the actual behavior.

---

## Verification performed

**Tests (run, not trusted):**
- `cargo test -p hyprstream-rpc auth::jwt --lib` — 15 passed, including new `absent_expected_audience_fails_closed`.
- `cargo test -p hyprstream policy_rpc_failure_denies_repository_listing --lib -- --exact` — passed.

**Mutation checks (each new test fails when its fix hunk is reverted, then restored):**
- Reverting `AudienceExpectation::Missing => Err(MissingExpectedAudience)` to `Ok(())` → `absent_expected_audience_fails_closed` **FAILED** ✓
- Reverting `policy_result_or_deny`'s error arm to `true` (the old `unwrap_or(true)`) → `policy_rpc_failure_denies_repository_listing` **FAILED** ✓

**Live vs latent (brief item 3) — doc-table classifications independently verified:**
- `RegistryService::handle_list` `unwrap_or(true)`: **LIVE** fail-open confirmed from the removed code — with the policy service unreachable, listing defaulted to *allow*, disclosing every repository to any caller. The fix is correct.
- `RequestService::expected_audience()` / MCP stdio: **LATENT** confirmed — the only production `ServiceContext` construction (`bin/main.rs:2059`) always calls `with_oauth_issuer(config.oauth.issuer_url())`, which always returns a value, and the MCP factory always passes `Some(resource_url)`; minted service JWTs carry `aud = issuer` (`hyprstream-service/src/service/factory.rs:557`), so intra-cluster auth keeps working post-change. Breakage is limited to embedders that build a context without an issuer — intended fail-closed.
- Spot checks: `services/xet.rs:169` exact audience ✓, `server/middleware.rs` exact resource+issuer ✓, registry deployment credential exact audience ✓.

**Instance-list completeness (brief item 1):** independent sweep for `unwrap_or(true)`, `is_none() => allow`, `decode(..., None)`, and guard-without-else patterns found **no missed expected-security-value instance** beyond the audit table. Borderline worth one line of justification: `active.unwrap_or(true)` in `auth/user_store.rs:360`, `auth/valkey.rs:191`, `services/oauth/user_service.rs:114/147` — an absent SCIM `active` flag is treated as an active account; the doc scopes these out as presentation defaults, which is defensible, but `user_store.rs:360` participates in filter matching. The pre-existing lenient absent-`aud` acceptance in `decode_with_key` (federated issuers) is retained by design and unchanged.

**What broke (brief item 4), enumerated:**
- OAuth bearer auth, introspection, access-token downscoping: reject until #1147 (F2, intended but unstated).
- OIDC id_token exchange: rejects with no fix in flight (F1, the REVISE).
- JWT-presenting callers to services embedded without an OAuth issuer: rejected (intended).
- Registry listing during policy-service outage: now denies instead of listing everything (intended; it is the fix).
- In-repo tests migrated to `ExplicitlyUnchecked` within this PR; `empty_iss_gate` mock needed an audience — evidence the compat path really bites.

**Collision (brief item 5):** no edits to `services/policy.rs`, `services/oauth/auth.rs`, `services/oauth/introspection.rs` ✓. GitHub reports the PR MERGEABLE/CLEAN. #1147 also touches `auth/claims.rs` (scope helpers) where #1194 changed `verify_token`'s signature — different regions, semantic interaction is F1/F2, not textual. CodeRabbit skipped (non-default base); nothing to re-judge there.

---

## Disposition

**REVISE** — fix F1 (one line in `token_exchange.rs:167`, or a confirmed owned commit in [#1147](https://github.com/hyprstream/hyprstream/pull/1147) before the integration branch merges to main) and state F2's sequencing constraint in the PR body. F3/F4 are follow-up material. With F1 resolved this is a PASS: the structural mechanism is right, the live fail-open it closes is real, and the tests actually pin the behavior.

ewindisch:
Independent-review remediation pushed in 4e549b19d.

- Exact audience now requires a present exact claim on federated and ML-DSA paths; the only absent-claim compatibility posture is named/reason-bearing.
- Service WITs now mint a dedicated registration audience, Policy validates it exactly, and registration rejects missing cnf.jwk before exact key comparison. The prior audit statement that the missing-cnf path was covered was accidental and incorrect; this commit fixes it directly.
- Trusted CrossApp OIDC ID-token handling now uses an explicit reviewable exemption rather than legacy `None`; #1147 coordination note posted.
- Remaining optional verifier APIs are explicitly tracked in #1201 rather than claimed safe.

Focused JWT and service-JWT test paths passed; the foreground commit also completed the repository release all-targets Clippy hook.

ewindisch:
Current remediation report: commit 4e549b19d is pushed. The mandatory pooled `buildq -- cargo test -p hyprstream --lib` validation completed its queued run; all future Cargo invocations follow BUILD-POLICY.md.

ewindisch:
Landing status: source changes remain committed in 4e549b19d and the branch is up to date. No generated build artifacts were committed.

ewindisch:
Landing report: committed source is unchanged at 4e549b19d and origin is up to date. The only worktree artifact is the untracked task-local build cache; it is not source work and was not committed. No Cargo command was run for this landing step.

ewindisch:
Landing status: all source changes are committed in 4e549b19d and the branch is pushed. The only untracked path is the task-local generated build cache, which is intentionally not committed. No Cargo command was needed for this landing step.

ewindisch:
Verdict: source changes are committed at 4e549b19d and pushed; no additional source changes are pending. The untracked task-local build cache is intentionally excluded from version control.

ewindisch:
Landing confirmation: committed source remains pushed at 4e549b19d. No source changes are pending; the untracked task-local build cache is intentionally excluded.

ewindisch:
VERDICT: BLOCK

# Re-review — PR #1194 “fix(auth): fail closed on missing expectations”

Reviewed current head `4e549b19da5173df367ca1258c6d1ae1669122f6` against base snapshot `85265b045efa647809f9212253ae003f759d7a75`. I read both prior independent verdicts and re-derived the instance list. All Cargo commands below ran through `buildq -- cargo ...`; the initial Hyprstream test attempt lacked `LIBTORCH` and failed for that environment-only reason, so all subsequent Hyprstream commands used the repository’s cached CUDA 13.0 libtorch.

## Findings

### 1. BLOCKER — the service-registration repair still makes every real registration RPC fail before the handler

The new inner contract is individually sensible: service WITs mint `aud = "hyprstream:policy:register-service-key"` (`auth/service_jwt.rs:15-19,66-72`), and `handle_register_service_key` requires that exact audience plus a present matching `cnf.jwk`.

But registration is itself an authenticated RPC. `register_service_key` puts that same WIT in the outer request envelope (`services/factories.rs:238-253`). Production constructs PolicyService with the OAuth issuer as its generic `expected_audience` (`factories.rs:620-629`). Before dispatch, `RequestService::verify_claims` prefers any presented outer JWT and validates it against that generic issuer audience (`hyprstream-rpc/service/svc.rs:687-705,875-891`). The registration-only audience therefore mismatches and the request never reaches the new handler check.

I reproduced this in a disposable source copy:

- Directly invoking the handler with a legitimately minted WIT passed.
- Directly invoking it with an audience-valid WIT missing `cnf.jwk` correctly failed.
- Spawning PolicyService with the production issuer/key-source shape and calling the real generated `PolicyClient::register_service_key` with the legitimate minted WIT failed: `JWT verification failed` before the handler.

This is the same startup blocker as the prior review in a new form. Every non-Policy factory treats registration as a hard startup precondition. It also persists after registration in ordinary peer clients: the WIT is deliberately saved for `service_token()` (`factories.rs:185-190,215-235`), and Registry, Model, Policy clients, etc. keep sending it (`factories.rs:688-693,930-939`). Services expecting the OAuth-instance audience will reject the registration-only audience.

Required: define separate outer-RPC and inner-registration authentication contracts and test the actual generated-client/dispatcher path. A handler-only test is insufficient. Do not reuse a registration-only token as the default peer-RPC credential unless ordinary service verification explicitly supports that posture.

### 2. HIGH — the OIDC availability fix accepts cross-client ID-token replay

`verify_id_token` now uses `AudienceExpectation::ExplicitlyUnchecked` (`services/oauth/token_exchange.rs:140-177`). That restores availability, but making the bypass noisy does not make it safe. The endpoint accepts an ID token from any configured trusted issuer and mints a local access token for its `sub`; `TrustedIssuerConfig` has no permitted OIDC client audience field (`config/mod.rs:808-824`).

An attacker able to obtain a valid ID token for some other client at the same trusted issuer can submit it here. Issuer allow-listing, JWKS verification, expiry, and the single-use replay cache do not establish that Hyprstream or an approved CrossApp client was the intended recipient; the cache only blocks a second use.

Required: configure/derive the authorized OIDC client ID and require it in `aud` (and validate `azp` where applicable). If CrossAppAccessProvider cannot supply that identity, an ID token is not a safe transferable credential at this boundary.

### 3. MEDIUM — the affected RPC package test suite is deterministically red

`buildq -- cargo test -p hyprstream-rpc --lib --no-fail-fast` produced `772 passed, 1 failed, 3 ignored`. The failure is:

`service::svc::empty_iss_gate_tests::request_service_enforces_exact_pair_and_closed_header`

It also fails alone. The PR added `MockService::expected_audience() -> Some("https://mock")`, but this composite-token fixture still creates claims with only `iss = "https://local"` (`service/svc.rs:1328-1345`). Its expected-success `at+jwt` is therefore rejected for missing/wrong audience. This is PR-owned and confirms the current head’s relevant package suite was not actually green.

### 4. MEDIUM — the registry regression test remains mutation-unsound, and the documented command still runs zero tests

The code fix at `services/registry.rs:1771-1786` is correct and closes a live repository-name disclosure when the policy RPC is unavailable. The test at `registry.rs:3469-3478`, however, calls only `policy_result_or_deny`.

I restored only the vulnerable handler call site to `.await.unwrap_or(true)` while leaving the helper/test intact. The test still passed. Thus the disclosure can be reintroduced without tripping the regression.

The PR body’s exact command:

`cargo test -p hyprstream policy_rpc_failure_denies_repository_listing --lib -- --exact`

still runs **zero tests** at current head. The fully qualified helper test passes, but it does not cover the vulnerable boundary.

Required: invoke `RegistryHandler::handle_list` with a policy client/transport that fails and assert the private repository is omitted; document the fully qualified test command.

### 5. MEDIUM — the audience decoder is structurally fail-closed, but the broader optional-expectation class remains open

The current audience mechanism is a genuine centralized improvement, not four point fixes:

- legacy `Option<&str>::None` maps to `Missing` and rejects;
- `Exact` now requires a present matching claim on EdDSA, federated, ML-DSA, and composite paths;
- deliberate bypasses are named.

However, the unsafe spelling still compiles through legacy `decode(..., None)`, `decode_with_key(..., None)`, `impl Into<AudienceExpectation>`, and `RequestService::expected_audience() -> Option<&str>`. The new `ExactOrMissing` is another absence-permitting posture; it is explicit and currently test-only, but the standard’s statement that `ExplicitlyUnchecked` is the only way to omit a check is no longer literally true.

The prior review’s other optional verifier seams are now honestly tracked in open issue #1201: DPoP `access_token: Option`, response signer pinning, conditional host/refresh bindings, and SPIFFE WIT binding. That closes the audit-accounting defect, not the structural work itself.

## Prior findings and instance classification

- **Prior registration outage:** not closed; reproduced on the real RPC path (finding 1).
- **`Exact` accepting absent `aud`:** closed. Both EdDSA/federated and ML-DSA strict tests are mutation-sound.
- **Missing-`cnf` registration bypass:** inner handler defect closed. My absent-`cnf` negative test failed when I restored the old optional branch, then passed after restoration. No PR-owned regression test covers it.
- **OIDC ID-token outage:** availability restored, but as a live cross-client replay weakness (finding 2).
- **Remaining optional expectations:** enumerated in #1201, still structurally open.
- **Registry disclosure:** live vulnerability fixed; regression test remains inadequate.
- **Missing configured service audience:** latent fail-open converted to intended fail-closed denial for embedders.
- **Federated tokens with absent audience:** live cross-resource replay class now rejected under `Exact`.

## What now breaks

- Every ordinary service’s real `registerServiceKey` RPC fails before dispatch when production PolicyService expects the OAuth issuer (unintended outage).
- Clients that continue sending the persisted registration-only WIT to ordinary service RPCs fail audience validation (unintended outage).
- ID-token exchange succeeds again, but for tokens issued to unrelated clients at a trusted issuer (unintended security regression).
- JWT-presenting embedded services with a key source but no configured audience reject (intended fail-closed change).
- Registry listing during policy-service outage denies/omits repositories (intended fix).
- Federated/ML-DSA tokens with an exact expectation but no `aud` reject (intended fix).

## #1147 / #1111 collision check

#1147 is merged into the target branch (`c91fd8207`). A three-way merge-tree check found no textual conflicts in the jointly touched `auth/claims.rs`, `oauth/token_exchange.rs`, or `services/policy.rs`; #1147 owns OAuth bearer/introspection/access-token migration, while findings 1–2 remain semantic defects in #1194. #1111’s event/epoch files are untouched. CodeRabbit did not review this PR because the base is non-default; its passing status is only the skip notice.

## Validation

- `git diff --check 85265b0...HEAD` — passed.
- `buildq -- cargo build -p hyprstream-rpc -p hyprstream` — passed.
- `buildq -- cargo test -p hyprstream-rpc auth::jwt --lib --no-fail-fast` — 15 passed.
- `buildq -- cargo test -p hyprstream-rpc --lib --no-fail-fast` — failed deterministically: 772 passed, 1 failed, 3 ignored (finding 3).
- `buildq -- cargo test -p hyprstream --lib --no-fail-fast` — ran but aborted in an unrelated `noq-proto` GSO-batch assertion/SIGABRT before a final summary. Relevant focused Hyprstream tests passed.
- ML-DSA strict audience test — passed at current head.
- Service-JWT tests — 3 passed at current head.
- Fully qualified registry helper test — passed; PR-body short exact filter ran zero tests.
- Real registration RPC happy-path review test — failed with `JWT verification failed`; direct valid-handler and missing-`cnf` tests passed.

Mutation checks, all restored afterward:

- `Missing => Ok(())` made `absent_expected_audience_fails_closed` fail.
- Permitting absent `aud` under `Exact` made both the federated EdDSA and ML-DSA strict tests fail.
- Removing the minted registration audience made the service-JWT assertion fail.
- Restoring optional missing-`cnf` acceptance made the review negative test fail.
- Restoring `.unwrap_or(true)` only at Registry’s vulnerable handler left the PR test green, proving the coverage defect.

The review worktree remained clean throughout. **BLOCK** until the real registration/peer-RPC happy path works, OIDC ID-token audience is bound, the RPC suite is green, and the registry test covers its handler.

ewindisch:
VERDICT: BLOCK

# Re-re-review — PR #1194 “fix(auth): fail closed on missing expectations”

Reviewed current head `4e549b19da5173df367ca1258c6d1ae1669122f6`. I read all three prior verdicts, re-ran the affected tests, mutation-tested the security-critical assertions in an isolated clone, traced the production registration and peer-RPC paths, and checked the merge against the current target head `c91fd8207` (#1147).

## Findings

### 1. BLOCKER — service registration is still impossible before the repaired handler

The inner handler contract is now locally coherent: service WITs mint `aud = "hyprstream:policy:register-service-key"` (`auth/service_jwt.rs:15-19,66-77`), and `handle_register_service_key` requires that exact audience and a present matching `cnf.jwk` (`services/policy.rs:1315-1342`).

The legitimate RPC path nevertheless fails at two earlier gates:

1. `register_service_key` sends that same registration-only WIT as the outer request JWT (`services/factories.rs:238-253`). Production constructs PolicyService with the OAuth issuer as its generic `expected_audience` (`factories.rs:618-629`). `RequestService::verify_claims` validates the outer JWT before dispatch against that issuer (`hyprstream-rpc/service/svc.rs:687-705,875-891`), so the registration audience mismatches and the handler is unreachable.
2. The WIT is encoded as classical `alg = EdDSA` (`hyprstream-rpc/auth/jwt.rs:361-368`), while production `RequestService` verification defaults to mandatory `CryptoPolicy::Hybrid` and rejects every JWT without an ML-DSA component before audience routing (`service/svc.rs:57-80,619-635,808`). This is an internal Hyprstream RPC, not an interop surface, so the operator's Hybrid requirement applies.

I exercised these boundaries in a disposable review-only test using the real service-WIT issuer and `RequestService::verify_claims`: the WIT passed with the registration audience under test-only Classical policy, failed with PolicyService's production issuer expectation, and failed under Hybrid even when the registration audience matched.

Registration is a hard startup precondition for every non-Policy factory (`factories.rs:170-264`), so Registry, Model, Worker, OAuth, MCP, TUI, Discovery, Metrics, and the remaining ordinary services still refuse to start. The same WIT is persisted as `service_token()` and reused by ordinary peer clients (`factories.rs:185-235,688-693,930-939`), where the registration-only audience is also wrong.

Required: separate the outer RPC credential from the inner registration assertion, make internal WITs Hybrid, and add an actual generated-client/dispatcher happy-path test. A handler-only contract is insufficient.

### 2. HIGH — the OIDC “fix” accepts cross-client ID-token replay

`verify_id_token` now uses `AudienceExpectation::ExplicitlyUnchecked` (`services/oauth/token_exchange.rs:140-177`). That makes the bypass visible but does not enforce that the token was issued to Hyprstream or an approved CrossApp client. `TrustedIssuerConfig` has no allowed OIDC client audience (`config/mod.rs:808-824`), and this path mints a local token for any valid `sub` from a trusted issuer.

An ID token issued to an unrelated client at the same trusted issuer therefore passes signature, issuer, expiry, and replay-cache checks and can be exchanged here. The replay cache prevents reuse; it does not establish intended recipient. The repository's ordinary OIDC callback correctly validates `aud` against `provider.client_id` (`oauth/oidc_callback.rs:302-337`), demonstrating the missing binding.

Required: configure or derive the authorized OIDC client ID, require it in `aud`, and enforce `azp` when a multi-audience token requires it. Labeling the exemption is not enforcement.

### 3. MEDIUM — the affected RPC package suite is still red

`buildq -- cargo test -p hyprstream-rpc --lib --no-fail-fast` deterministically reports `772 passed, 1 failed, 3 ignored`.

The failing test is `service::svc::empty_iss_gate_tests::request_service_enforces_exact_pair_and_closed_header`. The PR gave `MockService` an expected audience, but the expected-success composite `at+jwt` fixture still lacks that audience and is rejected at `service/svc.rs:1341`. The test also fails alone. This is PR-owned and means the relevant suite is not green.

### 4. MEDIUM — the registry test still does not cover the vulnerability, and the documented command runs zero tests

The implementation at `services/registry.rs:1771-1786` correctly changes the live policy-RPC failure from repository disclosure to denial. The test at `registry.rs:3469-3478`, however, calls only `policy_result_or_deny`; it never invokes `RegistryHandler::handle_list`.

Mutation proof: in the disposable clone I restored only the vulnerable `handle_list` call site to `.await.unwrap_or(true)` while leaving the helper and test unchanged. The fully qualified test still passed. The PR-body command with the short filter and `--exact` still runs zero tests; the fully qualified helper test runs one.

Required: make the policy client fail through `handle_list` and assert the private repository is omitted. The missing-`cnf` repair likewise has no PR-owned negative regression test even though the code now rejects absence correctly.

### 5. MEDIUM — the unsafe spelling is fail-closed for JWT audience, but the promised structural closure is not complete

The centralized audience behavior is materially improved:

- legacy `Option<&str>::None` maps to `Missing` and rejects;
- `Exact` now requires a present matching claim on EdDSA, federated, ML-DSA, and composite paths;
- `ExactOrMissing` is explicit and currently used only by a compatibility test.

But the unsafe spelling is not impossible: `decode(..., None)`, `decode_with_key(..., None)`, the `impl Into<AudienceExpectation>` decoder parameters, and `RequestService::expected_audience() -> Option<&str>` all remain compile-valid compatibility APIs. The diff explicitly adds legacy/migration shims despite the greenfield no-compatibility ruling.

The broader class is deferred, not closed, in open #1201: DPoP's optional access-token binding, response signer pinning, registered-host/refresh DPoP bindings, and SPIFFE WIT binding remain `Option`-shaped. The audit is internally contradictory: `docs/standards/expected-value-validation.md:29` calls optional response signer keys test-only, while production `RpcClientImpl` passes `server_verifying_key.as_ref()` to production `unwrap_response_with` (`rpc_client.rs:733-738`), whose `None` skips exact signer equality (`envelope.rs:2438-2453,2748-2761`). Lines 37-41 of the same document then correctly admit that this API remains open in #1201.

This is a centralized runtime fail-closed audience fix, not compile-time structural abolition of the unsafe API or the repo-wide expected-value bug class.

## Prior-finding closure

- Prior registration outage: **not closed**; reproduced at both pre-handler gates (finding 1).
- `Exact` accepting absent `aud`: **closed and mutation-sound** on federated EdDSA and ML-DSA.
- Missing-`cnf` registration bypass: **code path closed** by required extraction and equality; no PR regression test.
- OIDC ID-token outage: availability restored by removing enforcement, creating the cross-client replay in finding 2.
- Remaining optional expectations: honestly linked to #1201, but still structurally open and partly misclassified in the audit.
- Registry disclosure: implementation closed; regression test remains mutation-unsound.
- Missing configured service audience: now rejects, an intended fail-closed change for embedders.

## What the change breaks

- Every ordinary service's real `registerServiceKey` RPC and therefore startup: unintended outage.
- Ordinary peer RPCs that reuse the persisted registration-only WIT: unintended outage.
- Mandatory-Hybrid internal JWT verification of the new EdDSA-only service WIT: unintended policy violation/outage.
- OIDC ID-token exchange now accepts tokens for unrelated clients at a trusted issuer: unintended security regression.
- The full `hyprstream-rpc` library test suite: unintended regression.
- JWT-presenting embedded services with a key source but no configured audience: intended fail-closed denial.
- Registry listing during policy-service outage: intended denial/fix.
- Federated and ML-DSA tokens with an exact expectation but no `aud`: intended denial/fix.

## Integration check

The current target branch contains #1147 at `c91fd8207`. A three-way merge-tree check shows jointly changed `auth/claims.rs`, `oauth/token_exchange.rs`, and `services/policy.rs` but no textual conflict. #1147 repairs OAuth bearer validation, introspection, and access-token downscoping; it does not repair findings 1 or 2. #1111's event/epoch files are untouched. CodeRabbit did not review this non-default-base PR; its status is only a skip notice.

## Validation

- `git diff --check 85265b0...HEAD` — passed.
- `buildq -- cargo build -p hyprstream-rpc -p hyprstream` — passed with the cached CUDA 13.0 libtorch environment.
- `buildq -- cargo test -p hyprstream-rpc auth::jwt --lib --no-fail-fast` — 15 passed.
- `buildq -- cargo test -p hyprstream-rpc --lib --no-fail-fast` — failed: 772 passed, 1 failed, 3 ignored.
- Focused failing RPC test — failed alone.
- Service-JWT tests — 3 passed.
- ML-DSA exact-audience test — passed.
- PR-body registry command — 0 tests.
- Fully qualified registry helper test — 1 passed.
- Review-only service-WIT boundary probe — confirmed valid registration-only/Classical posture, production audience rejection, and Hybrid rejection.

Mutation checks in an isolated clone, restored after each:

- `Missing => Ok(())` made `absent_expected_audience_fails_closed` fail.
- Reintroducing absent-`aud` acceptance under `Exact` made both federated-EdDSA and ML-DSA strict tests fail.
- Removing the minted registration audience made the service-JWT assertion fail.
- Restoring `.unwrap_or(true)` only at Registry's vulnerable handler left the PR test green, proving the coverage defect.

The review worktree remained clean. **BLOCK** until the real Hybrid registration/peer-RPC path works, OIDC ID-token audience is bound, the RPC suite is green, and the registry test exercises its handler.

ewindisch:
VERDICT: BLOCK — head 4e549b19: the handler now requires present/equal cnf.jwk, but register_service_key still sends the EdDSA registration-only WIT as the outer JWT, which PolicyService rejects pre-dispatch under its Hybrid/default-issuer gates; verify_id_token still uses AudienceExpectation::ExplicitlyUnchecked and TrustedIssuerConfig has no client audience, so aud/azp remain unbound and cross-client replay remains (cargo metadata exit 0; buildq auth filters: hyprstream-rpc 207 passed/0 failed, hyprstream 347 passed/0 failed).

ewindisch:
Re-review BLOCK addressed in `8a524e364500893bf47f34313b30beb06b589472` (rebased onto `origin/main` at `3e2155f310ce92492c72c9aea5169502dfd30cb2`).

- Service registration now keeps the registration WIT exclusively in `RegisterServiceKey.service_jwt`; it is no longer reused as the outer RPC bearer credential. The Hybrid request-envelope signer is bound to both the supplied verifying key and the WIT `cnf.jwk`. A generated-client/real-dispatch regression proves present+equal succeeds and missing/mismatched `cnf.jwk` fail.
- OIDC ID-token exchange now requires a configured per-issuer `oidc_client_id`, requires that client in `aud`, requires matching `azp` for multi-audience tokens, and rejects any present mismatched `azp`. Missing client binding fails closed, preventing cross-client replay.
- Also repaired the stale audience fixtures exposed by the fail-closed verifier.

Validation (requested LibTorch environment):
- `cargo metadata --all-features --format-version 1` — exit 0
- `buildq -- cargo test -p hyprstream-rpc auth` — 243 matching tests passed, 0 failed
- `buildq -- cargo test -p hyprstream auth` — 426 matching tests passed, 0 failed
- repository pre-commit clippy — passed

Force-pushed with an explicit lease.

ewindisch:
VERDICT: PASS

Issue #1161, opened 2026-07-22
E1: did:plc resolver arm — federation intake, read-only forever


ewindisch:
Phase E1 of https://github.com/hyprstream/hyprstream/issues/1158. **No dependency — can run in parallel with all of Phase A/B.**

We mint did:web; we **accept federation from did:plc**. This is the intake path that makes an external bsky user work.

- Configurable PLC base URL (do not hardcode `plc.directory`)
- `doc.id` validation — the returned document must claim the DID that was asked for
- TTL cache
- **Fail-closed** on resolution failure, malformed document, or ambiguity
- Egress allowlist
- **Read-only forever** — we never write to a PLC directory. Make that structural, not a convention.

Historical verification of plc audit logs is E2 and depends on C1 verifier trait; do not build a second verifier here.

Issue #1175, opened 2026-07-22
E3: wire #1119 exchange audience to the host-form service did:web


ewindisch:
Phase **E3** of https://github.com/hyprstream/hyprstream/issues/1158. Dep: https://github.com/hyprstream/hyprstream/issues/1161 (E1), https://github.com/hyprstream/hyprstream/issues/1119.

Bind `aud` for `ai.hyprstream.identity.exchangeUcan` to the deployment's **host-form service did:web**, and document the discovery/binding path.

Note #1119's "server-side did:plc resolution" still stands — we **federate** did:plc even though we **mint** did:web. E1 is the read-only resolver arm this depends on.

#1119 is separately blocked on the #1146 Lane B scope-to-authority binding (https://github.com/hyprstream/hyprstream/pull/1147): until a minted token distinguishes granted authority, an exchange endpoint cannot attenuate from the granted scope set and would fall back to subject-wide authority — the confused-deputy outcome.

Issue #1174, opened 2026-07-22
E2: foreign handle resolution + plc audit-log historical verification


ewindisch:
Phase **E2** of https://github.com/hyprstream/hyprstream/issues/1158. Dep: https://github.com/hyprstream/hyprstream/issues/1161 (E1), C1, B4.

Foreign handle resolution by **both** methods (`/.well-known/atproto-did` and DNS TXT `_atproto.`) with **bidirectional checking** — reuse B4's implementation, do not write a second one.

Consume plc `/log/audit` through **C1's verifier trait**, so a federated bsky account's key history verifies through the same code path as our own. That shared trait is the reason C1 was specified with a plc-shaped interface.

PR #1195, opened 2026-07-23
feat(rpc): add read-only did:plc resolver

ewindisch:
Implements #1161 (E1 of #1158).\n\nAdds a configurable, read-only did:plc document resolver with egress allowlisting, TTL caching, subject binding, and fail-closed validation.


ewindisch:
# Verdict: PR #1195 — read-only did:plc resolver (E1 / #1161)

**Verdict: BLOCK** — one CRITICAL correctness finding. The security posture of the module is genuinely good (every hardening claim in the ticket was verified, including by reverting the checks and watching the tests fail), but the resolver cannot resolve any real did:plc: the derived document URL is wrong against the did:plc method spec and against the live reference directory. Merging this ships a dead-on-arrival intake path behind green tests.

PR: https://github.com/hyprstream/hyprstream/pull/1195 · Ticket: https://github.com/hyprstream/hyprstream/issues/1161 · Epic: https://github.com/hyprstream/hyprstream/issues/1158

---

## F1 · CRITICAL · BLOCK — URL derivation is wrong: msi-only path never resolves on a conformant PLC directory

`did_plc_url` derives `{base}/{msi}` (the bare 24-char identifier). The did:plc method resolves documents at `GET {base}/{full-did}` — the path segment includes the `did:plc:` prefix. Verified live against the reference implementation:

```
GET https://plc.directory/did:plc:ewvi7nxzyoun6zhxrhs64oiz  → 200 (DID document)
GET https://plc.directory/ewvi7nxzyoun6zhxrhs64oiz          → 404 {"message":"DID not registered: ewvi7nxzyoun6zhxrhs64oiz"}
```

Every legitimate resolution fails closed → this is exactly the repo prior: **a fail-closed conversion that denies the LEGITIMATE path is an outage, not a safer bug**. All 22 unit tests pass because the fixture fetcher asserts the same wrong URL (`https://plc.example/ewvi7nxzyoun6zhxrhs64oiz`) — green CI is not review.

Fix caution: `Url::join("did:plc:x…")` will parse the argument as an absolute URL with scheme `did` and *replace* the base — that is very likely why the msi-only form was chosen. Use `url.path_segments_mut().push(did)` (or equivalent explicit segment append) with the **full DID**, keep the existing origin-equality re-check, and change the fixture/test URLs to the full-DID form so the tests would have caught this.

## F2 · MEDIUM · REVISE — duplicated fetch/cache infra vs `did_web.rs`, already divergent in strictness

`read_capped`, `evict_for_insert`, `MAX_DID_DOC_BYTES`, `MAX_CACHE_ENTRIES`, and the hardened-client construction are copied nearly verbatim from `crates/hyprstream-rpc/src/did_web.rs`. They have already diverged: did:plc rejects duplicate JSON object keys (`NoDuplicateJson`) while did:web parses with plain `serde_json` and keeps the last duplicate — both feed the *same* admission gate (`admit_key_against_did` reads `verificationMethod` from either). This is the "two implementations over one format will diverge, and one will be the lenient one" failure mode the ticket warned about for verifiers, reproduced in the fetch layer. did:web is now the lenient arm. Factor the shared fetch/parse core (or at minimum lift duplicate-key rejection into did:web) before a third DID arm copies it again.

## F3 · LOW — `HttpPlcFetcher` is a public `DidDocFetcher` whose cache holds never-subject-validated documents

Subject binding (`validate_plc_document`) lives in `DidPlcResolver::resolve_document`, *above* the cache — correct, and it re-runs on every read including cache hits (verified: no bypass on hit/miss/error paths). But `HttpPlcFetcher` is `pub` and implements the shared `DidDocFetcher` trait; `DidWebResolver::resolve_document` performs **no** `doc.id` validation of its own. A future wiring that hands this fetcher (or its cached output) to any consumer other than `DidPlcResolver` gets documents with no subject binding. Not exploitable in this PR (nothing constructs it in production yet). Guard: narrow visibility, or a doc-contract line on the trait impl stating validation is the resolver's job, or validate in `fetch` as defense in depth.

## F4 · LOW / informational — #1145 pattern sweep ("absent expected value disables the check")

Grepped the whole diff for it, including inside `NoDuplicateJson`:

- `validate_plc_document` shape-checks `alsoKnownAs` / `verificationMethod` / `service` **only when present** — an absent `verificationMethod` yields a "validated" document with zero keys. Not exploitable today: the only consumer (admission stage 2) fails closed on an empty key set. But "validated" must never be read as "has usable keys" by a future consumer (E2 https://github.com/hyprstream/hyprstream/issues/1174, E3 https://github.com/hyprstream/hyprstream/issues/1175).
- `resp.content_length()` absent → the early reject is skipped, but the streaming cap in `read_capped` still enforces the 1 MiB ceiling — **not** an instance.
- `did.split(['#','?']).next().unwrap_or(did)` — `split` always yields ≥1 item; the `unwrap_or` is unreachable, not a disabled check.

No live instance of the #1145 class found.

## F5 · informational — real Bluesky did:plc identities publish secp256k1/p256 keys, not Ed25519

The `DidDocResolve` impl wires this resolver into an admission gate that extracts only Ed25519 verification methods (`verification_method_ed25519_keys`). A stock bsky account's atproto signing key is secp256k1, so it will *resolve* but never *admit* unless it publishes an additional Ed25519 VM. That is E2/E3 scope, not an E1 defect — noting it so nobody expects this PR alone to make "an external bsky user work" end-to-end. Also: the gate takes ONE resolver, so a deployment cannot admit both did:web and did:plc peers through one gate until a method-dispatching resolver exists (future wiring).

---

## Ticket checklist (each verified, not taken from the docs)

1. **`doc.id` validation** — ✅ present and non-bypassable on every path: validation runs in `resolve_document` above the cache, so hits, misses, and the admission-gate path all revalidate. Reverted the mismatch bail → `validate_rejects_substituted_subject` and `resolve_document_rejects_id_mismatch` FAILED; restored. Caveat F3 for direct-fetcher misuse.
2. **Configurable PLC base** — ✅ no default provided anywhere; `grep -r "plc.directory"` over `crates/` finds nothing; base URL is a required constructor argument, validated (https-only, no creds/query/fragment).
3. **Fail closed** — ✅ expired cache entries never served, failures never cached, malformed DID never fetches (NeverFetcher test), duplicate-key JSON rejected (reverted → test FAILED), non-object/missing-id/mismatch all `Err`. But see F1: the happy path fails closed too, which is the outage.
4. **Read-only structural** — ✅ genuinely structural within the module: the reqwest client is a private field, used for `.get()` only; the type exposes no write method and no client accessor. No POST/PUT anywhere in the diff.
5. **Egress allowlist / SSRF / size bounds** — ✅ origin allowlist checked before any I/O (reverted → `fetcher_rejects_origin_outside_allowlist` FAILED with a live DNS attempt to evil.example, proving the check is what prevents egress); redirects disabled (`Policy::none()`); https-only; 1 MiB body cap enforced both by Content-Length and by streaming; 10 s connect/total timeouts; msi charset (`a-z2-7`, exact 24) makes path smuggling into the derived URL impossible, with an origin-equality re-check as depth.
6. **No second plc audit-log verifier** — ✅ none built; the module fetches only the document endpoint and says so; audit-log verification correctly deferred to E2 through the C1 (https://github.com/hyprstream/hyprstream/issues/1167) verifier trait.

## What the change breaks

Nearly nothing, by construction: the diff is purely additive (new module + one `DidDocResolve` impl + a `pub mod` line). Nothing in production wiring constructs `DidPlcResolver` or `HttpPlcFetcher` yet, so no runtime behavior changes on merge. The one new surface is the admission-gate impl (F3/F5 above). It also breaks an *expectation*: anyone reading "E1 done" would believe did:plc intake works — it does not (F1).

## Build & test evidence

- `cargo test -p hyprstream-rpc --lib did_plc`: 22/22 pass.
- Full `--lib` suite: 794 pass / 0 fail on a clean run; intermittent SIGABRT at ~10 s from panics inside `noq-proto`/`noq` (moq) worker threads — **reproduced identically on base commit `15e0b5fd2`** (2 clean runs, 1 abort), so pre-existing flake, not attributable to this PR.
- `cargo clippy -p hyprstream-rpc --all-targets`: clean.
- Negative verification: three protections neutered (doc.id bail, egress allowlist, duplicate-key parse) → exactly the 4 covering tests failed; file restored to PR state afterward.
- `cargo fmt --check` ignored per known-broken-on-main (#1140).

*One of two independent reviewers; the other verdict was not consulted.*

ewindisch:
# Verdict: BLOCK

Reviewed PR #1195 at `7a773cc31ec1a1ef16bb338c0f0e2760bd172c63` against merge base `15e0b5fd201ef19ba603980a228467e94b689758`.

## Findings

1. **BLOCK — URL derivation omits `did:plc:` and makes the legitimate resolution path return 404.**

   `did_plc_url` validates the DID, extracts only the 24-character MSI, and joins that beneath the configured base (`crates/hyprstream-rpc/src/did_plc.rs:315-336`). The tests explicitly require `https://plc.example/ewvi7nxzyoun6zhxrhs64oiz` (`:611-638`). The did:plc specification instead defines the resolution endpoint as `https://plc.directory/:did`, where `:did` is the full identifier: `https://plc.directory/did:plc:ewvi7nxzyoun6zhxrhs64oiz` ([specification, DID Resolution](https://web.plc.directory/spec/v0.1/did-plc#did-resolution)).

   Live confirmation with the DID used by this PR's tests:

   ```text
   https://plc.directory/did:plc:ewvi7nxzyoun6zhxrhs64oiz status=200 type=application/did+ld+json bytes=554
   https://plc.directory/ewvi7nxzyoun6zhxrhs64oiz         status=404 type=application/json bytes=58
   ```

   No base URL can compensate for the missing prefix because this implementation always inserts `/` before the MSI. This breaks every standard PLC directory, so the new resolver and its admission-trait implementation cannot resolve an external Bluesky user. Derive `{base}/{full base DID}` and replace the current unit-test oracle with the protocol path; add a local HTTP integration test that asserts the received request target.

2. **HIGH — “malformed document” validation checks array containers but accepts malformed members.**

   `validate_plc_document` checks only that the top-level value is an object, `id` is a matching string, and three optional members are arrays (`crates/hyprstream-rpc/src/did_plc.rs:357-376`). Consequently documents such as `{"id": DID, "verificationMethod": [null]}` or verification methods with missing/non-string `id`, `type`, `controller`, and key material are returned as successfully validated. Duplicate semantic identifiers/controllers are not rejected either. This contradicts E1's explicit fail-closed requirement for malformed/ambiguous documents at an untrusted public parser boundary. The current admission extractor happens to skip unusable Ed25519 entries, but `resolve_document` is public and returns the malformed document to every other consumer.

   Parse and validate a typed DID-document shape before returning or caching it. At minimum validate each recognized array member and reject semantic ambiguity (including duplicate verification-method/service IDs), while preserving extensions needed by valid PLC documents.

3. **HIGH — the advertised cache TTL is not structurally coupled to the cache that enforces it.**

   `HttpPlcFetcher::new(&config_a)` copies `config_a.ttl()` into its private cache (`crates/hyprstream-rpc/src/did_plc.rs:450-468`), but `DidPlcResolver::new(fetcher, config_b)` independently accepts and reports a second config (`:394-403`). With the same origin and different TTLs, resolution succeeds while the fetcher can serve a document beyond `resolver.config().ttl()`. The API therefore permits the exact cached-but-expired behavior the ticket forbids and makes `PlcResolverConfig`'s documented “document-cache TTL” unreliable.

   Provide a constructor that creates the native fetcher from the resolver's single config, or otherwise make mismatched fetcher/config construction unrepresentable. Keep fixture injection separate from production construction.

4. **MEDIUM — redirects and validation failures are not fully fail-closed at the HTTP/cache boundary.**

   Redirect following is disabled, but `error_for_status()` at `crates/hyprstream-rpc/src/did_plc.rs:505` rejects only 4xx/5xx. A 3xx response with a valid JSON DID-document body is parsed, accepted, and cached. Require an actual 2xx status before reading the body.

   Separately, the fetcher inserts a parsed document into the cache at `:507-519` before subject/shape validation occurs in `DidPlcResolver::resolve_document` at `:411-415`. A mismatched or malformed document is rejected on every resolver return (so `doc.id` is not bypassed), but it is still cached; if the directory repairs the response, clients remain denied until TTL. That contradicts the module's “failures are never cached” guarantee. Validate before cache insertion, or cache only a validated type tied to the requested DID.

## What the change breaks

- Standard did:plc resolution and therefore the intended external-user intake path: the generated endpoint is a deterministic 404 on the reference directory.
- The configured cache-expiry contract when fetcher and resolver configs differ.
- Recovery after a transient validation failure, because the invalid parsed value is retained for the TTL.
- Strict redirect failure: a 3xx response body can be treated as a successful resolution.

The change does not hardcode `plc.directory`, add a PLC write API, or implement the out-of-scope audit-log verifier. The native client's HTTP handle is private and used only for GET; the origin check, HTTPS requirement, redirect non-following, timeouts, response cap, and bounded cache are otherwise present. `doc.id` is checked after every resolver fetch, including cache hits, with no absent-value bypass.

## Validation

- `HYPRSTREAM_SHARED_TARGET_DIR=/home/birdetta/.cache/cargo-target-rw1195b .../hyprstream-build.sh run -- cargo test -p hyprstream-rpc did_plc --lib` — **PASS**, 22 passed, 0 failed, 0 ignored; SHA `7a773cc31ec1a1ef16bb338c0f0e2760bd172c63`.
- `HYPRSTREAM_SHARED_TARGET_DIR=/home/birdetta/.cache/cargo-target-rw1195b .../hyprstream-build.sh run -- cargo clippy -p hyprstream-rpc --lib -- -D warnings` — **PASS**.
- Live reference-directory check shown above — full DID path 200; PR-generated MSI-only path 404.
- Mutation audit in a disposable detached worktree (restored and removed afterward): replacing `validate_plc_document` with `Ok(())` made all five subject/missing-ID/collection-shape rejection tests fail; removing the MSI, config, duplicate-key, and allowlist implementations made seven corresponding tests fail. The native cache/body/status behaviors have no causal tests to mutate. Most importantly, the two URL happy-path tests are causal to the implementation but encode the wrong protocol endpoint, so their passing result is evidence of the blocker rather than correctness.
- `cargo fmt --check` intentionally not run, per the review brief's known clean-main failure note.

ewindisch:
Fixed in 155ab5c59.

- Resolution now derives `{configured base}/{full did:plc DID}` using `path_segments_mut().push(did)`, retaining the origin check.
- Added a resolver test using an unchanged fixture captured from the live PLC reference-directory response for `did:plc:ewvi7nxzyoun6zhxrhs64oiz`; the fixture fetcher asserts the full-DID request target.
- Validates recognized document members before returning or caching: member object/string/key-material shape plus duplicate verification-method/service IDs. Native cache insertion now occurs only after subject/shape validation.
- Native construction now couples the fetcher cache TTL to the resolver config; 3xx/non-2xx responses fail closed.
- Moved duplicate-key DID JSON parsing into the shared did:web layer so both DID fetch paths reject ambiguous JSON.

Validation: `buildq -- cargo test -p hyprstream-rpc did_plc --lib`, `buildq -- cargo clippy -p hyprstream-rpc --all-targets -- -D warnings`, and the repository pre-commit workspace release Clippy gate all passed.

No CodeRabbit review threads are currently open; its non-threaded duplication note is addressed by the shared duplicate-key parser, while PLC cache validation remains method-specific.

ewindisch:
# Re-review: PR #1195 — read-only did:plc resolver

VERDICT: REVISE

Reviewed head `155ab5c59e1200646b7eb2cadcd32b08276541ed`.

The prior CRITICAL is fixed for a root-mounted PLC directory: URL derivation now uses the full DID path, the reference endpoint returns a real document, and the committed fixture has the same JSON value as the live directory response. However, the focused test suite is red at this head and exposes a second URL-derivation outage for a supported configuration.

## Findings

### 1. HIGH — a configured PLC base with a path derives a double-slash document URL

`did_plc_url` first changes `/mirror` to `/mirror/`, then calls `path_segments_mut().push(did)` (`crates/hyprstream-rpc/src/did_plc.rs:227-236`). With the `url` crate, the trailing empty segment is retained, so the push produces:

```text
https://plc.example/mirror//did:plc:ewvi7nxzyoun6zhxrhs64oiz
```

The documented and tested URL is:

```text
https://plc.example/mirror/did:plc:ewvi7nxzyoun6zhxrhs64oiz
```

This is not hypothetical: the PR's own `did_plc::tests::url_derivation_custom_base_with_path` failed on the clean requested head with exactly that left/right mismatch. `PlcResolverConfig` accepts path-prefixed bases and the public API explicitly documents them, so a PLC mirror mounted below its origin path will receive the wrong request target and will commonly return 404. The root-mounted `https://plc.directory` happy path is unaffected.

Fix by removing the trailing empty segment before pushing (for example, the appropriate `PathSegmentsMut` empty-segment operation), then rerun the entire `did_plc` filter. Keep test coverage for root, `/mirror`, `/mirror/`, and a port.

### 2. MEDIUM — malformed/ambiguous recognized members are still accepted

The prior malformed-document finding is only partially fixed (`crates/hyprstream-rpc/src/did_plc.rs:328-400`):

- `verification_method_has_key_material` accepts any nonempty string such as `"publicKeyMultibase":"zKey"` and any nonempty object such as `"publicKeyJwk":{"garbage":true}`; it does not validate the selected verification material's syntax.
- verification-method and service IDs are deduplicated as raw strings. Semantically equivalent DID URLs such as `#key` and `did:plc:ewvi7nxzyoun6zhxrhs64oiz#key` are accepted together even though both identify the same resource after relative-DID-URL resolution.
- a service endpoint object is considered valid merely because it is nonempty.

The native resolver validates before caching and again before return, but these values pass that validator, so `resolve_document` can still publicly return a document advertised as validated despite the ticket's fail-closed requirement for malformed documents and ambiguity. The current admission extractor independently rejects malformed Ed25519 material, which limits immediate admission impact; it does not repair the resolver contract for other consumers.

Use a typed recognized-member representation plus method-specific verification-material validation, and canonicalize DID URLs before uniqueness checks. Preserve extension members needed by conforming PLC documents.

## Re-confirmed requirements

- **Real root-directory happy path:** the did:plc specification defines `GET https://plc.directory/:did`; live `GET https://plc.directory/did:plc:ewvi7nxzyoun6zhxrhs64oiz` returned 200, `application/did+ld+json`, 554 bytes. The obsolete MSI-only path returned 404. The committed fixture and live response parse to identical JSON; their only byte difference is the fixture's trailing newline. The focused fixture resolver test passed. A temporary native end-to-end live test was queued but canceled at the user's delivery deadline before it acquired a build slot, so that additional probe is **unverified**.
- **Causal URL regression test:** temporarily reverting `.push(did)` to the old `.push(plc_msi(did)?)` made `did_plc::tests::url_derivation_configured_base` fail with MSI-only actual versus full-DID expected. The hunk was restored and `git diff --exit-code` passed.
- **`doc.id` binding:** mismatch and missing-ID tests passed; validation runs before native cache insertion, on native cache hits, and again in the resolver.
- **Configurable base:** no production default or hardcoded `plc.directory`; HTTPS, host, credential, query, and fragment checks remain. Finding 1 breaks the accepted path-prefix variant.
- **Fail closed/cache:** non-success HTTP status is rejected, redirects are disabled, duplicate JSON keys are rejected, invalid documents are validated before cache insertion, expired entries are not served, and the native constructor couples the cache TTL to its single config.
- **Read-only structural boundary:** the reqwest client is private, only `.get()` is used, and no write method/client accessor exists.
- **Egress/SSRF/DoS:** exact-origin allowlist, HTTPS-only fetch, no redirects, 10-second connect/request timeouts, 1 MiB streamed response cap, and bounded cache remain present.
- **No second audit-log verifier:** none was added; audit verification remains deferred to E2/#1174 through C1/#1167.
- **#1145 sweep:** optional `Content-Length` does not disable the cap because the streaming cap is unconditional. Optional recognized arrays do not admit a key by absence in the current admission consumer, although the raw validation contract remains too weak as described in Finding 2.

## What the change breaks

- PLC directory bases mounted at a path prefix derive a double-slash request target and can fail every legitimate resolution.
- `did:web` documents with duplicate JSON keys that previously parsed last-value-wins now fail closed through the newly shared duplicate-key parser. This is a deliberate security hardening but is still a compatibility change.
- The new PLC validation rejects some previously parseable loose/custom document shapes; intended malformed cases should remain rejected, but conforming extension shapes need typed coverage.
- No production code currently constructs this new resolver, so existing runtime behavior is otherwise unchanged. Also, stock Bluesky PLC documents generally publish secp256k1/P-256 rather than Ed25519 keys; the existing admission gate's Ed25519-only extraction means this PR alone does not make such an identity admit end to end.

## Validation evidence

- `buildq -- cargo test -p hyprstream-rpc did_plc --lib` — **FAIL**, 24 passed / 1 failed / 0 ignored; failing test: `did_plc::tests::url_derivation_custom_base_with_path`.
- Reverse-hunk causal run: `buildq -- cargo test -p hyprstream-rpc did_plc::tests::url_derivation_configured_base --lib -- --exact` — **expected FAIL**, 0 passed / 1 failed after restoring MSI-only derivation; source restored afterward.
- Live reference checks described above — full-DID 200, MSI-only 404, fixture/live semantic JSON equality.
- Remote PR checks were green, including Clippy and CodeQL, but were not treated as a substitute for the failing local test.
- Final source state: clean at `155ab5c59e1200646b7eb2cadcd32b08276541ed`.

ewindisch:
# Re-review: PR #1195 — read-only did:plc resolver

VERDICT: REVISE

Reviewed head `155ab5c59e1200646b7eb2cadcd32b08276541ed`.

The original CRITICAL is closed for a root-mounted PLC directory: the resolver now requests the full DID path, the native resolver successfully fetched the live reference document, and the committed fixture still parses identically to the live response. The prior re-review's two remaining findings are not closed, however. The focused suite is red on the clean PR head.

## Findings

### 1. HIGH — a configured PLC base with a path still derives a double-slash URL

`did_plc_url` changes `/mirror` to `/mirror/` and then calls `path_segments_mut().push(did)` (`crates/hyprstream-rpc/src/did_plc.rs:227-236`). The trailing empty path segment is retained, producing:

```text
https://plc.example/mirror//did:plc:ewvi7nxzyoun6zhxrhs64oiz
```

instead of:

```text
https://plc.example/mirror/did:plc:ewvi7nxzyoun6zhxrhs64oiz
```

This is an accepted and documented configuration, not an unsupported edge case. The clean-head focused run failed exactly at `did_plc::tests::url_derivation_custom_base_with_path`, with 24 tests passing and this test failing. A path-mounted PLC mirror will commonly return 404 for the extra-slash target, so the configurable-base promise remains broken.

Remove the trailing empty segment before pushing the full DID, and retain coverage for a root base, `/mirror`, `/mirror/`, and a non-default port.

### 2. MEDIUM — malformed and semantically ambiguous recognized members are still accepted

The recognized-member validator remains an ad hoc presence check (`crates/hyprstream-rpc/src/did_plc.rs:328-400`):

- A `Multikey` with `"publicKeyMultibase":"zKey"` is accepted even though it is not a valid Multikey public key. Any nonempty `publicKeyJwk` object, including `{"garbage":true}`, is likewise treated as valid, and multiple verification-material properties are not rejected.
- IDs are deduplicated as raw strings. `#key` and `did:plc:ewvi7nxzyoun6zhxrhs64oiz#key` are both accepted even though DID Core relative-URL resolution makes them the same resource.
- `alsoKnownAs`, verification-method IDs/controllers, and string service endpoints are checked only for non-emptiness, not the URI/DID syntax required by DID Core.

A temporary review regression test proved the first two concrete cases: both assertions that `validate_plc_document` reject invalid key material and equivalent relative/absolute IDs failed. The native live-reference test in the same run passed, so tightening this need not deny the legitimate reference-directory shape.

Use a typed or equivalently complete validation layer: resolve/canonicalize relative DID URLs before uniqueness checks; validate DID/URI fields; validate verification material according to its declared method (including JWK and Multikey syntax and exclusivity); and preserve conforming extensions.

## Prior findings and ticket requirements re-confirmed

- **Real happy path:** `GET https://plc.directory/did:plc:ewvi7nxzyoun6zhxrhs64oiz` returned 200, `application/did+ld+json`, 554 bytes. The obsolete MSI-only URL returned 404. The committed fixture is 555 bytes only because of its trailing newline and is semantically identical to the live JSON. A temporary native `DidPlcResolver` test against the live reference directory passed.
- **Full-DID URL fix is causal:** temporarily restoring `.push(plc_msi(did)?)` made the exact root-base happy-path test fail with MSI-only actual versus full-DID expected. The hunk was restored.
- **`doc.id` binding:** exact subject binding runs in the resolver and before native cache insertion/on cache hits. Temporarily removing it made `validate_rejects_substituted_subject` fail. The hunk was restored.
- **Configurable base:** no production default or hardcoded `plc.directory`; HTTPS, host, credentials, query, and fragment are validated. Finding 1 still breaks path-prefixed bases.
- **Fail closed/cache:** non-2xx status is rejected, redirects are disabled, duplicate JSON keys are rejected, invalid native responses are validated before cache insertion, failures are not cached, expired entries are not served, and native construction couples resolver configuration to fetcher TTL.
- **Read-only structurally:** the HTTP client is private, only `.get()` is used, and there is no write method or client accessor.
- **Egress/SSRF/DoS:** exact-origin allowlist before I/O, HTTPS-only URLs, no redirects, 10-second connect/request timeouts, a streamed 1 MiB cap even without `Content-Length`, and a 256-entry bounded cache remain present.
- **No duplicate audit-log verifier:** none was added; audit verification remains deferred to E2/#1174 through C1/#1167.
- **Absent-value sweep:** optional recognized arrays are legitimate DID Core properties and absence does not admit a key downstream; missing `Content-Length` does not disable the unconditional streaming cap. No caller-controlled or default-off security requirement was found.
- **Alert sweep:** no open CodeQL alert was returned for `refs/heads/feat/e1-plc-resolver`.

## What the change breaks

- PLC mirrors mounted below an origin path receive a double-slash target and can fail every resolution.
- `did:web` documents containing duplicate JSON object keys, previously parsed with last-value-wins behavior, now fail closed through the shared duplicate-key parser. This is intentional security hardening but is a compatibility change.
- The new PLC validation rejects some loose/custom recognized-member shapes that previously parsed as raw JSON. Rejection of malformed cases is intended; conforming extension shapes must remain interoperable.
- No production code constructs `DidPlcResolver`, so existing deployment behavior is otherwise unchanged. The live federation path still uses `DidWebResolver`.
- Stock Bluesky PLC documents generally publish P-256/secp256k1 signing keys, while the current admission gate extracts only Ed25519 methods. This PR resolves such documents but does not by itself make those identities admit end to end.

## Validation evidence

- `buildq -- cargo test -p hyprstream-rpc did_plc --lib -- --nocapture` — **FAIL**, 24 passed / 1 failed / 0 ignored; `url_derivation_custom_base_with_path` derived the double slash.
- Reverse full-DID hunk: `buildq -- cargo test -p hyprstream-rpc did_plc::tests::url_derivation_configured_base --lib -- --exact --nocapture` — **expected FAIL**, MSI-only actual versus full-DID expected.
- Reverse subject/raw-duplicate guards: `buildq -- cargo test -p hyprstream-rpc did_plc::tests::validate --lib -- --nocapture` — **expected FAIL**, substitution and duplicate-ID tests failed; 3 passed / 2 failed.
- Temporary current-code probes: `buildq -- cargo test -p hyprstream-rpc did_plc::tests::review_probe --lib -- --nocapture` — native live reference resolution **passed**; invalid key material and semantically duplicate IDs were both wrongly accepted, so those two rejection assertions **failed**.
- `buildq -- cargo clippy -p hyprstream-rpc --all-targets -- -D warnings` — **PASS**.
- Final source state is clean at `155ab5c59e1200646b7eb2cadcd32b08276541ed`; every temporary mutation/probe was removed.

ewindisch:
Both re-review findings addressed in `3c5269731` (pushed). Each guard ships with a test verified to fail without it (revert → red → restore → green). Reference-directory capture and all 25 prior `did_plc` tests still pass; `cargo clippy -p hyprstream-rpc --all-targets -- -D warnings` clean.

**Finding 1 (HIGH — double-slash URL).** `did_plc_url` no longer uses `path_segments_mut().push(did)`, which retained the trailing empty segment for `/mirror`-style bases. It now normalizes the base path by trimming any trailing `/` and joining with exactly one separator via `set_path("{trimmed}/{did}")`, so a base of `https://plc.example/mirror` derives `…/mirror/did:plc:…` (single slash). Covered for root, `/mirror`, `/mirror/`, and a non-default port by `url_derivation_custom_base_with_path` — which was red on the prior head and is now green. `set_path` does not percent-encode the DID's `:` (verified by the existing `url_derivation_configured_base` exact-string assertion), and `plc_msi`'s charset check makes an embedded `/` impossible.

**Finding 2 (MEDIUM — ad-hoc member validation).** Replaced the presence check with a typed layer:
- **Relative DID-URL canonicalization:** `#key` and `did:…#key` are resolved against the subject DID before the uniqueness check, so the same resource can no longer appear twice (`validate_rejects_semantically_duplicate_ids`).
- **DID/URI fields:** `alsoKnownAs`, verification-method `id` (must be a DID URL or relative ref), `controller` (must be a DID), and string `serviceEndpoint`s are now syntax-checked, not merely non-empty (`validate_rejects_non_did_or_non_uri_fields`).
- **Verification material per declared encoding:** `publicKeyMultibase` is decoded as a Multikey (base58btc multibase + multicodec varint + codec-specific key length), so `zKey` is rejected; `publicKeyJwk` requires a recognized `kty` with its coordinate members, so `{garbage:true}` is rejected. DID Core §5.1 exclusivity rejects methods carrying more than one known material property.
- **Conforming extensions preserved:** an unknown algorithm whose `publicKeyMultibase` carries a real (unknown-codec) key still resolves — this is a classical, read-only interop surface, so unknown algs are ignored rather than failing (`validate_preserves_unknown_algorithm_with_valid_material`).

Not merging — single review is the gate.

ewindisch:
Re: the CodeRabbit nitpick on `HttpPlcFetcher` sharing the no-redirect client / capped-body read / TTL-eviction with `did_web::HttpDidDocFetcher` — fair drift risk, but deferring it out of this change. The two verdict findings (URL double-slash, recognized-member validation) are the single-review gate here, and extracting the shared helper is a cross-crate refactor into `did_web` that would widen the diff's blast radius without changing this PR's behavior. Filed as #1231 so the dedup lands in its own focused change rather than riding a correctness fix. The PLC-specific layer (egress-origin allowlist, `plc_msi` re-check, `validate_plc_document`) stays where it is either way.

ewindisch:
VERDICT: BLOCK — at 3c526973 the path is structurally read-only (private client, GET only) and malformed/unreachable inputs fail closed, and `buildq -- cargo test -p hyprstream-rpc did_plc` passes 30/30; however `did_plc.rs:35-41` explicitly leaves PLC audit-history/signature verification out of scope, `resolve_document` trusts the directory’s current document after shape/`id` checks, and `admission.rs:339-345` admits on Ed25519 alone, so requirements (2) and (4) are unmet on this SPINE trust surface.

ewindisch:
VERDICT: PASS — the BLOCK is closed in `d0cfd9d55d47e69a6a551e10b94d775a1d793439`.

- `did:plc` resolution no longer trusts the directory current document. It fetches `/log/audit`, canonicalizes operations as DAG-CBOR, verifies operation CIDs and the genesis DID derivation, follows `prev` links, and verifies each canonical low-S ECDSA signature against the parent rotation keys before accepting a derived current document. Tests cover a captured real chain, signature tampering with a recomputed CID, and unsigned current-document substitution.
- Admission now requires an explicit trust surface plus verified signer assurance. Native/SPINE admission requires `PqHybrid`; classical assurance is accepted only through the explicit `ThirdPartyInterop` surface. A native classical downgrade is rejected before DID resolution.
- This implements verification of the presented current operation chain. Full nullified-fork recovery-window/priority auditing and historical key-at-time queries remain tracked in existing issue #1174, backed by the pure-verifier work in #1167; I did not open a duplicate issue.

Validation:
- `buildq -- cargo test -p hyprstream-rpc did_plc` — PASS, 34/34
- `buildq -- cargo test -p hyprstream-rpc admission::tests --lib` — PASS, 23/23
- `buildq -- cargo test -p hyprstream-pds at9p_resolver::tests::admission --lib` — PASS, 2/2
- `buildq -- cargo clippy -p hyprstream-rpc --all-targets -- -D warnings` — PASS
- Commit hook CI-mirroring clippy — PASS

ewindisch:
Fresh verification at `d0cfd9d55d47e69a6a551e10b94d775a1d793439`:

- `buildq -- cargo test -p hyprstream-rpc did_plc` — PASS, 34 passed, 0 failed, 0 ignored
- `git add -A` found no additional edits; the working tree was already clean and the fix was already committed at this SHA
- Commit-hook CI-mirroring clippy — PASS
- `git push origin HEAD:feat/e1-plc-resolver` — everything up to date

VERDICT: PASS.

Issue #1183, opened 2026-07-22
audit: 'exactly one published key' is an anti-pattern — it forecloses overlap rotation AND PQ-hybrid rollout


ewindisch:
Raised by Erica 2026-07-22, after the same defect surfaced independently in two unrelated subsystems on the same day.

## The pattern

**hyprstream repeatedly assumes exactly one key is published at a time.** That assumption makes *rotation-with-overlap* — publishing the new key alongside the old, the standard way to rotate anything without a flag day — structurally impossible, and it also forecloses PQ-hybrid rollout, which is the same mechanism.

Two confirmed instances:

1. **`ca_keys.len() == 1`** in the DID-anchored trust bootstrap (`crates/hyprstream-discovery/src/did_anchored.rs:198-203`), found by the independent review of #1143. Publishing the new deployment CA alongside the old makes `len() == 2` and **breaks bootstrap for every node in the cluster simultaneously** — a self-inflicted cluster-wide outage on the first rotation. Being fixed as part of the #1157 option-C work (selection becomes identity-based, so the cardinality rule is deleted rather than repaired).

2. **Single `#atproto` verification method** in `oauth::did_document`, tracked at #918. Historical commits stop verifying at each rotation boundary because only the active ES256 key is ever published. An earlier design comment ruled multi-key publication out as "reserved for did:at9p" — wrong, see below.

## Why the "one key" instinct is wrong

**Hybrid and overlapping key rotation work by publishing multiple keys and letting each verifier ignore what it does not understand.** In JWKS this is the `kid` mechanism: a set may carry a classical key and a post-quantum or composite key side by side; a verifier selects by `kid` and **skips entries whose `kid`/`alg` it does not recognize** rather than failing the whole set. Classical verifiers keep working against the classical entry while PQ-capable verifiers use the PQ entry, and neither needs to know about the other.

Erica verified this against each mainstream JWT library — behavior matches RFC guidance and PQUIP recommendations, with **one patch needed to ruby-jwt** where it was errant. So this is empirically established across the ecosystem, not merely specified.

[...]

Issue #1188, opened 2026-07-22
design(at9p): give subject_keys set semantics instead of primary-only semantics


ewindisch:
Relates #1183.

## Problem

The at9p capsule schema publishes `subject_keys: Vec<HybridKeyPair>`, but all live consumers treat index 0 as the sole identity/signing key.

- The schema permits every non-empty key list (`crates/hyprstream-pds/src/at9p.rs:401-477`).
- Rotation authorisation and signature verification select only `new_capsule_body.subject_keys.first()` (`crates/hyprstream-pds/src/at9p_chain.rs:250-275`).
- The GATE bridge projects only the first pair into `VerifiedAt9pKeys` (`crates/hyprstream-pds/src/at9p_resolver.rs:46-71`), and downstream `IdentityKeys` is scalar (`crates/hyprstream-rpc/src/identity_resolver.rs:263-274`).
- Native announcements likewise require the local signer to equal the first key (`crates/hyprstream-service/src/service/factory.rs:70-95`).

Thus a capsule containing an old and new subject pair has no specified/working overlap semantics; reordering changes authority and a signature by the later published pair is rejected. Tests construct/read index 0 only; none publish two pairs or verify rotation/announcement against the second.

## Required direction

Either make the wire model an explicit single `primary_subject_key` plus separately named role/key-set fields, or preserve the vector and define key ids/roles plus candidate verification/validity rules. In either case, make hybrid pairs atomic, prohibit order-as-authority, and add end-to-end two-pair overlap tests for capsule GATE, successor validation, identity resolution, and native announcements.

PR #1202, opened 2026-07-23
fix(identity): resolve DID verification key sets

### LINKED ISSUE #1187: fix(identity): replace first-VM DID identity resolution with named/candidate key sets
Relates #1183.

## Problem

`MethodDispatchResolver` treats a DID document's ordered verification-method arrays as one-key containers.

- `crates/hyprstream-rpc/src/identity_resolver.rs:208-238` selects the first Ed25519 VM and first ML-DSA-65 VM, then returns scalar `IdentityKeys`.
- The source comment explicitly acknowledges that multiple VMs get arbitrarily paired by document order (`:208-213`).
- `crates/hyprstream/src/services/oauth/did_document.rs:327-435` is already capable of publishing multiple named VMs, and user DIDs publish every registered key (`:565-615`), but this consumer cannot represent or try them.

A valid document with old+new Ed25519 VMs (or overlapping hybrid pairs) cannot be consumed safely through this resolver. No test exercises two VMs where the later key/pair is the relevant one.

## Required direction

Model DID verification material as an explicitly identified set (and hybrid pairs as explicit bindings), select by VM id/relationship where the protocol supplies one, and otherwise try compatible candidates rather than document order. Keep `did:key` as the intentional one-key exception. Add two-key classical and two-hybrid-pair overlap tests.

ewindisch:
Fixes #1187\n\nRelates #1183.\n\n- represent DID identity material as named candidates rather than positional scalars\n- bind hybrid candidates by VM id ( ↔ )\n- select the candidate matching the admission-verified signer\n- cover classical and hybrid overlap plus retired-key rejection

### DIFF
--- a/crates/hyprstream-rpc/src/auth/atproto_perimeter.rs
+++ b/crates/hyprstream-rpc/src/auth/atproto_perimeter.rs
@@ -439,13 +439,11 @@ mod tests {
     }
 
     fn hybrid_keys(ed25519: [u8; 32]) -> IdentityKeys {
-        IdentityKeys {
-            candidates: vec![IdentityKeyCandidate {
+        IdentityKeys::new(vec![IdentityKeyCandidate {
                 id: "did:web:peer.example#mesh".to_owned(),
                 ed25519,
                 ml_dsa_65: Some(an_ml_dsa_vk()),
-            }],
-        }
+            }])
     }
 
     /// Fixture resolver returning fixed key material — the #549 stand-in for #579,
@@ -526,8 +524,7 @@ mod tests {
     fn overlap_selects_each_admitted_signer_then_rejects_the_retired_candidate() {
         let old = [30u8; 32];
         let new = [31u8; 32];
-        let overlap = IdentityKeys {
-            candidates: vec![
+        let overlap = IdentityKeys::new(vec![
                 IdentityKeyCandidate {
                     id: "did:web:peer.example#mesh-old".to_owned(),
                     ed25519: old,
@@ -538,8 +535,7 @@ mod tests {
                     ed25519: new,
                     ml_dsa_65: Some(an_ml_dsa_vk()),
                 },
-            ],
-        };
+            ]);
         let old_peer = AtprotoPerimeterGateway::new(FixtureResolver(overlap.clone()))
             .enroll(&admitted(Some(DID_WEB), old))
             .expect("draining key remains accepted during overlap");
--- a/crates/hyprstream-rpc/src/auth/atproto_perimeter.rs
+++ b/crates/hyprstream-rpc/src/auth/atproto_perimeter.rs
@@ -127,38 +127,15 @@ impl<R: IdentityResolver> AtprotoPerimeterGateway<R> {
             .resolve_identity_keys(&did)
             .map_err(|e| anyhow!("perimeter enroll: DID {did} did not resolve: {e}"))?;
 
-        // Fail-closed on missing verifying key material for the asserted assurance:
-        // a `Classical`/`PqHybrid` assurance with no Ed25519 key, or a `PqHybrid`
-        // assurance with no ML-DSA-65 anchor, is an inconsistent resolution we will
-        // not enroll.
-        match keys.assurance {
-            Assurance::Unverified => {
-                return Err(anyhow!(
-                    "perimeter enroll: DID {did} resolved Unverified (no verifiable key) — fail-closed"
-                ));
-            }
-            Assurance::Classical => {
-                if keys.ed25519.is_none() {
-                    return Err(anyhow!(
-                        "perimeter enroll: DID {did} asserts Classical but published no Ed25519 key — fail-closed"
-                    ));
-                }
-            }
-            Assurance::PqHybrid => {
-                if keys.ed25519.is_none() {
-                    return Err(anyhow!(
-                        "perimeter enroll: DID {did} asserts PqHybrid but published no Ed25519 key — fail-closed"
-                    ));
-                }
-                if keys.ml_dsa_65.is_none() {
-                    return Err(anyhow!(
-                        "perimeter enroll: DID {did} asserts PqHybrid but published no ML-DSA-65 anchor — fail-closed"
-                    ));
-                }
-            }
-        }
-
-        let key_material = assurance_to_key_material(keys.assurance);
+        // Admission already verified `admitted.key` against the DID document.
+        // Select that exact candidate from the named set rather than granting
+        // document-order authority to another overlap/rotation slot.
+        let candidate = keys.candidate_for_ed25519(&admitted.key).ok_or_else(|| {
+            anyhow!(
+                "perimeter enroll: DID {did} no longer publishes the admitted Ed25519 signer — fail-closed"
+            )
+        })?;
+        let key_material = assurance_to_key_material(candidate.assurance());
         let derived = key_material.assurance();
 
         // Edge ceiling × clamp-DOWN to crypto-derived assurance (#548).
@@ -441,7 +418,7 @@ pub struct TranslatedAtprotoAuthorization {
 mod tests {
     use super::*;
     use crate::crypto::pq::ml_dsa_generate_keypair;
-    use crate::identity::{IdentityKeys, MlDsaVk};
+    use crate::identity::{IdentityKeyCandidate, IdentityKeys, MlDsaVk};
 
     const DID_WEB: &str = "did:web:peer.example";
 
@@ -455,6 +432,20 @@ mod tests {
 
     fn an_ml_dsa_vk() -> MlDsaVk {
         ml_dsa_generate_keypair().1
+    }
+
+    fn classical_keys(ed25519: [u8; 32]) -> IdentityKeys {
+        IdentityKeys::single_ed25519("did:web:peer.example#mesh", ed25519)
+    }
+
+    fn hybrid_keys(ed25519: [u8; 32]) -> IdentityKeys {
+        IdentityKeys {
+            candidates: vec![IdentityKeyCandidate {
+                id: "did:web:peer.example#mesh".to_owned(),
+                ed25519,
+                ml_dsa_65: Some(an_ml_dsa_vk()),
+            }],
+        }
     }
 
     /// Fixture resolver returning fixed key material — the #549 stand-in for #579,
@@ -476,11 +467,7 @@ mod tests {
 
     #[test]
     fn classical_peer_enrolls_at_classical_assurance() {
-        let keys = IdentityKeys {
-            ed25519: Some([7u8; 32]),
-            ml_dsa_65: None,
-            assurance: Assurance::Classical,
-        };
+        let keys = classical_keys([7u8; 32]);
         let gw = AtprotoPerimeterGateway::new(FixtureResolver(keys));
         let peer = gw
             .enroll(&admitted(Some(DID_WEB), [7u8; 32]))
@@ -497,11 +484,7 @@ mod tests {
 
     #[test]
     fn pqhybrid_peer_enrolls_at_pqhybrid_assurance() {
-        let keys = IdentityKeys {
-            ed25519: Some([9u8; 32]),
-            ml_dsa_65: Some(an_ml_dsa_vk()),
-            assurance: Assurance::PqHybrid,
-        };
+        let keys = hybrid_keys([9u8; 32]);
         let gw = AtprotoPerimeterGateway::new(FixtureResolver(keys));
         let peer = gw
             .enroll(&admitted(Some(DID_WEB), [9u8; 32]))
@@ -513,11 +496,7 @@ mod tests {
 
     #[test]
     fn unverified_resolution_fails_closed() {
-        let keys = IdentityKeys {
-            ed25519: None,
-            ml_dsa_65: None,
-            assurance: Assurance::Unverified,
-        };
+        let keys = IdentityKeys::default();
         let gw = AtprotoPerimeterGateway::new(FixtureResolver(keys));
         assert!(gw.enroll(&admitted(Some(DID_WEB), [0u8; 32])).is_err());
     }
@@ -531,36 +510,55 @@ mod tests {
     #[test]
     fn missing_did_fails_closed() {
         // Admission matched via JWKS fallback only — no DID to resolve.
-        let keys = IdentityKeys {
-            ed25519: Some([2u8; 32]),
-            ml_dsa_65: None,
-            assurance: Assurance::Classical,
-        };
+        let keys = classical_keys([2u8; 32]);
         let gw = AtprotoPerimeterGateway::new(FixtureResolver(keys));
         assert!(gw.enroll(&admitted(None, [2u8; 32])).is_err());
     }
 
     #[test]
-    fn pqhybrid_assertion_without_ml_dsa_fails_closed() {
-        // Inconsistent resolution: claims PqHybrid but published no PQ anchor.
-        let keys = IdentityKeys {
-            ed25519: Some([3u8; 32]),
-            ml_dsa_65: None,
-            assurance: Assurance::PqHybrid,
-        };
+    fn candidate_set_without_admitted_signer_fails_closed() {
+        let keys = classical_keys([4u8; 32]);
         let gw = AtprotoPerimeterGateway::new(FixtureResolver(keys));
         assert!(gw.enroll(&admitted(Some(DID_WEB), [3u8; 32])).is_err());
+    }
+
+    #[test]
+    fn overlap_selects_each_admitted_signer_then_rejects_the_retired_candidate() {
+        let old = [30u8; 32];
+        let new = [31u8; 32];
+        let overlap = IdentityKeys {
+            candidates: vec![
+                IdentityKeyCandidate {
+                    id: "did:web:peer.example#mesh-old".to_owned(),
+                    ed25519: old,
+                    ml_dsa_65: None,
+                },
+                IdentityKeyCandidate {
+                    id: "did:web:peer.example#mesh-new".to_owned(),
+                    ed25519: new,
+                    ml_dsa_65: Some(an_ml_dsa_vk()),
+                },
+            ],
+        };
+        let old_peer = AtprotoPerimeterGateway::new(FixtureResolver(overlap.clone()))
+            .enroll(&admitted(Some(DID_WEB), old))
+            .expect("draining key remains accepted during overlap");
+        let new_peer = AtprotoPerimeterGateway::new(FixtureResolver(overlap))
+            .enroll(&admitted(Some(DID_WEB), new))
+            .expect("new key is accepted during overlap");
+        assert_eq!(old_peer.assurance, Assurance::Classical);
+        assert_eq!(new_peer.assurance, Assurance::PqHybrid);
+
+        let retired = IdentityKeys::single_ed25519("did:web:peer.example#mesh-new", new);
+        let gateway = AtprotoPerimeterGateway::new(FixtureResolver(retired));
+        assert!(gateway.enroll(&admitted(Some(DID_WEB), old)).is_err());
     }
 
     #[test]
     fn classical_peer_cannot_obtain_pqhybrid_context() {
         // A classical-key peer, even resolved at the PqHybrid-capable edge, must
         // clamp to Classical and fail to dominate a PqHybrid object.
-        let keys = IdentityKeys {
-            ed25519: Some([4u8; 32]),
-            ml_dsa_65: None,
-            assurance: Assurance::Classical,
-        };
+        let keys = classical_keys([4u8; 32]);
         let gw = AtprotoPerimeterGateway::new(FixtureResolver(keys));
         let peer = gw.enroll(&admitted(Some(DID_WEB), [4u8; 32])).unwrap();
 
@@ -572,11 +570,7 @@ mod tests {
 
     #[test]
     fn enrollment_store_insert_get_roundtrip() {
-        let keys = IdentityKeys {
-            ed25519: Some([5u8; 32]),
-            ml_dsa_65: None,
-            assurance: Assurance::Classical,
-        };
+        let keys = classical_keys([5u8; 32]);
         let gw = AtprotoPerimeterGateway::new(FixtureResolver(keys));
         let peer = gw.enroll(&admitted(Some(DID_WEB), [5u8; 32])).unwrap();
 
@@ -601,11 +595,7 @@ mod tests {
         let native = crate::envelope::KeyedPqTrustStore::new();
         assert!(native.is_empty());
 
-        let keys = IdentityKeys {
-            ed25519: Some([6u8; 32]),
-            ml_dsa_65: Some(an_ml_dsa_vk()),
-            assurance: Assurance::PqHybrid,
-        };
+        let keys = hybrid_keys([6u8; 32]);
         let gw = AtprotoPerimeterGateway::new(FixtureResolver(keys));
         let peer = gw.enroll(&admitted(Some(DID_WEB), [6u8; 32])).unwrap();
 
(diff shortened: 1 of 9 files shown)


ewindisch:
# PR #1202 review — BLOCK

Reviewed commit `17771aed4516a563ad1f200fd0a4a3ec1858f756`.

## Findings

### BLOCK — The production DID publisher is still a singleton, and the “bounded drain” test has no bound or production publisher

`root_did_document` still constructs exactly one Ed25519 VM from the scalar `state.signing_key`:

- `crates/hyprstream/src/services/oauth/did_document.rs:454-461,518-531`
- `crates/hyprstream/src/services/oauth/state.rs:399,731-742`

The new resolver test at `identity_resolver.rs:546-576` calls a test-only `RotatingDocs::publish` to replace one JSON value with another. No clock advances, no `nbf`/`exp` or drain deadline exists, no observation condition exists, and the production root-DID publisher is not exercised. The gateway overlap test at `atproto_perimeter.rs:526-555` likewise injects an `IdentityKeys` fixture rather than resolving a document emitted by the production publisher.

This does not meet the acceptance bar from #1183/#1187: production still cannot publish lead + active/drain simultaneously, so a root identity rotation remains a flag day. In contrast, the working JWKS path publishes every real drain/active/lead slot with `kid`, `nbf`, and `exp` (`services/oauth/jwks.rs:50-66`, `auth/key_rotation.rs:1-11,1074-1088`).

Required: give the root DID path a named, bounded rotation set (or an equivalent lifecycle), publish the new VM before use, retain the old VM through a real bounded drain, and add a production-bound test that accepts signatures from both during overlap and rejects the retired key only after an injected clock passes the bound.

### BLOCK — The `#key-1` → `#mesh` flag-day rename breaks rolling upgrades and existing DID-URL references

Before this PR, the root document publishes Ed25519 as `#key-1` and ML-DSA as `#mesh-pq`. The new resolver binds PQ only when the PQ id is exactly `"{ed_id}-pq"` (`identity_resolver.rs:321-329`), while this PR changes the producer to emit Ed25519 as `#mesh` in the same commit (`did_document.rs:518-521`).

A new resolver talking to an old-but-valid producer therefore resolves `#key-1` without its published `#mesh-pq` companion and silently downgrades that peer from `PqHybrid` to `Classical`. `AtprotoPerimeterGateway::enroll` then clamps the enrolled context to that lower assurance (`atproto_perimeter.rs:130-147`), denying legitimate access to PQ-hybrid-labelled objects during an ordinary rolling deployment. Separately, consumers holding the previously published stable DID URL ending in `#key-1` lose that reference immediately.

There is no overlap/migration test for old producer + new consumer. Publishing both ids as aliases is not currently a safe bridge either: `candidate_for_ed25519` rejects two named candidates carrying the same Ed25519 bytes as ambiguous (`identity.rs:309-317`).

Required: define a non-flag-day fragment migration (or retain the stable fragment and add a compatible explicitly named PQ companion), and test both mixed-version directions plus retirement after the compatibility window.

### BLOCK — A malformed `authentication` relationship disables the relationship check

`did_web_key_candidates` parses `authentication` with:

```rust
doc.get("authentication").and_then(Value::as_array)
```

and treats the resulting `None` as “relationship absent; accept the whole `verificationMethod` set” (`identity_resolver.rs:279-303`). Thus a present but malformed value such as `"authentication": {}` or `"authentication": "bad"` is indistinguishable from an omitted relationship and authorizes every compatible VM. This is the known “absent expected value disables the check” failure shape, inside the new type/path intended to make selection safe.

Required: distinguish absent from present-but-malformed. If omission retains the documented compatibility fallback, malformed relationship data must fail closed. Add negative tests for wrong types and decide explicitly how legal embedded relationship methods are handled.

### HIGH — The new key-set type still exposes positional selection

`IdentityKeys` exposes `pub candidates: Vec<IdentityKeyCandidate>` (`identity.rs:302-305`). Any consumer can still call `.first()`, `.next()`, or index it; the PR itself indexes `keys.candidates[0]` in a resolver test. The comments prohibit the original defect, but the API preserves its exact spelling, unlike the audit’s proposed `PublishedKeySet` surface with only named/candidate lookup.

Required: keep the collection private and expose purpose/id/signer candidate operations that do not permit authority to be selected by position.

## Validation

All compiling commands were run through `buildq`; the reviewed tree was restored clean after each causal check.

- `buildq -- cargo test -p hyprstream-rpc --lib --no-fail-fast` — **PASS**: 775 passed, 3 ignored.
- `env LIBTORCH=.../appimage/libtorch-cache/cuda130/libtorch LD_LIBRARY_PATH=.../lib LIBTORCH_BYPASS_VERSION_CHECK=1 OPENSSL_NO_VENDOR=1 buildq -- cargo test -p hyprstream --lib services::oauth::did_document::tests` — **PASS**: 19 passed.
  - The first attempt, without the repository’s cached libtorch environment, stopped in `torch-sys` before source compilation; this was an environment-only prerequisite failure.
- Temporarily restoring first-only resolver reduction (`.take(1)`) made `did_web_classical_overlap_accepts_both_then_rejects_retired_key_after_drain_window` fail with 1 candidate instead of 2.
- The same first-only reduction made `did_web_two_hybrid_pairs_remain_independent_overlap_candidates` fail because the new hybrid candidate was absent.
- Temporarily restoring first-candidate gateway selection made `overlap_selects_each_admitted_signer_then_rejects_the_retired_candidate` fail because the new signer received the old candidate’s `Classical` assurance instead of `PqHybrid`.
- Final `git diff --exit-code` and `git status --short` were clean at the reviewed SHA.

The three new tests are causally sensitive to the local selection fixes, but they do not cover the missing bounded production publication lifecycle or rolling-version compatibility; those omissions are blocking.

ewindisch:
# Verdict — PR #1202 `fix(identity): resolve DID verification key sets` — **REVISE**

Reviewer: fable (one of two independent reviewers). Read-only review of
https://github.com/hyprstream/hyprstream/pull/1202 (fixes
https://github.com/hyprstream/hyprstream/issues/1187, child of
https://github.com/hyprstream/hyprstream/issues/1183), judged against the audit at
`.fleet-coord/keyaudit/report.md` and the in-tree JWKS reference
(`crates/hyprstream/src/services/oauth/jwks.rs:50-130`, `crates/hyprstream/src/auth/key_rotation.rs`).

**Summary.** The core replacement is right and the acceptance bar is met: identity material is now a
named candidate set, the consumer selects by the admission-verified signer (never document order),
hybrid pairs are explicit per-candidate bindings, and two-key overlap plus retired-key rejection are
tested at both the resolver and perimeter layers — all four new discriminating tests **revert-verified**
(details at bottom). It also fixes a real latent escalation the old code had: the first-ed/first-PQ
mispairing could attach *another key's* ML-DSA anchor to a classical signer and enroll it PqHybrid.
But the PR ships an untested new `authentication`-relationship filter that rejects spec-legal documents
while being bypassable by omission (#1145 shape), has no compatibility story with the fleet's own
already-published `#key-1` documents, and leaves the publisher half of #1187's audit row (root
singleton) undischarged. REVISE, not BLOCK: nothing is exploitable (all failure directions are
fail-closed) and neither the resolver nor the perimeter gateway has a production construction site yet
(verified: zero non-test references to `MethodDispatchResolver`/`AtprotoPerimeterGateway`/`DidDocumentProvider`).

---

## Findings (severest first)

### F1 — MAJOR: the new `authentication` filter rejects legitimate documents, is bypassable by omission, and is untested in every branch

`did_web_key_candidates` (`crates/hyprstream-rpc/src/identity_resolver.rs:498-518`) newly drops any VM
whose full `id` string is not a *string* entry of `authentication`. Nothing in #1187 asked for this, and
it misfires on spec-legal shapes:

- **Relative fragment refs** — a document whose `authentication` is `["#mesh"]` (DID-core-legal relative
  DID URLs, common in the wild) matches nothing against the VM's full `did:web:…#mesh` id → **every VM
  is filtered → resolution fails entirely** for a valid document. Per the review brief: a fail-closed
  conversion that denies the legitimate path is an outage, not a safer bug.
- **Embedded VM objects** in `authentication` (also DID-core-legal) are silently dropped by
  `filter_map(Value::as_str)` — same total-rejection outcome.
- **Silent hybrid downgrade** — a PQ VM listed under `assertionMethod` but not `authentication` is
  filtered, so a hybrid identity resolves Classical with no error.
- **#1145 shape, and therefore security-vacuous** — when `authentication` is *absent* the check is
  disabled entirely, so a malicious document bypasses it by omitting the field. The filter contributes
  no enforcement while adding the breakage above. (The brief asked to grep the diff for exactly this
  pattern, including inside new types meant to abolish it: `authentication.as_ref().is_some_and(|ids| !ids.contains(id))`
  is it.)
- **Inconsistent with admission** — `crates/hyprstream-rpc/src/admission.rs:329` extracts VMs with no
  relationship filter, so a signer can be *admitted* and then fail *enrollment* on the same document.
- **Zero tests** cover the filter: not present-and-filtering, not relative-ref, not embedded-object, not
  the absent-field fallback.

**Fix:** drop the filter (restoring the previous relationship-agnostic behavior — sufficient for #1187),
or normalize relative refs + handle embedded VMs + align admission + add tests for all four branches.

### F2 — MAJOR: no compatibility with the fleet's own already-published documents — rolling upgrade silently downgrades hybrid; live `#key-1` id renamed without a bridge

Every deployed node today publishes `#key-1` (+ `#key-1-jwk`) alongside `#mesh-pq`
(pre-PR `root_did_document`). The new pairing rule requires `{ed_id}-pq`, so an upgraded verifier
resolving a not-yet-upgraded peer builds candidate `#key-1`, looks for `#key-1-pq`, finds nothing, and
ignores the published `#mesh-pq`: **the peer resolves Classical**. During any mixed-version window the
whole fleet's PqHybrid assurance is silently lost — the exact mechanism #1183 exists to protect (hybrid
rollout = overlap rotation). The PR neither shims the legacy shape (e.g. honoring a lone `#mesh-pq`
when the document has exactly one Ed25519 VM) nor mentions the migration.

Secondarily, the rename changes the **live** `/.well-known/did.json` verification-method ids
(`crates/hyprstream/src/services/oauth/did_document.rs:521`). In-tree consumers select by key bytes, not
id, so nothing in-tree breaks — but any external verifier or stored reference pinned to
`did:web:…#key-1` / `#key-1-jwk` breaks with no deprecation window. At minimum the PR description
should state the compatibility decision explicitly.

Severity is MAJOR-not-CRITICAL only because the resolver/perimeter pair has no production wiring yet;
this becomes the outage at wiring time if merged as-is and forgotten.

### F3 — MEDIUM: `candidate_for_ed25519` rejects duplicate publications of the *same* key — denying a legitimately admitted signer

`crates/hyprstream-rpc/src/identity.rs` (`candidate_for_ed25519`): two candidates carrying the **same**
Ed25519 key under different VM ids → `None` → perimeter enroll fails for a signer admission just
verified. Publishing one key under both the old and new id is the natural bridge for exactly the
`key-1`→`mesh` rename this PR performs (and the F2 shim). Ambiguity is only dangerous when the *hybrid
bindings differ*; identical or absent bindings should collapse to one candidate (or the
hybrid-consistent one). The duplicate-rejection branch is also untested.

### F4 — MEDIUM: the publisher half of #1187's audit row is not discharged — root still publishes exactly one mesh key

Audit row (report.md:19) assigned the root did:web defect — "No rotation-slot input or named lead/drain
VM is exposed for the root entity key" — to #1187 *together with* its scalar consumer. This PR fixes the
consumer and renames the slot, but `root_did_document` still hardcodes
`vec![("mesh".to_owned(), vk)]`: overlap for the root entity key remains **impossible to produce**, so
end-to-end the mesh cannot actually rotate this key with overlap. Judged against the JWKS reference,
the consumer side now matches the working pattern, but the publisher side has no drain/active/lead
analogue (contrast `jwks.rs:50-130`, which publishes all slots with `nbf`/`exp`). Relatedly, the
`#mesh` ↔ `#mesh-pq` convention exists as unlinked string literals in two crates with **no test that our
own published document round-trips through `did_web_key_candidates`** — a rename on either side would
silently downgrade hybrid again. Either add publisher slots + a producer↔consumer round-trip test in
this PR, or file the follow-up ticket before "Fixes #1187" closes the audit row.

### F5 — MEDIUM (largely pre-existing): retired-key rejection only applies to *fresh* enrollments; candidates carry no validity bounds

The new tests prove a fresh resolution rejects a removed key — but `EnrollmentStore` pins the derived
context forever and the #548 S2 monitor is documented to read the cache and never re-resolve. A peer
enrolled under a key later withdrawn from its document stays enrolled indefinitely: the "bounded
window" exists only at enroll time. `IdentityKeyCandidate` also carries none of the audit's durable-rule
fields (`not_before`/`not_after`, explicit hybrid-binding id) that the JWKS reference publishes per
slot. The pin-until-re-enroll design predates this PR, so this is a follow-up, not a defect introduced
here — but the ticket family is precisely about this lifecycle, so it should be tracked, not assumed.

### F6 — MINOR: at9p arm stays single-candidate — sequencing with open PR #1210

`resolve_did_at9p` maps the scalar `VerifiedAt9pKeys` to one candidate. Open PR
https://github.com/hyprstream/hyprstream/pull/1210 (#1188) makes at9p capsules multi-key and was
reviewed with the note that multi-key resolve hard-fails "until #1187 lands" — that is only true if
whichever PR lands second maps *every* verified subject key to a candidate. As written, #1202 does not;
the outage noted in #1210's review would persist. Coordinate the merge order and the follow-up.

### F7 — NIT

Drive-by rustfmt hunks in untouched code (`identity.rs`, several `identity_resolver.rs` error strings) —
harmless, repo is not fmt-clean (#1140), but they pad the diff. Misindented JSON literal inside
`did_doc_with_two_hybrid_pairs` (`"type"`/`"controller"` of `#mesh-new-pq`).

---

## What the change breaks (enumeration)

1. Third-party did:web documents using relative-fragment or embedded-object `authentication` entries:
   previously resolved (Classical/PqHybrid), now **rejected entirely** (F1).
2. Documents whose PQ VM is not in `authentication`, or not named `{ed_id}-pq`: previously PqHybrid,
   now silently Classical — includes **every currently deployed node's own document** (`#key-1` +
   `#mesh-pq`) when resolved by upgraded code (F1/F2).
3. External consumers pinned to the published `#key-1`/`#key-1-jwk` ids (F2).
4. Documents publishing the same Ed25519 key under two VM ids: previously usable, now the signer is
   rejected as ambiguous (F3).
5. Enrollment assurance semantics tightened (deliberate, good): a signer absent from the resolved set
   now fails enrollment where old code enrolled on document-level assurance without signer binding —
   the old behavior was the escalation bug, so this "break" is the fix.

## What was verified

- `buildq -- cargo test -p hyprstream-rpc --lib identity` / `atproto_perimeter` /
  `identity_resolver`: all pass at PR head (49 + 15 + resolver suites, 0 failures).
- **Revert verification** (emulate pre-fix behavior against the new tests, then restore):
  - Perimeter selection reverted to `keys.candidates.first()` →
    `candidate_set_without_admitted_signer_fails_closed` and
    `overlap_selects_each_admitted_signer_then_rejects_the_retired_candidate` **FAIL** (2/15).
  - Resolver reverted to scalar first-ed + arbitrary-PQ pairing →
    `did_web_classical_overlap_accepts_both_then_rejects_retired_key_after_drain_window` and
    `did_web_two_hybrid_pairs_remain_independent_overlap_candidates` **FAIL** (2/15).
  Both experiments restored; worktree clean.
- Wiring check: no non-test construction site of `MethodDispatchResolver`, `DidDocumentProvider`, or
  `AtprotoPerimeterGateway` anywhere in the workspace (bounds F1/F2 blast radius today).
- Producer check: `build_did_document` emits full-form `authentication` ids and `#mesh-pq`, so the
  renamed root document does self-resolve as one hybrid candidate under the new consumer.
- CI green at head; the `build` job is merge_group-only (known trap), so local test runs above are the
  actual evidence.

## Acceptance bar

Met at the unit level, with the revert-verified overlap + retired-rejection tests required by the brief
(resolver and perimeter both). Not yet met end-to-end: the mesh's real publisher cannot produce the
overlap those tests consume (F4), and the new filter can reject the very overlap documents the thesis
requires (F1). Hence **REVISE**: fix F1 (drop or complete the filter), state/ship the F2 compatibility
story, and resolve F4's ticket accounting; F3 strongly recommended in the same pass.

ewindisch:
VERDICT: BLOCK

# PR #1202 review — BLOCK

Reviewed commit `17771aed4516a563ad1f200fd0a4a3ec1858f756`.

The local first-candidate defect is repaired: the resolver retains every compatible
Ed25519 candidate, binds each ML-DSA key by VM id, and enrollment selects the
admission-verified signer. The new tests are causally sensitive to those changes.
That is not sufficient to merge. This is the same SHA reviewed in the prior
BLOCK/REVISE verdicts, with no subsequent fix commit, and the production overlap
lifecycle, relationship parsing, rolling-version behavior, and positional API
findings remain present.

## Findings

### BLOCK — The required bounded production overlap lifecycle still does not exist

The #1183 audit assigns both the singleton root publisher and scalar consumer to
#1187. The consumer is now a candidate set, but the production root DID publisher
still reads one scalar `state.signing_key` and constructs exactly one Ed25519 VM:

- `crates/hyprstream/src/services/oauth/did_document.rs:454-461,518-531`
- `crates/hyprstream/src/services/oauth/state.rs:397-399,719-742`

There is no root-DID drain/active/lead state, validity interval, injected clock,
promotion rule, consumer-observation condition, or production code capable of
publishing old and new root keys simultaneously.

The test named
`did_web_classical_overlap_accepts_both_then_rejects_retired_key_after_drain_window`
does not test a bounded window. At
`identity_resolver.rs:546-575`, test-only `RotatingDocs` starts with two hand-built
JSON keys and then immediately replaces the document with another JSON value.
No bound is represented or advanced, and neither `root_did_document` nor
`OAuthState` participates. The perimeter test at
`atproto_perimeter.rs:526-555` likewise replaces an injected `IdentityKeys`
fixture.

This invents a weaker pattern than the working reference. `jwks_json` publishes
every real slot with `kid`, `nbf`, and `exp`
(`services/oauth/jwks.rs:50-66`), while `KeySlots` and `rotate_once` implement
lead publication, active promotion, bounded drain retention, and expiry
(`auth/key_rotation.rs:1043-1087,1257-1314`).

Required: give the root identity a real named rotation set, publish lead before
use, retain drain through an explicit bound/observation condition, and add a
production-bound clocked test proving both signatures are accepted during
overlap and the retired signature is rejected only after the bound.

### BLOCK — The new `authentication` check is both bypassable by malformed/absent data and rejects legitimate DID documents

`did_web_key_candidates` uses:

```rust
doc.get("authentication")
    .and_then(Value::as_array)
```

and applies the filter only when that produces `Some`
(`identity_resolver.rs:279-303`). A present wrong type therefore collapses to
the same state as an absent property and disables the check, authorizing every
compatible `verificationMethod`. This is the exact “absent expected value
disables the check” shape called out by the review brief.

The present-array path is also an outage for valid documents: it accepts only
absolute string references. W3C DID Core permits relative DID URL references
such as `"#key-1"` and embedded verification-method objects in
`authentication`; both are silently dropped here, potentially filtering every
valid signer. See the W3C DID Core sections on
[relative DID URLs and verification relationships](https://www.w3.org/TR/did-core/#verification-relationships).
No test covers absent, malformed, relative-reference, or embedded-object
branches, and admission uses a different relationship-agnostic extractor.

Required: either remove this newly introduced filter and retain the prior
relationship-agnostic contract, or implement one authoritative DID relationship
parser shared by admission and enrollment that distinguishes absent from
malformed, resolves relative references, supports embedded methods, and has
positive and negative tests for every shape.

### BLOCK — `#key-1` to `#mesh` is a flag-day change that silently removes mandatory hybrid assurance during rolling deployment

Before this PR the public root DID endpoint emits Ed25519 as `#key-1` and
ML-DSA as `#mesh-pq`. The new resolver only pairs `{ed_id}` with
`{ed_id}-pq` (`identity_resolver.rs:321-329`), while this PR changes the
publisher to `#mesh` (`did_document.rs:518-521`).

An upgraded verifier resolving an existing producer therefore sees
`#key-1`, searches for `#key-1-pq`, ignores the published `#mesh-pq`, and
returns a Classical candidate. On the internal mesh, where PQ-hybrid is
mandatory, this denies legitimate access rather than safely degrading.
Existing DID URL references to `#key-1`/`#key-1-jwk` also disappear
immediately. No mixed-version overlap test or bounded retirement exists.

The obvious overlap bridge—publishing the same Ed25519 key under both ids—does
not work with this API: `candidate_for_ed25519` treats two matching candidates
as ambiguity and returns `None` (`identity.rs:309-317`), rejecting the signer.

Required: retain a stable fragment/pairing contract or implement a bounded,
tested fragment transition that preserves mandatory hybrid assurance in both
mixed-version directions and retires the old id only after the bound.

### HIGH — The replacement type still exposes the exact positional-selection spelling the audit prohibits

`IdentityKeys` exposes `pub candidates: Vec<IdentityKeyCandidate>`
(`identity.rs:302-305`). Any downstream consumer can select `.first()`, index,
or iterate-to-first; the PR itself uses `keys.candidates[0]` in a resolver test
(`identity_resolver.rs:766`). The type comment says not to select by position,
but the structural fix requested by #1183 was a key-set API with only named,
purpose, signer, or verify-any-compatible operations.

Required: make the collection private and expose only non-positional
construction and lookup/verification operations. There should be one
authoritative implementation of candidate selection.

### HIGH — “Retired-key rejection” applies only to new enrollment attempts

`EnrolledPeer` deliberately pins assurance and the monitor never re-resolves
(`atproto_perimeter.rs:50-58`). An identity enrolled with the old candidate
before document replacement remains enrolled indefinitely. Consequently the
gateway test proves only that a fresh enrollment fails against a hand-edited
fixture; it does not prove the retired key stops authorizing after a bounded
window.

This is partly pre-existing lifecycle debt, but it directly contradicts the
acceptance claim in this PR. The production-bound test must exercise the actual
authorization consumer after retirement, not only a fresh resolver call.

## What the change breaks

1. A new verifier resolving the existing `#key-1` + `#mesh-pq` root document
   silently derives Classical assurance, causing mandatory-hybrid internal
   authorization failures.
2. External/stored DID URL references to the existing `#key-1` and
   `#key-1-jwk` fragments stop resolving after publisher upgrade.
3. Valid DID documents using relative `authentication` references or embedded
   authentication methods can be rejected entirely.
4. Documents that publish one Ed25519 key under multiple names are rejected as
   ambiguous even when the bindings are identical.
5. The intentional break is correct: an admitted signer absent from the
   resolved candidate set now fails enrollment instead of inheriting another
   candidate’s assurance.

## Prior-verdict reconciliation

- Production singleton / fake bounded-drain test: **open**.
- `#key-1` fragment flag day and mixed-version hybrid downgrade: **open**.
- Malformed `authentication` disabling the check: **open**.
- Relative/embedded `authentication` outage and admission inconsistency:
  **open**.
- Public positional `Vec`: **open**.
- Duplicate-publication rejection: **open**.
- Indefinitely pinned retired enrollment: **open**.
- The prior at9p single-candidate sequencing note is owned by #1188/PR #1210,
  not fixed at this SHA and not used as an additional blocker here.

No new commit followed either prior verdict, so none of these could have been
closed by the requested fix pass.

## Validation

All Cargo commands ran through `buildq`. The worktree was restored clean after
each mutation.

- `buildq -- cargo test -p hyprstream-rpc --lib --no-fail-fast` — **PASS**:
  775 passed, 3 ignored.
- Cached libtorch/OpenSSL environment +
  `buildq -- cargo test -p hyprstream --lib services::oauth::did_document::tests -- --nocapture`
  — **PASS**: 19 passed.
- Happy paths exercised by those suites include classical and hybrid did:web
  resolution, classical and hybrid enrollment, exact candidate selection, and
  the producer’s single `#mesh`/`#mesh-pq` document construction.
- Temporary resolver regression: adding first-only `.take(1)` made
  `did_web_classical_overlap_accepts_both_then_rejects_retired_key_after_drain_window`
  and `did_web_two_hybrid_pairs_remain_independent_overlap_candidates`
  **FAIL** (6 passed, 2 failed). Restored.
- Temporary gateway regression: replacing exact signer lookup with
  `keys.candidates.first()` made
  `candidate_set_without_admitted_signer_fails_closed` and
  `overlap_selects_each_admitted_signer_then_rejects_the_retired_candidate`
  **FAIL** (13 passed, 2 failed). Restored.
- Final `git diff --exit-code` and `git status --short` — clean.
- Reviewed and remote PR head both equal
  `17771aed4516a563ad1f200fd0a4a3ec1858f756`.

The unit-level selection fix is real, but the mandatory production overlap
acceptance bar is not met. Do not queue this PR for merge.

Issue #1157, opened 2026-07-22
security(at9p): mutual alias binds DID strings, not key material — #1143 F1


ewindisch:
## Summary

The `did:at9p` mutual-alias pin binds identifier **strings**, not the key material actually
installed. Surfaced as the **F1 BLOCK** on PR #1143 (DID-anchored trust bootstrap) and
independently reproduced.

`verify_did_anchored_document` (`crates/hyprstream-discovery/src/did_anchored.rs:173-222`)
takes **both** pieces of deployment trust material from the **unsigned `did:web` JSON
document**:

- `ca_verifying_key` ← `verification_method_ed25519_keys(document)` (:198-205)
- `discovery_transport` ← `preferred_transport(document, None)` (:207-215)

The `did:at9p` capsule is consulted only via `At9pAliasResolver::resolve_authoritative`
(:189-196), and the ratified alias rule (`crates/hyprstream-pds/src/at9p_alias.rs:82-131`,
#905 §2/§6) compares **DID strings only** — leg 1: `classical_aka_at9p == verified.did()`;
leg 2: the capsule's `also_known_as` contains the `did:web` string. **Neither leg commits to
the document's `verificationMethod` or `service` contents.**

**Consequence.** An adversary controlling the `did:web` origin (DNS + a DV cert; no DNSSEC,
no cert pinning) replays the genuine, hash-pinned capsule byte-for-byte (it is public data)
and swaps the CA key and reach for their own. Both documents verify identically. On a
same-host UDS bootstrap the socket carries most of the assurance and this is survivable; on
the **multi-host bootstrap that Erica ruled mandatory (#1135)** the network document *is* the
trust material, so this escalates from a defect to a hard blocker.

Today the attack is blunted only by the root-owned, `kid`-bound `registry-service.jwt` — the
exact file #1136's enrollment increment exists to delete. So the sole compensating control is
scheduled for removal: F1 must be fixed before enrollment, not after.

## Independent reproduction

Two `#[ignore]`d evidence tests committed to `crates/hyprstream-discovery/src/did_anchored.rs`
(branch `f1-alias-binding-evidence`, off #1143 head `9081eef84`):

- `f1_repro_string_only_alias_does_not_bind_document_contents` — security repro; replays the
  genuine capsule verbatim and swaps only the CA + reach. Both are accepted:
  `F1 CONFIRMED: did:at9p pin did not bind did:web document contents; CA and reach were both
  substituted and accepted`.
[...]

Issue #1137, opened 2026-07-22
bug: stacks/hyprstream cannot boot — production resolver fails closed, stack provisions neither trust file


ewindisch:
## The metal Quadlet stack cannot start on current `main`

`stacks/hyprstream` (metal repo) is merged and was about to be applied as the first production deployment. It **fails closed at startup**. Verified end to end:

**1. The production resolver install is unconditional and runs before service dispatch.**
`crates/hyprstream/src/bin/main.rs:1893-1902`:
```rust
install_process_production_resolver(&signing_key)
    .context("Failed to install checkpoint-backed production resolver")?;
```
Only `wizard`, `mac genesis`, and `pds join` dispatch earlier (`bin/main.rs:1841-1900`). `service start <svc> --foreground --ipc` — what the Quadlets run — does **not**.

**2. That install requires two root-owned files.**
`hyprstream-discovery/src/service.rs:1497-1499` + `read_os_owned_file` (`:1869-1899`): both `/etc/hyprstream/trust/deployment-ca.ed25519` and `/run/hyprstream/credentials/registry-service.jwt`, **and every ancestor directory up to `/`**, must be regular files (not symlinks), `uid == 0`, and not group/world-writable.

**3. The stack provisions neither.**
`grep -i "deployment-ca|registry-service.jwt|/etc/hyprstream/trust"` across `stacks/hyprstream/` → **zero hits**. The Quadlets run as user `rocky` under `systemctl --user`, e.g.:
```
Network=host
Volume=hyprstream-ipc.volume:/run/hyprstream
Exec=service start model --foreground --ipc
```

**Net effect:** even once an image publishes (blocked on #1133), `tofu apply` produces an instance whose services fail closed at boot. The documented deploy path — "pin a verified image digest, then apply" — does not work.

## Why this wasn't caught

The stack was reviewed for infrastructure correctness (VPC, EBS, EIP, Quadlet wiring, TLS) but never against the binary's startup preconditions. There is no CI or acceptance check that boots the image with the stack's actual invocation.

## Complicating factor: the userns loophole

[...]

PR #1209, opened 2026-07-23
fix(discovery): source DID-anchored CA + reach from GATE-verified capsule (#1157 option C)

ewindisch:
## What this is

Implements **option C** for the #1143 **F1 BLOCK**, per Erica's decision on #1157
(2026-07-22). The DID-anchored deployment CA and discovery reach are now sourced
from the **GATE-verified `did:at9p` capsule**, not the unsigned `did:web`
document. Branched off `feat/1136-did-anchored-trust` (the #1143 branch); does
**not** rewrite #1143.

## The bug (F1)

`verify_did_anchored_document` took both the deployment CA
(`verification_method_ed25519_keys`) and the reach (`preferred_transport`) from
the **unsigned** `did:web` document, while the mutual-alias rule (#905 §2/§6)
compared only DID identifier **strings**. An adversary controlling the `did:web`
origin (DNS + DV cert) could replay the genuine, hash-pinned capsule byte-for-byte
and substitute their own CA + reach — both documents verify identically. On the
mandatory multi-host bootstrap (#1135) the network document *is* the trust
material, so this was a trust-root compromise, not a survivable defect.

## The fix (option C)

- **CA** ← capsule `body.subject_keys[0]` (the hybrid key that GATE-signed the
  capsule). **Reach** ← capsule `#ns` `NinePExport` service entry, dialed by its
  independent iroh `nodeId` carrier.
- The `did:web` document is used **only** for the #905 §2 leg-2 reciprocal
  identifier vouch (`alsoKnownAs` must name the configured `did:at9p`). Its keys
  and reach are advisory hints, never installed. This is what ratified **#905 §8**
  already requires — *"GATE is authority; addresses select; verification asserts."*

### Carried with C

- **F5 / #556 (classical floor fixed, not deferred).** The CA is a hybrid subject
  key, so `PqHybrid` is *carried*, not reconstructed. `authenticate_did_anchored_bootstrap`
  fails closed if the installed authority is ever not `PqHybrid`. The deployment CA
  anchors the registry key → audit key → accepted-state checkpoints; a classical
  root would have chained the composite issuance (`services/policy.rs`), #1034 and
  #1058 to a classical SPOF.
- **F4 subsumed.** The `ca_keys.len() == 1` document rule is **deleted**. Selection
  is by identity from the capsule, so a document publishing overlapping (old + new)
  CAs during rotation no longer breaks bootstrap for every node at once.
- **#1031 honored.** Reach is the service entry's explicit `nodeId` carrier, never
  derived from the subject identity key; a carrier-less entry fails closed. The reach
  stays an untrusted carrier — response is pinned to the separately resolved Discovery
  key and a signed `ping` is still required before the resolver installs.

## Tests

Every fix has a test that fails without its hunk. Verified by reverting to
document-sourcing (6 tests fail); restored. In `hyprstream-discovery` lib
(`did_anchored::tests`):

| test | guards |
|---|---|
| `substituted_document_ca_and_reach_are_ignored` | F1 attack neutralized |
| `ca_is_bound_to_verified_capsule_not_document` | former #1143 `#[ignore]`'d expected-fail target, now **passing** (un-ignored) |
| `installed_authority_is_pqhybrid_not_classical` | F5 / #556 regression guard |
| `rotation_with_overlapping_document_cas_still_bootstraps` | F4 |
| `reach_without_carrier_node_id_fails_closed` | #1031 |
| `f1_capsule_carries_the_material_the_document_should_not_own` | evidence (retained) |

## F3 / F6 / F7 re-assessed against the new shape

- **F3** (did:web document unsigned, no freshness/rollback) — **largely dissolved.**
  The document is no longer trust material; it carries only the reciprocal identifier
  string. A rollback/replay of the document can no longer reinstate a retired CA
  because the CA comes from the hash-pinned capsule. **Residual:** capsule *currentness*
  (a rolled-back genesis capsule) is a separate concern owned by the accepted-state /
  duplicity machinery (#1004) and the KERI chain, not this path. No new ticket; tracked
  under existing rotation work (#1123/#1126) and #1004.
- **F6** (`dial_local_transport_bootstrap` chokepoint bypassed) — **still real,
  unchanged by C.** The DID path still calls `hyprstream_rpc::dial::dial` directly with
  a network config, sidestepping the `Inproc|Ipc|SystemdFd` guard. C changes *where the
  reach comes from*, not *how it's dialed*. Needs its own follow-up (widen the guard with
  an explicit `for_did_anchored_bootstrap` constructor, or add a source-control test).
  **Recommend a tracking ticket** — out of scope here.
- **F7** (liveness authenticates a locally-provisioned Discovery key, not the DID
  identity) — **narrowed but still real.** Reach now comes from the capsule, but the
  `ping` is still verified against the `discovery` key from the local trust store
  (`global_trust_store().resolve_one("discovery")`), not one derived from the DID. So the
  liveness proof still binds the capsule-derived *address* to a locally-provisioned
  identity. This is the #1136 §3 "no local Discovery dependency" goal, still unmet.
  Distinct from F1; **recommend folding into the enrollment / #1136 next increment** and
  tracking there. Not a regression introduced by C.

## Scope / coordination

- Touched: `crates/hyprstream-discovery/src/{did_anchored.rs, at9p_alias.rs, service.rs}`
  and `docs/deployment-registry-trust.md`.
- Stayed out of the #1137 sibling's files (`feat/1137-did-trust-e2e` /
  `feat/1137-capsule-serving`): the capsule-serving well-known endpoint and the e2e boot
  harness. `HttpWellKnownCapsuleSource` is unchanged here.
- **F2 / #1137 is independent and NOT addressed** — the root-owned `/run` JWT requirement
  remains; C does not unblock the boot.

## Verification

- `cargo test -p hyprstream-discovery -p hyprstream-pds` — all pass (discovery 89, pds 199,
  integration suites green).
- `cargo clippy --workspace --all-targets --release -- -D warnings` — clean.
- All builds run through `buildq` per the fleet build policy. `cargo fmt --check` fails on
  a clean base (#1140), not from this change.

Refs #1157 #1143 #1136 #1135 #556


ewindisch:
# Review — PR #1209 “fix(discovery): source DID-anchored CA + reach from GATE-verified capsule (#1157 option C)”

Reviewer: sol (independent reviewer)

Reviewed head: `d943cc17b91bbf667556d97c65a8f38e897f4b9f`

## VERDICT: BLOCK

The patch does close the original `did:web`-origin substitution at the local
`DidAnchoredTrust` helper: document `verificationMethod` and `service` contents no
longer select the CA or reach, while the configured capsule CID still passes the
canon/hash/hybrid-signature GATE and supplies both values. A document-origin
attacker can make bootstrap unavailable by withholding the reciprocal vouch, but
cannot choose the helper's returned CA or carrier.

It does not, however, meet either other load-bearing acceptance condition. The
installed registry-credential root is still cryptographically Ed25519-only even
though it is labeled `PqHybrid`, and the new capsule consumer immediately reduces
`subject_keys` to `.first()`. There is no capsule-key overlap, no accepted-state
rotation path, and no bounded retirement test. The advertised overlap test only
puts two ignored keys in the `did:web` document.

## Findings (severest first)

### F1 — CRITICAL / BLOCKING: `PqHybrid` is propagated as a label, but the installed root discards the ML-DSA-65 key and verifies an EdDSA-only credential

`ca_key_from_capsule` selects `subject_keys.first()` and copies only
`primary.ed25519_pub` into an `ed25519_dalek::VerifyingKey`
(`did_anchored.rs:199-221`). `DidAnchoredTrust` has no ML-DSA-65 key or atomic
hybrid pair, only that scalar Ed25519 key plus an `Assurance` enum
(`did_anchored.rs:102-115`).

The enum is inherited from `AuthoritativeIdentity` because the *capsule* passed a
hybrid GATE (`did_anchored.rs:309-315`). At the actual trust installation boundary,
`authenticate_did_anchored_bootstrap` compares the enum to `PqHybrid` and then
immediately passes only `trust.ca_verifying_key` into
`TrustedRegistryDeploymentCredentials` (`service.rs:2030-2048`). The credential
profile requires `alg == "EdDSA"` and verifies with
`jwt::decode(token, ca_verifying_key, ...)`; the capsule's ML-DSA-65 half never
participates (`service.rs:1757-1785,1853-1860`). The resulting registry verifier
is also a scalar Ed25519 key.

GATE proving that a capsule was hybrid-signed does not make a later EdDSA-only
credential signature hybrid. A CRQC attacker that forges the selected Ed25519 CA
can mint a conforming credential whose `cnf` installs an attacker registry key;
the `PqHybrid` enum check still succeeds because it describes the earlier capsule
verification, not the credential verification. That leaves the registry, audit,
checkpoint, composite issuance, and hybrid-KEM chain rooted in the classical
single point of failure that #556/#1157 explicitly require this patch to remove.

The new test only asserts the enum. Mutation evidence confirms it does not cover
the production floor: deleting just the new `trust.assurance == PqHybrid` check
from `service.rs` leaves the entire affected package suite green (89/89).

Required: carry the selected atomic Ed25519↔ML-DSA-65 pair (or set of pairs) to
the credential-verification boundary and require a credential proof that actually
verifies under pinned Hybrid/composite policy. Derive/carry `PqHybrid` from that
performed verification, not merely from the preceding alias/capsule GATE.

### F2 — HIGH / BLOCKING: the mandatory overlap behavior is absent; `.first()` recreates the singleton anti-pattern on `subject_keys`

The deleted document check was:

```rust
ca_keys.len() == 1
```

Its replacement is:

```rust
verified.capsule().body.subject_keys.first()
```

followed by a scalar `ca_verifying_key`. Thus a capsule publishing `[old, new]`
does not make both CAs usable: only position zero can authenticate the registry
credential. There is no `kid`-based candidate selection, no try-each-compatible
verification, and no key-set type at the trust boundary. This is weaker than the
in-tree JWKS reference, which publishes drain/active/lead and selects candidates
by `kid`.

The path also reads only the hash-pinned genesis capsule returned by
`At9pAliasResolver`; it never consumes the at9p KERI/`AcceptedAt9pState` rotation
chain that the comments and documentation claim will rotate these keys. The
configured genesis CID is immutable, so a rotated CA published in accepted state
cannot reach this verifier. No expiry/drain bound is represented in
`DidAnchoredTrust`.

`rotation_with_overlapping_document_cas_still_bootstraps` is not the required
test. It publishes two keys in the `did:web` document, proves that both are
ignored, and asserts only the unchanged capsule key. It never:

1. publishes two capsule/accepted-state CA keys,
2. authenticates a credential under each while both are live, or
3. rejects the retired key after its bounded window.

The review brief makes that missing test independently blocking. Required: make
the deployment CA a named/current key set, select by the credential's `kid` (or
try each compatible atomic pair), source the live set from checkpointed accepted
state, and add the full two-live/old-retired acceptance test.

### F3 — HIGH / BLOCKING: the coordinated origin-substitution exploit is not tested at the authority-installation boundary

`substituted_document_ca_and_reach_are_ignored` calls only
`verify_did_anchored_document` and compares its returned CA/transport. It never
authenticates a registry credential, constructs `ProcessBootstrapAuthority`, or
attempts to install an attacker registry verifier. Its document “CA” fixtures
also use raw repeated-byte arrays such as `[0x07; 32]` and `[0x66; 32]`; under the
old vulnerable implementation these fail as malformed Edwards points before an
attacker authority can be installed, so the reverse-hunk failure is not a
faithful reproduction of the valid-key attack.

Static tracing supports that this PR's helper now ignores origin-selected key and
reach material, but #1192's CRITICAL was a coordinated document plus credential
attack against the production boundary. Clearing it requires a valid attacker
keypair, an attacker document, and a credential signed by that same attacker key,
fed through the real DID bootstrap/authority installation path and refused
because only capsule/accepted-state hybrid authority is usable.

This matters operationally because #1192 is not yet compatible with this patch:
its current capsule fixture has no `nodeId`, and its registry credential is
signed by a CA separate from the capsule subject key. Stacking this patch without
updating and rerunning that e2e path rejects the fixture before it demonstrates
the security property.

### F4 — MEDIUM: reach selection silently narrows the previously accepted transport contract

The old code accepted the document's preferred `Iroh` or `Quic` transport. The
new `reach_from_capsule` accepts only an exact `#ns` + `NinePExport` + `Iroh`
entry with a separate `nodeId`; its later `EndpointType::Quic` match is
unreachable. The capsule schema itself permits `Transport::Quic`, and
`ServiceEndpoint::new` defaults `node_id` to `None`. Consequently:

- every existing capsule/fixture that carries only `address`/`relay` now fails;
- a signed QUIC `NinePExport` that was previously a legitimate network reach now
  fails; and
- rollout requires republishing the capsule (therefore changing its CID/configured
  anchor) and reissuing the registry credential under the capsule CA.

Rejecting identity-derived iroh reach is correct for #1031, but option C's source
change did not itself authorize dropping legitimate signed QUIC reach. Either
support the capsule's typed QUIC carrier or document an explicit operator
decision and migration requirement. In all cases, update the dependent e2e
fixtures before claiming this unblocks #1192.

## What this change breaks or changes

- `did:web` CA and reach rotation no longer changes installed trust material.
  This is intentional option-C behavior.
- Existing deployment credentials signed by the former document CA fail unless
  that key equals the selected capsule key; credentials must be reissued.
- Capsules without exact `#ns`/`NinePExport`/iroh/`nodeId` reach fail bootstrap,
  including the current #1192 fixture.
- Capsule QUIC reach no longer works despite the prior Iroh-or-QUIC contract.
- A non-first or accepted-state-rotated capsule CA cannot authenticate bootstrap,
  so overlap/rotation currently causes rejection rather than drain/lead behavior.
- Missing or malformed document key/service fields no longer block bootstrap,
  because those fields are advisory only. The document `id` and reciprocal
  `alsoKnownAs` checks remain mandatory.

## Validation

All Cargo commands were run through `buildq`.

- `buildq -- cargo test -p hyprstream-discovery did_anchored --lib -- --nocapture`
  — PASS: 10 passed, 0 failed, 0 ignored.
- `buildq -- cargo test -p hyprstream-discovery --lib --tests --no-fail-fast`
  — PASS: 89 unit tests passed; `tests/podspec_k8s.rs` contained 0 tests.
- `buildq -- cargo clippy -p hyprstream-discovery --all-targets -- -D warnings`
  — PASS.
- `git diff --check 9081eef84dfafeab4ae89f652015d771c4fdd9b2..d943cc17b91bbf667556d97c65a8f38e897f4b9f`
  — PASS.

Mutation checks were performed in isolated temporary clones; the reviewed
worktree remained clean:

- Reintroduced the pre-option-C document CA/reach path while retaining the PR's
  tests. Result: 6 failed, 4 passed. Five new option-C behavior tests failed, as
  did the modified positive test; the newly added
  `f1_capsule_carries_the_material_the_document_should_not_own` evidence test
  still passed without the fix.
- Deleted only the new production `PqHybrid` assurance check. Result: the full
  package suite still passed 89/89, demonstrating that no test exercises that
  enforcement hunk.

Reviewed source remained at `d943cc17b91bbf667556d97c65a8f38e897f4b9f`
with a clean `git status`.

ewindisch:
# Verdict: PR #1209 — REVISE

**PR:** https://github.com/hyprstream/hyprstream/pull/1209 (fix/1157-capsule-sourced-trust → feat/1136-did-anchored-trust)
**Ticket:** https://github.com/hyprstream/hyprstream/issues/1157 (option C, Erica's ruling 2026-07-22) · parent https://github.com/hyprstream/hyprstream/issues/1183
**Reviewer:** fable (one of two independent reviewers). Read-only review; build + tests run through `buildq`.

**REVISE, not BLOCK: the code is correct and implements the operator-ruled shape exactly — every revise ask is a tracking/accuracy gap, not a code change.** The F1 substitution attack is genuinely neutralized and I verified it empirically, not by reading.

## What I verified (ran, not read)

- `buildq -- cargo test -p hyprstream-discovery --release` — **89/89 lib tests pass** on the PR head (plus 21 integration tests).
- **Revert-the-fix check performed.** I restored the pre-option-C document-sourcing body of `verify_did_anchored_document` (document `verificationMethod` CA + `preferred_transport` reach + `len()==1` rule), keeping the new tests, and re-ran: **exactly the 6 claimed tests fail** — `substituted_document_ca_and_reach_are_ignored`, `ca_is_bound_to_verified_capsule_not_document`, `installed_authority_is_pqhybrid_not_classical`, `rotation_with_overlapping_document_cas_still_bootstraps`, `reach_without_carrier_node_id_fails_closed`, `mutual_alias_accepts_at9p_as_authoritative` (4 pre-existing pass either way). Restored; tree clean; green rerun confirmed. The PR's "6 tests fail without the fix" claim is accurate.
- **Attacker-controlling-did:web analysis (the #1192 CRITICAL scenario).** With option C, an origin adversary (DNS + DV cert) can serve any document and any capsule bytes, and can install **nothing**: the document contributes only its `id` (must equal the configured `did:web`) and the `alsoKnownAs` string vouch (must name the configured `did:at9p`) — both operator-pinned values the attacker can only echo. Capsule bytes must hash to the configured cid512 and pass the GATE, so only the genuine capsule verifies; CA and reach both come from inside it (`did_anchored.rs:209-267,297-300`). Residual attacker power is denial of service only. Confirmed there is no fallback path: `DeploymentTrustSource` has no variant fallback and any failure is terminal.
- **CA = the key that verified the capsule, not merely position 0.** `at9p_sign::verify_capsule` verifies the composite signature against `body.subject_keys[0]` (`at9p_sign.rs:211-227`), and `sign_capsule` refuses to sign unless the signer IS `subject_keys[0]` (`:182-207`). So `ca_key_from_capsule`'s `.first()` selects the signature-bound GATE authority by construction — it is not the #1145 or arbitrary-singleton pattern in this consumer today (but see F1 below for how this composes with #1188).
- **`ca_keys.len() == 1` is DELETED, not repaired** — no cardinality rule anywhere in the new path; `rotation_with_overlapping_document_cas_still_bootstraps` proves a document publishing old+new CAs no longer bricks bootstrap fleet-wide (the pre-C failure mode).
- **PqHybrid propagated and fail-closed.** The GATE pins Hybrid unconditionally (`verify_record`: `require_pq = true` always, plus an explicit both-layers assertion — `at9p_sign.rs:159-178`), `resolve_authoritative_alias` returns `Assurance::PqHybrid` only after walking the capsule leg, `DidAnchoredTrust.assurance` carries it, and `authenticate_did_anchored_bootstrap` (`service.rs:2040-2048`) re-asserts `PqHybrid` as a fail-closed floor before the CA anchors the registry→audit→checkpoint chain. A classical trust root is impossible on this path. The consumer ensure is currently tautological (assurance is `PqHybrid` by construction on the only path), which is the right kind of tautology — it catches any future classical alias leg.
- **#1145 grep ("absent expected value disables the check") — clean.** Every absent value fails closed: empty `subject_keys` → error; missing/non-`NinePExport`/non-iroh `#ns` entry → error; absent `nodeId` carrier → error (`#1031` honored — reach never derived from the identity key); malformed key bytes → error; absent `content_length` still enforced by the streaming cap; partial anchor config → rejected, no silent downgrade. Service ids are validated unique (`validate_unique_services`), so the `#ns` `find` cannot be ambiguous.
- **Judged against the JWKS reference** (`services/oauth/jwks.rs:50-130`, `auth/key_rotation.rs`): this PR did not invent a weaker key-set consumer — it **removed the mutable key-set surface from this path entirely**. The trust root is now a content-addressed, immutable genesis capsule whose one verifying key is signature-selected. That is a legitimately different shape from drain/active/lead, and the right one for a hash-pinned anchor; the set/overlap semantics now live where they belong, in the at9p update chain (#1188 / PR #1210). Which leads to:

## Findings (severest first)

### F1 — MAJOR: the retirement half of the overlap acceptance bar does not exist at this layer, and its deferral is not concretely tracked
The acceptance bar was: two keys published, both accepted, retired one **rejected after its bounded window**. Post-C the "both accepted" half is structurally inapplicable (only the signature-bound genesis key can ever verify; document keys are ignored, not selected among), but the **retirement half is a real gap**: `resolve_did_anchored_trust` GATE-verifies the **genesis** capsule only and never consults the update/KERI chain or accepted state. A deployment that rotates its subject key via the update chain still has every bootstrapping node install the **retired genesis key as CA, forever**. There is no bounded window and no rejection. The PR discloses this honestly (F3 residual, "capsule currentness") but defers it to #1123/#1126 (which are **#atproto** rotation tickets, not at9p capsule currentness) and #1004 (**already CLOSED**). Net: the deferral points at tickets that don't own it.
**Ask:** file a concrete follow-up — "did-anchored bootstrap must consume accepted-state/update-chain currentness; a rotated-away genesis subject key must be rejected after a bounded window" — and note the composition constraint with #1188/PR #1210: when `subject_keys` gains working-set semantics, this consumer must select the authority key by the chain's rule, not `first()`. (The #1210 review lane independently flagged this composition.) No code change required in this PR.

### F2 — MEDIUM: F6 chokepoint bypass persists and the promised tracking ticket was not filed
The DID path still calls `dial` directly with a network config, sidestepping the `dial_local_transport_bootstrap` `Inproc|Ipc|SystemdFd` guard. The PR says "recommend a tracking ticket — out of scope here"; I searched and **no such ticket exists**. Disclosed-but-untracked is how known gaps get lost.
**Ask:** file it before merge.

### F3 — MEDIUM: F7 (liveness authenticates a locally-provisioned Discovery key, not the DID identity) — deferred without a named home
Narrowed but real, as the PR says: the signed `ping` is still verified against `global_trust_store().resolve_one("discovery")` — note `resolve_one` is itself the #1186 arbitrary-singleton — so the capsule-derived address is bound to locally-provisioned identity, leaving the #1136 §3 goal unmet. #1136 is open; **confirm this residual is written into #1136's body** (or file a child) rather than living only in this PR's description.

### F4 — MINOR: PR-body accuracy — the "former `#[ignore]`'d expected-fail target, now un-ignored" claim
`ca_is_bound_to_verified_capsule_not_document` never existed on the base branch (`git grep` at the base head finds nothing); the ignored evidence tests lived on the separate `f1-alias-binding-evidence` branch per #1157. The test in this PR is new. Cosmetic, but reviewers verify claims like this — say "reimplements the evidence-branch target" instead.

### F5 — INFO: what the change BREAKS (enumerated)
1. **New capsule-content contract**: bootstrap now hard-requires the capsule to publish a `#ns` `NinePExport` **iroh** service entry **with an independent `nodeId` carrier**. A capsule without one fails closed. There is no production capsule-minting tooling in-tree yet (all `sign_capsule` callers are test fixtures), so nothing existing breaks — but **#1192's stack provisioning must mint capsules to this contract** or the legitimate path is an outage at first boot. The docs update (`docs/deployment-registry-trust.md`) states the requirement; #1192 should link it.
2. **Reach is iroh-only** from the capsule (`reach_from_capsule` only emits `TransportConfig::iroh`); the pre-C document path nominally admitted QUIC. Pre-release feature, acceptable narrowing, but the `Iroh|Quic` ensure at `did_anchored.rs:301-307` is now half-dead code.
3. **`registry-service.jwt` provisioning contract moved**: `kid` must now be the RFC 7638 thumbprint of the **capsule's primary subject key**, not a document key. Docs updated correctly; any credential minted against a document key stops working (again: pre-release, nobody deployed).
4. API surface: `resolve_authoritative` is preserved as a wrapper over `resolve_authoritative_with_capsule`, so the other three alias consumers are untouched. `verification_method_ed25519_keys`/`preferred_transport` keep their other callers (admission, identity_resolver — #1187's scope). Workspace clippy green per CI.

## Bottom line
Option C is implemented exactly as ruled: CA from GATE-verified `subject_keys[0]` (signature-bound), reach from the `#ns` service's independent iroh carrier, document demoted to the leg-2 string vouch, `len()==1` deleted, overlap-document bootstrap tested, PqHybrid carried and floor-enforced, everything absent fails closed. The origin-substitution attack that produced #1192's CRITICAL is dead, verified by revert. **REVISE** solely for the untracked deferrals: file the capsule-currentness/retired-key ticket (F1) and the F6 chokepoint ticket (F2), anchor F7 in #1136, and correct the PR-body test-history claim. With those filed, this is a PASS.

ewindisch:
Addressed in 4d7e5af5c.

- F1: `DidAnchoredTrust` now carries atomic Ed25519 + ML-DSA-65 pairs, and DID bootstrap requires `compact-jwt~base64url(COSE composite)` verification under the selected pair. A plain EdDSA JWT cannot install authority; the regression test fails when the Hybrid-proof hunk is removed.
- F2: capsule `subject_keys` is consumed as a key set, selected by authenticated credential `kid`, not position. The overlap test authenticates both live pairs and rejects the retired pair after it is removed. #1218 tracks wiring that live set to checkpointed accepted-state/update-chain currentness.
- F3: added a real valid-attacker-key/document/credential test at the registry-authority installation boundary; it proves the attacker verifier is refused. The option-C source reversion makes that test fail.
- F4: signed `#ns` QUIC carriers are accepted again; iroh still requires an independent `nodeId` carrier.

Tracking requested by the other verdict is filed: #1218 (currentness/retirement), #1219 (dial chokepoint), and #1220 (DID-bound liveness, linked from #1136). The earlier PR-body wording about the evidence target was inaccurate: this PR reimplements that evidence-branch target; it was not previously an ignored test on this base branch.

CodeRabbit has no actionable review thread here: it posted only the auto-review-skipped status, so there is nothing to resolve.

ewindisch:
VERDICT: BLOCK

# Re-review — PR #1209

Reviewed current head: `4d7e5af5c2483dc83d6c870ac27c9c9b2ea662ee`

The fix round genuinely closes the earlier classical-floor, coordinated
did:web-origin substitution, and QUIC-regression findings. It does not close the
load-bearing rotation finding: production still installs the immutable genesis
capsule's key set forever, while the new retirement test swaps hand-built vectors
and has neither accepted-state currentness nor a bounded window. The fix also
identifies a Hybrid pair by only its Ed25519 component, unlike the in-tree
composite-key reference, so overlap fails when only the PQ component rotates.

## Findings

### F1 — HIGH / BLOCKING: the production path still cannot retire a capsule CA; the acceptance test simulates a state transition that no production code performs

`resolve_did_anchored_trust` fetches and GATE-verifies only the capsule named by
the configured immutable genesis CID, then copies that capsule's complete
`body.subject_keys` into `DidAnchoredTrust` (`did_anchored.rs:316-373`).
`authenticate_did_anchored_bootstrap` consumes that vector directly
(`service.rs:2115-2133`). It never reads `AcceptedAt9pState`, the verified update
chain, an overlap deadline, or a drain-expiry condition.

The new
`did_anchored_hybrid_ca_overlap_accepts_both_then_rejects_retired_key` test
authenticates against `[old, new]` and then directly calls the private helper
against `[new]` (`service.rs:2760-2789`). No production transition produces the
second vector, and no time/window is exercised. Thus both live keys work within a
static set, but an old key removed by accepted current state remains accepted by
real bootstrap indefinitely because bootstrap keeps rereading genesis.

This is not merely a follow-up enhancement: it is the explicit acceptance bar in
the review brief and the earlier blocking F2. Newly filed issue #1218 confirms
the missing wiring in its own title/body: bootstrap still must consume
checkpointed accepted state so a rotated-away key is rejected after its bounded
window. Filing that issue does not make this PR's advertised rotation behavior
work.

Required: source the live CA set from authenticated accepted current state (with
the genesis capsule only as its seed), carry the bounded overlap/drain rule, and
make the acceptance test drive the production state transition: old and new both
authenticate while live, then old fails after currentness advances past its
bounded window.

### F2 — HIGH / BLOCKING: the new key-set selector is not an atomic Hybrid-pair selector

The credential `kid` and deployment domain are computed solely from the
Ed25519 key (`service.rs:1731-1735,1887-1905`). To make that scalar identifier
unambiguous, `ca_keys_from_capsule` rejects any two pairs sharing the same
Ed25519 key (`did_anchored.rs:239-245`), even when their ML-DSA-65 keys differ.
Therefore the ordinary rotation `[Ed + old-PQ, Ed + new-PQ]` cannot overlap at
all: bootstrap rejects the entire published set before either credential is
checked.

That is weaker than the required in-tree reference. The working composite path
computes `composite_kid(ml_dsa, ed25519)` over both components
(`hyprstream-rpc/src/auth/jwt.rs:214-225`) and publishes exactly accepted atomic
pairs (`services/oauth/jwks.rs:123-130`). The PR's overlap test uses two different
Ed25519 keys, so it does not cover component-only PQ rotation and would remain
green if this outage persisted.

Required: give each exact Ed25519 + ML-DSA-65 pair a composite identifier derived
from both components, use it for authenticated selection, and test overlapping
pairs that share one component. Do not reject a valid published pair set merely
because one component is intentionally retained across rotation.

## Prior findings re-checked

- **Classical floor: closed.** `HybridDeploymentCa` carries both key halves;
  DID credentials require exactly one non-empty `~` composite proof;
  `verify_composite(..., Some(pq), ..., require_pq = true)` and the explicit
  `eddsa && ml_dsa` check both succeed before authority is minted. A plain EdDSA
  JWT is rejected.
- **Origin substitution: closed.** CA pairs and reach come only from the
  hash-pinned, GATE-verified capsule. The valid attacker document/key/credential
  test reaches the registry-authority authentication boundary and is rejected.
  Reintroducing document CA selection together with the former classical
  credential verification makes the test fail by installing the attacker
  registry verifier.
- **Reach regression: closed.** Signed capsule QUIC carriers are accepted again;
  iroh still requires an independent `nodeId`.
- **`ca_keys.len() == 1`: deleted.** Capsule subject keys are decoded as a set and
  credential selection uses a mandatory authenticated `kid`; missing/empty
  values fail closed. No #1145-style absent-value bypass was found.

## What the change breaks or changes

1. DID-anchored `registry-service.jwt` is no longer a JWT despite the filename:
   it must be `compact-jwt~base64url(COSE-composite)`. Every existing plain JWT
   on that path now fails startup and must be reprovisioned; the repository has
   test minting and documentation but no production minting tool.
2. A capsule publishing two Hybrid pairs with the same Ed25519 component is now
   rejected, so PQ-only overlap rotation is an outage (F2).
3. Bootstrap still trusts every CA in the immutable genesis capsule indefinitely;
   accepted-state retirement cannot take effect (F1).
4. did:web `verificationMethod` and `service` changes no longer alter installed
   trust or reach. This is the intended option-C behavior.
5. Iroh capsules without an independent `nodeId` fail closed. Capsule QUIC reach
   must be a signed `quic://IP:port` carrier and use the IP as TLS server name;
   DNS-form carriers are rejected.
6. All non-primary capsule subject pairs are now decoded and validated. A
   malformed secondary pair that older primary-only consumption ignored now
   rejects bootstrap, which is appropriate fail-closed behavior.

## Validation

All Cargo commands ran through `buildq`; no source changes were made in the
review worktree.

- `buildq -- cargo test -p hyprstream-discovery --lib --tests --no-fail-fast`
  — PASS: 94 unit tests, 0 failures; integration target had 0 tests.
- `buildq -- cargo build -p hyprstream-discovery` — PASS.
- `buildq -- cargo clippy -p hyprstream-discovery --all-targets -- -D warnings`
  — PASS.
- `git diff --check origin/feat/1136-did-anchored-trust...HEAD` — PASS.
- Reviewed worktree remained clean at the pinned SHA.

Security-critical mutation checks were run in an isolated temporary clone and
restored after each run:

- restore positional first-key selection → overlap test FAILS on the new CA;
- remove the mandatory Hybrid proof → classical-only rejection test FAILS;
- restore document CA selection plus the former classical verification boundary
  → origin-substitution test FAILS by installing attacker authority;
- truncate capsule CA decoding to one key and remove QUIC support → the two new
  capsule-set/QUIC tests both FAIL.

The isolated clone was clean after restoration and was moved to trash.

ewindisch:
Addressed the re-review findings in a6740bec1e005c380be4e5f372db610a2084f770.

F1: DID bootstrap now treats the immutable GATE capsule as a seed, verifies the daemon checkpointed accepted state with the authenticated registry identity, enforces current-state freshness, and projects both the live CA set and Discovery reach from `AcceptedAt9pState.current`. The production reader regression constructs a GATE-seeded, accepted successor that retires the old pair and verifies the concrete checkpointed-PDS reader returns only the new state; `accepted_current_state_retires_genesis_ca_and_updates_reach` verifies the corresponding trust projection. Reverting that live-state projection made the latter fail (old genesis key remained).

F2: each Hybrid pair now uses `composite_kid(ml_dsa, ed25519)` for `kid`/domain selection and duplicate detection. This permits a bounded overlap that retains Ed25519 while rotating ML-DSA-65, selects the complete matching pair, and rejects the retired pair. `did_anchored_hybrid_ca_overlap_allows_pq_only_rotation` covers exactly that case. Reverting selection to the Ed25519-only domain made it fail with `credential kid does not name a current DID-anchored CA`.

Validation: `buildq -- cargo test -p hyprstream-discovery --lib` (97 passed); `buildq -- cargo check -p hyprstream-discovery`; `buildq -- cargo clippy -p hyprstream-discovery --all-targets -- -D warnings`; targeted mutation checks above. The foreground commit hook also completed its workspace release Clippy gate.

CodeRabbit has no actionable review thread on this PR (only its auto-review-skipped status), so there was nothing to resolve. Not merged.

ewindisch:
VERDICT: BLOCK

# Review — PR #1209

Reviewed head: `a6740bec1e005c380be4e5f372db610a2084f770`

## F1 — HIGH / BLOCKING: post-retirement DID bootstrap is an unavoidable credential-ordering outage

The new accepted-state projection is correct in isolation, but the production
bootstrap cannot use it after a CA retires.  In
`authenticate_did_anchored_bootstrap`, the only registry credential is first
authenticated against the immutable genesis seed (`service.rs:2182-2188`) in
order to obtain the registry verifier needed to open the checkpoint reader.
Only after that does it load `AcceptedAt9pState.current` and authenticate the
same credential against the current CA set (`service.rs:2203-2217`).

After a complete rotation from `old` to `new`, no legitimate credential can
satisfy both steps:

- a credential signed by `new` fails the seed/genesis check before the current
  state is reachable;
- a credential signed by `old` can open the checkpoint but then fails the
  current-set check after `old` is retired.

Thus a restart after the accepted state retires `old` denies the legitimate
deployment.  This is the prior blocking retirement finding in a different
form, and it is an outage rather than a safe fail-closed boundary.

The added tests do not exercise that production transition.
`checkpointed_current_state_retires_a_ca_on_the_production_read_path` supplies
the checkpoint reader with a hand-selected registry verifier; it never calls
the two-stage bootstrap.  `accepted_current_state_retires_genesis_ca_and_updates_reach`
only calls `DidAnchoredTrust::from_accepted_state` directly.  Neither test can
show that a post-retirement credential reaches the current-state reader and
then authenticates.

Required: provide a bootstrap trust chain that can authenticate the accepted
state without first requiring the same credential to verify under a retired
genesis CA (or carry the required bounded overlap proof), and add a production
causal test: start with the old credential/state, advance through overlap,
retire old, restart with a new-only credential, and prove bootstrap succeeds;
prove the old credential then fails.

## Prior findings rechecked

- **Hybrid floor: closed.** `HybridDeploymentCa` carries the complete atomic
  Ed25519 + ML-DSA-65 pair, `composite_kid` names the pair, and DID credentials
  require a verified composite proof.  A plain EdDSA JWT cannot mint the DID
  authority.
- **Atomic PQ-only rotation selector: closed.** Pair identity includes both
  components and the dedicated PQ-only-overlap test selects the exact pair.
- **did:web origin substitution: closed.** CA pairs and reach are read from the
  GATE-verified capsule/accepted state; the document-origin attacker test reaches
  the registry-authority boundary and is refused.
- **QUIC reach regression: closed.** Signed capsule QUIC carriers are accepted;
  iroh still requires an independent `nodeId`.
- **`ca_keys.len() == 1`: deleted.** There is no replacement singleton rule;
  empty/malformed key material and absent required reach fail closed.

## What this changes or breaks

1. **Blocking outage:** a deployment whose accepted state has retired the
   genesis CA cannot restart with either its old or its new sole credential, as
   described in F1.
2. DID-anchored registry credentials now require
   `compact-jwt~base64url(COSE-composite)`; prior plain JWT provisioning fails
   closed.
3. `did:web` verification methods and services no longer alter installed CA or
   reach.  This is the intended option-C behavior.
4. Iroh capsule reach requires an explicit independent `nodeId`; malformed or
   absent carriers fail closed.  Signed QUIC carriers remain supported.

## Validation

All Cargo commands ran through `buildq` on the reviewed SHA:

- `buildq -- cargo test -p hyprstream-discovery did_anchored --lib -- --nocapture` — PASS.
- `buildq -- cargo test -p hyprstream-discovery --lib --tests --no-fail-fast` — PASS.
- `buildq -- cargo clippy -p hyprstream-discovery --all-targets -- -D warnings` — PASS.
- `git diff --check 9081eef84dfafeab4ae89f652015d771c4fdd9b2..HEAD` — PASS.

Security-critical revert checks were made in isolated temporary worktrees and
fully restored; the reviewed worktree stayed clean:

- Replaced `from_accepted_state`'s live CA/reach projection with `Ok(seed)`.
  `accepted_current_state_retires_genesis_ca_and_updates_reach` failed, proving
  the test detects resurrection of the genesis authority.
- Replaced the DID `PqHybrid` credential branch with the former scalar Ed25519
  profile.  `did_anchored_authority_rejects_a_classical_only_credential` failed
  by accepting the EdDSA-only JWT, proving the composite-proof enforcement test
  is effective.

The source change is substantial and closes several real security defects, but
F1 keeps the legitimate rotated deployment from booting.  Do not merge.

ewindisch:
Fixes the blocking post-retirement bootstrap cycle in 659f0f091.

- F1: the registry credential is now authenticated only after replaying the append-only accepted update history from the immutable GATE genesis. Bootstrap then verifies the checkpointed state with that current credential and requires the replayed and checkpointed states to match. A retired genesis CA therefore cannot strand a restart, while a stale/forked history or checkpoint fails closed.
- Production writer: every accepted successor is atomically retained as a replayable history record alongside the state envelope and checkpoint; genesis remains the immutable GATE anchor rather than mutable history.
- #1210 composition: this path consumes the complete current `subject_keys` set through `DidAnchoredTrust` and does not select an Ed25519/PQ half independently. It must land with #1210’s set-semantic successor/persisted-record changes so non-first published hybrid pairs remain valid end-to-end.

Validation: `buildq -- cargo test -p hyprstream-discovery --lib --tests --no-fail-fast`; focused production writer test; strict discovery clippy; commit hook release clippy. Mutation checks: removing the persisted-history write fails the production writer test; removing history replay fails the post-retirement replay test.

No open review or CodeRabbit threads were present on the PR.

PR #1210, opened 2026-07-23
fix(at9p): give subject_keys set semantics instead of primary-only (#1188)

### LINKED ISSUE #1188: design(at9p): give subject_keys set semantics instead of primary-only semantics

ewindisch:
## Summary

Gives the at9p capsule `subject_keys` field **set semantics** instead of
primary-only (position-0) semantics. Fixes #1188; part of the #1183 audit
family ("exactly one published key" is an anti-pattern — it forecloses overlap
rotation AND PQ-hybrid rollout).

The schema published `subject_keys: Vec<HybridKeyPair>`, but every consumer
selected `subject_keys.first()` and treated index 0 as the sole authority. That
made rotation-with-overlap and hybrid rollout structurally impossible: a capsule
carrying an old + new subject pair had no working overlap semantics, reordering
changed authority, and a signature by the later-published pair was rejected.

Per #1183's durable rule and the working in-tree JWKS reference (drain/active/
lead by `kid`, caching every usable candidate), published verification material
is a **named set**: a consumer selects a member by its atomic Ed25519↔ML-DSA-65
identity and skips what it does not match, never by position.

## Changes

- **schema** (`at9p.rs`): reject an ambiguous set — a duplicate hybrid pair or a
  repeated Ed25519 half across pairs (application-signer selection keys on the
  Ed25519 half, so a repeat would be ambiguous). Add `subject_key_for_ed25519`.
- **`VerifiedAt9pKeys`** (`identity_resolver.rs`): carry the whole verified set;
  select the atomic pair by Ed25519 via `for_ed25519`; preserve the GATE-mint
  provenance invariant. `new_gate_verified_set` fails closed on empty/ambiguous.
- **GATE bridge** (`at9p_resolver.rs`): project **all** subject keys.
- **admission** (`admission.rs`): the app signer must match **some** published
  key; bind that matched key's atomic ML-DSA-65 half (never a positional one).
- **successor-check** (`at9p_chain.rs`): authorize the rotation signer by trying
  each **pre-committed** revealed candidate and accepting the one that verifies,
  never position 0. An uncommitted revealed key never gains authority.
- **genesis sign/verify** (`at9p_sign.rs`): a self-certifying capsule may be
  signed by any published subject key; verify tries each.
- **persisted-update re-verify** (`at9p_duplicity.rs`), **native announcements**
  (`service/factory.rs`, `services/factories.rs`), **service selection**
  (`discovery/service.rs`): match the local/selected signer against **any**
  published key.

The scalar `IdentityKeys` resolve arm still projects only single-key capsules
and fails closed on a multi-key set (no selector at resolve time); set-aware
identity resolution is tracked separately in #1187.

## Tests

Every fix ships a test that fails without it (verified by reverting the hunk,
watching it fail, and restoring). Includes the required **overlap tests** — two
keys published simultaneously, both accepted, and the retired one rejected after
its bounded window:

- **successor validation**: overlap successor signed by the non-first
  pre-committed key is accepted; uncommitted set members still fail closed;
  retired rotation rejected after its bounded `expires_at` window.
- **capsule GATE / self-cert**: capsule signed by the second published key
  verifies; a key it does not publish is rejected; duplicate Ed25519 rejected.
- **admission**: admits a peer presenting the second published key and binds
  that key's atomic pair (nothing under the first); non-member signer rejected.
- **native announcements**: admits a service holding the non-first published
  key; non-member rejected.

## Verification

- `cargo build` (workspace, with libtorch) — clean.
- `cargo test -p hyprstream-pds -p hyprstream-rpc -p hyprstream-service` lib
  tests — all pass (pds 207, rpc 772).
- `cargo clippy -p hyprstream-pds -p hyprstream-rpc -p hyprstream-service
  -p hyprstream-discovery -p hyprstream --all-targets -- -D warnings` — clean.

Note: `cargo fmt --check` fails on a clean `main` (#1140) and touches these
files with pre-existing drift; new code is written to match surrounding style
rather than introduce full-file rustfmt churn.

Relates #1183. Do not merge without review.

### DIFF
--- a/crates/hyprstream-discovery/src/service.rs
+++ b/crates/hyprstream-discovery/src/service.rs
@@ -638,16 +638,17 @@ impl ResolvedService {
         let current_key = state
             .current
             .subject_keys
-            .first()
-            .ok_or_else(|| anyhow::anyhow!("accepted state has no current response key"))?;
-        anyhow::ensure!(
-            selected.response_verifying_key().to_bytes().as_slice()
-                == current_key.ed25519_pub.as_slice(),
-            "selected response key is not accepted current key"
-        );
+            .iter()
+            .find(|key| {
+                key.ed25519_pub.as_slice()
+                    == selected.response_verifying_key().to_bytes().as_slice()
+            })
+            .ok_or_else(|| {
+                anyhow::anyhow!("selected response key is not one of the accepted current keys")
+            })?;
         anyhow::ensure!(
             selected.response_ml_dsa65() == current_key.mldsa65_pub.as_slice(),
-            "selected PQ response key is not accepted current key"
+            "selected PQ response key is not the atomically-bound accepted current key"
         );
         anyhow::ensure!(
             state
--- a/crates/hyprstream-pds/src/at9p.rs
+++ b/crates/hyprstream-pds/src/at9p.rs
@@ -436,6 +436,28 @@ impl CapsuleBody {
         Ok(body)
     }
 
+    /// Select the published subject key whose Ed25519 half equals `ed25519_pub`,
+    /// if any — the **set-semantics** accessor (#1188 / #1183).
+    ///
+    /// `subjectKeys` is a set of currently-usable identity keys, not an ordered
+    /// list with a positional "primary". A consumer that holds an Ed25519
+    /// application-signer key (admission #894, native announcements #1004) finds
+    /// *its* hybrid pair by identity, and takes the atomically-bound ML-DSA-65
+    /// half from the same entry. The schema guarantees at most one match
+    /// ([`validate_unique_subject_keys`] forbids a repeated Ed25519 half), so
+    /// this is unambiguous. Returns `None` when no published key matches — the
+    /// caller then rejects fail-closed, never falling back to a positional key.
+    pub fn subject_key_for_ed25519(&self, ed25519_pub: &[u8]) -> Option<&HybridKeyPair> {
+        self.subject_keys
+            .iter()
+            .find(|key| key.ed25519_pub.as_slice() == ed25519_pub)
+    }
+
+    /// Whether `ed25519_pub` is one of the published subject keys.
+    pub fn publishes_ed25519(&self, ed25519_pub: &[u8]) -> bool {
+        self.subject_key_for_ed25519(ed25519_pub).is_some()
+    }
+
     /// Canonical DAG-CBOR bytes of the capsule body.
     ///
     /// This is the signed payload for an at9p capsule: the composite signature
@@ -461,6 +483,19 @@ impl CapsuleBody {
         for key in &self.subject_keys {
             key.validate()?;
         }
+        // subjectKeys is a *set*, not an ordered list with a positional
+        // "primary" (#1188 / #1183): consumers select a member by its atomic
+        // hybrid identity (Ed25519↔ML-DSA-65 pairing), never by index. Two
+        // identical pairs would be a meaningless repeat, and — worse — an
+        // ambiguous key-selection input. Reject duplicates by their
+        // commitment digest (the same BLAKE3-512 binding both halves that
+        // `next_key_commitments` uses), so overlap/rotation publishes *distinct*
+        // usable keys. A member sharing one half with another but differing on
+        // the other half is a distinct pair and is allowed (each is atomically
+        // bound), but an Ed25519 half MUST NOT repeat across pairs: selection by
+        // application signer key (admission, native announcements) keys on the
+        // Ed25519 half, so a repeated Ed25519 would reintroduce ambiguity.
+        validate_unique_subject_keys(&self.subject_keys)?;
         // next_key_commitments MAY be empty (#879 §5.3: a declared-immutable
         // identity with rotation frozen forever), but two identical commitments
         // are meaningless (the same next key pre-committed twice) and would
@@ -1292,6 +1327,37 @@ fn validate_unique_digests(digests: &[[u8; H512_LEN]], field: &str) -> Result<()
     let mut seen = BTreeSet::new();
     for digest in digests {
         ensure!(seen.insert(digest.as_slice()), "duplicate {field} entry");
+    }
+    Ok(())
+}
+
+/// Reject an ambiguous `subjectKeys` *set* (#1188 / #1183).
+///
+/// `subjectKeys` publishes every currently-usable identity key so overlap
+/// rotation and PQ-hybrid rollout work by publishing the new key alongside the
+/// old and letting each verifier select the member it understands (the JWKS
+/// `kid` discipline, applied to hybrid pairs). For that selection to be
+/// unambiguous:
+///
+/// - no two entries may be the *same* atomic pair (identical commitment
+///   digest) — a meaningless repeat; and
+/// - no Ed25519 half may repeat across entries — application-signer selection
+///   (admission #894, native announcements #1004) keys on the Ed25519 half, so
+///   a repeated Ed25519 would make "which hybrid pair authorizes this signer?"
+///   ambiguous, exactly the positional-authority failure #1188 removes.
+fn validate_unique_subject_keys(keys: &[HybridKeyPair]) -> Result<()> {
+    let mut seen_pairs = BTreeSet::new();
+    let mut seen_ed25519 = BTreeSet::new();
+    for key in keys {
+        ensure!(
+            seen_pairs.insert(key.commitment_digest()),
+            "duplicate subjectKeys entry (identical hybrid pair)"
+        );
+        ensure!(
+            seen_ed25519.insert(key.ed25519_pub.clone()),
+            "duplicate subjectKeys Ed25519 half across hybrid pairs — \
+             ambiguous key selection"
+        );
     }
     Ok(())
 }
--- a/crates/hyprstream-pds/src/at9p_chain.rs
+++ b/crates/hyprstream-pds/src/at9p_chain.rs
@@ -20,12 +20,14 @@
 //!
 //! # The four gates ([`validate_successor`])
 //!
-//! 1. **Signer authorization** — the update-record's signing key
-//!    (`new_capsule_body.subject_keys[0]`, revealed by the rotation) must have
-//!    its [`HybridKeyPair::commitment_digest`] present in the predecessor's
-//!    `next_key_commitments`. This is the key-selection decision B1 owns: it
-//!    picks *which* key [`verify_update_record`] must verify against, rather than
-//!    trusting a caller-supplied key.
+//! 1. **Signer authorization** — the update-record's signing key (one of the
+//!    revealed `new_capsule_body.subject_keys`, which is a published SET, not a
+//!    positional singleton — #1188 / #1183) must have its
+//!    [`HybridKeyPair::commitment_digest`] present in the predecessor's
+//!    `next_key_commitments`, and must actually have signed the record. This is
+//!    the key-selection decision B1 owns: it tries each pre-committed revealed
+//!    candidate and accepts the one that verifies, rather than trusting position
+//!    0 or a caller-supplied key.
 //! 2. **Linkage** — `prev_record_digest` equals the predecessor's record digest
 //!    (`H512` of its canonical bytes), chaining the segment back toward genesis.
 //! 3. **Monotonic epoch** — strictly increasing epoch.
@@ -247,32 +249,59 @@ pub fn validate_successor(
         return Err(SuccessorError::BrokenLinkage);
     }
 
-    // Gate 1 (signer authorization): the rotation reveals its now-current keys in
-    // new_capsule_body.subject_keys; the primary key is the authorizing signer.
-    // B1 selects THIS key for verification rather than trusting any caller input.
-    let signer: &HybridKeyPair =
-        candidate
-            .new_capsule_body
-            .subject_keys
-            .first()
-            .ok_or_else(|| {
-                SuccessorError::Malformed(anyhow::anyhow!(
-                    "update capsule body has no subject keys"
-                ))
-            })?;
-    let signer_digest = signer.commitment_digest();
-    if !predecessor
-        .next_key_commitments
+    // Gate 1 (signer authorization + signature): the rotation reveals its
+    // now-current keys in `new_capsule_body.subject_keys` — a published SET, not
+    // an ordered list with a positional "primary" (#1188 / #1183). The
+    // authorizing signer is whichever revealed key was BOTH pre-committed by the
+    // predecessor (its `commitment_digest` present in `next_key_commitments`) AND
+    // actually signed this record. B1 selects that key by trying each pre-rotation
+    // candidate rather than trusting position 0 or any caller input.
+    //
+    // This is the same discipline the durable rule (#1183) prescribes for a
+    // published key set with no wire-level selector: try each compatible
+    // candidate, accept the one that verifies, never let position confer
+    // authority. A predecessor may pre-commit to several next keys (hybrid /
+    // overlap), and the successor set may publish several; exactly one of them
+    // signs, and that one must have been pre-committed.
+    let precommitted: Vec<&HybridKeyPair> = candidate
+        .new_capsule_body
+        .subject_keys
         .iter()
-        .any(|commitment| commitment == &signer_digest)
-    {
+        .filter(|key| {
+            let digest = key.commitment_digest();
+            predecessor
+                .next_key_commitments
+                .iter()
+                .any(|commitment| commitment == &digest)
+        })
+        .collect();
+    if precommitted.is_empty() {
+        // No revealed subject key was pre-rotated: none is authorized to rotate
+        // this identity, regardless of whether the signature is otherwise valid.
         return Err(SuccessorError::SignerNotCommitted);
     }
 
-    // Gate 1 (signature): the pre-committed key must actually have signed the
+    // Gate 1 (signature): a pre-committed key must actually have signed the
     // record. Pinned-Hybrid (EdDSA + ML-DSA-65) verification (#939); done last as
-    // the most expensive check.
-    verify_update_record(candidate, signer).map_err(SuccessorError::SignatureInvalid)?;
+    // the most expensive check. Try each authorized candidate and accept the
+    // first whose composite verifies — set-selection by "which published key
+    // signed", not by index. If none verifies, the record was not signed by any
+    // authorized (pre-committed) key.
+    let mut last_err = None;
+    let signed = precommitted
+        .iter()
+        .any(|signer| match verify_update_record(candidate, signer) {
+            Ok(()) => true,
+            Err(e) => {
+                last_err = Some(e);
+                false
+            }
+        });
+    if !signed {
+        return Err(SuccessorError::SignatureInvalid(last_err.unwrap_or_else(
+            || anyhow::anyhow!("no pre-committed subject key verified the update record"),
+        )));
+    }
 
     // Gate 4 (freshness): now < expires_at.
     if !datetime_before(now, &candidate.expires_at).map_err(SuccessorError::Malformed)? {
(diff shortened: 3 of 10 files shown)


ewindisch:
@coderabbitai review

ewindisch:
@coderabbitai full review

ewindisch:
# Verdict: BLOCK

Reviewed PR #1210 at `58021440d932c0d6b17f496a4a4b7287cc41559a`.

## Findings

1. **BLOCK — the live admission path makes every genesis-published overlap key immortal, so the required bounded retirement does not exist.**

   `At9pGateResolver::verify_bytes` verifies the immutable genesis capsule named by the `did:at9p` CID and projects every `subject_key` (`crates/hyprstream-pds/src/at9p_resolver.rs:47-77`). `admit_key_against_did` then accepts any matching projected key without consulting the accepted-current-state chain, an update epoch, or an expiry (`crates/hyprstream-rpc/src/admission.rs:296-324`). Once two keys are published in that genesis capsule, the old one remains admissible forever.

   The alleged retirement test does not test key retirement. `overlap_retired_after_bounded_window_rejected` re-validates the same update after that **entire update record** has expired (`crates/hyprstream-pds/src/at9p_chain.rs:836-866`). It never proves “old rejected, replacement still accepted”; after the tested expiry neither key/state is accepted. The positive overlap test also exercises only the second signer, not both keys followed by selective retirement (`at9p_chain.rs:750-779`).

   The acceptance bar explicitly requires two simultaneously published keys, both accepted, followed by rejection of the retired key after its bounded window while the replacement remains usable. That end-to-end lifecycle is absent. This is materially weaker than the OAuth/JWKS drain/active/lead reference, which has currentness and bounded drain state.

2. **BLOCK — the change deliberately turns legitimate overlap into an identity-resolution outage.**

   `resolve_did_at9p` now rejects every multi-key capsule with `keys.len() != 1` (`crates/hyprstream-rpc/src/identity_resolver.rs:338-367`). This is not harmless ambiguity handling: `AtprotoPerimeterGateway::enroll` calls `resolve_identity_keys` and treats the error as failed enrollment (`crates/hyprstream-rpc/src/auth/atproto_perimeter.rs:116-128`). A peer may pass the newly set-aware admission path and then be denied at enrollment solely because it is in the required old/new overlap window.

   Issue #1188 explicitly required a two-pair identity-resolution test. Deferring set-aware resolution to #1187 leaves #1188 incomplete. A fail-closed conversion that denies the legitimate rotation path is an outage, not working set semantics.

3. **BLOCK — PR #1210 does not compose with option C in PR #1209; option C remains positional.**

   At PR #1209 head `d943cc17b91bbf667556d97c65a8f38e897f4b9f`, `ca_key_from_capsule` explicitly installs the “primary” `subject_keys[0]` using `.first()` (`crates/hyprstream-discovery/src/did_anchored.rs:199-221`). Merging #1210 does not change that consumer or give it a non-positional selection rule. Reordering the same valid set changes the deployment CA, and during overlap only the first key can validate the registry credential. Option C therefore inherits the cardinality bug it was intended to remove.

4. **HIGH — unbounded sets now cause unbounded composite verification work on peer-presented bytes.**

   `CapsuleBody::validate` requires only a non-empty `subject_keys` vector; it imposes no count bound (`crates/hyprstream-pds/src/at9p.rs:473-498`). `verify_capsule` tries a full Ed25519 + ML-DSA verification for every member until one succeeds (`crates/hyprstream-pds/src/at9p_sign.rs:218-243`), and the admission trait accepts an unconstrained peer-provided byte slice. A valid capsule signed by its last member—or by none—turns one admission attempt into attacker-selected `O(n)` expensive PQ verifications. Add a protocol cardinality/byte bound or a signed key selector before widening verification this way.

5. **REVISE — the stated mutation-test claim is false.**

   I temporarily restored the pre-PR primary-only signing logic and removed the new uniqueness validation, ran the added tests through `buildq`, then restored the tree. `capsule_signed_by_non_first_subject_key_verifies` failed as expected, and `duplicate_ed25519_subject_key_rejected` failed as expected. However, the newly added `capsule_signed_by_unpublished_key_rejected` remained green under the reverted implementation. Thus “every new test fails without its fix” is not true. Several changed production consumers also receive no direct added regression test.

## What this change breaks

- A multi-key `did:at9p` identity cannot enroll through the scalar identity resolver during its legitimate overlap window.
- A retired key from a peer-presented genesis capsule remains admissible indefinitely.
- PR #1209 deployment bootstrap still changes CA authority when `subject_keys` order changes.
- Worst-case GATE/signature verification cost scales with an attacker-controlled, unbounded key count.

## Verification

- `buildq -- cargo test -p hyprstream-pds -p hyprstream-rpc -p hyprstream-service --lib`
  - `hyprstream-pds`: 207 passed, 0 failed, 0 ignored.
  - `hyprstream-rpc`: 772 passed, 0 failed, 3 ignored.
  - `hyprstream-service`: 29 passed, 0 failed, 1 ignored.
- Mutation:
  - `buildq -- cargo test -p hyprstream-pds at9p_sign::tests::capsule_signed --lib -- --nocapture`
    - non-first positive failed; unpublished-key rejection still passed.
  - `buildq -- cargo test -p hyprstream-pds at9p_sign::tests::duplicate_ed25519_subject_key_rejected --lib -- --exact --nocapture`
    - failed under the reverted validation hunk.
- GitHub Clippy, CodeQL, WASM, cargo-deny, and analysis checks are green. Green CI does not cover the blockers above.
- The worktree was restored to the reviewed SHA and is clean (`git diff --exit-code`, `git diff --check`).

ewindisch:
# Verdict — PR #1210 (at9p `subject_keys` set semantics, #1188)

**Reviewer:** fable (one of two independent reviewers) · 2026-07-23 · worktree `.worktrees/ks1210b` at PR head `58021440d`
**PR:** https://github.com/hyprstream/hyprstream/pull/1210 · **Ticket:** https://github.com/hyprstream/hyprstream/issues/1188 · **Parent audit:** https://github.com/hyprstream/hyprstream/issues/1183

## VERDICT: REVISE

The core change is correct, matches the working JWKS reference discipline (select by identity, try each candidate when no selector exists, atomic hybrid binding, fail-closed everywhere), and meets the overlap acceptance bar with tests I independently confirmed fail against the pre-fix behavior. Two things keep it out of PASS: it does **not compose** with PR #1209 as written (the tasked cross-check), and four of the ten production hunks ship with **no test at all**, contradicting the PR description's "every fix ships a test that fails without it."

## Verification performed

- Baseline at PR head: `hyprstream-pds` 207 ✓, `hyprstream-rpc` 772 ✓, `hyprstream-discovery` 79 ✓, `hyprstream-service` 29 ✓, and the torch-bound `hyprstream` crate lib 1099 ✓ (the crate whose build job is merge_group-only, so this is not covered by PR CI). All via `buildq`.
- Revert-check: I temporarily restored the pre-fix positional behavior in all five behavior-carrying sites (`at9p.rs` uniqueness, `at9p_chain.rs` Gate 1, `at9p_sign.rs` sign+verify, `admission.rs` match, `factory.rs` membership) and re-ran. All six positive-path new tests failed exactly as claimed:
  - `overlap_successor_signed_by_non_first_committed_key_accepted` FAILED
  - `overlap_retired_after_bounded_window_rejected` FAILED
  - `capsule_signed_by_non_first_subject_key_verifies` FAILED
  - `duplicate_ed25519_subject_key_rejected` FAILED
  - `admission_admits_second_published_subject_key_and_binds_its_pair` FAILED
  - `native_announcement_admits_non_first_published_service_key` FAILED
  - (Negative-path tests — uncommitted set member, unpublished signer, non-member — pass both ways, as expected: they assert fail-closed behavior that also held pre-fix.)
  Worktree restored to PR head afterward; `git status` clean.
- Grepped the diff for the #1145 pattern ("absent expected value disables the check"): every empty/absent case in the new code errors — empty `precommitted` → `SignerNotCommitted`; empty set in `verify_capsule`/`from_persisted_update` → error via `last_err.unwrap_or_else`; `new_gate_verified_set` rejects empty and duplicate; the `factories.rs`/`service.rs` `find`/`any` conversions terminate in `ok_or_else`/`ensure`. **No instance found.**
- Residual `subject_keys.first()`/`[0]`: only in `#[cfg(test)]` fixtures. Production is clean.

## Findings (severest first)

### F1 — MAJOR (composition, blocks the **pair**, primarily #1209): option C still selects the CA positionally, and #1210 falsifies its stated invariant

The tasking's predicted hazard is real. PR #1209's `ca_key_from_capsule` (https://github.com/hyprstream/hyprstream/pull/1209, `crates/hyprstream-discovery/src/did_anchored.rs`) does:

```rust
let primary = verified.capsule().body.subject_keys.first()...
```

documented as *"the deployment CA is the capsule's primary hybrid subject key (`subject_keys[0]`) — the same key that GATE-verified the capsule signature."* After #1210, **both halves of that sentence are false**:

1. `verify_capsule` now accepts a signature by **any** published subject key, so `subject_keys[0]` is no longer "the key that GATE-verified the capsule." A two-key capsule self-certified by the second key (legal under #1210, tested) makes #1209 install a CA key that signed nothing.
2. CA-from-position is precisely the positional-authority/cardinality anti-pattern #1183 names and option C was meant to *dissolve*, not relocate: an overlap rotation of the deployment identity changes the effective CA with list order, and every consumer pinned to "the CA" breaks on the first overlap publication — the #1157/F4 outage shape, moved from the did:web document into the capsule.

Contributing API gap in **#1210**: `verify_capsule` returns `Ok(())` without reporting *which* key verified, so a consumer that legitimately needs the self-cert signer (as #1209 claims to) cannot obtain it. Fix direction: either (a) `verify_capsule` returns the verified `&HybridKeyPair` (and #1209 uses it), or (b) #1209 accepts leaf certs chaining to **any** published subject key (true set semantics — the better option per the durable rule). Either way, **#1209 must not merge in its current form on top of #1210**; #1210 itself needs at most the small API affordance. Sequencing/ownership note for the queue, filed here because the tasking required the composition check.

### F2 — MEDIUM: the anonymous resolve path now hard-fails on any multi-key capsule — overlap is not actually deployable end-to-end until #1187

`identity_resolver.rs::resolve_did_at9p` newly rejects `keys.len() != 1`. Pre-fix, a multi-key capsule signed by key[0] was valid and resolved (to key[0]); post-fix the same capsule errors. Consumer: `auth/atproto_perimeter.rs:127`. Consequence: the moment an identity begins overlap rotation — the exact capability this PR ships at sign/GATE/admission — its perimeter resolution breaks until https://github.com/hyprstream/hyprstream/issues/1187 lands. The choice is defensible under the #1183 durable rule (no selector → reject ambiguity, never take `keys[0]`) and it is honestly documented in-code, but per the repo prior, *a fail-closed conversion that denies the legitimate path is an outage*. Required: (a) a test asserting the multi-key rejection (this new fail-closed branch is currently untested), and (b) an explicit operational note (PR description or #1187) that **no production identity may publish a multi-key capsule until #1187 merges**. Ideally #1187 is treated as a co-requisite before anyone exercises overlap.

### F3 — MEDIUM: four hunks ship untested; the PR description's test claim overreaches

"Every fix ships a test that fails without it (verified by reverting the hunk...)" is true for the six tests above but **false for four production hunks**, which have no new test anywhere in the diff:

- `at9p_duplicity.rs::from_persisted_update` — the overlap case this hunk exists for (persisted update signed by a non-first published key round-trips) is unexercised.
- `discovery/service.rs::ResolvedService::mint` — no test admits a selected service under the second accepted current key (or rejects a cross-pair PQ mismatch against the matched entry).
- `hyprstream/src/services/factories.rs::with_checkpointed_native_announcements` — no test; also sits in the crate whose build job is merge_group-only (the #1196 lesson: red tests hide there). I ran the crate's 1099 lib tests locally — green — but green-on-single-key proves nothing about the two-key behavior these hunks changed.
- `identity_resolver.rs` multi-key fail-closed branch (see F2).

Per the tasking's own bar: untested overlap is untested, not working. Add two-key tests for these consumer sites.

### F4 — MINOR: no cardinality cap on `subject_keys` → try-each verification amplification on untrusted input

`validate()` requires non-empty and unique but sets no upper bound. `verify_capsule` now performs up to N composite (Ed25519 + ML-DSA-65) verifications over attacker-supplied capsule bytes at admission (`admit_key_against_did` → GATE); pre-fix cost was exactly 1. The DAG-CBOR decoder's header hardening bounds *claimed* sizes, not real entries — a real 200-key capsule (~400 KB of ML-DSA pubkeys) decodes fine. Same shape applies to Gate 1's pre-committed candidate loop. Recommend a small `MAX_SUBJECT_KEYS` (and `MAX_NEXT_KEY_COMMITMENTS`) in `validate()` — overlap/hybrid needs 2–4, not 200.

### F5 — INFO: uniqueness is Ed25519-half-keyed only, by design

A repeated ML-DSA-65 half across distinct pairs is explicitly allowed. No current consumer selects by PQ half, so no ambiguity today — but this is a standing constraint: any future consumer keying on the ML-DSA half must add its own uniqueness guard or this becomes #1188 again with the halves swapped. Worth one line in the `validate_unique_subject_keys` doc (it's implied, not stated).

### F6 — INFO: persisted-update re-verify drops the pre-commitment leg

`from_persisted_update` now verifies "signed by some *published* key" where the old code pinned keys[0]; authorization-by-pre-commitment is not re-checked (documented: it was enforced at `validate_successor` acceptance; this path only re-confirms local bytes integrity). Acceptable given the trust model of locally persisted, previously-accepted records; noting it so the weakened re-check is a recorded decision, not an accident.

## What the change breaks (enumeration, as tasked)

1. Multi-key capsules on the anonymous resolve path: previously resolved to key[0], now error (F2) — a behavior regression on any already-published multi-key capsule.
2. The contract "capsule signer == `subject_keys[0]`" for every out-of-tree consumer — most concretely PR #1209 (F1). "GATE-verified" no longer identifies a particular key.
3. Capsules with duplicate subject-key entries or a repeated Ed25519 half now fail `validate()` (hence decode/`to_dag_cbor`/GATE). Such capsules were pathological but *were* previously valid; any persisted one becomes unreadable. Almost certainly none exist.
4. Error text/variant drift: several rejections now report aggregate try-each messages, and the mixed uncommitted-signer shape surfaces as `SignatureInvalid` where a caller might have expected `SignerNotCommitted` (the new test deliberately accepts either — fine, but callers matching variants should know).
5. Per-verification cost on hostile input scales with set size (F4).

## Acceptance bar

Met for the core: two keys published simultaneously, both usable (chain: signed-by-second accepted alongside existing first-key tests; admission: second key admitted with its atomic pair, first-key path covered by existing tests), and the retired rotation rejected after its bounded `expires_at` window (`overlap_retired_after_bounded_window_rejected`). Uncommitted set members gain no authority (both shapes tested). Fail-closed preserved throughout; no #1145 instance; no security regression found — the change strictly widens *which published, pre-committed* key may act, never admits an unpublished or uncommitted one.

## Required for PASS

1. Resolve F1 with #1209 (preferably: #1209 accepts any published subject key as CA; or #1210 exposes the verified signer from `verify_capsule` and #1209 uses it). Do not merge #1209 unrevised after #1210.
2. F3 tests: two-key coverage for `from_persisted_update`, `ResolvedService::mint`, `with_checkpointed_native_announcements`; a test for the multi-key resolve rejection (F2).
3. F2 operational note: multi-key publication is blocked on #1187; state it where an operator will see it.
4. (Recommended, non-blocking) F4 cardinality cap.

ewindisch:
# VERDICT: REVISE — PR #1210 "fix(at9p): give subject_keys set semantics instead of primary-only" (#1188)

**Reviewer:** kiro (sole reviewer, operator ruling 2026-07-23) · reviewed PR head `58021440d932c0d6b17f496a4a4b7287cc41559a` (unchanged since both prior verdicts) · read-only worktree `.worktrees/b1210`, all cargo via `buildq`.

The core cryptographic change is **correct and mutation-verified**: set semantics is implemented consistently across sign / verify / GATE / admission / successor-check / native-announcements; it strictly widens *which published, pre-committed* key may act and never admits an unpublished or uncommitted one; fail-closed is preserved throughout and there is no #1145 "absent value disables the check" instance. It is **not a PASS** for four reasons, severest first — the tasked composition check with #1209 **fails as written**, the PR description's blanket test claim is **false** (four production hunks ship no test), the new multi-key resolve rejection is an untested latent outage on the legitimate overlap path, and the live admission boundary gains an uncapped try-each verification loop.

Prior verdicts on this exact SHA: sol **BLOCK**, fable **REVISE**. I independently re-derived every prior finding below (not deferred to their write-ups) and land on **REVISE** — most of sol's BLOCK rests on paths that are **not wired live yet** (so they are forward-looking gaps to fix before those paths light up, not present regressions), while the test-coverage falsehood and the composition failure are real and squarely #1210's to fix.

---

## Findings (severest first)

### F1 — HIGH (the tasked composition check): #1210 does NOT compose with #1209 option C as written

This was the explicit acceptance question ("**Verify the two compose.**"). They do not.

- #1210 makes `verify_capsule` (`at9p_sign.rs`) accept a composite signature by **any** published subject key, and `sign_capsule` accept **any** published key as the self-cert signer. This is correct set semantics — and it **falsifies the invariant #1209 relies on**. #1209's `ca_key_from_capsule` selects the deployment CA as `verified.capsule().body.subject_keys.first()`, documented as "the same key that GATE-verified the capsule signature." After #1210 that sentence is false: a two-key capsule self-certified by its *second* key (legal and tested under #1210) makes #1209 install a CA key that signed nothing, and CA-by-position is exactly the positional-authority/cardinality anti-pattern #1183/#1188 exist to dissolve.
- `did_anchored.rs` (the #1209 consumer) **does not exist in #1210's tree** (base = `main`; confirmed). So merging #1210 to main breaks nothing extant — the hazard is a constraint on **#1209**, which must not merge unrevised on top of #1210.
- Contributing gap **in #1210**: `verify_capsule` returns `Ok(())` without reporting *which* key verified, so a consumer that legitimately needs the self-cert signer (as #1209 does) cannot obtain it non-positionally.

**Required:** either (a) `verify_capsule` returns the verified `&HybridKeyPair` and #1209 uses it, or (b) #1209 selects the CA by trying every published subject key (true set semantics — preferred per the durable rule). #1210 itself needs at most the small API affordance in (a); #1209 must not land unrevised.

### F2 — MEDIUM: the PR description's central test claim is FALSE — four production hunks ship with no test

The PR states: *"Every fix ships a test that fails without it (verified by reverting the hunk, watching it fail, and restoring)."* I verified the six covered tests **do** fail under mutation (see Verification). But four changed production hunks have **no new test anywhere in the diff** (confirmed by grep):

- `at9p_duplicity.rs::from_persisted_update` — the overlap case it exists for (persisted update signed by a non-first published key round-trips) is unexercised.
- `discovery/service.rs::ResolvedService::mint` — no test admits a selected service under the second accepted current key, nor rejects a cross-pair PQ mismatch against the matched entry.
- `hyprstream/src/services/factories.rs::with_checkpointed_native_announcements` — no test; also sits in the `hyprstream` crate whose build job is `merge_group`-only (the #1196 lesson — red tests hide there). Its sibling `factory.rs` hunk *is* tested (`native_announcement_admits_non_first_published_service_key`, passing), but that proves nothing about this `factories.rs` copy.
- `identity_resolver.rs::resolve_did_at9p` multi-key fail-closed branch (F3) — untested.

Per the tasking's own bar ("untested overlap is untested, not working") and the repo prior ("labeling is not enforcement"), a false coverage claim over a security boundary is a REVISE. **Required:** add two-key regression tests for these four sites, or correct the PR description to scope the claim honestly.

### F3 — MEDIUM: multi-key resolve rejection is a latent outage on the legitimate overlap path (untested, unreachable today)

`resolve_did_at9p` newly rejects any capsule with `keys.len() != 1`; the consumer `auth/atproto_perimeter.rs::enroll` treats a resolver error as failed enrollment. So the moment an identity begins overlap rotation — the exact capability this PR ships at sign/GATE/admission — its perimeter enrollment breaks, until #1187 lands set-aware resolution.

**Why this is MEDIUM, not BLOCK (sol rated it BLOCK):** the production resolver `At9pGateResolver::resolve()` returns `Err` unconditionally (mainline locator #890 not wired), and `.with_at9p(...)` is wired **only in test fixtures** — so `resolve_did_at9p` never succeeds in production regardless of key count today. The rejection is also defensible under the #1183 durable rule (no selector → reject ambiguity, never take `keys[0]`) and is honestly documented in-code. It is a *latent* gap, not a present outage. **Required:** (a) a test asserting the multi-key rejection (this new fail-closed branch is untested — folds into F2), and (b) an explicit operational note (PR body or #1187) that **no production identity may publish a multi-key capsule until #1187 merges**; ideally #1187 is a co-requisite before anyone exercises overlap.

### F4 — LOW/MEDIUM: no cardinality cap on `subject_keys` → O(N) try-each verification on capsule bytes

`CapsuleBody::validate()` requires non-empty + unique but sets **no upper bound**. `verify_capsule` and Gate 1's pre-committed loop now perform up to N composite (Ed25519 + ML-DSA-65) verifications over capsule bytes; pre-fix cost was exactly 1. A valid capsule signed by its last member (or none) turns one GATE attempt into attacker-chosen O(N) expensive PQ work.

Bounded in practice: the DAG-CBOR decoder caps *claimed* sizes so a real ~200-key capsule (~400 KB) still decodes; the live at9p admission arm is **not wired** (`FederationAdmissionGate::admit` passes `None` for did:at9p — reachable only via the test-only direct `admit_key_against_did`); and the stage-1 origin policy gate (`federation:register`) precedes the capsule GATE, so this is an authenticated-origin amplification, not anonymous. Hence LOW today, MEDIUM once the arm goes live. **Recommended (fix before live):** add `MAX_SUBJECT_KEYS` (and `MAX_NEXT_KEY_COMMITMENTS`) to `validate()` — overlap/hybrid needs 2–4, not 200.

### F5 — INFO (recorded decisions, not asks)

- Uniqueness is **Ed25519-half-keyed only** by design (a repeated ML-DSA-65 half across distinct pairs is allowed; no consumer selects by PQ half today). Standing constraint: any future PQ-half selector must add its own guard or this becomes #1188 with the halves swapped. Worth one doc line.
- `from_persisted_update` now re-verifies "signed by *some* published key" and drops the pre-commitment leg (documented: authorization-by-pre-commitment was enforced at `validate_successor` acceptance; this path only re-confirms retained-byte signature integrity). Acceptable for locally-persisted, previously-accepted records; noted so the weakened re-check is a recorded decision.
- sol's F1 "immortal genesis keys" (admission admits any genesis-published key forever, never consulting accepted-state/expiry): real, but a **pre-existing property of the not-yet-live D2 admission path** at N=1, not a regression #1210 introduces — it widens cardinality, it does not create the immortality. Bounded retirement *does* exist at the layer that owns it (the successor-check `expires_at` freshness gate, tested). Capsule-currentness on the admission path is the separately-tracked concern (#1187/#1218), not #1188's scope. This is why I do not treat it as a #1210 blocker.

---

## What the change BREAKS (enumerated, as tasked)

1. **Multi-key capsules on the anonymous resolve path**: previously resolved to `key[0]`, now error (F3). Behavior regression on any already-published multi-key capsule — but the resolve path is unwired in prod, so no live path hits it today.
2. **The contract "capsule signer == `subject_keys[0]`"** for every out-of-tree consumer — concretely #1209 (F1). "GATE-verified" no longer identifies a particular key.
3. **Capsules with a duplicate hybrid pair or a repeated Ed25519 half** now fail `validate()` (hence decode / `to_dag_cbor` / GATE). Pathological and previously valid; almost certainly none exist.
4. **Error-variant drift**: several rejections now report aggregate try-each messages; the mixed uncommitted-signer shape surfaces as `SignatureInvalid` where a caller might have expected `SignerNotCommitted` (the new test deliberately accepts either — fine, but variant-matching callers should know).
5. **Per-verification cost on hostile capsule bytes** scales with set size (F4), bounded by decode size and gated behind origin policy.

No security regression: the change never admits an unpublished or uncommitted key (mutation- and negative-test-verified).

---

## Verification

All cargo via `buildq`; no source left modified (tree clean at the reviewed SHA, `git diff --exit-code` clean).

**Baseline at head** (`buildq -- cargo test -p hyprstream-pds -p hyprstream-rpc -p hyprstream-service --lib`):
- `hyprstream-pds`: **207 passed**, 0 failed.
- `hyprstream-rpc`: **772 passed**, 0 failed, 3 ignored.
- `hyprstream-service`: **29 passed**, 0 failed, 1 ignored (incl. new `native_announcement_admits_non_first_published_service_key`).

**Mutation (revert-the-hunk, watch it fail, restore) — I performed these:**
- Reverted **Gate 1** in `at9p_chain.rs` to positional `subject_keys.first()` → `overlap_successor_signed_by_non_first_committed_key_accepted` **FAILED** and `overlap_retired_after_bounded_window_rejected` **FAILED** (both: EdDSA verification against the wrong/first key). Restored; retested green.
- Reverted **`verify_capsule`** in `at9p_sign.rs` to positional-only → `capsule_signed_by_non_first_subject_key_verifies` **FAILED**. Restored; retested green.
- Confirmed sol's F5 observation independently: `capsule_signed_by_unpublished_key_rejected` stays **green** under the verify mutation. This is **correct and expected** — it guards fail-closed rejection of an unpublished signer, a property that also held pre-fix; it is not a positional-fix regression test. So "every new test fails without its fix" is imprecise, but the negative test is doing its intended job.

**Structural checks:**
- `At9pGateResolver::resolve()` returns `Err` unconditionally (locator #890 unwired); `.with_at9p(...)` appears only in `#[cfg(test)]`. → F3 unreachable in prod today.
- `did_anchored.rs` absent from this tree (base `main`). → F1 gates #1209, not #1210.
- No production `subject_keys.first()` / `[0]` consumer remains outside `#[cfg(test)]` (grep clean).
- `#1145` grep of the diff: every empty/absent case errors (`precommitted.is_empty()` → `SignerNotCommitted`; empty set in `verify_capsule`/`from_persisted_update` → `last_err.unwrap_or_else` error; `new_gate_verified_set` rejects empty + duplicate; `find`/`any` conversions terminate in `ok_or_else`/`ensure`). **No instance.**

---

## Required for PASS

1. **F1:** resolve composition with #1209 — expose the verified signer from `verify_capsule` (and #1209 consumes it) **or** #1209 selects CA by trying every published key. Do not merge #1209 unrevised on top of #1210.
2. **F2/F3:** add two-key regression tests for `from_persisted_update`, `ResolvedService::mint`, `with_checkpointed_native_announcements`, and the `resolve_did_at9p` multi-key rejection — or correct the PR description's coverage claim.
3. **F3:** operational note that multi-key capsule publication is blocked until #1187 (state it where an operator will see it); treat #1187 as co-requisite for overlap.
4. **F4 (before the at9p admission arm goes live):** add `MAX_SUBJECT_KEYS` / `MAX_NEXT_KEY_COMMITMENTS` bounds in `validate()`.

ewindisch:
Coordination note from #1209 fix 659f0f091: DID bootstrap now replays the accepted update history from GATE genesis and authenticates the registry credential against the resulting full current hybrid key set. Please keep #1210’s non-positional subject-key verification/persisted-update changes in the landing sequence; #1209 must not land without them, because a non-first valid hybrid signer must survive replay and current-CA selection atomically.

Issue #1185, opened 2026-07-22
fix(federation): make FederationKeyResolver JWKS rotation-aware


ewindisch:
Relates #1183.

## Problem

The legacy federation path reduces a published JWKS to its first Ed25519 entry before it sees the JWT's `kid`.

- `crates/hyprstream/src/auth/federation.rs:129-245`: `get_key` has no `kid` parameter and caches one `VerifyingKey` per issuer.
- `crates/hyprstream/src/auth/federation.rs:343-367`: `fetch_ed25519_key` returns the first OKP/Ed25519 key.
- `crates/hyprstream-rpc/src/auth/key_source.rs:231-243`: `FederatedKeySource` accepts `kid` but drops it when delegating to `FederationKeySource`.
- Live HTTP middleware then verifies the token only with that one key (`crates/hyprstream/src/server/middleware.rs:442-450`).

Therefore an issuer publishing old+new Ed25519 keys cannot rotate safely: tokens signed by any non-first key fail, despite a valid published `kid`. There is no two-key overlap test for this resolver.

## Required direction

Change the federation resolver contract/cache to resolve by `(issuer, kid)` (or return a verified candidate set and make the verifier try it), retain all usable JWKS entries, and refetch on an unknown `kid`. Add an end-to-end test for a two-key JWKS where the token is signed by the second key. Do not retain a first-key fallback.

ewindisch:
Closing: the fix landed in main via #1214 (merged 2026-07-24, body says "Closes #1185" but the auto-close never fired). Verified the merge commit is an ancestor of main. Reopen if rotation still picks a stale key.

PR #1213, opened 2026-07-23
fix(auth): try all JWKS candidates for kid-less JWTs

### LINKED ISSUE #1184: fix(auth): reject or try all compatible JWKS entries for kid-less JWTs
Relates #1183.

## Problem

Two generic JWT/JWKS consumers silently choose one candidate when a token has no `kid`, so overlap rotation becomes order-dependent instead of being verified against the published set.

- `crates/hyprstream/src/auth/id_token_verify.rs:111-140`: `find_matching_key` returns the first same-algorithm key, then ultimately `keys.first()`. An externally issued id_token without `kid`, signed by the second simultaneously published RS256/ES256/EdDSA key, fails although its key is present.
- `crates/hyprstream-rpc/src/auth/key_source.rs:536-549`: `JwksKeySource::get_key(..., None)` returns an arbitrary `HashMap::values().next()` key. It cannot know that the returned key verifies the token, and the result is nondeterministic across runs.

Both code paths have single-key tests only (`key_source.rs:741-754`); neither tests two simultaneously published same-algorithm keys and a kid-less token signed by the non-first key.

## Required direction

Make the safe default structural: a verifier receiving a JWKS must either require `kid` at this boundary or retain/filter all algorithm-compatible candidates and try them until one verifies. Never turn an unordered published key set into an arbitrary single key. Add two-key overlap tests (including non-first signer and unknown/unsupported entries).

ewindisch:
Closes #1184. Relates #1183.

## Problem

Two generic JWT/JWKS consumers collapsed a valid published key SET to a positional or arbitrary singleton, foreclosing both overlap rotation and PQ-hybrid rollout (#1183):

- `crates/hyprstream/src/auth/id_token_verify.rs` — `find_matching_key` returned the first same-algorithm key, then ultimately `keys.first()`. An externally issued id_token without `kid`, signed by the second simultaneously published RS256/ES256/EdDSA key, failed although its key was present.
- `crates/hyprstream-rpc/src/auth/key_source.rs` — `JwksKeySource::get_key(..., None)` returned an arbitrary `HashMap::values().next()` key. It could not know that the returned key verifies the token, and the result was nondeterministic across runs.

Both paths had single-key tests only; neither tested two simultaneously published same-algorithm keys with a kid-less token signed by the non-first key.

## Fix

Per the audit's durable rule — *"a consumer MUST select by the protocol's key id/relationship; when no selector exists, it MUST try each compatible candidate or reject ambiguity"* — the safe default is now structural: **retain every algorithm-compatible candidate and try each until one verifies**. Matches the in-tree reference implementation in `services/oauth/client_auth.rs` (which already did this).

### `hyprstream-rpc`
- **`JwtKeySource::get_keys(issuer, kid)`** — new trait method returning `Vec<VerifyingKey>`; default delegates to `get_key` for single-key sources. `JwksKeySource` overrides it to return every Ed25519 entry when `kid` is `None`, or the single kid-matched key when present.
- **`JwksKeySource::get_key(None)`** is now fail-closed when the JWKS publishes more than one key: it refuses to pick arbitrarily and points the caller at `get_keys()`. Single-key resolution is unchanged.
- **`jwt::decode_with_any_key`** (strict aud, mirrors `decode`) and **`decode_with_any_key_lenient`** (mirrors `decode_with_key`) try each candidate; per-candidate failures are `debug!`-level, unknown/unsupported algorithms are skipped rather than fatal to the set.
- `svc.rs` EdDSA path uses `get_keys` + `decode_with_any_key_lenient` (preserving the original lenient-aud choice).

### `hyprstream`
- `id_token_verify`: `find_matching_key`/`determine_algorithm` replaced by **`candidate_keys`** (kid-filtered or alg-compatible set) + **`algorithm_from_header`**. **`verify_id_token_with_keys`** is a pure (HTTP-free) verification core that iterates candidates; `verify_id_token` composes fetch + core. The old "fallback: return first key" is gone.
- `factories.rs` MCP path uses `get_keys` + `decode_with_any_key` (strict aud preserved — was `decode`).

## Tests

Every new test **fails without its fix hunk** (verified by reverting and re-running):

| Test | Asserts |
|---|---|
| `jwks_key_source_no_kid_multi_key_get_key_refuses_arbitrary` | `get_key(None)` errors when >1 key published (no arbitrary pick) |
| `jwks_key_source_no_kid_multi_key_get_keys_tries_all` | `get_keys(None)` returns BOTH published keys |
| `jwks_key_source_overlap_non_first_signer_verifies` | a kid-less token signed by the 2nd published key verifies via `get_keys` + `decode_with_any_key_lenient` |
| `id_token_kid_less_signer_is_non_first_published_key` | non-first signer verifies against a two-key JWKS |
| `id_token_overlap_both_signers_accepted_unknown_rejected` | both signers accepted; unknown signer rejected |
| `id_token_skips_unsupported_alg_and_accepts_classical` | an `ML-DSA-65-Ed25519` (AKP) entry is skipped, classical Ed25519 still verifies |

## Verify

```
buildq -- cargo build -p hyprstream-rpc -p hyprstream
buildq -- cargo test  -p hyprstream-rpc --lib key_source        # 18 ok
buildq -- cargo test  -p hyprstream --lib auth::id_token_verify # 14 ok
buildq -- cargo test  -p hyprstream-rpc --lib service::svc      # 23 ok
buildq -- cargo test  -p hyprstream --lib client_auth           # 11 ok
buildq -- cargo clippy --workspace --all-targets --release -- -D warnings
```

`cargo fmt --check` fails on a clean main (#1140) — not introduced here.

CodeRabbit threads will be handled on the PR.

### DIFF
--- a/crates/hyprstream-rpc/src/auth/jwt.rs
+++ b/crates/hyprstream-rpc/src/auth/jwt.rs
@@ -387,6 +387,62 @@ pub fn decode_with_key(
     decode_inner(token, verifying_key, expected_aud, true)
 }
 
+/// Try verifying a JWT against each candidate key, returning the first
+/// successful decode. Mirrors [`decode`] (strict audience: an absent `aud`
+/// is rejected when `expected_aud` is `Some`).
+///
+/// This is the safe verifier shape for a kid-less JWT: when the JOSE header
+/// carries no `kid`, the verifier cannot know which published key signed the
+/// token, so it MUST try every algorithm-compatible candidate rather than
+/// collapse the published set to a positional singleton (#1183 / #1184).
+/// Per-candidate failures are expected in the happy path (a JWKS with several
+/// keys has exactly one real verifier), so they are logged at `debug!`; the
+/// final failure surfaces `InvalidSignature`.
+///
+/// `candidates` MUST be non-empty; an empty slice is an `InvalidSignature`
+/// error rather than a panic.
+pub fn decode_with_any_key(
+    token: &str,
+    candidates: &[VerifyingKey],
+    expected_aud: Option<&str>,
+) -> Result<Claims, JwtError> {
+    decode_with_any_key_inner(token, candidates, expected_aud, false)
+}
+
+/// Lenient-audience twin of [`decode_with_any_key`], mirroring
+/// [`decode_with_key`]: a wrong `aud` is still rejected, but an absent `aud`
+/// is accepted when `expected_aud` is `Some`. Use this for federated tokens
+/// from issuers that do not set `aud`.
+pub fn decode_with_any_key_lenient(
+    token: &str,
+    candidates: &[VerifyingKey],
+    expected_aud: Option<&str>,
+) -> Result<Claims, JwtError> {
+    decode_with_any_key_inner(token, candidates, expected_aud, true)
+}
+
+fn decode_with_any_key_inner(
+    token: &str,
+    candidates: &[VerifyingKey],
+    expected_aud: Option<&str>,
+    lenient_aud: bool,
+) -> Result<Claims, JwtError> {
+    if candidates.is_empty() {
+        return Err(JwtError::InvalidSignature);
+    }
+    let mut last_err = JwtError::InvalidSignature;
+    for vk in candidates {
+        match decode_inner(token, vk, expected_aud, lenient_aud) {
+            Ok(claims) => return Ok(claims),
+            Err(e) => {
+                tracing::debug!(error = %e, "JWT candidate key did not verify; trying next");
+                last_err = e;
+            }
+        }
+    }
+    Err(last_err)
+}
+
 /// Internal decode implementation shared by `decode` and `decode_with_key`.
 ///
 /// `lenient_aud`: when true, accepts tokens with no `aud` claim even when
--- a/crates/hyprstream-rpc/src/auth/key_source.rs
+++ b/crates/hyprstream-rpc/src/auth/key_source.rs
@@ -55,6 +55,24 @@ pub trait JwtKeySource: Send + Sync + 'static {
     /// Returns error if the issuer is untrusted or key resolution fails.
     async fn get_key(&self, issuer: &str, kid: Option<&str>) -> Result<VerifyingKey>;
 
+    /// Resolve **all** algorithm-compatible candidates for an issuer/kid pair.
+    ///
+    /// This is the safe default for verifying a kid-less JWT: when no `kid`
+    /// selector is present the verifier cannot know which published key signed
+    /// the token, so it MUST retain every compatible candidate and try each
+    /// until one verifies — never collapse the published set to a positional
+    /// singleton (`first`/`next`/`values().next()`). See #1183 / #1184.
+    ///
+    /// Single-key sources return a one-element vec. JWKS-backed sources
+    /// override this to return every Ed25519 entry when `kid` is `None`.
+    ///
+    /// Callers pair this with [`crate::auth::jwt::decode_with_any_key`] to
+    /// try each candidate against the JWT signature.
+    async fn get_keys(&self, issuer: &str, kid: Option<&str>) -> Result<Vec<VerifyingKey>> {
+        let key = self.get_key(issuer, kid).await?;
+        Ok(vec![key])
+    }
+
     /// Check if an issuer is trusted (before attempting key fetch).
     ///
     /// Returns `true` for local issuers and configured federated issuers.
@@ -473,6 +491,74 @@ impl JwksKeySource {
 
         Ok(())
     }
+
+    /// Resolve a single key by `kid`, consulting the offline local-CA path,
+    /// the JWKS cache, the negative cache, and (on miss) a refetch. The
+    /// `kid` selector makes the choice unambiguous — contrast with the
+    /// kid-less path in [`JwtKeySource::get_keys`] which must return every
+    /// candidate.
+    async fn resolve_by_kid(&self, issuer: &str, kid_str: &str) -> Result<VerifyingKey> {
+        // Authoritative local CA key (offline): for a LOCAL issuer whose
+        // `kid` matches the on-disk CA key thumbprint, resolve without any
+        // network fetch. This is the service-to-service WIT JWT path; the
+        // HTTP /oauth/jwks endpoint may not be up yet during startup, and
+        // #441 makes service-key registration a hard precondition, so this
+        // resolution must not depend on it. Never overrides a JWKS-published
+        // (rotated) key because it's keyed on the exact CA thumbprint.
+        if self.is_local(issuer) {
+            if let (Some(ca_kid), Some(ca_key)) = (self.local_ca_kid.as_deref(), self.local_ca_key) {
+                if ca_kid == kid_str {
+                    return Ok(ca_key);
+                }
+            }
+        }
+        {
+            let cache = self.cache.read().await;
+            if let Some(entry) = cache.get(kid_str) {
+                return Ok(entry.verifying_key);
+            }
+        }
+
+        // Check negative cache
+        {
+            let neg = self.negative_cache.read().await;
+            if let Some(&ts) = neg.get(kid_str) {
+                if ts.elapsed() < self.negative_ttl {
+                    anyhow::bail!("Key not found for kid={} (negative cached)", kid_str);
+                }
+            }
+        }
+
+        // On-miss refetch
+        if let Err(e) = self.fetch_and_cache(issuer).await {
+            tracing::warn!("JWKS fetch failed for issuer={}: {}", issuer, e);
+        }
+
+        // Retry from cache
+        let cache = self.cache.read().await;
+        if let Some(entry) = cache.get(kid_str) {
+            return Ok(entry.verifying_key);
+        }
+
+        // Add to negative cache
+        {
+            let mut neg = self.negative_cache.write().await;
+            neg.insert(kid_str.to_owned(), std::time::Instant::now());
+        }
+
+        anyhow::bail!("Key not found in JWKS for kid={}", kid_str)
+    }
+
+    /// Ensure the JWKS cache has been populated for `issuer` (refetch on
+    /// empty). Used by the kid-less resolution paths before they read every
+    /// candidate out of the cache.
+    async fn ensure_cached(&self, issuer: &str) {
+        if self.cache.read().await.is_empty() {
+            if let Err(e) = self.fetch_and_cache(issuer).await {
+                tracing::warn!("JWKS fetch failed for issuer={}: {}", issuer, e);
+            }
+        }
+    }
 }
 
 #[async_trait::async_trait]
@@ -482,71 +568,52 @@ impl JwtKeySource for JwksKeySource {
             anyhow::bail!("Untrusted issuer: {}", issuer);
         }
 
-        // Try cache first
+        // A presented `kid` is unambiguous: resolve that exact slot.
         if let Some(kid_str) = kid {
-            // Authoritative local CA key (offline): for a LOCAL issuer whose
-            // `kid` matches the on-disk CA key thumbprint, resolve without any
-            // network fetch. This is the service-to-service WIT JWT path; the
-            // HTTP /oauth/jwks endpoint may not be up yet during startup, and
-            // #441 makes service-key registration a hard precondition, so this
-            // resolution must not depend on it. Never overrides a JWKS-published
-            // (rotated) key because it's keyed on the exact CA thumbprint.
-            if self.is_local(issuer) {
-                if let (Some(ca_kid), Some(ca_key)) = (self.local_ca_kid.as_deref(), self.local_ca_key) {
-                    if ca_kid == kid_str {
-                        return Ok(ca_key);
-                    }
-                }
-            }
-            {
-                let cache = self.cache.read().await;
-                if let Some(entry) = cache.get(kid_str) {
-                    return Ok(entry.verifying_key);
-                }
-            }
-
-            // Check negative cache
-            {
-                let neg = self.negative_cache.read().await;
-                if let Some(&ts) = neg.get(kid_str) {
-                    if ts.elapsed() < self.negative_ttl {
-                        anyhow::bail!("Key not found for kid={} (negative cached)", kid_str);
-                    }
-                }
-            }
-
-            // On-miss refetch
-            if let Err(e) = self.fetch_and_cache(issuer).await {
-                tracing::warn!("JWKS fetch failed for issuer={}: {}", issuer, e);
-            }
-
-            // Retry from cache
-            let cache = self.cache.read().await;
-            if let Some(entry) = cache.get(kid_str) {
-                return Ok(entry.verifying_key);
-            }
-
-            // Add to negative cache
-            {
-                let mut neg = self.negative_cache.write().await;
-                neg.insert(kid_str.to_owned(), std::time::Instant::now());
-            }
-
-            anyhow::bail!("Key not found in JWKS for kid={}", kid_str)
-        } else {
-            // No kid: fetch JWKS and return the first Ed25519 key (fallback)
-            if self.cache.read().await.is_empty() {
-                if let Err(e) = self.fetch_and_cache(issuer).await {
-                    tracing::warn!("JWKS fetch failed for issuer={}: {}", issuer, e);
-                }
-            }
-
-            let cache = self.cache.read().await;
-            cache.values()
-                .next()
-                .map(|e| e.verifying_key)
-                .ok_or_else(|| anyhow::anyhow!("No Ed25519 keys in JWKS"))
+            return self.resolve_by_kid(issuer, kid_str).await;
         }
+
+        // No kid: refuse to collapse a multi-key published set to a positional
+        // singleton. `get_keys` is the safe API for kid-less tokens — it
+        // returns every Ed25519 candidate so the caller can try each. Here we
+        // only succeed when the JWKS publishes exactly one key, which is the
+        // unambiguous case. Two or more published keys is overlap rotation /
+        // PQ-hybrid rollout territory and MUST NOT be resolved to "the first
+        // one" (#1183 / #1184).
+        self.ensure_cached(issuer).await;
+        let cache = self.cache.read().await;
+        let mut iter = cache.values();
+        match (iter.next(), iter.next()) {
+            (Some(first), None) => Ok(first.verifying_key),
+            (None, _) => anyhow::bail!("No Ed25519 keys in JWKS for issuer {}", issuer),
+            (Some(_), Some(_)) => anyhow::bail!(
+                "Issuer {} publishes multiple Ed25519 keys and the JWT carries no kid; \
+                 use get_keys() and try each candidate (overlap rotation / hybrid rollout)",
+                issuer
+            ),
+        }
+    }
+
+    async fn get_keys(&self, issuer: &str, kid: Option<&str>) -> Result<Vec<VerifyingKey>> {
+        if !self.is_trusted(issuer) {
+            anyhow::bail!("Untrusted issuer: {}", issuer);
+        }
+
+        // A presented `kid` selects one slot.
+        if let Some(kid_str) = kid {
+            let vk = self.resolve_by_kid(issuer, kid_str).await?;
+            return Ok(vec![vk]);
+        }
+
+        // No kid: return EVERY Ed25519 candidate so the caller can try each
+        // until one verifies. Never positional selection (#1183 / #1184).
+        self.ensure_cached(issuer).await;
+        let cache = self.cache.read().await;
+        let v: Vec<VerifyingKey> = cache.values().map(|e| e.verifying_key).collect();
+        if v.is_empty() {
+            anyhow::bail!("No Ed25519 keys in JWKS for issuer {}", issuer);
+        }
+        Ok(v)
     }
 
     fn is_trusted(&self, issuer: &str) -> bool {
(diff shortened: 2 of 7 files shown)


ewindisch:
# Verdict: PR #1213 — fix(auth): try all JWKS candidates for kid-less JWTs

**PR:** https://github.com/hyprstream/hyprstream/pull/1213 · **Ticket:** https://github.com/hyprstream/hyprstream/issues/1184 · **Epic:** https://github.com/hyprstream/hyprstream/issues/1183
**Reviewer:** fable (one of two independent reviewers) · **Date:** 2026-07-23 · Read-only review at head `78f39fb63`.

## VERDICT: REVISE

The design decision is correct and the tests are real (all six revert-verified by me), but the PR
cannot merge as-is: it breaks the wasm32 CI build, and on the `JwksKeySource` path the widened
accepting set is **not bounded** — the candidate cache never drops a retired key and never picks up
a new one, so the acceptance criterion "retired key REJECTED after its bounded window" is
structurally false there and untested.

## The ticket's central question: reject, or try-all — and is it safe?

The PR chose **both, split by API**, which is exactly the audit's durable rule ("try each compatible
candidate **or** reject ambiguity"):

- **Verifying consumers try all**: `svc.rs` EdDSA, the MCP HTTP path in `factories.rs`, and
  `id_token_verify` now collect every algorithm-compatible candidate and try each
  (`get_keys` + `decode_with_any_key`/`_lenient`, `verify_id_token_with_keys`).
- **The legacy singleton API rejects ambiguity**: `JwksKeySource::get_key(None)` now fails closed
  when more than one key is published, pointing callers at `get_keys`. No production
  `get_key(kid=None)` callers remain in-tree (I enumerated: `svc.rs` and the MCP path migrated;
  remaining `get_key` callers are kid-ful, tests, or the separate `FederationKeySource` trait).

**Is try-all a widening, and is it intended?** Yes and yes — *in principle*. Accepting a token
verifiable by any currently-published key is the definition of overlap rotation; the audit
(`keyaudit/report.md`) ratifies it and the in-tree reference (`services/oauth/client_auth.rs`,
called "the desired verifier shape" by the audit; publisher side `services/oauth/jwks.rs` +
`auth/key_rotation.rs` with bounded drain expiry) already does it. The widening is bounded **iff
the consumer's candidate set tracks the issuer's current publication**. On the `id_token_verify`
path it does: the JWKS is fetched per verification, so retirement = absence from the fetched set,
and the `unknown_rejected` test genuinely covers retirement there. On the `JwksKeySource` path it
does not — see F2. The PR matched the reference's *verifier* shape but not its *bounded-window*
property.

## Findings (severest first)

### F1 — BLOCKER (trivial fix): wasm32 build broken by unguarded `tracing::debug!`
`crates/hyprstream-rpc/src/auth/jwt.rs:438` (`decode_with_any_key_inner`) calls
`tracing::debug!` unconditionally. `tracing` is a `cfg(not(target_arch = "wasm32"))` dependency of
`hyprstream-rpc` (Cargo.toml:115/143), and the two pre-existing `tracing::` calls in this same file
(lines 298–299, 350–351) are `#[cfg(not(target_arch = "wasm32"))]`-guarded for exactly this reason.
The **WASM (browser client)** check fails on the PR head with
`error[E0433]: cannot find module or crate 'tracing'` →
https://github.com/hyprstream/hyprstream/actions/runs/29978728532/job/89115945830.
This is a hard compile failure introduced by this PR (not the known `cargo fmt`/#1140 noise, and
not a queue flake). Fix: guard the call like its neighbors.

### F2 — MAJOR: the `JwksKeySource` kid-less candidate set is frozen at first fetch — the overlap window is unbounded, not bounded
`JwksKeySource.cache` is insert-only: `fetch_and_cache`
(`crates/hyprstream-rpc/src/auth/key_source.rs:431-483`) inserts per-key and **never evicts** — no
`clear()`, no `retain()`, and both `soft_ttl` and `CachedKey.fetched_at` are `#[allow(dead_code)]`.
Contrast `kid_alg_map` in the *same function*, which IS atomically replaced as a snapshot each
fetch (line 486). Meanwhile the new kid-less paths (`get_keys(None)`, `get_key(None)`) call
`ensure_cached`, which fetches **only when the cache is empty**, and have no
refetch-on-verification-failure analogous to `resolve_by_kid`'s on-miss refetch. Consequences:

1. **Retired keys are accepted forever** (until process restart): once cached, a key removed from
   the JWKS remains a candidate. The ticket's acceptance bar — "the retired one REJECTED after its
   bounded window" — is false on this path, and no test exercises it (a test would fail).
   Trying-all makes the accepting set = *the accumulated cache*, not the published set, so the
   widening the ticket asked about is unbounded here. (Kid-ful acceptance of a stale cached key is
   pre-existing; the PR is what turns the whole cache into the kid-less candidate set.)
2. **New keys published after the first fetch are never seen** by kid-less verification: a
   legitimate kid-less token signed by a newly-leaded key is denied until some kid-ful request
   happens to trigger a refetch, or restart. So overlap "works" only if both keys were published
   before the process's first fetch — timing-dependent, which is precisely the "untested overlap is
   untested" trap: the PR's tests all use a fresh source with a single fetch and cannot see this.
3. **`get_key(None)`'s fail-closed check counts cached entries, not published ones**: after one
   rotation observed via a kid-ful fetch, a single-key issuer looks permanently multi-key, so any
   future `get_key(None)` caller (the trait is public API) hits the hard "publishes multiple
   Ed25519 keys" error even though the issuer publishes one. Latent today (no in-tree callers),
   but it converts the fail-closed guard into a false positive — outage-shaped, per the repo prior.

Required: make the cache a per-fetch snapshot (as `kid_alg_map` already is) or evict by
`fetched_at`/`soft_ttl`, add a bounded refetch (negative-cache-style throttle) when all kid-less
candidates fail, and add the retirement test: fetch {A,B} → verify B → refetch {B} → token signed
by A REJECTED. If the maintainers prefer to defer this to a follow-up ticket, that ticket must
exist and be linked before merge, because the PR's stated claim ("bounded by the publisher's drain
window") does not hold on this path without it.

### F3 — MINOR: last-error masking in `decode_with_any_key_inner`
`crates/hyprstream-rpc/src/auth/jwt.rs:443-462` returns the **last** candidate's error. If the
real signer's key verifies the signature but the claims fail (expired, wrong `aud`), a later
candidate's `InvalidSignature` overwrites it, so an expired token surfaces as "invalid signature".
Diagnostics only — the deny is still correct. Same shape exists in `verify_id_token_with_keys`.
Consider preferring the most-specific error (claims errors over signature errors).

### F4 — INFO: behavior changes enumerated (what the change breaks)
- `get_key(None)` with ≥2 published keys: was a nondeterministic ~50%-correct singleton, now a hard
  error. Correct per the rule; no in-tree callers affected; external `JwtKeySource` users would see
  new failures (plus the F2.3 false positive).
- `id_token_verify`: the `keys.first()` fallback and infer-alg-from-kty path are **gone** — a
  kid-less token with a missing or unsupported header `alg` is now rejected (previously could
  verify via inference). RFC 7515-correct and fail-closed, but a noncompliant IdP that omits `alg`
  would break. The kid-matched wrong-`kty` case now fails with a generic "candidate did not verify"
  instead of the old descriptive kty-mismatch error (the kid path deliberately no longer
  alg-filters, to allow composite+classical entries under one kid; mismatches still fail at decode).
- Audience semantics preserved per call site: MCP kept strict (`decode` → `decode_with_any_key`),
  `svc.rs` kept lenient (`decode_with_key` → `decode_with_any_key_lenient`) — verified in the diff;
  no strict→lenient regression.
- Out of scope, unchanged, correctly so: the federation resolver path
  (`middleware.rs:447`, `factories.rs:1496`, `FederatedKeySource`) still collapses to a singleton —
  owned by https://github.com/hyprstream/hyprstream/issues/1185. Note `svc.rs`'s fix does **not**
  reach tokens verified through `FederationKeySource`.
- DoS surface: candidate count is unbounded → O(n) signature checks per request; bounded by the
  trusted issuer's JWKS (and, until F2 is fixed, inflated by cache accretion). Acceptable.

### F5 — INFO: #1145-shape grep came back clean
Every "absent expected value" in the new code fails closed: empty candidate slice →
`InvalidSignature`; `candidate_keys` empty → error; missing header `alg` → no candidates → error;
missing `kty` → `unwrap_or("")` → excluded; `ensure_cached` fetch failure → empty cache → error.
The `get_key(None)` singleton — the #1145 instance this ticket targets — is genuinely dead.

## Verification performed
- Built and ran all suites via `buildq`: `hyprstream-rpc --lib auth::` **178 passed**,
  `service::svc` **23 passed**, `hyprstream --lib auth::id_token_verify` **14 passed**,
  `client_auth` **11 passed**.
- Revert-verified the fixes (working tree only, restored after):
  - `id_token` loop → `.take(1)`: `id_token_kid_less_signer_is_non_first_published_key` and
    `id_token_overlap_both_signers_accepted_unknown_rejected` **FAIL**; the skip-unsupported-alg
    test rightly still passes (it is guarded by `candidate_keys` filtering, not the loop).
  - `key_source` → restored `values().next()` + `.take(1)`:
    `...get_key_refuses_arbitrary` and `...get_keys_tries_all` **FAIL deterministically**;
    `...overlap_non_first_signer_verifies` failed **4/6 runs** (HashMap-order nondeterminism —
    matching the PR's own "roughly half the time" claim).
- Confirmed production wiring: `cluster_key_source()`
  (`crates/hyprstream-service/src/service/factory.rs:707-722`) returns a real `JwksKeySource`, so
  the `get_keys` override is live in both changed consumers.
- CodeRabbit was rate-limited on this PR (no review to weigh); judged independently.

## Bottom line
Right rule, right reference, honest tests — REVISE for the wasm32 compile break (F1, one-line fix)
and for bounding the `JwksKeySource` overlap window (F2: snapshot/evict the cache, refetch on
kid-less miss, retirement test — or an explicit linked deferral). F3 is take-it-or-leave-it.

ewindisch:
# PR #1213 — BLOCK

The PR chose **try all compatible candidates** for kid-less JWTs. That is the intended choice for #1184: during a deliberate overlap window, a token verifiable by either currently published key is supposed to be accepted. The widening is safe only if the candidate set is an issuer-scoped, current JWKS snapshot and retirement removes the old key after a bounded drain. The external ID-token path fetches a fresh snapshot per verification, but the RPC path does not preserve either bound.

## Findings

### [BLOCK] The RPC try-all set is append-only, so a retired key remains accepted indefinitely

`JwksKeySource::fetch_and_cache` inserts every fetched Ed25519 key into the existing map but never clears or retains the map against the new JWKS snapshot (`crates/hyprstream-rpc/src/auth/key_source.rs:431-483`). `ensure_cached` fetches only while the entire cache is empty (`:552-561`), and the new kid-less path returns every cached value (`:597-617`). `CachedKey::fetched_at` and `soft_ttl` are explicitly dead (`:311-340`).

Consequently, after the working OAuth producer removes its expired drain key from `/oauth/jwks`, a long-lived RPC verifier neither refreshes the nonempty cache nor evicts that key. A kid-less token signed by the retired key remains in the accepting set for the life of the process. This is precisely the unbounded version of the acceptance-set widening the review brief warned about.

The required overlap acceptance test is absent. `id_token_overlap_both_signers_accepted_unknown_rejected` supplies one static two-key vector and checks an unrelated unknown key; it never transitions `{active, drain}` to `{active}` and proves the retired signer is rejected. The RPC tests likewise use one immutable mock response. Per the stated acceptance bar, this alone blocks the PR.

The fix should make a successful fetch atomically replace an **issuer-scoped** snapshot (or apply an explicit bounded stale policy), refresh it according to a real cache lifetime, and test: both published signers accepted, publisher retires drain, verifier refreshes, retired signer rejected.

### [BLOCK] The change breaks the supported browser/WASM build

The new all-candidate helper calls `tracing::debug!` unconditionally at `crates/hyprstream-rpc/src/auth/jwt.rs:438`, while `hyprstream-rpc` declares `tracing` only under `cfg(not(target_arch = "wasm32"))`. Both CI and the required local gate fail:

```text
error[E0433]: cannot find module or crate `tracing` in this scope
  --> crates/hyprstream-rpc/src/auth/jwt.rs:438:17
```

This is a source regression introduced by the PR, not the known clean-main formatting failure.

### [HIGH] The new candidate set is not scoped by issuer in federated mode

`JwksKeySource` owns one `HashMap<String, CachedKey>` keyed only by `kid` (`key_source.rs:325-336`). After any issuer populates it, `ensure_cached(other_issuer)` sees a nonempty map and does not resolve or fetch the other issuer; `get_keys(other_issuer, None)` returns all keys from the first issuer. The callers derive `issuer` from unverified claims and the EdDSA decoder validates signature/audience but does not bind an expected issuer.

Thus the new API can verify a kid-less token claiming issuer B with a key cached from issuer A, and it can bypass B's resolver altogether. `JwksMode::Federated` is part of this type's supported design even if current in-tree construction is isolated-only. The cache must be keyed by issuer (then by `kid`) and tests must prove candidates never cross issuer boundaries.

### [MEDIUM] The claimed fail-without-fix coverage is false for the unsupported-algorithm test

I temporarily restored first-compatible selection while retaining the new tests. The two ID-token overlap tests failed, as did the candidate-count helper, but `id_token_skips_unsupported_alg_and_accepts_classical` still passed. Its first entry has `kty: "AKP"` and the pre-fix selector already skipped it for an EdDSA header before choosing the following `OKP` key. It therefore does not distinguish this patch from its parent despite the commit claim that every named test fails when reverted.

This test should exercise an entry that the old behavior actually mishandles, and the production callers in `svc.rs` and `factories.rs` should have boundary tests rather than relying only on helper-level coverage.

## What the change breaks or deliberately changes

- **Breaks:** `wasm32-unknown-unknown` compilation for `hyprstream-rpc`.
- **Breaks security retirement:** removed JWKS keys continue to authenticate kid-less RPC JWTs indefinitely after first cache population.
- **Breaks issuer isolation in supported federated mode:** a nonempty cache suppresses resolution/fetch for later issuers and exports prior issuers' candidates.
- **Deliberate compatibility change:** `JwksKeySource::get_key(issuer, None)` now rejects a multi-key set; migrated in-tree try-all callers use `get_keys`, but downstream callers of the public trait method will fail closed during overlap.
- **Deliberate acceptance widening:** while two compatible keys are published, either may authenticate a kid-less token. That is intended by #1184, but only within the publisher's bounded overlap window; the RPC implementation currently makes it unbounded.
- **Deliberate malformed-token tightening:** external ID tokens without a header `alg` are no longer inferred from the JWK and are rejected.

## Comparison with the working reference

The in-tree OAuth reference publishes named drain/active/lead slots and removes drain after `exp + drain_secs`. PR #1213 correctly adopts set-wise verification, but its RPC consumer invents a weaker append-only cache instead of consuming the producer's current bounded set. It therefore does not complete the working rotation pattern.

## Validation

Reviewed detached HEAD `78f39fb6345871ae61e108615d71acb9723b9616` against parent/base `96f3083dcf04a53e41894a3afcf3176558397fc2`.

- `buildq -- cargo test -p hyprstream-rpc --lib jwks_key_source_ -- --nocapture` — PASS, 9 passed.
- Cached-libtorch `buildq -- cargo test -p hyprstream --lib id_token_ -- --nocapture` — PASS, 16 passed.
- `RUSTFLAGS=--cfg=web_sys_unstable_apis buildq -- cargo check --target wasm32-unknown-unknown -p hyprstream-rpc` — FAIL, source `E0433` at `auth/jwt.rs:438`.
- RPC pre-fix singleton mutation — the three named RPC regression tests all FAILED; restored.
- ID-token pre-fix first-compatible mutation — both named overlap tests FAILED, while `id_token_skips_unsupported_alg_and_accepts_classical` PASSED; restored.
- Final `git diff --exit-code` and `git status --porcelain=v1` — clean.

Verdict: **BLOCK**.

ewindisch:
# PR #1213 — sole merge-gate review

**Reviewed head:** `78f39fb6345871ae61e108615d71acb9723b9616`

VERDICT: BLOCK

The PR chooses the correct interop rule in principle: verifying consumers try every compatible
candidate for a kid-less JWT, while the legacy singleton `get_key(None)` API rejects a multi-key
set. Accepting a token signed by either currently published key is the intended widening during an
overlap window. This implementation does not bound that widening to a current, issuer-scoped JWKS
snapshot, and it also breaks the supported browser/WASM build.

## Findings

### BLOCK — the RPC accepting set is append-only, unrefreshed, and not issuer-scoped

`JwksKeySource::fetch_and_cache` inserts fetched Ed25519 keys into one global
`HashMap<kid, CachedKey>` but never clears, retains, or replaces that map
(`crates/hyprstream-rpc/src/auth/key_source.rs:431-490`). `fetched_at` and `soft_ttl` remain unused.
The new kid-less path calls `ensure_cached`, which fetches only while the global map is empty, then
returns every accumulated value (`:552-617`).

This leaves all related prior findings open:

- A key removed from the publisher's JWKS remains accepted by kid-less RPC verification for the
  process lifetime. The intended overlap window is therefore unbounded.
- A newly published key is not fetched for a kid-less token after the first population, so the
  legitimate post-rotation path can fail until an unrelated kid-ful miss or process restart.
- `JwksMode::Federated` shares the same map across issuers. Once issuer A populates it,
  `ensure_cached(B)` sees a nonempty map and `get_keys(B, None)` exports A's keys without resolving
  B. Because callers take `iss` from unverified claims and the EdDSA decoder does not bind an
  expected issuer, a token claiming B can authenticate with A's cached key.
- `get_key(None)` counts accumulated cache entries rather than the current published snapshot, so a
  now-single-key issuer can remain permanently classified as ambiguous.

The required transition test is absent: no test publishes `{active, drain}`, verifies both, refreshes
to `{active}`, and rejects the retired signer. `id_token_overlap_both_signers_accepted_unknown_rejected`
uses one static set and rejects an unrelated unknown key; it does not test retirement.

Required fix: maintain an issuer-scoped snapshot that is atomically replaced after each successful
fetch, apply a real bounded refresh policy (including retry/refetch for kid-less verification), and
test retirement plus cross-issuer isolation.

### BLOCK — the PR breaks `wasm32-unknown-unknown`

`decode_with_any_key_inner` unconditionally calls `tracing::debug!` at
`crates/hyprstream-rpc/src/auth/jwt.rs:438`, but `hyprstream-rpc` declares `tracing` only for
non-WASM targets. The current GitHub **WASM (browser client)** check fails, and the required local
gate reproduced:

```text
error[E0433]: cannot find module or crate `tracing` in this scope
  --> crates/hyprstream-rpc/src/auth/jwt.rs:438:17
```

Guard the log call consistently with the pre-existing WASM-safe calls in the same file.

### MEDIUM — one claimed regression test does not fail without the fix

With ID-token verification temporarily reduced to the first compatible candidate,
`id_token_kid_less_signer_is_non_first_published_key` and
`id_token_overlap_both_signers_accepted_unknown_rejected` failed, but
`id_token_skips_unsupported_alg_and_accepts_classical` still passed. Its leading `AKP` entry was
already skipped by the old same-algorithm selector, so it does not distinguish this patch from its
parent. The commit/PR claim that all six named tests fail without their fix is false.

### MINOR — candidate iteration masks a more specific claims error

Both `decode_with_any_key_inner` and `verify_id_token_with_keys` retain only the last candidate
error. If the real signer verifies the signature but fails expiry/audience, a later wrong key can
replace that result with `InvalidSignature`. This remains deny-by-default, but degrades diagnostics.

## Behavior changes / breakage enumerated

- Broken unintentionally: browser/WASM compilation.
- Broken security bound: retired RPC JWKS keys remain accepted indefinitely.
- Broken availability: newly published kid-less RPC signers may be rejected after initial cache
  population.
- Broken trust isolation in supported federated mode: candidates can cross issuer boundaries.
- Intentional and sound once the cache is fixed: public `get_key(None)` now errors for a true
  multi-key snapshot; migrated in-tree verifying callers use `get_keys`.
- Intentional fail-closed tightening: ID tokens missing/using an unsupported header `alg` are no
  longer inferred from JWK type.
- Preserved: MCP remains strict about an absent audience; the service RPC path remains lenient as
  before. Empty candidate sets and missing key fields fail closed. No new `Option::None`/default
  disables a check.

## Verification

- Current head and reviewed worktree: `78f39fb6345871ae61e108615d71acb9723b9616`;
  final `git diff --exit-code`, `git diff --check`, and `git status --porcelain=v1` are clean.
- Read both prior verdicts under `.fleet-coord/`; the PR head has not changed, and their cache,
  issuer-isolation, WASM, and test-quality findings remain present.
- `buildq --slots 3 -- cargo test -p hyprstream-rpc --lib jwks_key_source_ -- --nocapture`:
  **9 passed**.
- Cached-libtorch `buildq --slots 3 -- cargo test -p hyprstream --lib id_token_ -- --nocapture`:
  **16 passed** (legitimate ID-token/MCP happy paths exercised).
- `buildq --slots 3 -- cargo test -p hyprstream-rpc --lib service::svc -- --nocapture`:
  **23 passed**.
- `RUSTFLAGS=--cfg=web_sys_unstable_apis buildq -- cargo check
  --target wasm32-unknown-unknown -p hyprstream-rpc`: **failed** with source error E0433 above.
- Pre-fix RPC singleton mutation: all three named RPC regressions **failed**; restored.
- Pre-fix ID-token first-candidate mutation: two overlap regressions **failed**, while the claimed
  unsupported-alg regression **passed**; restored.

This is not mergeable at the reviewed head. No PASS is issued, so this review must not queue the PR.

ewindisch:
**Shared BLOCK with #1214 — same finding, same files, will conflict.**

Both PRs edit `auth/jwt.rs`, `key_source.rs`, `auth/mod.rs` and both BLOCK on the identical issue: a JWT `kid` is a hint, not authorization — the verifier accepts a token signed by a *different* published key than its `kid` names (sol proved it live on #1214: `kid-new` token signed by `kid-old`, accepted). #1213's kid-less candidate-trying is the one defensible case, but each candidate must still be a full pinned-algorithm verification, not first-that-parses.

These should be **one** fix on the shared verification path, then the other rebased on top. Fix guidance recorded as invariant I2 in `.fleet-coord/AUTH-INVARIANTS.md`. Holding a consolidated fix lane pending operator call on which PR carries it.

ewindisch:
Consolidated into #1234 on `fix/jwks-consolidated`: https://github.com/hyprstream/hyprstream/pull/1234

The kid-substitution BLOCK is closed: named tokens verify only the exact key their `kid` selects; candidate-trying is restricted to kid-less tokens, with full pinned-algorithm verification per candidate. Validation: hyprstream-rpc auth::jwt 19/19; hyprstream auth 357/357.

Issue #1215, opened 2026-07-23
v0 (#1090 §4.5): HostId migration + HostResolver + ReachAttestation + admit() — Plane A, unblocks #1078


ewindisch:
First rung of the https://github.com/hyprstream/hyprstream/issues/1090 §4.5 phase ladder. **Zero segment machinery.**

**Why this is the priority rung:** https://github.com/hyprstream/hyprstream/pull/1078 landed a capacity-aware inter-host pipeline planner with **no callers** — every public item in `crates/hyprstream/src/runtime/interhost_pipeline.rs` (`InterhostPipelinePlanner`, `MicrobatchSizing`, `PipelineJobRecovery`, `RestartDirective`) is a type nothing constructs. v0 is the consumer that makes it live code rather than a frozen API surface.

## Scope

- **`HostId` migration.**
- **`HostResolver`** — expose the production discovery engine through the rpc client seam (OQ-A4). **Dependency: the rpc-side resolver client surface must be un-gated.**
- **`ReachAttestation`.**
- **The `admit()` step.**
- **Ambient-network fencing defaults** (§3.4).

Planner types stay pure; the runtime segment stays implicit (`seg: None`). No `SegmentId`, no `SegmentManager`, no guest attachment — those are v1.

## Constraints

- **MAC PEP wiring is assumed present** — actively underway, a named v0 dependency rather than an open question. If it turns out not to be present, that is a blocker to report, not something to work around.
- §2/§4-v0 is the **build-now spec**; §3/§5 (segments, RDMA, hardware tiers) is a design track that iterates **without blocking this**. Do not pull segment machinery forward.
- Fail closed on an unresolvable host or a missing/invalid `ReachAttestation` — never admit on absent evidence. That is the #1145 pattern and it has produced a CRITICAL and five BLOCKs in this repo in the last day.

## Acceptance

`#1078`\s planner is actually invoked on a real inter-host path, and a test proves an unattested or unresolvable host is REFUSED rather than admitted.

Refs https://github.com/hyprstream/hyprstream/issues/325, https://github.com/hyprstream/hyprstream/issues/1083, https://github.com/hyprstream/hyprstream/pull/1078

ewindisch:
## Unblocked by restacking, not by waiting

[...]

Issue #1090, opened 2026-07-18
Epic: capability-scoped networking — Segments + Attachments (kernel/runtime reach + zero-trust guest VPC)


ewindisch:
> **Epic.** This is the #1083 design spike (rev 6) promoted to a living epic body. Revise HERE; child tickets spin out of §4.5's phase ladder once the body settles.
> **Provenance:** authored by the fleet spike agent; two independent reviews (Fable, Kimi-k3) incorporated at rev 5; operator ratifications (deterministic SegmentId; dual guest surface virtio-net + 9P; kernel/hw data paths mandatory; MAC assumed live) at rev 6.
> **Unblocks:** #1078 / #325 (planner), #324 (MoQ transport), #327 (failover). Spike issue: #1083.

---

# Design Spike #1083 — General-purpose worker/kernel networking (VPC / virtual networking)

**Status:** Rev 5 — dual independent review (Fable, Kimi-k3) incorporated; for operator review/iteration
**Blocks:** #1078 / #325 (inter-host pipeline planning), #324 (inter-host transport), #327
**Author:** networking spike (fleet), 2026-07-18
**Scope:** design only — no production code, no PR, no worktree changes.

> **Revision note (post-review-1):** the first draft answered only the
> **kernel/runtime plane** (hyprstream host processes reaching each other) and
> treated MoQ+envelopes as the whole answer. Operator feedback corrected two
> things: (1) pipeline stages / workers run in a **mix** of host-process *and*
> sandboxed guests, and (2) general-purpose **container/VM-to-VM IP networking** is
> in scope and must be a **custom, zero-trust, capability-bound overlay —
> namespaced to a capability, not an ambient subnet.** This revision restructures
> the whole doc around two planes unified by one abstraction (**Segment +
> Attachment**). The Plane-A / MoQ material from draft 1 is retained (still correct)
> but is now *one attachment type* under the capability-scoped model, not the
> ceiling.
>
> **Revision note (post-review-2):** operator set the RDMA posture — **Tier 1 with
> MAC** — and named the fleet (RoCE / EFA / DPU / NVIDIA GPUDirect; **no native
> InfiniBand**). RDMA moved from "defer" to an accepted `Endpoint::Rdma` attachment
> whose isolation is a hardware property made explicit via the MAC **Assurance
> axis** (`EnforcementTier`, §3.1a/§4.1), with heterogeneous Tier-1 enforcement per
[...]

PR #1222, opened 2026-07-23
feat: admit verified inter-host pipeline plans

ewindisch:
## Summary

Stacked on #1078 (base `fleet/issue325-interhost-failover`) so the planner contract and its first production consumer land together. Implements #1215, the v0 Plane A rung of #1090, in the #325 inter-host failover track.

- Changes #1078 planner identities from strings/LAN labels to `HostId` (`did:at9p`) plus non-authoritative placement attributes; discovery and admission now establish the identity/reach boundary.
- Ungates the RPC resolver client surface and adds `RpcHostResolver`, `ReachAttestation`, explicit MAC-PEP boundary authorization, and fail-closed `admit()` / `plan_and_admit()` paths. The planner is now invoked on the real admission path.
- Defaults nspawn and OCI workloads to isolated networking.
- Addresses #1078 review findings: exact-epoch recovery acknowledgements, manifest capacity metadata is authoritative when present, and same-epoch completed microbatch IDs cannot be reused (ABA).
- Restores browser currentness dispatch protection during the stacked rebase.

## Validation

- `buildq -- cargo build`
- `buildq -- cargo test -p hyprstream`
- `buildq -- cargo clippy --all-targets -- -D warnings`
- Mutation check: temporarily weakened attestation-expiry validation; the missing/invalid-attestation test failed, then the exact check was restored.

Refs #1215 #1090 #1078 #325

### DIFF
--- a/crates/hyprstream-rpc/src/envelope.rs
+++ b/crates/hyprstream-rpc/src/envelope.rs
@@ -1160,30 +1160,8 @@ pub fn global_pq_store() -> Option<std::sync::Arc<dyn PqTrustStore>> {
     VERIFY_CONFIG.get().and_then(|c| c.pq_store.clone())
 }
 
-#[cfg(not(target_arch = "wasm32"))]
-static BROWSER_CURRENTNESS_VERIFIER: std::sync::OnceLock<
-    std::sync::Arc<dyn crate::browser_provisioning::BrowserCurrentnessVerifier>,
-> = std::sync::OnceLock::new();
-
-/// Install the checkpoint-backed authority used at WebTransport dispatch.
-/// First write wins; there is deliberately no test/global fallback.
-#[cfg(not(target_arch = "wasm32"))]
-pub fn install_browser_currentness_verifier(
-    verifier: std::sync::Arc<dyn crate::browser_provisioning::BrowserCurrentnessVerifier>,
-) -> Result<()> {
-    BROWSER_CURRENTNESS_VERIFIER
-        .set(verifier)
-        .map_err(|_| anyhow!("browser currentness verifier already installed"))
-}
-
-#[cfg(not(target_arch = "wasm32"))]
-pub fn global_browser_currentness_verifier(
-) -> Option<std::sync::Arc<dyn crate::browser_provisioning::BrowserCurrentnessVerifier>> {
-    BROWSER_CURRENTNESS_VERIFIER.get().cloned()
-}
-
-/// Apply the mandatory process-global policy and trust store to an
-/// `UnwrapOptions`.
+/// Apply the process-global verify configuration to an `UnwrapOptions`,
+/// unless it has already been explicitly downgraded/overridden.
 #[cfg(not(target_arch = "wasm32"))]
 pub fn apply_global_verify_config<'a>(
     mut opts: UnwrapOptions<'a>,
--- a/crates/hyprstream-rpc/src/envelope.rs
+++ b/crates/hyprstream-rpc/src/envelope.rs
@@ -1160,8 +1160,30 @@ pub fn global_pq_store() -> Option<std::sync::Arc<dyn PqTrustStore>> {
     VERIFY_CONFIG.get().and_then(|c| c.pq_store.clone())
 }
 
-/// Apply the process-global verify configuration to an `UnwrapOptions`,
-/// unless it has already been explicitly downgraded/overridden.
+#[cfg(not(target_arch = "wasm32"))]
+static BROWSER_CURRENTNESS_VERIFIER: std::sync::OnceLock<
+    std::sync::Arc<dyn crate::browser_provisioning::BrowserCurrentnessVerifier>,
+> = std::sync::OnceLock::new();
+
+/// Install the checkpoint-backed authority used at WebTransport dispatch.
+/// First write wins; there is deliberately no test/global fallback.
+#[cfg(not(target_arch = "wasm32"))]
+pub fn install_browser_currentness_verifier(
+    verifier: std::sync::Arc<dyn crate::browser_provisioning::BrowserCurrentnessVerifier>,
+) -> Result<()> {
+    BROWSER_CURRENTNESS_VERIFIER
+        .set(verifier)
+        .map_err(|_| anyhow!("browser currentness verifier already installed"))
+}
+
+#[cfg(not(target_arch = "wasm32"))]
+pub fn global_browser_currentness_verifier(
+) -> Option<std::sync::Arc<dyn crate::browser_provisioning::BrowserCurrentnessVerifier>> {
+    BROWSER_CURRENTNESS_VERIFIER.get().cloned()
+}
+
+/// Apply the mandatory process-global policy and trust store to an
+/// `UnwrapOptions`.
 #[cfg(not(target_arch = "wasm32"))]
 pub fn apply_global_verify_config<'a>(
     mut opts: UnwrapOptions<'a>,
--- a/crates/hyprstream-rpc/src/lib.rs
+++ b/crates/hyprstream-rpc/src/lib.rs
@@ -253,14 +253,10 @@ pub use hyprstream_rpc_derive::{authorize, service_factory, FromCapnp, ToCapnp};
 
 #[cfg(not(target_arch = "wasm32"))]
 pub use resolver::{
-    NetworkDiscoveryResolver, Resolver, ResolverProfile, ServiceQuery,
-};
-
-#[cfg(all(not(target_arch = "wasm32"), test))]
-pub use resolver::{
     retry_validated_candidates, select_service_candidate, select_service_candidates,
     AcceptedStateEvidence, AnchoredKemRecipient, CandidateDecision, ResolutionEvidence,
-    SelectedService, ServiceCandidate,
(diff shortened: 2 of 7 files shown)


ewindisch:
# PR #1222 frozen-API and admission review

Reviewed head: `3eb1b439f92931dd21ae0acb0f177e008d996324`

VERDICT: BLOCK

## Findings

### BLOCK — #1215's acceptance path is still model-only; no production path enforces admission

The named acceptance criterion is that #1078's planner be invoked on a **real inter-host path** and that the resulting plan cannot run without admission. That does not happen here.

Repository-wide caller search finds no construction or call of `RpcHostResolver`, `plan_and_admit`, the plan-level `admit`, `AdmissionRecord`, or a production `NetworkView` outside `interhost_pipeline.rs`; the only external references to the planner/recovery types are re-exports from `runtime/mod.rs`. The test named `real_admission_path_invokes_planner_and_records_verified_reach` is entirely synthetic: it uses `MockHostResolver` and `MockNetworkView` (`interhost_pipeline.rs:1398-1518`). The only `NetworkView` implementation in the repository is that mock, whose “MAC PEP” is a boolean (`:1438-1443`). No production implementation calls the installed MAC policy machinery, and no runtime consumer requires a live `AdmissionRecord` before weights, transport, or execution.

This leaves the combined #1078/#1222 contract in the same no-caller state #1215 was filed to end. The types are better, but the consumer is absent. Wire `plan_and_admit` into the actual host-process pipeline launch path, provide the real `HostResolver`/reach/MAC-PEP implementation, and test that the launch itself is refused when any planned boundary lacks valid evidence.

### BLOCK — `ReachAttestation` carries epoch and nonce but admission has no expected values for either

`ReachAttestation::verify_for` checks source, target, non-empty transport, wall-clock expiry, descriptor expiry, and the hybrid signature (`interhost_pipeline.rs:699-730`). It never checks `epoch` against the resolved descriptor's `ResolutionEvidence.accepted_state_epoch`, and it never checks `nonce` against an admission-generated challenge. `HostDescriptor.evidence` is otherwise unused by admission.

The positive-control fixture makes the defect concrete: every descriptor has `accepted_state_epoch: 1` (`:1496-1500`), while every accepted attestation is signed with epoch `7` and the fixed nonce `[7; 32]` (`:1459-1469`). The “real admission path” test accepts that mismatch. Thus a previously captured, still-unexpired signature can be replayed after the accepted resolution epoch changes; the fields intended to prove freshness silently disable their checks because no expected epoch/nonce exists. This is the exact “absent expected value disables the check” class the brief called out.

Admission must generate/bind a nonce and compare it, bind the attestation epoch to the current resolved evidence (or another explicit current reach epoch), and add stale-epoch/replayed-nonce negative tests plus a matching legitimate happy path.

### HIGH — the claimed production `RpcHostResolver` cannot resolve a legitimate multi-host plan

`RpcHostResolver` stores one fixed `service_name` (`interhost_pipeline.rs:567-582`). Every `resolve_host(host, ...)` call submits the same service query (`:593-600`) and only afterward requires the single resolved DID to equal the requested host (`:602-606`). One resolver instance is passed across all candidates/stages by `plan_and_admit` and `admit`.

For a real inter-host plan containing two distinct `HostId`s, a deterministic service resolver returns the same service authority for both identical queries, so at most one host can pass and the legitimate second host is refused. The tests hide this outage behind the map-backed mock resolver. Resolution must be keyed by the requested `HostId` (or otherwise expose a host-indexed production query), and the production resolver needs a two-distinct-host positive-control test.

### HIGH — the current stacked PR merge surface is not the reviewed seven-file feature commit

The intended head commit changes seven files (723 additions, 83 deletions) and contains no `SegmentId`, `SegmentManager`, membership directory, guest attachment, or RDMA machinery. That part respects v0's Plane-A scope.

However, GitHub currently reports 166 changed files and 23,731 additions / 2,544 deletions from the PR base, because `base..head` contains 82 commits. Merging the PR as currently stacked also advances #1078's branch through a large unrelated mainline delta. Restack/update the base so the merge surface is the reviewed feature change before this can be merge-ready.

## Contract and folded #1078 findings

- The planner identity migration is real: stage/capacity identities are `HostId(Did)`, planning rejects non-`did:at9p` identities, `lan_id` is deleted, and affinity/bandwidth remain non-authoritative attributes.
- Missing/invalid shard `metadata.total_size` now fails closed when the manifest exists.
- Recovery control calls carry exact expected epochs.
- Completed microbatch IDs are retained for the epoch, preventing same-epoch ABA completion.
- The intended head commit contains zero segment/guest/RDMA machinery.
- These corrections make the pure types more usable, but the combined contract is not yet consumable as a live inter-host boundary because the production admission path is absent and the provided production resolver cannot resolve two hosts.

## What the change breaks

- **Current pipeline runtime behavior:** no new pipeline behavior is activated; there is still no production caller. This is the principal acceptance failure.
- **Legitimate multi-host resolution:** the supplied `RpcHostResolver` refuses the second distinct host for a fixed service query.
- **Public API compatibility:** `String` host IDs become `HostId`, `HostCapacity` moves VRAM into `HostAttributes`, `lan_id` is removed, and recovery control methods now require an epoch. There are no in-repo callers, but downstream callers must migrate.
- **Checkpoint compatibility:** an existing shard index without a positive integer `metadata.total_size` is now rejected instead of analytically estimated. This is an intentional fail-closed break.
- **Microbatch compatibility:** a completed ID cannot be reused within the same epoch. This is an intentional correctness break.
- **Guest networking defaults:** nspawn now defaults to a private loopback-only namespace and OCI to `--network=none`; workloads that relied on implicit veth/pasta/slirp connectivity lose network access until explicitly opted in. This is the intended ambient-network fence, but it is operationally breaking.
- **Stack merge surface:** without restacking, merging #1222 into #1078 also carries the 82-commit unrelated base delta described above.

## Validation

- Clean worktree; head pinned before and after review to `3eb1b439f92931dd21ae0acb0f177e008d996324`.
- With the cached CUDA 13.0 libtorch environment, `buildq -- cargo test -p hyprstream interhost_pipeline --lib -- --nocapture`: **19 passed, 0 failed, 1,100 filtered**.
- `buildq -- cargo test -p hyprstream-rpc resolver --lib -- --nocapture`: **34 passed, 0 failed, 741 filtered**.
- `buildq -- cargo test -p hyprstream-workers --features nspawn,oci --lib -- --nocapture`: **175 passed, 0 failed**.
- The first focused attempt omitted `LIBTORCH` and failed in `torch-sys`; that was an environment-selection failure, not a source diagnostic. The identical queued command passed with the repository's cached libtorch.
- Mutation evidence, all through `buildq`, then byte-for-byte restored:
  - weakened attestation-expiry validation: `missing_or_invalid_reach_attestation_refuses_admission` failed;
  - weakened authoritative manifest handling: `manifest_without_authoritative_total_size_is_refused` failed;
  - weakened replacement epoch fencing: `delayed_recovery_control_messages_are_epoch_fenced` failed;
  - removed completed-ID retention: `same_epoch_microbatch_id_reuse_is_refused_to_prevent_aba_completion` failed.
- After restoration, the source blob matches `HEAD` (`88aad8584dd8bee138c3924f7965321c648a597b`) and `git diff --exit-code` / `git status --short` are clean.
- `cargo fmt --check` was not run, per the brief's clean-main #1140 exception.

ewindisch:
Update at `141a80b71` (restacked directly on `fleet/issue325-interhost-failover`; the merge surface is now 4 files / 879 additions, not the prior unrelated mainline delta).

**Epoch + nonce** — fixed. Admission now generates a fresh mandatory `ReachChallenge` per boundary, passes it to `NetworkView::reachable`, binds its epoch to the target descriptor's accepted-state epoch, and rejects either mismatch before accepting the signature. Added positive and stale-epoch/replayed-nonce refusal coverage. I temporarily removed both comparisons; `stale_epoch_and_replayed_nonce_refuse_admission` failed, then restored them.

**Multi-host RPC resolution** — fixed. `RpcHostResolver` now requires an explicit `HostId -> service record` mapping and resolves each requested host through its own record. Added a two-distinct-host positive-control test that verifies both identities and both queries.

**Production launch-path finding** — precisely confirmed, but not falsely claimed as fixed: no inter-host host-process activation path exists in this stack. `ModelService::load_model_inner` only spawns one local `InferenceService`; its `RouteDecision::Remote` arm explicitly falls back to the local client. The current replica state exposes opaque `ReplicaId`s, not `HostId`/capacity candidates; there is no production `NetworkView`, no reach-attestation RPC handler, and no stage transport/executor. The interhost module itself documents activation as #324 and continuous stage scheduling as #329. Without those production components there is no safe caller for `plan_and_admit`; adding a wrapper here would remain model-only. This PR therefore still needs that host-process launch boundary before the #1215 acceptance criterion can be closed.

Validation: pre-commit `buildq -- cargo clippy --workspace --all-targets --release -- -D warnings` passed (5m51s); `rustfmt` was applied only to edited files. `cargo fmt --check` was not run because the known clean-main #1140 failure remains.

ewindisch:
Follow-up validation on `141a80b71`: `buildq -- cargo test -p hyprstream interhost_pipeline --lib -- --nocapture` passed — 21 passed, 0 failed. There are no actionable CodeRabbit review threads; its only PR message is the automated skipped-review notice.

PR #1226, opened 2026-07-23
feat(mac): enforce labels on every 9P object access

ewindisch:
## Summary

- activate a mandatory resolver-backed MAC decider on the shared 9P translator path
- retain each fid's absolute walked path so every attach, walk, open, read, getattr, readdir, and write reaches the PEP with the exact object reference
- deny unlabeled/unresolvable objects, apply SecurityContext::can_access dominance to read-class actions, and fail every write closed
- record every decision through the existing hash-chained MAC audit sink; audit failure downgrades permits to denial
- remove the permissive translator default; lower layers without the concrete Hyprstream resolver explicitly deny

Refs #699
Refs #547

## Authoritative PEP placement

The 9P translator seam is the single authoritative PEP. Every supported carrier (TCP, UDS, hybrid/raw vsock, WebSocket, and WebTransport) converges on the same translator dispatch before either MountBackend or an RPC Backend sees an object operation. Placing the check here covers both backend families and preserves the walked fid path needed by ObjectLabelResolver.

The VFS Mount trait is deliberately not a second PEP. Duplicating enforcement at Mount would create two independently evolving decisions, and the more permissive one would eventually become a bypass. Mount continues to thread Subject for its existing policy responsibilities, while the translator owns MAC mediation.

## Write direction

Writes fail closed because the IFC direction has not been ratified. Recommendation: adopt Bell-LaPadula for confidentiality-bearing labels (no write down: the target must dominate the subject's effective label), while treating integrity/Biba as a separate lattice or explicitly composed policy rather than reversing the same confidentiality order. Until that decision is approved, no 9P write is permitted.

## Validation

- buildq -- cargo test -p hyprstream mac::pep::tests --lib
- buildq -- cargo test -p hyprstream-9p --lib --tests
- buildq -- cargo build
- buildq -- cargo clippy --all-targets -- -D warnings
- foreground pre-commit release Clippy gate

Mutation check: replaced the intrinsic dominance result with unconditional permit; the insufficient-clearance assertion failed. Restored SecurityContext::can_access and reran the focused test successfully.

### DIFF
--- a/crates/hyprstream-9p/examples/serve_mount_uds.rs
+++ b/crates/hyprstream-9p/examples/serve_mount_uds.rs
@@ -44,5 +44,11 @@ async fn main() -> anyhow::Result<()> {
     let subject = Subject::new("tenant-a");
 
     eprintln!("serving 9P mount at {path}");
-    serve_mount_uds(mount, subject, &path).await
+    serve_mount_uds(
+        mount,
+        subject,
+        Arc::new(hyprstream_9p::DenyAllDecider),
+        &path,
+    )
+    .await
 }
--- a/crates/hyprstream-9p/src/lib.rs
+++ b/crates/hyprstream-9p/src/lib.rs
@@ -46,11 +46,9 @@ pub mod memory;
 // The browser/wasm build reaches the backend through the DMA/Wanix client path.
 #[cfg(not(target_arch = "wasm32"))]
 pub mod translator;
-// 9P reference-monitor contract (S2 / #568): the attach-time token seam, the
-// per-op mediation composition (`ReferenceMonitor`), and fail-closed defaults.
-// Dormant groundwork — `Translator` runs without a monitor unless the
-// application installs one (activation is blocked on #698). Native-only
-// alongside `translator`, its only consumer today.
+// 9P reference-monitor contract (S2 / #568): attach-time token verification,
+// optional full token/IFC composition, and the mandatory per-op access-decider
+// seam. Native-only alongside `translator`, its only consumer today.
 #[cfg(not(target_arch = "wasm32"))]
 pub mod mac_seam;
 // `MountBackend` bridges a VFS `Mount` (+ `Subject`) to the `Backend` seam so
--- a/crates/hyprstream-9p/src/mac_seam.rs
+++ b/crates/hyprstream-9p/src/mac_seam.rs
@@ -101,6 +101,7 @@ pub use hyprstream_rpc::auth::mac::{ObjectLabelResolver, ObjectRef};
 /// deliberately not mediated.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub enum Action {
+    Attach,
     Walk,
     Open,
     Read,
@@ -304,28 +305,31 @@ impl ObjectLabelResolver for DenyUnlabeledResolver {
     }
 }
 
-/// The policy decision point behind the mandatory token and IFC gates.
+/// Authorizes one operation against a subject's cached [`SecurityContext`] and
+/// exact walked object reference.
 ///
-/// `check` is called only after the monitor has resolved a non-optional
-/// object label, checked the attached token's operation/ceiling/TTL, and
-/// enforced `subject.ctx ⊒ object.label`. The concrete application
-/// implementation is a local AVC lookup (with a PDP miss); it must not perform
-/// UCAN chain validation or Casbin matching per operation.
+/// Resolution belongs to the concrete implementation in `hyprstream`, which
+/// owns the production content-truth label resolver and tamper-evident audit
+/// sink. Passing the object reference—not an optional caller-supplied
+/// label—keeps label resolution and auditing inside the authoritative PEP.
 pub trait AccessDecider: Send + Sync {
-    fn check(&self, ctx: &SecurityContext, object_label: &SecurityLabel, action: Action) -> bool;
+    fn check(&self, ctx: &SecurityContext, object: ObjectRef<'_>, action: Action) -> bool;
 }
 
-/// The fail-closed default policy monitor. It permits nothing.
+/// Fail-closed fallback for layers that cannot construct the production
+/// resolver-backed, audited decider.
 #[derive(Debug, Default, Clone, Copy)]
 pub struct DenyAllDecider;
 
 impl AccessDecider for DenyAllDecider {
-    fn check(
-        &self,
-        _ctx: &SecurityContext,
-        _object_label: &SecurityLabel,
-        _action: Action,
-    ) -> bool {
+    fn check(&self, ctx: &SecurityContext, object: ObjectRef<'_>, action: Action) -> bool {
+        tracing::warn!(
+            target: "hyprstream.mac.audit",
+            ?object,
+            ?action,
+            clearance = ?ctx.clearance(),
+            "9P access denied: production MAC decider unavailable"
+        );
         false
     }
 }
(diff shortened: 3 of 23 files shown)


ewindisch:
VERDICT: REVISE

Reviewed head `6cfcb53c21292f033c642658b0d328bd8ac18842`.

## Findings

1. **[P0] The production Wanix and Kata 9P exports are dead endpoints, and their denials bypass `mac::audit`.** `inject_9p_socket` installs `DenyAllDecider` unconditionally (`crates/hyprstream-workers/src/runtime/wanix_workload.rs:318-322`), as does the Kata tenant-VFS vsock server (`crates/hyprstream-workers/src/runtime/kata_backend.rs:1260-1266`). That decider only emits a tracing warning and returns `false`; it has no `AuditSink` and creates no hash-chained `AuditRecord` (`crates/hyprstream-9p/src/mac_seam.rs:100-115`). I exercised the legitimate production path by temporarily adding a test that passed a real tenant `Mount` and non-anonymous `Subject` to `inject_9p_socket`, then connected with `Remote9pMount::connect_uds`. `Tversion` succeeded and `Tattach` failed with `errno 1`. The existing “live server” test stops after `Tversion` (`wanix_workload.rs:744-783`), so it masks the outage. This breaks OCI/nspawn Wanix namespace import and Kata guest tenant-VFS attach/read, while leaving no MAC audit record for the denial. Inject the concrete resolver-backed audited decider through these construction paths (or fail server construction); do not advertise/start an endpoint that denies every legitimate client.

2. **[P0] Verified mount-ticket identity and clearance never reach the authoritative PEP.** All three translator constructors still install `AnonymousAuthenticator` (`crates/hyprstream-9p/src/translator.rs:157-165,188-203,219-239`), and the only `with_authenticator` callers are tests. `handle_attach` therefore evaluates the root and caches every fid as `Public/Unverified`, even after the backend has validated a WebTransport ticket (`translator.rs:454-479`). The WS route likewise verifies a ticket to a `Subject` and passes that only to `MountBackend`; it never supplies the PEP's `SecurityContext` (`crates/hyprstream/src/server/routes/ninep.rs:276-292`). Consequently an enrolled, PQ-verified, cleared subject is indistinguishable from anonymous to MAC: exact Public structural paths can pass, while every Internal/Confidential object is denied regardless of the caller's real clearance, and audit records attribute the anonymous floor rather than the verified principal. Make attach-time verification return/cache the authority-derived `SecurityContext` used by the single translator PEP; a separate VFS `Subject` check is not a substitute.

3. **[P1] Activation makes an explicitly `Unverified` production assurance fallback authoritative.** `SitePolicy::conservative` labels Public, Internal, and Confidential nodes with `Assurance::Unverified`, with the comment that requiring `PqHybrid` would make objects unreachable before assurance binding is wired (`crates/hyprstream/src/mac/genesis.rs:236-258`). The PEP now actively consumes those labels while the production translator supplies an `Unverified` subject. That is the forbidden compatibility softening for this review: PQ-hybrid is mandatory, and missing assurance wiring must fail closed rather than lower the requirement so the anonymous default can pass. Wire verified hybrid key material into the attach context and remove the production `Unverified` policy fallback.

## Authoritative PEP and breakage

The PR correctly names the 9P translator seam as the single authoritative PEP; I found no second MAC decision added to `Mount`. Unresolvable labels deny, and writes fail closed with `WriteDirectionUndecided`, which is the right behavior while write direction is unratified.

The PR description still needs to state the actual outages plainly:

- OCI/nspawn Wanix Direction A and Kata tenant-VFS cannot attach at all because their production decider is deny-all.
- WS/WT ticket holders are evaluated as anonymous by MAC; only exact Public genesis nodes can work.
- Every write is denied intentionally.
- Every CAS CID is denied because production uses `NoManifests`; PR #1224 is still open and absent.
- Namespace carrier PR #1228 is still open and absent, so non-genesis/dynamic namespace objects remain unresolvable and denied.

This is fail-closed, but it is not a working legitimate path and must not be described only as “lower layers explicitly deny.”

## Validation

All Cargo commands ran through `buildq -- cargo ...`. The first Hyprstream PEP attempt stopped before source compilation because the build slot lacked `LIBTORCH`; rerunning with the cached AppImage libtorch, `OPENSSL_NO_VENDOR=1`, and `LIBTORCH_BYPASS_VERSION_CHECK=1` passed.

- `cargo test -p hyprstream-9p --lib --tests --no-fail-fast`: 30 passed.
- `cargo test -p hyprstream mac::pep::tests --lib`: 2 passed after final restoration.
- `cargo test -p hyprstream resolver_cid_unlabeled_manifest_is_denied --lib`: 1 passed.
- `cargo test -p hyprstream-workers runtime::wanix_workload::tests --lib`: 7 passed, but the production server test exercises only `Tversion`.
- `cargo build -p hyprstream-9p -p hyprstream-workers -p hyprstream`: passed.

Mutation checks, each restored afterward:

- `DenyAllDecider: false -> true` made `fallback_decider_denies_every_action` fail.
- Restoring present-but-unlabeled manifest fallback to `Some(floor)` made `resolver_cid_unlabeled_manifest_is_denied` fail.
- Replacing `ctx.can_access(&label)` with unconditional permit made `reads_fail_closed_and_every_decision_is_audited` fail on insufficient clearance.
- Removing the undecided-write deny branch made `writes_fail_closed_and_are_audited` fail.
- Bypassing `AuditSink::record` made `reads_fail_closed_and_every_decision_is_audited` fail with 0 records instead of 4.

Final worktree status is clean at the reviewed SHA.

ewindisch:
BLOCK — `mac/pep.rs:42-44,91-128` reads the compiled policy only for audit metadata and permits read-class ops solely via `ctx.can_access(label)` (no signed-attenuation/Casbin-compiled TE/IFC join), while `hyprstream-9p/src/translator.rs:155-177,194-242,454-479` leaves production attach MAC context on `AnonymousAuthenticator` rather than a verified credential (though cached); durable labels fail closed and all object ops are gated, and `buildq -- cargo test -p hyprstream mac::pep` (2 pass) plus `-p hyprstream-9p` (30 pass) pass.

ewindisch:
P0 verdict: FIXED in `701c8ea7c`.

- Removed unconditional `DenyAllDecider` construction from Wanix UDS and Kata vsock 9P servers. Both paths now require an injected `Arc<dyn AccessDecider>`.
- Added production construction of the resolver-backed `NinePAccessDecider` with a mandatory Hybrid-capable signer and `WalAuditStore` (`mac-audit/ninep-worker`); worker/backend construction fails if the audited PEP cannot be built.
- Threaded the mandatory decider through `BackendCtx`, `KataBackend`, `inject_9p_socket`, and `prepare_wanix_workload`.
- Extended the live Wanix server regression from `Tversion/Rversion` through policy-controlled `Tattach/Rattach`, so a deny-all outage fails the test.

Validation via `buildq`:
- `cargo test -p hyprstream-workers runtime::wanix_workload::tests --lib -- --nocapture` — 7 passed.
- `cargo check -p hyprstream-workers --all-targets --features kata` — passed.
- `cargo check -p hyprstream --lib --bins` — passed.
- `cargo clippy -p hyprstream-workers --all-targets --features kata -- -D warnings` — passed.
- Repository pre-commit release-profile Clippy gate — passed.

ewindisch:
VERDICT: APPROVE — prior P0 is FIXED at 701c8ea7: Wanix and Kata inject the resolver-backed audited 9P decider or fail construction, legitimate Tattach reaches Rattach, and denied decisions flow to the hash-chained WAL.

Issue #569, opened 2026-06-29
feat(security) S3: Scope-vocabulary unification + baseline TE policy in schema


ewindisch:
**Epic:** #547. **Touches** #539 (annotation/codegen seam). Schema breakage in scope.

## Goal
Rename `mcpScope` → `$scope`/`$capability`, unify it with the structured `Scope` vocabulary, make it
**mandatory**, and treat the `$scope`/TE annotations as the **compile-time baseline TE policy + node
labels**.

## Scope
- Rename the capnp annotation; **unify the action vocabulary** (the `ScopeAction` enum ↔ the `Scope`
  action strings — today they are two parallel, divergent systems).
- **Mandatory**: a method with no scope = **build error** (no silently-public files).
- The `$scope`/TE annotations carry: the node's **object-type/label** + the `(subject-type, op)` it permits
  = the baseline type-enforcement relations (the "policy module" shipped with the schema).
- **Compile-time guards**: a `query`-scope method must not map to a `ctl`; a `write|manage` method must not
  map to a read-only file (9p read = side-effect-free). Contradiction = build error.

## Dependencies / relates
#547 · #539 (the generated-mount + annotation codegen this extends).

## Acceptance
- `mcpScope` renamed and unified with `Scope`; missing scope = build error.
- `$scope`/TE annotations compile to the baseline TE policy + node labels.
- Read/write-vs-scope contradictions fail the build.

ewindisch:
**Review pass complete (commit `c553fd497`).** Merged the latest `feature/security-mac-ucan` tip (S1 #598, S5 milestones #594/#595/#590) into this branch for an up-to-date review and pushed.

**Verified against acceptance criteria:**
1. ✅ `mcpScope` fully renamed to `$scope`/`$capability`, unified with the single `ScopeAction` vocabulary — confirmed zero remaining `mcpScope` references in any `.capnp` file; `annotations.capnp`'s doc comment states "there is no second, parallel vocabulary."
2. ✅ Mandatory scope = build error, genuinely enforced (not just documented) — `validate_mandatory_scope` in `cgr_reader.rs` recurses correctly through scoped-client dispatchers and nested dispatchers, with `$scopeExempt` as the sole audited escape hatch. 8 unit tests cover the dispatcher/nested-dispatcher/exempt cases.
[...]

Issue #556, opened 2026-06-29
S5: Fail-closed mandatory-hybrid posture (remove classical floor + escape hatch)


ewindisch:
Part of #550. Relates to #547, #441.

**S5 — Fail-closed mandatory-hybrid posture (remove the classical floor).**

> **Defensive context (no safeguard removed).** Every change below *strengthens* the fail-closed posture by removing an in-band downgrade/fallback path so the mandatory hybrid suite cannot be silently weakened at runtime. No production security control is disabled, no authentication is bypassed, and no offensive capability is requested; the `CryptoPolicy::Classical` path is retained only as an explicitly gated test-only mode. The sequencing below ensures key distribution precedes the flip so no live peer is locked out.

The current posture is deliberately migration-friendly (WNS, per-identity `require_pq`, classical fallback for unanchored signers, `HYPRSTREAM_ENVELOPE_POLICY=classical` escape hatch). Greenfield + the #547 crypto floor ("hybrid PQ, never Ed25519-only, no in-band downgrade") → make hybrid mandatory and fail-closed.

### Deliverables
- Remove the per-identity `require_pq = uses_pq() && anchored.is_some()` downgrade — require both anchored PQ **sign and KEM** keys for every peer; unanchored → reject (not classical-floor).
- Remove the `HYPRSTREAM_ENVELOPE_POLICY=classical` escape hatch and the `CryptoPolicy::Classical` runtime path (or gate it to test-only).
- Crypto-agility becomes a **policy-selected pinned suite id** with **no in-band downgrade** — a peer supports the pinned suite fully or the handshake fails.
- Update `docs/cryptography-architecture.md` (the "Verification rule — per-identity, NOT blanket fail-closed" section and the "Key Exchange" section, which currently lists only classical Ristretto/P-256).

### Risk / ordering
- Land AFTER S0–S4 + S1 key distribution so every in-process + mesh peer already has anchored KEM keys; otherwise fresh installs lock themselves out (the exact regression the WNS floor was added to prevent — see #277). Sequence: distribution first, then flip fail-closed.

## Carrier-profile downgrade rule

Mandatory hybrid posture applies to the application security layer on every untrusted carrier. A standards-compatible classical MOQT relay hop is permitted only under the explicitly selected `standard-public-relay` profile and only when end-to-end application hybrid KEM, identity/current-state admission, and payload AEAD are already mandatory.

### Additional acceptance criteria

[...]

PR #1228, opened 2026-07-23
mac: namespace label carriers — (a) schema-derived + (c) bind-time (#699)

### LINKED ISSUE #699: feat(mac): object-label resolution — content-bound SecurityLabels for the 9P/VFS surface (MAC activation prereq)

ewindisch:
## What

Two of the three object-label carriers the 9P/VFS namespace needs before MAC enforcement can turn on. Closes #699 scope item 1 · epic #547 · relates #569 (S3 baseline).

The third carrier (**b** — the `security_label` field on the content-addressed CAS manifest) is a sibling lane (`.worktrees/mac-ingest-label`); `storage/cas/` is **not touched** here. The three carriers meet at `ObjectLabelResolver`, keyed on the walked path/CID — the same `Tread` resolving differently per object is exactly why schema annotation alone (carrier a) cannot cover CAS (carrier b).

A sibling lane is turning enforcement ON; anything left unlabeled becomes inaccessible. **This PR delivers the label carriers and their resolution, not the 9P PEP** (`hyprstream-9p` `AccessDecider` still defaults to `AllowAll` — wiring the resolver into the decider is the enforcement lane).

## Carrier (a) — schema-derived labels on generated mount nodes (static `/srv/*`)

The label is a property of the node's TYPE, derived from a new `$vfsMac` schema annotation at codegen — the same path `$vfsKind`/`$vfsPath` already flow:

- `annotations.capnp`: `annotation vfsMac(field) :Text`
- `VfsNode.mac` + `VfsNode::mac_label() -> Option<SecurityLabel>` (decoder; format `level[:assurance[:compBits]]`)
- codegen (`cgr_reader` → `types::UnionVariant` → `vfs.rs`) captures and emits it onto the generated `VfsNode` table

A node whose schema declares no `$vfsMac` decodes to `None` ⇒ **unlabeled ⇒ deny ⇒ a genesis-coverage finding** (`GeneratedNodeCoverage`). No permissive default; absence is denial.

## Carrier (c) — bind-time labels for synthetic VFS mounts, via the previously-unused `bind_time_label`

New `BindLabelMap` (`hyprstream-rpc::auth::mac::bind`):

- `bind(binder_floor, declared)` — both labels are `SecurityLabel` (not `Option`); **there is no label-less bind** (a mount must carry a label). Effective label = `bind_time_label` = the restrict-only join, always `⊒ binder_floor` and `⊒ declared`.
- `resolve(components)` — nearest enclosing mount's effective label, the floor descendants clamp to. Unbound ⇒ `None` ⇒ deny.
- `clamp_descendant(mount_floor, node_label)` — restrict-only join; a node declaring a lower label cannot lower below the mount floor.

## No escape hatch

Neither carrier has one: no `Option` whose `None` means skip, no `NotApplicable`, no default that turns a check off, no feature flag. `None` always means **deny**, never skip.

## Tests — every behavior has one that FAILS without it (revert-verified)

| behavior | test | verified by reverting |
|---|---|---|
| clamp restrict-only | `clamp_descendant_is_restrict_only` | returning `node_label` (no join) → left Public / right Secret, fails ✅ |
| empty `$vfsMac` ⇒ deny | `mac_label_empty_is_unlabeled_deny` | defaulting empty to `Some(bottom())` → fails ✅ |
| codegen emits `mac` | `vfs_mac_annotation_flows_onto_emitted_node` | dropping the field from the literal → fails ✅ |
| generated-node coverage | `unannotated_generated_node_is_a_finding_not_a_default` | — |
| mount must carry a label | `unbound_prefix_resolves_to_none_deny` + type-level `bind` signature | — |
| descendant clamps to mount floor | `descendant_inherits_mount_effective_label` | — |
| effective never below binder floor | `effective_label_never_below_binder_floor` | — |

`VERIFY`: `buildq -- cargo build -p hyprstream`; carrier tests in `hyprstream-rpc` (`auth::mac::bind`, `_metadata`), `hyprstream-rpc-derive` (`codegen::vfs`), `hyprstream` (`mac::genesis`); `buildq clippy --all-targets -D warnings` on all four crates (and the pre-commit full-workspace release clippy passed). `cargo fmt --check` fails on a clean main (#1140) — not introduced here.

---

## Namespace inventory — every object class reachable through the 9P/VFS surface

| # | object class | examples | carrier | label source today | status |
|---|---|---|---|---|---|
| 1 | structural export-root container dirs | `/srv`, `/worktree`, `/stream` | (a) static | `SitePolicy::conservative` → **Public** (listing only; its label **does not** inherit to descendants — anti-laundering rule in `CompositeObjectLabelResolver`) | labeled |
| 2 | generated service mount nodes (schema-backed) | `/srv/policy`, `/srv/oauth`, `/srv/ledger`, `/srv/model`, `/srv/registry`, `/srv/mcp`, `/srv/tui`, `/srv/discovery`, `/srv/metrics`, `/srv/inference` | (a) schema-derived | `SitePolicy` table: control-plane (`policy`/`oauth`/`ledger`) → **Confidential**; data services → **Internal**. **Carrier (a) mechanism now lets the schema `$vfsMac` be the source instead of the site table.** | labeled (site table); schema-annotation content is the S3/#569 follow-up |
| 3 | generated method nodes under a service | `/srv/model/{name}/status`, `…/ctl`, `…/health` | (a) per-node `$vfsMac` on the service's `vfs_nodes()` table | none annotated yet → `mac=""` → `None` | ⚠️ **finding** (carrier delivered; annotation content TBD) |
| 4 | services with **no** schema / no generated VfsNode table | `/srv/event`, `/srv/worker`, `/srv/oai`, `/srv/xet`, `/srv/flight`, `/srv/streams` | (a) — none emitted | no table → no node → unlabeled | ⚠️ **finding** (needs a schema or an explicit site-policy rule) |
| 5 | standard non-service synthetic mounts | `/bin`, `/env`, `/lang/tcl` | **(c) bind-time** | `BindLabelMap.bind(binder_floor, declared)` — floor × declared | ⚠️ **finding** until bound through the labeled-bind surface |
| 6 | subject-composed synthetic mounts | any mount a subject binds into its namespace | **(c) bind-time** | `bind_time_label(binder_floor, declared)` | carrier delivered; enforcement lane wires it at bind |
| 7 | content-addressed data objects (CAS blobs/manifests) | `/worktree/...` blob content | **(b) CAS manifest `security_label`** | the manifest field | 🚫 **sibling lane** (`mac-ingest-label`); untouched here |
| 8 | stream pipes | `/stream/{...}` (MoQ plane) | TBD | — | ⚠️ **finding** (stream-plane labeling is a separate scope) |
| 9 | browser doc mounts | `/srv/{svc}/doc` | (a) structural | Public | labeled (structural) |

**Findings to close before enforcement turns on** (reported, not defaulted): rows 3, 4, 5, 8. None are defaulted to a permissive label — each is `None` ⇒ deny ⇒ surfaced by the genesis coverage gate.

### DIFF
--- a/crates/hyprstream-rpc-build/src/schema/cgr_reader.rs
+++ b/crates/hyprstream-rpc-build/src/schema/cgr_reader.rs
@@ -727,6 +727,7 @@ fn extract_union_variants(
         let vfs_kind_id = annotation_id_by_short_name(node_map, "vfsKind");
         let vfs_bulk_id = annotation_id_by_short_name(node_map, "vfsBulk");
         let vfs_hidden_id = annotation_id_by_short_name(node_map, "vfsHidden");
+        let vfs_mac_id = annotation_id_by_short_name(node_map, "vfsMac");
         let vfs_path = extract_annotation_text(
             field.get_annotations().map_err(|e| format!("{e}"))?,
             vfs_path_id,
@@ -745,6 +746,12 @@ fn extract_union_variants(
             field.get_annotations().map_err(|e| format!("{e}"))?,
             vfs_hidden_id,
         );
+        // $vfsMac (#699 carrier (a)): the MAC label this generated node carries.
+        // Empty when absent ⇒ the node is unlabeled ⇒ a genesis finding.
+        let vfs_mac = extract_annotation_text(
+            field.get_annotations().map_err(|e| format!("{e}"))?,
+            vfs_mac_id,
+        );
 
         variants.push(UnionVariant {
             name,
@@ -758,6 +765,7 @@ fn extract_union_variants(
             vfs_kind,
             vfs_bulk,
             vfs_hidden,
+            vfs_mac,
         });
     }
 
@@ -1656,6 +1664,7 @@ mod mandatory_scope_tests {
             vfs_kind: String::new(),
             vfs_bulk: false,
             vfs_hidden: false,
+            vfs_mac: String::new(),
         }
     }
 
--- a/crates/hyprstream-rpc-build/src/schema/types.rs
+++ b/crates/hyprstream-rpc-build/src/schema/types.rs
@@ -53,6 +53,12 @@ pub struct UnionVariant {
     /// `$vfsHidden`: exclude this method from the generated mount.
     #[serde(default)]
     pub vfs_hidden: bool,
+    /// `$vfsMac` annotation text — the MAC security label this generated node
+    /// carries (#699 carrier (a); a property of the node's type, derived at
+    /// generation). Empty = unset ⇒ the node is unlabeled ⇒ deny ⇒ a
+    /// genesis-coverage finding (no permissive default).
+    #[serde(default)]
+    pub vfs_mac: String,
 }
 
 /// Payload carried by a tagged-union arm.
@@ -292,6 +298,7 @@ mod vfs_metadata_tests {
             vfs_kind: "file".into(),
             vfs_bulk: false,
             vfs_hidden: false,
+            vfs_mac: String::new(),
         };
         let json = serde_json::to_string(&v).expect("serialize");
         let back: UnionVariant = serde_json::from_str(&json).expect("deserialize");
--- a/crates/hyprstream-rpc-build/src/ts_codegen/client.rs
+++ b/crates/hyprstream-rpc-build/src/ts_codegen/client.rs
@@ -493,6 +493,7 @@ mod browser_binding_tests {
             vfs_kind: String::new(),
             vfs_bulk: false,
             vfs_hidden: false,
+            vfs_mac: String::new(),
         }
     }
 
--- a/crates/hyprstream-rpc-build/src/ts_codegen/parsers.rs
+++ b/crates/hyprstream-rpc-build/src/ts_codegen/parsers.rs
@@ -1686,6 +1686,7 @@ struct EmbedImagesResponse {
                 vfs_kind: String::new(),
                 vfs_bulk: false,
                 vfs_hidden: false,
+                vfs_mac: String::new(),
             }],
             structs: vec![],
             scoped_clients: vec![],
--- a/crates/hyprstream-rpc-derive/src/codegen/handler.rs
+++ b/crates/hyprstream-rpc-derive/src/codegen/handler.rs
@@ -2413,6 +2413,7 @@ mod browser_method_dispatch_tests {
             vfs_kind: String::new(),
             vfs_bulk: false,
             vfs_hidden: false,
+            vfs_mac: String::new(),
         }
     }
 
--- a/crates/hyprstream-rpc-derive/src/codegen/vfs.rs
+++ b/crates/hyprstream-rpc-derive/src/codegen/vfs.rs
@@ -403,6 +403,21 @@ pub fn generate_mount(service_name: &str, resolved: &ResolvedSchema) -> TokenStr
             (#service_name, NODES)
         }
 
+        // #699 carrier (a) inventory: register this table so the genesis
+        // coverage gate can walk every reachable generated node (annotated or
+        // not) at startup. Submitted at module scope via the `inventory`
+        // re-export on `hyprstream_rpc::metadata` (wasm-safe; same pattern as
+        // `ScopeDefinition`), so consumer crates need not declare an `inventory`
+        // dep directly. The gate iterates this to build the startup activation
+        // gate; a node missing a `$vfsMac` surfaces as a finding (unlabeled ⇒
+        // deny) rather than being silently absent.
+        hyprstream_rpc::metadata::inventory::submit! {
+            hyprstream_rpc::metadata::VfsNodeTable {
+                name: #service_name,
+                nodes_fn: vfs_nodes,
+            }
+        }
+
         #(#scoped_tables)*
     }
 }
@@ -634,7 +649,7 @@ mod tests {
         // `mac` field carries the annotation text verbatim — the label is a
         // property of the node's type, derived at generation.
         let mut health = variant("healthCheck", "Void", "query");
-        health.vfs_mac = "internal".into();
+        health.vfs_mac = "internal:pq-hybrid".into();
         let mut clone = variant("clone", "CloneRequest", "write");
         clone.vfs_mac = "secret:pq-hybrid:0".into();
         // A node with NO `$vfsMac` carries the empty string (→ unlabeled → deny).
@@ -648,7 +663,7 @@ mod tests {
 
         // Annotated labels land verbatim on their nodes.
         assert!(
-            out.contains("mac : \"internal\""),
+            out.contains("mac : \"internal:pq-hybrid\""),
             "health node carries its $vfsMac label: {out}"
         );
         assert!(
--- a/crates/hyprstream-rpc-derive/src/codegen/vfs.rs
+++ b/crates/hyprstream-rpc-derive/src/codegen/vfs.rs
@@ -293,6 +293,10 @@ fn build_node_entries(
         let method_snake = r.method_snake.as_str();
         let scope = v.scope.as_str();
         let bulk = v.vfs_bulk;
+        // $vfsMac (#699 carrier (a)): schema-derived MAC label for this generated
+        // node. Empty string when the annotation is absent (→ unlabeled → deny →
+        // a genesis finding); decoded at runtime by `VfsNode::mac_label`.
+        let mac = v.vfs_mac.as_str();
 
         node_entries.push(quote! {
             hyprstream_rpc::metadata::VfsNode {
@@ -301,6 +305,7 @@ fn build_node_entries(
                 kind: #kind_tokens,
                 scope: #scope,
                 bulk: #bulk,
+                mac: #mac,
             }
         });
     }
@@ -504,6 +509,7 @@ mod tests {
             vfs_kind: "bogus".into(),
             vfs_bulk: false,
             vfs_hidden: false,
+            vfs_mac: String::new(),
         };
         let err = resolve_kind(&v, &CapnpType::Struct("CloneRequest".into()), false).unwrap_err();
         assert!(err.contains("bogus"));
@@ -525,6 +531,7 @@ mod tests {
             vfs_kind: "query".into(),
             vfs_bulk: false,
             vfs_hidden: false,
+            vfs_mac: String::new(),
         };
         let k = resolve_kind(&v, &CapnpType::Text, false).expect("valid kind");
         assert_eq!(k, NodeKind::Query);
@@ -547,6 +554,7 @@ mod tests {
             vfs_kind: String::new(),
             vfs_bulk: false,
             vfs_hidden: false,
+            vfs_mac: String::new(),
         }
     }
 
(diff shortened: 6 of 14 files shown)


ewindisch:
@coderabbitai review

ewindisch:
@coderabbitai review

ewindisch:
VERDICT: REVISE

Reviewed head: `cf3078ec60462c937a2b5e682e59621c40b88467`

## Findings

### HIGH — neither carrier reaches the namespace/resolver it is meant to protect

`BindLabelMap` and `clamp_descendant` have no production caller outside the new
module.  `GeneratedNodeCoverage::for_service` is likewise only exercised with
hand-built `VfsNode` fixtures; it is not constructed from the real service
inventory, installed at genesis, or consulted by `CompositeObjectLabelResolver`.
The resolver still only composes the pre-existing genesis map and manifest
source.  Consequently this does not make a bind without a label fail to
construct, nor does it apply the floor clamp to a walked VFS object.

The real schemas make the outage concrete: the only `$vfsMac` occurrence is the
new annotation declaration itself.  Thus every generated method node currently
emits `mac: ""`, resolves to `None`, and will be denied when #1226 enables the
PEP.  That includes legitimate generated service paths such as registry,
discovery, workflow, and tui.  The added inventory merely documents the gap;
it neither produces an inventory over the actual 9P/VFS surface nor prevents
the incomplete inventory from being installed.  A fixture-only happy path is
not an end-to-end VFS happy path.

Wire the real generated-node inventory and bind operation into the authoritative
object-label resolution path, make a missing label reject the actual bind or
mount construction, and add an integration test that walks one annotated real
node and one unannotated real node (the latter must deny).  The inventory must
cover all reachable object classes and be a startup/activation gate before the
PEP can enable, rather than a manually callable reporting helper.

### HIGH — `$vfsMac` silently permits non-PQ labels on internal VFS paths

`VfsNode::mac_label` accepts a level-only annotation and assigns it
`Assurance::Unverified`; it also explicitly accepts `classical`
([`_metadata.rs`, lines 131–135]).  These are internal generated 9P/VFS
surfaces, not one of the stated interop exceptions.  A classical subject
dominates an `Unverified` object label, so an annotation such as
`$vfsMac("internal")` creates precisely the non-PQ path the operator ruled
out.  This is a default which weakens a configured object label, not a
fail-closed absence rule.

Require `pq-hybrid` for this non-interop carrier (reject omitted,
`unverified`, and `classical` assurances), then add negative tests proving
each rejected form is unlabeled/denied.

## Validation

- `buildq -- cargo build -p hyprstream` — PASS.
- `git diff --check 89befb09c..cf3078ec` — PASS; checkout restored clean after
  temporary negative-test mutations.
- The first focused invocations accidentally combined a partial filter with
  `--exact` and ran zero tests; I detected the runner's `791 filtered out`
  report and did not count them as validation.  A correctly filtered test
  listing/rerun is serialized behind the shared `buildq` queue at review time.
- Negative mutation attempted: replacing `clamp_descendant`'s join with
  `node_label`, and replacing empty `$vfsMac` denial with a Public/Unverified
  label.  Both mutations were restored immediately.  The second invocation
  also used the erroneous exact partial filter and ran zero tests, so this is
  not claimed as successful negative-test evidence.

No prior human review verdict for PR #1228 exists under `.fleet-coord`; the
only PR comments before this review are CodeRabbit bot activity.

ewindisch:
Fix pushed (`7fc5cd7fb`). Both HIGH findings addressed; every fix has a test that fails without it.

## HIGH — neither carrier reaches the namespace/resolver it protects

Wired both carriers into the authoritative resolution path and made the generated-node inventory a startup gate rather than a reporting helper.

**Carrier (a) — schema-derived labels reach the resolver:**
- New `VfsNodeTable` inventory type (`hyprstream_rpc::metadata`), submitted by `generate_rpc_service!` for *every* service with generated VFS nodes (same `inventory::collect!`/`submit!` pattern as `ScopeDefinition`). `inventory` is re-exported on `hyprstream_rpc::metadata` so consumer crates (incl. `hyprstream-discovery`) stay dep-free.
- `GeneratedNodeCoverage::for_all_services()` walks the real registry — not a fixture.
- `GenesisGate::build` folds the real inventory into the startup gate: labeled generated nodes are materialized into the genesis map; **an unannotated generated node makes the report incomplete**, so enforcement cannot be enabled until every reachable object class is annotated. Exposed via `gate.generated_coverage()`.

**Carrier (c) — bind-time labels reach the resolver:**
- `CompositeObjectLabelResolver` now carries a `BindLabelMap` and consults it per walk: the nearest enclosing synthetic mount's floor is D2-joined (restrict-only, `clamp_descendant`) with any genesis/generated label.

**Missing label rejects the bind/mount construction:**
- A generated node whose schema declared no `$vfsMac` (or a non-`pq-hybrid` one) lands in the resolver's `generated_unlabeled` deny-set → the **exact path denies**. It does *not* inherit its service mount's label (the silent hole). `BindLabelMap::bind` takes both labels as `SecurityLabel` (no `Option`, no label-less construction); an unbound prefix resolves to `None` ⇒ deny.

**Integration test** (`resolver_resolves_real_annotated_generated_node` / `resolver_denies_real_unannotated_generated_node`): walks the *real* inventory end-to-end. Annotated registry `health` node (`internal:pq-hybrid`) resolves; unannotated registry `{name}` node denies. Plus `resolver_bind_time_floor_clamps_descendant_carrier_c` / `resolver_bind_floor_joins_genesis_label_restrict_only` for carrier (c).

**Negative-test discipline:** removing the deny-set early-return makes `resolver_denies_real_unannotated_generated_node` fail (the node inherits the mount label) — confirmed, then restored.

## HIGH — `$vfsMac` silently permits non-PQ labels on internal VFS paths

`parse_mac_annotation` now requires exactly `pq-hybrid`. Omitted, `unverified`, and `classical` assurances are all rejected (⇒ `None` ⇒ unlabeled ⇒ deny ⇒ a genesis finding) — the internal non-interop carrier must not reintroduce the classical floor #556 removed. Negative tests (`mac_label_level_only_is_denied_assurance_is_mandatory`, `mac_label_unverified_assurance_is_denied`, `mac_label_classical_assurance_is_denied`, `non_pq_hybrid_assurance_is_a_finding`) cover each rejected form. The `annotations.capnp` doc now states the assurance is mandatory.

## One-way door note

Annotated one real schema node to make the end-to-end test real: registry `healthCheck` → `$vfsMac("internal:pq-hybrid")`. A health-check is the canonical low-sensitivity read-only surface, and `internal:pq-hybrid` is the floor the operator ruling sets for internal VFS. The remaining generated nodes are intentionally left unannotated — they surface as findings in the production gate (correctly incomplete), which is the activation evidence the gate is meant to produce.

**Not merging** — single-review gate.

Issue #1164, opened 2026-07-22
B2: canonical sealed DID-document artifact (sealed artifact, not a computed response)


ewindisch:
Phase **B2** of https://github.com/hyprstream/hyprstream/issues/1158. Dep: B1.

**The DID document is a sealed artifact, not a computed response.** Byte-stable canonicalization; sealed into CAS with its CID recorded in the operation.

Content must be atproto-profile-valid: exactly one `#atproto` P-256 `Multikey`, an `#atproto_pds` service, `alsoKnownAs at://<handle>`, plus the `DidOperationLog` service entry.

**Rejects the host `did:at9p` in `alsoKnownAs` by construction** — one-way door #7. Once published and relied upon, that entanglement cannot be withdrawn safely. Enforce it in the artifact builder, with a test.

Issue #1162, opened 2026-07-22
A3: account zone — consume a DEPLOYMENT-CONFIGURED zone, wildcard DNS, DNS-01 wildcard TLS


ewindisch:
Phase A3 of https://github.com/hyprstream/hyprstream/issues/1158. Dep: A2 (https://github.com/hyprstream/hyprstream/issues/1160). **Unblocked** — Q1 answered 2026-07-22.

**The account zone is deployment-relative configuration**, chosen in the metal repo when deploying hyprstream. Do **not** bake a zone name into hyprstream. This ticket consumes one.

1. **Configuration surface** for the account zone + the DNS-01 credential. A deployment that configures no account zone must **degrade sanely** — account minting disabled with a clear error, not a panic and not a synthesized default.
2. **Extend `TlsConfig`** from its current single-domain HTTP-01 shape to **DNS-01 wildcard issuance** with a zone-scoped credential.
3. **Renewal, monitoring, and an expiry alarm.** One cert failing breaks **all** account resolution for that deployment — this is a total-outage surface, so the alarm is part of the deliverable, not a follow-up.
4. Wildcard A/AAAA for the delegated zone; label depth must match A2 (wildcard covers exactly one label).

Out of scope: custom user domains (tiers T2/T3 in the Q1 answer). Those need proof-of-control verification and, per that analysis, **C5 witness anchoring first** — a domain that expires or transfers is an identity takeover, and only a witnessed op chain makes that detectable rather than silent.

Also out of scope: PowerDNS or any specific DNS implementation. hyprstream defines the interface; managed DNS is a product layer above it. If hyprstream requires a particular DNS server, self-hosters cannot run it.

Issue #1160, opened 2026-07-22
A2: account label grammar, normalization, and allocation policy (permanent rules)


ewindisch:
Phase A2 of https://github.com/hyprstream/hyprstream/issues/1158. No dependency — parallel with A1.

These rules are a **one-way door**: a homograph or non-normalized label minted once is permanent, and labels are **never reusable**. Get them strict from day one; loosening later is possible, tightening is not.

A pure library, no networking, with an exhaustive test corpus:
- LDH single-label, <= 63 chars (wildcard TLS covers exactly one label depth)
- NFKC normalize **or reject** — decide and enforce one, do not silently transform
- confusable and mixed-script rejection
- punycode handling
- reserved list
- **never-reuse** allocation record that makes retirement permanent

The test corpus is the deliverable as much as the code — include known homograph attacks, mixed-script pairs, and normalization edge cases.

PR #1229, opened 2026-07-23
B1: seal user-controlled hosted-account genesis

### LINKED ISSUE #1163: B1: account record + genesis DidOp + mint path (decides the permanent genesis format)

ewindisch:
## Summary

- seal genesis `DidOp` v1 from operation #1 with canonical DAG-CBOR and a frozen wire-format digest
- make the Q2 option-B ordering structural: required user key at priority 0, explicit recovery enrollment/decline at priority 1, optional host key only at priority 2
- require a pinned Ed25519 + ML-DSA-65 user signature before producing an account record
- generate a distinct per-account ES256 `#atproto` key and persist the secret, genesis bytes, and account record through an exclusive never-overwrite mint boundary

## Permanent genesis shape

`rotation_keys` is a fixed three-slot array. Every slot carries its role; declined recovery and absent host slots carry an explicit null key. The API has no constructor that accepts a host key in the user slot, and the doctest proves that misuse does not compile.

Genesis v1 seals `version`, `prev = null`, `seq = 0`, `did`, `doc_cid`, the ordered rotation slots, deliberate `head_at_op`, and the pinned-Hybrid signature. Unknown fields, reordered roles, missing PQ material, non-canonical bytes, and path-form did:web all fail closed.

## Dependency seams

- A2 #1160 supplies an atomically reserved, never-reusable account label.
- A3 #1162 / PR #1182 supplies the deployment-zone-derived host-form DID.
- B2 #1164 consumes the generated `#atproto` public key, seals the canonical DID document, and supplies its CID before the user signs genesis.
- A1 #1159 / PR #1181 remains the separate path-form freeze.

## Validation

- `buildq -- cargo test -p hyprstream-pds --lib --tests --no-fail-fast` (214 unit + 6 integration tests passed)
- `buildq -- cargo test -p hyprstream-pds --doc` (compile-fail custody test passed)
- `buildq -- cargo clippy -p hyprstream-pds --all-targets -- -D warnings`
- repository pre-commit release-profile Clippy gate

Closes #1163
Refs #1158

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7160,6 +7160,7 @@ dependencies = [
  "parking_lot 0.12.4",
  "proptest",
  "rand 0.8.5",
+ "rustix 1.1.4",
  "serde_json",
  "sha2 0.10.9",
  "tempfile",
--- a/crates/hyprstream-pds/Cargo.toml
+++ b/crates/hyprstream-pds/Cargo.toml
@@ -24,6 +24,7 @@ hyprstream-rpc = { path = "../hyprstream-rpc" }
 hyprstream-crypto = { path = "../hyprstream-crypto" }
 # Ed25519 signing/verifying keys for the at9p composite signature path.
 ed25519-dalek = { workspace = true }
+rustix = { version = "1.1.4", features = ["fs"] }
 blake3 = { workspace = true }
 # Non-poisoning Mutex for the in-memory duplicity watermark store (#886).
 parking_lot = { workspace = true }
--- a/crates/hyprstream-pds/Cargo.toml
+++ b/crates/hyprstream-pds/Cargo.toml
@@ -9,6 +9,8 @@ description = "atproto PDS record store — DAG-CBOR records, MST, signed commit
 [dependencies]
 # P-256 ES256 signing for atproto commits (#atproto verification method).
 p256 = { workspace = true }
+# OS-backed entropy for per-account #atproto key generation.
+rand = { workspace = true }
 # SHA-256 — signing digest + CIDv1 dag-cbor multihash for all repo blocks.
 sha2 = { workspace = true }
 # Error handling.
@@ -33,8 +35,7 @@ tracing = { workspace = true }
 # tag-42 links) that serde-driven CBOR cannot guarantee. No CBOR/CID crate dep.
 
 [dev-dependencies]
-# Random signing keys for round-trip / tamper tests.
-rand = { workspace = true }
+# Property tests for deterministic codecs.
 proptest = { workspace = true }
 # at9p_resolver tests exercise the admission arm (admit_key_against_did) which is
 # generic over DidDocResolve — an async trait — against a never-call test double.
--- a/crates/hyprstream-pds/src/did_op.rs
+++ b/crates/hyprstream-pds/src/did_op.rs
@@ -1013,6 +1013,26 @@ mod tests {
     }
 
     #[test]
+    fn tampered_mldsa_component_cannot_downgrade_genesis_to_classical_only() {
+        let user = signer();
+        let unsigned = unsigned(
+            &user,
+            RecoveryKeyEnrollment::Declined,
+            HostKeyEnrollment::Absent,
+        );
+        let signature = sign_genesis(&unsigned, &user.ed, &user.pq).unwrap();
+        let mut tampered_pq = signature.mldsa65().to_vec();
+        tampered_pq[0] ^= 1;
+        let tampered =
+            DidOpSignature::from_components(signature.ed25519().to_vec(), tampered_pq).unwrap();
+
+        assert!(
+            GenesisDidOp::seal(unsigned, tampered).is_err(),
+            "genesis sealing must require a valid ML-DSA-65 signature as well as Ed25519"
+        );
+    }
+
+    #[test]
     fn duplicate_custody_keys_are_rejected() {
         let user = signer();
         let error = GenesisRotationKeys::new(
--- a/crates/hyprstream-pds/src/did_op.rs
+++ b/crates/hyprstream-pds/src/did_op.rs
@@ -1,0 +1,1075 @@
+//! The permanent genesis operation for a hyprstream-hosted `did:web` account.
+//!
+//! This module deliberately implements only operation zero. Successor-chain
+//! validation, fork selection, and historical key-validity queries belong to
+//! C1 (#1167). The bytes emitted here are nevertheless the first record in that
+//! future log, so their grammar is a one-way door for every minted account.
+//!
+//! # Rotation custody is structural
+//!
+//! [`GenesisRotationKeys`] is not a `Vec`. It has exactly three ordered slots:
+//!
+//! 0. a required [`UserRotationKey`];
+//! 1. an enrolled recovery key or an explicit
+//!    [`RecoveryKeyEnrollment::Declined`];
+//! 2. an enrolled host key or an explicit [`HostKeyEnrollment::Absent`].
+//!
+//! A [`HostRotationKey`] therefore cannot be passed as priority zero:
+//!
+//! ```compile_fail
+//! use hyprstream_pds::{
+//!     GenesisRotationKeys, HostKeyEnrollment, HostRotationKey,
+//!     RecoveryKeyEnrollment,
+//! };
+//! # fn example(host: HostRotationKey) {
+//! let _ = GenesisRotationKeys::new(
+//!     host,
+//!     RecoveryKeyEnrollment::Declined,
+//!     HostKeyEnrollment::Absent,
+//! );
+//! # }
+//! ```
+//!
+//! The wire form preserves those slots as a fixed three-element
+//! `rotation_keys` array. A declined/absent slot is encoded as a role-bearing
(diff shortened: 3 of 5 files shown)


ewindisch:
## Operator refinement (Erica, 2026-07-23): YAGNI — do not build host-key-at-priority-0 at all

Stronger than "make it unrepresentable". **Do not build the capability, then guard it. Do not build it.**

`rotation_keys` should not be a general ordered list where any key may occupy any priority, policed by a validation rule that rejects the escrow arrangement. It should be the specific shape we actually support:

- **priority 0 — user-held, REQUIRED**
- priority 1 — user-held **recovery** key, optional, offered at mint
- priority 2 — host key, optional

If a host key at priority 0 is never constructible, there is nothing to validate, nothing to test for, and nothing for a future reader to relax because they did not know why the guard existed. A validation rule is a claim someone must keep believing; an absent capability is not.

If a case for host-priority-0 ever arises — enterprise escrow, recovery-as-a-service, something unforeseen — **it gets built then, with its own decision recorded**. Building it now on speculation means shipping an escrow mechanism nobody asked for, in the one place where the whole point of the design is that we cannot seize an account.

**For the review:** if this PR implements a general ordered-key structure plus a guard rejecting host-at-0, that is a **finding** — not a security defect, but the wrong shape, and worth correcting now while the genesis format is still unsealed. The genesis operation is a one-way door per account; the structure it seals is the thing that is hard to change later.

ewindisch:
VERDICT: REVISE

Reviewed PR #1229 at `2c47957a5a4487d7967723391f04a99cd210961f` against base `69eda14916c0870c16f101eb4147f637e09f2043`, issue #1163, Q2 option B, and the operator's YAGNI refinement.

## Findings

### 1. The never-overwrite writer can permanently brick a legitimately allocated account after an interrupted write

`DirectoryHostedAccountStore::write_new` creates the final label directory first (`hosted_account.rs:441-451`) and then writes the three final filenames in place. `write_new_file` opens each final name with `create_new(true)` before `write_all`/`sync_all` (`hosted_account.rs:525-539`). In particular, `account-record.cbor` becomes visible as the publication marker as soon as `open` succeeds, before its bytes are complete.

Two outage paths follow:

- A crash or I/O error after the final label directory is created leaves that directory behind. Every retry then fails at `create_dir`, even if no public record was ever completed.
- A crash during the final record write can leave a visible, truncated `account-record.cbor`. The label is still unretryable.

A2 makes labels never reusable, so this is not a harmless fail-closed conversion: one transient disk/full/fsync/process failure can permanently deny the legitimate mint path for that reserved account. The happy-path test cannot exercise this because it has no fault injection.

Please make incomplete mint state recoverable and make publication atomic. For example, stage and fsync under an unpublished name, publish with an atomic no-replace transition, and define safe retry/resume behavior for an allocated label with no completed publication marker. Add failure injection at each create/write/sync/publication boundary and prove both that partial state is never published and that the legitimate mint can recover without label reuse.

### 2. The mandatory PQ verification has no causal regression test

The production verifier is currently correct: it supplies the anchored ML-DSA key, sets `require_pq = true`, and requires both result bits (`did_op.rs:757-768`). But an isolated mutation changed that call to `expected_pq_vk = None`, `require_pq = false`, and accepted EdDSA alone. All 9 `did_op::tests` still passed.

That means the permanent genesis boundary can be weakened to classical-only without this new suite detecting it. This is exactly the non-interop surface on which PQ Hybrid is mandatory. Add a focused test that starts with a valid Ed25519 component and a missing/invalid/tampered ML-DSA-65 component and proves `GenesisDidOp::seal` fails. The test should fail under the downgrade mutation above.

## Q2 / one-way-door review

- The YAGNI shape is correct. `GenesisRotationKeys` is a fixed private-field structure `{ user, recovery, host }`, not a general ordered list plus a policy guard. The only emitted shape is priority 0 required user, priority 1 recovery enrollment/decline, priority 2 host enrollment/absence. Decoder role checks reject hostile wire data; they do not expose a host-at-0 construction capability.
- Ordering is explicit in the sealed three-element array, with role-bearing entries and explicit nulls for declined recovery/absent host.
- `version = 1`, `prev = null`, and `seq = 0` are present in operation zero's signed bytes. Unknown fields and non-canonical encodings fail closed. The frozen unsigned-wire digest protects this permanent grammar.
- Recovery absence is an explicit `RecoveryKeyEnrollment::Declined`; host absence is explicit `HostKeyEnrollment::Absent`. Neither has `Default`.
- The `Option` uses found by grep are projections of explicit null wire slots (`RotationSlotRef`) and the strict parser result; `None` does not disable a check. There is no `NotApplicable`, permissive default, disabling feature flag, or configured-requirement/caller-field conjunction.
- Genesis sealing verifies against the required priority-zero key. Rotation keys and signatures require Ed25519 plus ML-DSA-65 in production code.
- The per-account operational `#atproto` key remains a separate generated P-256 ES256 key. This change does not turn it into a rotation key or alter the existing operational-key distinction.
- No legacy/migration/compatibility shim was added.

## What the change breaks

- Intended: path-form/port/percent-encoded/non-LDH `did:web` values, label/DID mismatches, missing priority-zero user material, reordered or missing slots, unknown fields, duplicate custody composites, non-canonical bytes, invalid signatures, and duplicate label-directory publication are rejected.
- Existing production callers: none found outside the new module exports; this is a greenfield API.
- Legitimate happy path: mint, Hybrid-sign, seal, round-trip, and persist succeeds in the uninterrupted case.
- Unintended: as described in finding 1, interruption after final-directory creation can permanently deny that legitimate path for the allocated label.

## Validation

All Cargo commands used `buildq -- cargo ...`; the reviewed worktree stayed clean.

- `buildq -- cargo test -p hyprstream-pds did_op::tests:: --lib -- --nocapture` — 9 passed.
- `buildq -- cargo test -p hyprstream-pds hosted_account::tests:: --lib -- --nocapture` — 6 passed, including the mint/persist happy path.
- `buildq -- cargo test -p hyprstream-pds --lib --tests --no-fail-fast` — 214 unit + 6 integration passed; 0 ignored.
- `buildq -- cargo test -p hyprstream-pds --doc` — compile-fail custody test passed.
- `buildq -- cargo clippy -p hyprstream-pds --all-targets -- -D warnings` — passed.

Mutation checks ran in a disposable isolated clone and were restored afterward:

- Removing the decoder role check made `host_role_at_priority_zero_is_rejected_on_decode` fail.
- Replacing the required-user error path with an unchecked panic made `missing_priority_zero_user_key_is_rejected` fail.
- Replacing the explicit recovery role made `declined_recovery_is_explicit_in_signed_bytes` fail.
- Swapping reported priorities in an isolated mutation made `sealed_ordering_roundtrips_byte_exactly` fail with `(2,user),(1,recovery),(0,host)`.
- Emitting version 0 made both version tests and the frozen digest fail.
- Removing seal-time signature verification made `tampering_a_signed_field_fails_closed` fail.
- Removing the signer/key match checks made `account_record_is_not_produced_before_valid_user_signature` fail.
- Weakening only PQ verification to classical-only unexpectedly left all 9 genesis tests green; this is finding 2.

ewindisch:
Addressed both findings in 6299cfd56.

1. Crash-safe publication: account material is written and fsynced in a private staging directory, then published with Linux `renameat2(RENAME_NOREPLACE)` via Rustix. The final label is never made visible before all three files are durable. A retry removes only invalid/incomplete residue, returns success for the same already-complete account, and refuses a different complete account. `durable_writer_recovers_from_every_interrupted_boundary` injects failures after every staging/file create, write, sync, publish, and root-sync boundary; it proves the final path is either absent or complete and that retry succeeds.

2. PQ regression: `tampered_mldsa_component_cannot_downgrade_genesis_to_classical_only` keeps the valid Ed25519 component but tampers ML-DSA-65, and asserts `GenesisDidOp::seal` fails. I applied the cited downgrade mutation (`expected_pq_vk = None`, `require_pq = false`, Ed25519-only acceptance); this test failed, then I restored the production verification.

Validation: `buildq -- cargo test -p hyprstream-pds --lib --tests --no-fail-fast` (216 unit + 6 integration), `buildq -- cargo test -p hyprstream-pds --doc`, and `buildq -- cargo clippy -p hyprstream-pds --all-targets -- -D warnings` all pass.

Issue #1152, opened 2026-07-22
tech-debt: the 'local bootstrap' vocabulary lies, and it is causing reviewer/implementor regressions — quick wins


ewindisch:
Raised by Erica 2026-07-22, off the https://github.com/hyprstream/hyprstream/issues/1135 decision that multi-host bootstrap is mandatory: the localhost-only assumptions baked in from the single-host era are now actively producing confusion and regressions in implementation AND review. This issue is the **cheap** half — the mechanical wins that reduce the confusion surface without waiting on the `for_local_bootstrap` replacement design.

Evidence for the framing: the constructor documents itself as *"Legacy explicit local/bootstrap adapter... this is not production resolution and carries no closure evidence"* (`hyprstream-rpc-derive/src/codegen/client.rs:619-621`) — while being the actual production constructor for the policy, registry, model, oai, worker, and tui clients in `services/factories.rs`, plus `oauth/mod.rs:419,437`, `inference.rs:2543`, and `bin/main.rs:1160,1194,2238`. **A reviewer who trusts the doc comment concludes these call sites are non-production. They are the production path.** That single mismatch explains a lot of the drift.

---

## W1 — Rename it, and make the doc true (mechanical, ~20 call sites)

`for_local_bootstrap` says *local* and *legacy*; it means *co-located, pre-authentication*. Rename to something that survives the distributed reading — `for_colocated_bootstrap` is the honest one — and rewrite the doc to state **where it is legitimately used** rather than implying it is vestigial. Mechanical, no behavior change, and it stops the next reviewer from mis-scoping a security question.

## W2 — Collapse four near-identical constructors into one (small)

`for_local_bootstrap`, `for_local_endpoint_bootstrap`, `for_local_transport_bootstrap`, and `dial_local_transport_bootstrap` all funnel into the same `anyhow::ensure!`. **None is marked `#[deprecated]`.** A reviewer currently has to know all four spellings to judge whether a call site is safe. Keep one public constructor over a typed `TransportConfig`; make the rest thin `#[deprecated]` shims.

## W3 — Make the refusal a type, not a runtime check (highest payoff per line)

[...]

PR #1243, opened 2026-07-24
fix(ci): make loopback burn-down gate cfg(test)-aware

ewindisch:
The loopback-literal burn-down gate (#1152 W4) was a file-level scan that counted `#[cfg(test)]` literals in src files — its own docstring flagged this as a known limitation. Result: it went **perpetually red** on ordinary test additions (`127.0.0.1:0` ephemeral listeners, `localhost` URL-parse assertions), so it was non-blocking noise inherited by every open PR (e.g. via `auth/valkey.rs`'s test listener, oauth test fixtures).

## Change
- Adds a lightweight, **string/comment-aware** brace-matching skipper (`strip_cfg_test`) that removes `#[cfg(test)]`-gated items/blocks before counting. `#[cfg(not(test))]` is preserved (gates production).
- Rebanks the baseline **274 → 86** — the test-only literals drop out; production loopbacks remain. Scan scope intact (705 source files, unchanged).
- Docstring/methodology updated to match.

## Why not just `--update` or disable?
Disabling loses the real guard (production loopbacks silently break multi-host reach, #1135). `--update` alone just papers over it until the next test adds a localhost. Making it cfg(test)-aware **keeps the production guard and ends the noise permanently.**

## Verified
- Gate passes at the new baseline (exit 0), idempotent.
- **Production guard intact:** injecting a new prod `127.0.0.1` literal still fails the gate (`GREW`).
- e.g. `auth/key_source.rs` 40→1 (39 were JWKS test fixtures), `oauth/registration.rs` 23→6.


ewindisch:
**VERDICT: APPROVE** — Operator (Erica) approved making the loopback burn-down cfg(test)-aware (2026-07-24). Implementation validated: gate passes at the rebanked 274→86 baseline (exit 0), idempotent across runs, and the **production-loopback guard is intact** — a newly-injected prod `127.0.0.1` literal still fails the gate (verified: `10→11 GREW`, then reverted). String/comment-aware brace skipper; `#[cfg(not(test))]` preserved. Ends the perpetual advisory-red inherited by every open PR.

Issue #1253, opened 2026-07-24
Scrub/redact prompts from process logs


ewindisch:
## Problem

Tokenization logs the complete raw prompt at INFO (`crates/hyprstream/src/runtime/torch_engine.rs:914`). The same error path can embed the original text when tokenization produces no tokens (`crates/hyprstream/src/runtime/torch_engine.rs:923`). In a shared inference process, those records place caller content in process-wide logs visible outside the tenant's request boundary.

## Scope

- Remove prompt text from INFO, WARN, and ERROR records. Log only bounded metadata such as token count, byte length, model identifier, request correlation ID, and a tenant-safe opaque identifier.
- Do not move raw prompts to ordinary DEBUG logging. If payload diagnostics are ever enabled, require an explicit single-tenant/development-only profile, short retention, access controls, and a conspicuous startup warning.
- Audit the inference, templating, tokenization, TTT, adapter, streaming, and error paths for prompt text, token text, generated content, bearer tokens, and serialized request bodies.
- Add tests or a tracing capture assertion that a distinctive prompt secret does not appear in logs on success, tokenization failure, cancellation, or stream error.
- Preserve useful observability with tenant-tagged structured fields derived from the verified tenant claim, without logging the claim token itself.

## Depends on

None.

## Refs

#1128

---
Part of Epic #1246 — Inference multi-tenant hardening.

Issue #1246, opened 2026-07-24
Epic: Inference multi-tenant hardening — make the engine a measured isolation plane


ewindisch:
## Problem

The inference engine is not yet a fail-closed tenant-isolation plane. KV ownership distinguishes sessions, stateless requests, and training runs but carries no tenant key (`crates/hyprstream/src/runtime/kv_cache.rs:46`). The engine also stores one process-wide active cache owner (`crates/hyprstream/src/runtime/torch_engine.rs:140`) and one mutable `base_delta` that is composed into every subject's effective delta (`crates/hyprstream/src/services/inference.rs:416`). Delta eviction asks the KV registry to invalidate tenant dependencies (`crates/hyprstream/src/training/delta_pool.rs:462`), but cache creation only calls `get_or_create` and installs the returned cache without registering the active subject delta (`crates/hyprstream/src/runtime/torch_engine.rs:359`; `crates/hyprstream/src/runtime/kv_cache.rs:370`). Full prompt text is emitted at INFO during tokenization (`crates/hyprstream/src/runtime/torch_engine.rs:914`).

The isolation radius is also process-wide: model loading selects `ServiceSpawner::threaded()` for each inference service (`crates/hyprstream/src/services/model.rs:773`), while a subprocess mode already exists as a separate spawner choice (`crates/hyprstream-service/src/service/spawner/service.rs:419`). Inference authorization uses the resolved subject but hardcodes the Casbin domain to `"*"` (`crates/hyprstream/src/services/inference.rs:1895`), and generation tolerates missing verified claims by assigning a default expiry (`crates/hyprstream/src/services/inference.rs:1923`). These behaviors leave the engine below the isolation bar measured by #1128.

## Scope

Make the engine the fail-closed tenant-enforcement authority: require a verified tenant claim before touching tenant-scoped state, deny by default when the claim is absent or invalid, and validate that contract first on the RPC path, where the handler already receives the verified `EnvelopeContext` subject and claims (`crates/hyprstream/src/services/inference.rs:1910`; `crates/hyprstream-rpc/src/service/svc.rs:387`). The HTTP/OAI path must then carry the same verified claim through the model proxy to satisfy that contract.

[...]

PR #1260, opened 2026-07-24
fix(inference): register TTT delta dependency on session KV cache (#1254)

### LINKED ISSUE #1254: Wire TTT→KV invalidation (call register_delta_dependency)
## Problem

`KVCacheRegistry::register_delta_dependency` records which tenant delta produced a cache, and `invalidate_for_tenant` removes only owners found in that dependency map (`crates/hyprstream/src/runtime/kv_cache.rs:370`; `crates/hyprstream/src/runtime/kv_cache.rs:382`). Delta removal, eviction, and clear paths call `invalidate_for_tenant` (`crates/hyprstream/src/training/delta_pool.rs:462`; `crates/hyprstream/src/training/delta_pool.rs:662`; `crates/hyprstream/src/training/delta_pool.rs:905`).

The serving cache-selection path currently obtains or creates the owner cache and installs it into the model, but does not register the active tenant delta between those operations (`crates/hyprstream/src/runtime/torch_engine.rs:359`; `crates/hyprstream/src/runtime/torch_engine.rs:361`). As a result, invalidation cannot discover those caches and stale KV can survive a TTT delta change once session reuse is active.

`CacheConfig.eviction_exempt` is configurable but eviction bypasses the field and instead hardcodes an exemption for `CacheOwner::Training` (`crates/hyprstream/src/runtime/kv_cache.rs:93`; `crates/hyprstream/src/runtime/kv_cache.rs:295`).

## Scope

- At the request-scoped cache create/select point, call `register_delta_dependency` with the tenant whose effective delta was used; explicitly register `None` when no tenant delta contributes.
- Update or clear the dependency whenever the effective delta changes, the cache is recomputed under different weights, the owner is released, or the cache is removed.
- Make registration and cache installation atomic enough that generation cannot publish output from an untracked cache.
- Add tests that create KV under tenant A's delta, mutate/remove/evict that delta, and prove the matching cache is cleared while tenant B's cache remains intact.
- Add a no-delta-to-delta and delta-to-no-delta transition test so stale cache reuse cannot cross weight versions.
- Either make `eviction_exempt` govern candidate selection or remove it and document the explicit training-owner rule.

## Depends on

The tenant-keyed cache-owner change must land before cross-tenant isolation is complete, but the missing registration bug can be fixed and tested independently on the current RPC Subject path.

## Refs

#1128

---
Part of Epic #1246 — Inference multi-tenant hardening.

ewindisch:
## Summary

Fixes #1254 (Epic #1246 — Inference multi-tenant hardening).

The serving cache-selection path (`TorchEngine::swap_session_cache`) obtained/installed the owner KV cache but **never called `KVCacheRegistry::register_delta_dependency`** between those steps. `invalidate_for_tenant` only evicts owners found in the dependency map, so it could not discover those caches — meaning stale KV survived a TTT delta change once session reuse was active.

## Changes

- **`runtime/torch_engine.rs`** — `swap_session_cache` now takes the effective tenant id (`None` when no tenant delta contributes) and registers it via `register_delta_dependency` **before** `set_kv_cache`, so a cache is tracked for invalidation before generation can read a token from it. `tenant_id` is threaded from the verified RPC subject through `generate_with_delta` → `TextStream::new_with_delta` → `swap_session_cache`.
- **`runtime/kv_cache.rs`**
  - `register_delta_dependency` now also clears the cache **in place when the effective tenant changed since its last compute** (no-delta↔delta, tenant-A↔tenant-B), so stale KV cannot cross a weight boundary. Returns whether it cleared.
  - `release()` (Stateless) and `evict_to_budget()` (tier-2) drop the delta dependency alongside the cache — no orphan dependency entries.
  - Removed the dead `CacheConfig.eviction_exempt` field + `with_eviction_exempt` builder and documented the explicit `CacheOwner::Training` exemption rule in `evict_to_budget` (the field was configurable but never consulted).
- **`services/inference.rs`** — `resolve_delta` now also returns the tenant id (`Some(subject)` only when a per-tenant delta contributed), derived from the existing verified request subject — no new identity source.

## Tests (`runtime::kv_cache`)

- Cross-tenant isolation: invalidate tenant-a clears only A's cache; B's stays intact (token IDs + layer KV).
- no-delta↔delta and tenant-A↔tenant-B transitions clear the stale KV; re-registering the same tenant preserves it.
- `release` drops the delta dependency (no orphan).
- A no-tenant cache survives any tenant invalidation.

## Build notes

Local gate green: `cargo check` / `cargo test runtime::kv_cache` (19 passed, 0 failed) / `cargo clippy --tests` for `hyprstream`, built with `--no-default-features --features gittorrent,xet,otel` (the box has no `libsystemd` devel headers, so the default `systemd` feature can't link locally). The changed code is feature-agnostic; CI validates the default-feature build.

## Merge coordination

Touches `runtime/torch_engine.rs` + `runtime/kv_cache.rs` — file overlap with **D7 (#1253)**, which also edits `torch_engine.rs`. Stack/serialize at merge to avoid conflicts.

Closes #1254
Epic #1246

### DIFF
--- a/crates/hyprstream/src/runtime/kv_cache.rs
+++ b/crates/hyprstream/src/runtime/kv_cache.rs
@@ -89,8 +89,6 @@ pub struct CacheConfig {
     pub max_seq_len: usize,
     /// Quantization type for memory efficiency
     pub quant_type: KVQuantType,
-    /// Whether this cache is exempt from eviction (e.g., for training)
-    pub eviction_exempt: bool,
     /// Use paged block storage (PagedAttention-style) instead of contiguous tensors.
     /// Requires a BlockPool to be initialized on the registry.
     pub paged: bool,
@@ -103,7 +101,6 @@ impl CacheConfig {
             num_layers,
             max_seq_len,
             quant_type: KVQuantType::None,
-            eviction_exempt: false,
             paged: false,
         }
     }
@@ -111,12 +108,6 @@ impl CacheConfig {
     /// Set quantization type
     pub fn with_quant_type(mut self, quant_type: KVQuantType) -> Self {
         self.quant_type = quant_type;
-        self
-    }
-
-    /// Mark as eviction exempt
-    pub fn with_eviction_exempt(mut self, exempt: bool) -> Self {
-        self.eviction_exempt = exempt;
         self
     }
 
@@ -260,13 +251,15 @@ impl KVCacheRegistry {
     pub fn release(&self, owner: &CacheOwner) {
         match owner {
             CacheOwner::Stateless(_) => {
-                // Stateless caches are removed immediately
+                // Stateless caches (and their delta dependency) are removed immediately.
                 self.caches.remove(owner);
+                self.delta_dependencies.remove(owner);
                 tracing::debug!("Released stateless cache: {:?}", owner);
             }
             CacheOwner::Session(_) | CacheOwner::Training { .. } => {
-                // Session and training caches are kept for reuse
-                // They will be evicted by LRU if memory pressure occurs
+                // Session and training caches are kept for reuse; their delta
+                // dependency is retained so a later tenant-delta eviction can
+                // still invalidate them. They fall to LRU eviction under memory pressure.
                 tracing::debug!("Marked session cache for potential reuse: {:?}", owner);
             }
         }
@@ -284,14 +277,17 @@ impl KVCacheRegistry {
             return;
         }
 
-        // Collect eviction candidates (skip eviction-exempt and training caches)
+        // Collect eviction candidates. `CacheOwner::Training` caches are exempt
+        // from eviction by rule — they hold validation/eval state for an active
+        // training run and must survive memory pressure (the run owns their
+        // lifecycle). All other owners (Stateless/Session) are evictable.
         let mut candidates: Vec<(CacheOwner, u64, usize)> = self
             .caches
             .iter()
             .filter_map(|entry| {
                 let owner = entry.key().clone();
 
-                // Skip training caches
+                // Skip training caches (exempt by rule — see comment above).
                 if matches!(owner, CacheOwner::Training { .. }) {
                     return None;
                 }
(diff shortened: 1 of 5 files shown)


ewindisch:
## F1 fix: per-tenant weight epoch (247b85afc)

Implements the owner-approved epoch design from the F1 ML reviews (both FIX-NOW, epoch preferred over blunt eager invalidation).

**Problem:** `register_delta_dependency` detected staleness by comparing `tenant_id` only. In-place re-adaptation (`adapt_tenant` / `train_step` / rollback / `reset`) changes the effective weights without changing the tenant id, so same-tenant session KV computed under the old weights was preserved — a hybrid old/new-weight computation on every adapted turn of a TTT session (including the generation in the same request that just adapted).

**Fix:**
- `TenantDelta` gains a `weight_epoch` version counter, bumped on every in-place weight mutation: `adapt_tenant` and `train_step` (on actual gradient steps), `load_state_dict` (rollback / snapshot rehydrate), and `reset`.
- The KV dependency is now `(tenant_id, weight_epoch)`; `register_delta_dependency` treats an epoch mismatch exactly like a tenant change — clear-in-place before install, so generation can never publish from weight-stale KV. Bump and compare are both O(1); no per-adaptation DashMap scan.
- The epoch is threaded `resolve_delta` → `generate_with_delta` → `swap_session_cache` → `register_delta_dependency`. `resolve_delta` reads the epoch after adaptation, so the first post-adaptation generation in the same request re-prefills under the new weights.
- No-op turns (skipped adaptation, 0 steps) don't bump, so they keep their prefix-cache hits.

**F2 comment:** `swap_session_cache` now documents that register→install atomicity relies on the single-threaded cooperative-scheduling invariant (InferenceService on a `current_thread` runtime + `LocalSet`; the function is await-free). If InferenceService ever moves to a multi-threaded runtime, a spanning lock or epoch handshake is needed. No lock added — the race is theoretical today.

**Tests:**
- New: `test_register_clears_stale_kv_on_weight_epoch_advance` — same tenant, epoch advanced → stale KV cleared; same epoch → preserved; post-clear reuse at recorded epoch → preserved.
- New: `test_weight_epoch_bumps_on_in_place_mutation` — fresh delta at epoch 0; `reset` and `load_state_dict` each bump.
- Updated: tenant-transition, cross-tenant isolation, release, and no-tenant tests for the new signature — all pass (20 kv_cache + 17 tenant_delta + 10 ttt lib tests green; clippy `-D warnings` clean).

PR #1261, opened 2026-07-24
fix(inference): scrub prompt/token text from process logs (#1253)

ewindisch:
> **Revision (OTel focus):** this PR is redirected from log-only redaction to emitting **token burn as OpenTelemetry metrics** + trimming the now-redundant count-bearing logs. The no-content security property and hermetic canary test from the first revision are preserved.

## Summary

Emit **token burn** (prompt + generated tokens per request) as **OpenTelemetry metrics**, and trim the now-redundant count-bearing logs (the counts live in metrics, not logs). This is accounting- and ops-relevant. The authoritative spend ledger is tracked in **#1264**; the metrics here are the metering/ops signal.

### OTel metrics — new `runtime/token_metrics`
- **`inference_tokens_total`** — `Counter<u64>` (unit `tokens`), cumulative tokens consumed, attrs `kind` = `prompt` | `generated`, `model`, and an **opaque hashed** tenant id (never the raw subject).
- **`inference_request_tokens`** — `Histogram<u64>` (unit `tokens`), per-request prompt + generated total.
- Prompt tokens recorded at the tokenization boundary (`TextStream::new_with_delta`); generated tokens + per-request total recorded at `TextStream` completion — `Drop` runs on every exit path (EOS, stop, max, timeout, error).
- Instruments build from the global OTel meter, so recording is always safe (no-op without a provider, and compiled out entirely when the `otel` feature is off).

### Meter provider wiring — `bin/main.rs` `init_telemetry`
- Registers a `SdkMeterProvider` (OTLP or stdout, mirroring the tracer transport) via `set_meter_provider`, so the metrics actually export to a collector instead of being no-ops. Adds the `metrics` feature to `opentelemetry-otlp` / `opentelemetry-stdout`.

### Log trimming (counts now in metrics)
- Removed the `bytes/chars/tokens` "Tokenized prompt" INFO line.
- Removed the per-chunk generated-text DEBUG count.
- **Kept** the content-free diagnostics — empty-token-sequence warn + tokenize/decode roundtrip-mismatch warn (bounded lengths, never text) — real tokenizer-bug signals, not accounting.
- Dropped raw `tenant=?tenant_id` from the stop-sequence warning; tenant identity in metrics/logs is opaque/hashed only (#1253).

### Tests
- Preserved the no-leak canary (`tokenize_redacted_never_logs_prompt_content`); it now asserts the success path no longer emits a metering log line (counts moved to metrics) while still never leaking prompt content.
- Added hermetic meter tests (`InMemoryMetricExporter` + `PeriodicReader`) asserting token burn lands on the right instruments with opaque tenant attrs; plus `opaque_tenant_id` stability / anonymous tests.

## Build

`--no-default-features --features "gittorrent,xet,otel"` locally (no libsystemd devel headers on this box; CI validates defaults). Gate green: `cargo check --tests`, `cargo clippy --tests -- -D warnings`, and the new meter tests + canary pass (rocm71 backend).

Refs #1253. Complement: **#1264** (authoritative spend ledger).

---
Stacked on #1260 (`ewindsch/d8-kv-invalidation`) — GitHub auto-retargets to `main` when #1260 lands. Do not merge / self-approve.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -95,9 +95,9 @@ tracing-opentelemetry = { version = "0.31" }
 # OpenTelemetry
 opentelemetry = { version = "0.30" }
 opentelemetry_sdk = { version = "0.30", features = ["rt-tokio"] }
-opentelemetry-otlp = { version = "0.30", features = ["grpc-tonic", "trace"] }
+opentelemetry-otlp = { version = "0.30", features = ["grpc-tonic", "trace", "metrics"] }
 opentelemetry-semantic-conventions = { version = "0.30" }
-opentelemetry-stdout = { version = "0.30", features = ["trace"] }
+opentelemetry-stdout = { version = "0.30", features = ["trace", "metrics"] }
 
 # Utilities
 uuid = { version = "1.0", features = ["v4", "v5", "serde"] }
--- a/crates/hyprstream/Cargo.toml
+++ b/crates/hyprstream/Cargo.toml
@@ -156,6 +156,18 @@ ratatui = { version = "0.30", default-features = false }  # TUI framework for pa
 critical-section = { version = "1", features = ["std"] }
 vte = "0.13"  # VT100/ANSI escape sequence parser for TUI pane ingestion
 
+# OpenTelemetry (0.30 + OTLP/stdout exporters + tokio) is not wasm32-compatible.
+# The `otel` feature stays in the default set, but the dependency family only
+# resolves for non-wasm targets; the OTel code paths are cfg-gated to match and
+# fall back to a no-op on wasm32 (#1253/#1261, browser client).
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+tracing-opentelemetry = { workspace = true, optional = true }
+opentelemetry = { workspace = true, optional = true }
+opentelemetry_sdk = { workspace = true, optional = true }
+opentelemetry-otlp = { workspace = true, optional = true }
+opentelemetry-semantic-conventions = { workspace = true, optional = true }
+opentelemetry-stdout = { workspace = true, optional = true }
+
 [target.'cfg(target_os = "linux")'.dependencies]
 lru.workspace = true  # LRU cache for model caching and Git repository handles
 dashmap.workspace = true  # Concurrent HashMap for thread-safe repository caching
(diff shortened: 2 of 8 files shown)


ewindisch:
Revision pushed (784702477) per owner direction: refocused from log-metadata to **OpenTelemetry metrics for token burn**. Adds `runtime/token_metrics.rs` instruments; records prompt + generated token counts at tokenize + stream completion; trims the now-redundant count-bearing log lines (counts live in metrics). Keeps the no-content security property + the hermetic canary test. Clippy-clean locally; full CI running. The authoritative ledger-spend complement is #1265 (issue #1264).

ewindisch:
**Revise-2** (`1e1fe1988` → `4abae12f8`, rebased onto current `origin/main`): wasm-gate + rebase + review fixes.

**1. OTel target-gated off wasm32 (owner's call — `otel` stays in default features).** The opentelemetry family (0.30 + OTLP/stdout exporters + tokio) is not wasm32-compatible, so the six crates (plus the SDK `testing` dev-dep) moved under `[target.'cfg(not(target_arch = "wasm32"))'.dependencies]` and every OTel code path (`bin/main.rs` telemetry, `runtime/token_metrics.rs`) now compiles only for `all(feature = "otel", not(target_arch = "wasm32"))`. The no-op `TokenBurnMeter` covers wasm and `otel`-off builds, so the browser client pays no metric cost. Verified: `cargo tree --target wasm32-unknown-unknown -p hyprstream` resolves **zero** opentelemetry crates (x86_64: 19), and the non-wasm `--features otel` build/instruments work.

**2. Rebased onto `origin/main`.** Clears the stale-base `tracing`/`FederationKey` wasm failure in `hyprstream-rpc/src/auth/jwt.rs` (fixed on main). Note: the CI "WASM (browser client)" job (`cargo check --target wasm32-unknown-unknown -p hyprstream-rpc`) currently fails **on main itself** — `inventory` was moved to native-only deps in hyprstream-rpc while `_metadata.rs:198` still uses `inventory::collect!` unconditionally (from #1228, unrelated to this PR; D7 touches no hyprstream-rpc files).

**3. Review findings addressed.**
- **Token IDs redacted** from all process logs (KV-cache debug, EOS, stop-match, buffering TRACE, decode-stream error) — logs carry positions/counts only; token IDs are reversible to text and the stop-token ID identifies a caller stop string. The raw `tenant_id` in the `kv_cache.rs` invalidation log is also replaced with the opaque id.
- **Tenant pseudonym is no longer enumerable**: `opaque_tenant_id` uses a process-random keyed SipHash (`RandomState` in a `OnceLock`) instead of unkeyed `DefaultHasher` — stable within the process, not precomputable offline.
- **Instruments built once process-wide**: `once_cell::Lazy` `global_meter()` shared by both recording sites; `from_meter` remains as the test seam.
- (LOW finding) Unit is now UCUM `{token}`, the per-request histogram carries only `model`/`tenant`, and tests assert exact per-point counter values (prompt=10 / generated=7), exact attribute sets, histogram attrs, and units.

**Gates (local, rocm71 libtorch, `--no-default-features --features gittorrent,xet,otel`):** `cargo clippy -p hyprstream --tests -- -D warnings` clean; `token_metrics` 3/3 + `tokenize_redacted_never_logs_prompt_content` 1/1 pass; `otel`-off check (`gittorrent,xet`) compiles. Not merged.

Issue #1065, opened 2026-07-17
feat(resource): canonical ResourceIntent, ownership manifests, attestations, and vectors


ewindisch:
# Goal

Define and implement the canonical, versioned `ResourceIntent`, ownership manifest, MAC attestation, ledger attestation, and receipt-envelope formats for #1064.

## Required model

- Separate stable `resource_id`, per-version `content_cid`, `manifest_cid`, and deterministic `operation_id`.
- Typed owner/controller/payer references supporting identified DIDs, pairwise DIDs, commitments, and anonymous entitlement/capability commitments.
- Exact operation, resource kind, expected predecessor/version, policy/state epochs, capability root, allocation/entitlement commitment, resource unit, maximum charge, nonce, expiry, carrier/profile context, and assurance inputs.
- Domain-separated canonical DAG-CBOR digest. Signatures cover the intent digest; the final manifest references signatures to avoid a CID/signature cycle.
- Versioned encrypted detailed receipts plus public commitments/checkpoint projection.

## Acceptance criteria

- [ ] Canonical CDDL and Rust types reject unknown critical fields, non-canonical encodings, bounds violations, suite/profile confusion, and version ambiguity.
- [ ] MAC and ledger verifiers independently derive the identical intent digest.
- [ ] Cross-resource, cross-operation, cross-origin, cross-unit, cross-policy, cross-predecessor, and replay substitutions fail.
- [ ] No public schema requires or leaks a holder DID, raw anonymous token, stable client key, or linkable entitlement identifier in anonymous profiles.
- [ ] Owner, controller, payer, issuer, and ledger operator cannot be confused by field reuse.
- [ ] Deterministic positive vectors and mutation-effective negative vectors cover every normative field.
- [ ] Browser/WASM and native implementations derive byte-identical IDs and digests.

## Artifacts

CDDL, generated/reference Rust types, vectors, schema documentation, compatibility/versioning policy, and the encoding sections consumed by `draft-hyprstream-resource-attestation-00`.

## Dependencies

#1064 architecture child, #699 content-bound labels, #923 ledger identifiers, #1039 accepted-current authority, and #1061 binding transcript. Blocks registrar, MAC, ledger bridge, CAS, and 9P finalization.

## #1058 coordination boundary

[...]

Issue #1064, opened 2026-07-17
Epic: dual-attested resource ownership — MAC title, cellular accounting, anonymous entitlements, and 9P/CAS integration


ewindisch:
# Objective

Deliver dual-attested resource ownership for Hyprstream resources: MAC establishes title/control authority, cellular ledgers attest economic entitlement and usage, and a durable resource registrar finalizes one canonical resource transition. Support identified, pairwise, and commitment-based privacy profiles without turning payment into title or anonymous authorization into a fabricated identity.

This is the resource-lifecycle implementation sibling of the cellular-ledger epic #922, rooted in the self-sovereign-resource design #921. It consumes the native MAC epic #547, PQ-hybrid anonymous-capability epic #1058, and 9P/CAS substrate #809/#705 without absorbing their scopes.

## Authority boundaries

- **MAC/capability authority:** whether a principal may create, control, transfer, mutate, or delete a resource under the current policy and content-bound label.
- **Ledger authority:** whether the identified or anonymous payer has a valid resource entitlement and what was reserved, consumed, released, or compensated.
- **Resource registrar:** the serialized, durable state machine that joins both attestations and selects the single canonical successor manifest.
- **Resource store/namespace:** bytes and namespace projection; it does not independently invent title.
- **PDS/proof plane:** manifests, receipts, checkpoints, and selective-disclosure material; it is not a synchronous filesystem lock service.

## Non-negotiable invariants

1. No final resource manifest exists without MAC and ledger attestations over the identical canonical `ResourceIntent` digest.
2. Payment does not confer title; MAC authorization does not prove payment. Owner, controller, payer, issuer, and ledger operator are distinct typed roles.
3. At most one finalized successor may advance a resource version. Every transition carries an expected predecessor/version and is finalized by compare-and-swap with a fencing token.
4. Stable resource identity, content CID, manifest CID, and operation ID are separate. No cyclic hash/signature construction is permitted.
5. Every request and external effect is idempotent by deterministic operation ID. Every reservation eventually posts, voids, or enters explicit quarantine/manual review.
6. Provisional material is unavailable through the normal namespace. Visibility is a property of registrar state, not merely byte existence.
7. Reads remain MAC-only. Ledger/proof-plane I/O is not placed on ordinary reads or fine-grained 9P writes; bounded local leases settle at defined commit points.
8. Identified, pairwise, committed-owner, and anonymous-controller/payer profiles remain distinct. Anonymous authorization never fabricates a DID.
[...]

Issue #985, opened 2026-07-10
security(ledger): bind spend-authz to the minted transfer_id before the enforcer activates (#925 N1 — pre-activation gate)


ewindisch:
From the fable review of PR #981 (ledger Phase-1 enforcer). **Not exploitable today** (enforcer is default-off, deny-all verifier, no live PEP caller) but **MUST be fixed before item-1.5 UCAN/PDS wiring flips enforcement on.**

**Gap:** `enforcer.rs` admit() verifies `authz.grant_cid`, `authz.host`, and `amount <= authz.max_amount`, then checks the signature — but `SpendAuthorization.transfer_id` (which IS folded into the signed digest) is **never compared** to the transfer_id minted at spend time (`mint_transfer_id(grant_cid, req.nonce)`). So one signed authorization is **replayable across many distinct transfers** (different nonces), up to `max_amount` each, for the whole `exp` window. The grant balance cell still floors total spend against the subject's own credit, so it's replay-amplification / per-transfer-intent loss, not cross-principal escalation — but it undercuts the per-transfer authorization and the #681 dual-principal receipt semantics.

**Fix:** after minting `transfer_id`, require `authz.transfer_id == transfer_id` (or fold `req.nonce` into the digest and check it). Add a negative test for a mismatched-authz transfer_id (the current happy-path test reuses the same nonce, masking it).

Also from the same review (lower priority nits, fold in): N2 grant cache not generation-scoped (doc overclaims); N3 PQ downgrade silent+unlabeled (the 'labeled not rejected' Decision-D half isn't recorded); N4 post_actual releases remainder regardless of post outcome; N5 factory block_on panics on current-thread rt; N6 hyprstream-crypto should be optional under the ledger feature.

## Resource-intent binding dependency

#1064/#1065/#1072 generalize the required fix: authorization and ledger spend must bind the exact canonical `ResourceIntent` digest, including operation ID, resource commitment, predecessor/version, origin, unit, amount ceiling, policy epoch, nonce, and assurance/profile context. Fixing the existing minted-transfer comparison remains the immediate prerequisite; the resource bridge must not activate on a weaker binding.

ewindisch:
[...]

PR #1265, opened 2026-07-24
feat(inference): post token-spend to cell ledger on generation completion (#1264)

### LINKED ISSUE #1264: feat(inference): post token-spend to hyprstream-ledger on generation completion (quota burn)
## Problem

Inference consumes tokens (prompt + generated) per verified subject but never posts a spend to the
ledger, so there is **no quota-burn accounting for inference**. The primitives are already on main:

- `hyprstream-ledger` (#941) exposes a single-phase spend API with overdraft checking on the debit
  side (`crates/hyprstream-ledger/src/backend.rs:41`, `engine.rs:49`) and dual-signed usage receipts
  carrying an optional `transfer_id` (`crates/hyprstream-ledger/src/journal.rs:81`).
- Spend-authz is bound to a minted `transfer_id` (#985, landed via #1129).
- A Phase-1 local enforcer + CreditGate + ledger service exist (#981); the `ledger/spend` ability
  is the grant verb an enforcer attenuates against (`crates/hyprstream-k8s/src/grant.rs:80`).

`InferenceService` already assembles a `GenerationQualityMetrics` record at completion
(`crates/hyprstream/src/services/inference.rs:2639`) and holds the verified caller subject
(`ctx.subject()`, `inference.rs:1910`). The missing piece is the **emit**: turn
`(subject, model, prompt_tokens, generated_tokens)` into a posted ledger spend bound to a
`transfer_id`.

## Scope

- At generation completion (near `inference.rs:2639`), post a **single-phase spend** to
  `hyprstream-ledger` for the request: debit the verified subject's account for
  `prompt_tokens + generated_tokens` on `model`, bound to a `transfer_id`.
- Use the ledger crate's **native** spend API (`LedgerBackend` / engine single-phase spend, journal
  receipt). Do **not** depend on `ResourceIntent` (#1065, not yet implemented); if a
  ResourceIntent-shaped record is the intended future abstraction, define a minimal adapter now and
  leave the full type to #1065.
- Source the **subject** from the verified `ctx.subject()`. Source **prompt_tokens** from
  tokenization and **generated_tokens** from the stream/completion count.
- **Bind the spend to a `transfer_id`** per #985. If one must be minted at admission (pre-generation),
  wire the minimal seam to obtain/mint it and carry it to completion; if the ledger service mints on
  spend, use that path.
- **Fail-safe:** an accounting/ledger failure must not break a generation that already produced
  output — emit a content-free error metric/log and continue (or gate per explicit policy). Never
  silently drop a spend without a signal. Decide and document the failure mode.
- **Tests:** two-subject — each subject's spend posts to its own account only; an overdraft/declined
  case behaves per the local enforcer's policy; a missing/invalid subject fails closed (no spend, no
  content leak).

## Depends on

None unmerged — `hyprstream-ledger` (#941) and `transfer_id` spend-authz (#1129) are on main.

## Refs

#1064 · #923 · #985 · #981 · #525 · #1246

Part of Epic #1064 (dual-attested resource ownership / cellular accounting); also serves the
inference multitenant epic #1246 (per-tenant quota is an isolation mechanism) and the scheduler's
per-Subject quota (#525).

ewindisch:
Closes #1264 · Epic #1064 · refs #1246, #923, #985, #981

## What

At generation completion the inference service now posts a **single-phase
spend** to the cell ledger (`hyprstream-ledger`): it debits the verified
caller's account for `prompt_tokens + generated_tokens` of the model's token
unit, crediting the cell usage account, bound to a `transfer_id`. This is the
accounting emit inference previously never produced — so there was **no
quota-burn accounting for inference**.

## How

- **`services/ledger/inference_spend.rs` (new)** — `InferenceSpendEmitter`
  over a `LedgerHandle`. `post_generation_spend` mints a completion
  `transfer_id` and calls the ledger crate's **native single-phase spend**
  (`LedgerHandle::debit`, overdraft-checked on the debit side). Every backend
  `Outcome` is classified into a fail-safe `SpendResult`
  (`Posted` / `Skipped` / `Declined` / `Failed`); the method **never returns a
  `Result` the caller propagates** — a ledger/accounting failure can never
  break a generation that already produced output, and every non-posted outcome
  is a **content-free signal** (log), never a silent drop. Anonymous caller ⇒
  fail closed (no spend, no leak). Does **not** depend on `ResourceIntent`
  (#1065, not merged); `SpendInput` is the minimal completion-shaped record.
- **`services/inference.rs`** — thread the verified caller's self-certifying
  **pairwise DID** (`ctx.authenticated_pairwise_did()`, not the Casbin label
  `ctx.subject()`) through `PendingWork::Generation`; add a feature-gated,
  default-**inert** emitter field + an `attach_ledger_spend` seam; post the
  spend in `execute_stream` **after** the client-visible completion frame.
- **`services/ledger/mod.rs`** — export the emitter + result types.

## `transfer_id` (#985) — which path

`Transfer.id` is **client-supplied** (it *is* the ledger's idempotency key; the
backend never mints one — a replayed id returns the original `Outcome`
verbatim). The admission path (`LocalEnforcer::admit`) mints it **at admission**
via `mint_transfer_id(grant_cid, nonce)` so the #985 spend-authorization binds to
it. **There is no admission on the inference completion path yet** — the
two-phase reserve/post enforcer is gated off and not wired into inference (see
the `services::ledger` module docs). So this emit mints the transfer id **at
completion**, under a distinct domain tag so it cannot collide with the
admission path's id space. Cross-replay idempotency (a retried request
reproducing the same id) is a documented follow-up: it needs the envelope
`request_nonce` threaded to completion; today the per-stream `stream_id` is the
entropy.

## Fail-safe (decided)

Chosen failure mode: **best-effort, fire-and-forget, fail-open-for-the-stream /
fail-closed-for-accounting**. A ledger/accounting failure **must not** break a
generation that already produced output — the spend is posted *after* the
client-visible completion frame, and the emitter converts every outcome to an
observable `SpendResult` the caller only logs. There is no path by which a
ledger error propagates as a stream error or drops a spend silently. An
overdraft / unknown-account is `Declined` (the durable ledger floor INV-1
upholds; the in-memory `CreditGate` optimistic hold is not on this path since
there's no admission).

## Gating

Behind the `ledger` cargo feature (default off), and inert unless an operator
attaches an emitter (`attach_ledger_spend`). The ledger subsystem's full
factory wiring (obtaining the sibling `LedgerService` handle at inference
startup) is the documented follow-up — same status as the enforcer-not-yet-
wired-into-the-scheduler. Native spend API only; no `ResourceIntent` (#1065).

## Tests

`cargo test -p hyprstream --features ledger --lib inference_spend` → **7 pass**:
two-subject account isolation; overdraft declined by the ledger floor;
missing/unknown subject account fails closed (no spend, no leak); zero-token
skip; replay idempotency; transfer-id determinism/distinctness; `build_transfer`
shape. `cargo clippy --all-targets -D warnings` clean.

> Note: the two pre-existing enforcer happy-path tests
> (`admitted_then_post_and_void_roundtrip`,
> `insufficient_credit_rejects_retryable_not_queues`) fail when the `services::ledger`
> subset is run in isolation — they depend on a process-global PQ trust store
> (`install_verify_config`) populated by tests outside that module. They pass in
> the full `cargo test --workspace` run. This PR does not touch enforcer/PQ-store
> code.

## Merge note

Edits `services/inference.rs` (completion path / `execute_stream`). That region
overlaps the D8 (#1260) and D7 (#1261) edits to the same file → **serialize at
merge** (rebase whichever lands second; the conflict surface is the
`execute_stream` closure and `PendingWork::Generation`).

### DIFF
--- a/crates/hyprstream/src/services/inference.rs
+++ b/crates/hyprstream/src/services/inference.rs
@@ -57,9 +57,12 @@ use tokio::runtime::Handle;
 use tokenizers::Tokenizer;
 use tracing::{debug, error, info, trace, warn};
 
-// #1264: completion-time ledger spend (behind the `ledger` feature, default off).
+// #1264/#1265: completion-time ledger spend (behind the `ledger` feature,
+// default off).
 #[cfg(feature = "ledger")]
-use crate::services::ledger::{observe_spend_result, InferenceSpendEmitter, SpendInput};
+use crate::services::ledger::{
+    observe_spend_result, InferenceSpendEmitter, SpendInput, SpendResult,
+};
 
 
 /// Pending work to be executed after REP response is sent.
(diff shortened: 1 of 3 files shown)


ewindisch:
-

Issue #1287, opened 2026-07-24
feature: Add FP8 E4M3FN native KV quantization


ewindisch:
## Scope

Add FP8 E4M3FN native KV quantization: wire format/API/config plus implementation and architecture-specific numerical/performance validation. Broad plumbing: model.capnp, generated bindings, model service mapping, CLI, config, and kv_cache.rs. When added, extend and validate BOTH contiguous and paged encodings.

## Depends on
#1, #2q (merge last; not part of the initial paged+quant delivery).

## Lane / merge order
Serialized kv_cache.rs lane — every issue stacks on the preceding one; do NOT develop in parallel.
Merge order: #1 -> #2 -> #2q -> #4a -> #3 -> #4b -> #5 -> #6 -> #7 -> #8 -> #9 (serialized kv_cache.rs lane; stack, do not parallelize). This issue's position: 11.

## Validation
CPU build/test via the hyprstream-build shared target lease (--no-default-features --features
"gittorrent,xet,otel", no libsystemd locally; CI validates defaults). GPU build only if a change
requires GPU tensor validation. Route ALL cargo through the shared lease; never a bare cargo build.

## Rules
Commit as the configured git identity, NO AI-attribution trailers. Independent review (kiro-cli
gpt-5.6-sol) must clear before merge. Do not touch ds4 or ~/ds4-worktrees. hyprstream-KV only.

---
Part of Epic #1276 — Native inference KV cache reuse and durability.

Issue #1278, opened 2026-07-24
feature: Add explicit paged-KV runtime configuration and pool geometry derivation


ewindisch:
## Scope

Add opt-in paged-KV runtime configuration and pool geometry derivation: validate incompatible combinations; derive block count from the budget and model geometry (KV-head count, head dim, dtype, layers, K/V factor, BLOCK_SIZE=256). Define whether the budget means reserved pool bytes, live allocated bytes, or both. Touches RuntimeConfig, service/local init, torch_engine.rs. Must NOT select paged storage yet.

## Depends on
#1 (recommended).

## Lane / merge order
Serialized kv_cache.rs lane — every issue stacks on the preceding one; do NOT develop in parallel.
Merge order: #1 -> #2 -> #2q -> #4a -> #3 -> #4b -> #5 -> #6 -> #7 -> #8 -> #9 (serialized kv_cache.rs lane; stack, do not parallelize). This issue's position: 2.

## Validation
CPU build/test via the hyprstream-build shared target lease (--no-default-features --features
"gittorrent,xet,otel", no libsystemd locally; CI validates defaults). GPU build only if a change
requires GPU tensor validation. Route ALL cargo through the shared lease; never a bare cargo build.

## Rules
Commit as the configured git identity, NO AI-attribution trailers. Independent review (kiro-cli
gpt-5.6-sol) must clear before merge. Do not touch ds4 or ~/ds4-worktrees. hyprstream-KV only,
decoupled from ds4.

---
Part of Epic 1276 — Native inference KV cache reuse and durability.

Issue #1276, opened 2026-07-24
feature: Native inference KV cache reuse and durability


ewindisch:
## Problem
hyprstream's native Torch inference KV cache has per-owner contiguous storage with single-session
linear prefix reuse, a two-tier GPU->CPU LRU, Int8/NF4/FP4 quant (bnb-gated), and paged block
storage that is implemented but never selected in production. There is no cross-request prefix
sharing, no durable persistence, and no cache compatibility identity. This epic brings the KV cache
to production reuse + durability — hyprstream-KV only, decoupled from ds4.

## Scope (serialized kv_cache.rs lane; merge order below)
- [ ] #1277 — #1 Define native KV cache compatibility key + rejection policy (foundational)
- [ ] #1278 — #2 Add paged-KV runtime config + pool geometry derivation
- [ ] #1279 — #2q Add quantized paged-KV storage (resolve the paged+quant mutex)
- [ ] #1280 — #4a Make paged-KV exhaustion transactional before activation (gates #3)
- [ ] #1281 — #3 Wire paged block pool into native serving init (production flip; #2q+#4a+#3 ship together)
- [ ] #1282 — #4b Paged-KV accounting, observability, benchmarks, load coverage
- [ ] #1283 — #5 Refactor paged block ownership for immutable shared-prefix references
- [ ] #1284 — #6 Compatibility-scoped full-block prefix index + attach path (the reuse win)
- [ ] #1285 — #7 Native KV snapshot codec + compatibility validation
- [ ] #1286 — #8 Local durable native KV store + serving lifecycle
- [ ] #1287 — #9 FP8 E4M3FN native KV quantization

## Merge order
#1 -> #2 -> #2q -> #4a -> #3 -> #4b -> #5 -> #6 -> #7 -> #8 -> #9. All touch kv_cache.rs -> stack,
do not parallelize. #2q + #4a + #3 are ONE release deliverable (no FP/BF16-only paging release).

## Key invariants
- Paged activation (#3) is GATED on transactional all-layer exhaustion (#4a) AND paged+quant (#2q).
  LayerKVCache::update today rolls back only the just-allocated key block; layers update
  sequentially so exhaustion can diverge layer seq positions — wait/retry is unsafe until #4a.
- Cross-tenant prefix sharing requires MAC + the #1246 verified tenant claim; tenant-safe reuse is
  not claimed before those land.
- #1260 (TTT->KV invalidation) is already done (PR #1260); any new shared/persisted cache must
[...]

PR #1289, opened 2026-07-24
kv-cache: define native compatibility fingerprint + rejection policy

### LINKED ISSUE #1277: feature: Define native KV cache compatibility key and rejection policy
## Scope

Add a versioned, testable KV-cache compatibility descriptor/fingerprint. Authoritative inputs include effective base weights + adapter/delta generation, model identity, architecture, layer/KV-head/head-dim geometry, context/RoPE settings, dtype, KV encoding/quant mode, BLOCK_SIZE, tokenizer identity, and format version. Define a rejection policy: KV from an incompatible config is never reused. New small module + torch_engine.rs + config/model metadata. Land first.

## Depends on
None (foundational — all reusable/durable lookups consume it).

## Lane / merge order
Serialized kv_cache.rs lane — every issue stacks on the preceding one; do NOT develop in parallel.
Merge order: #1 -> #2 -> #2q -> #4a -> #3 -> #4b -> #5 -> #6 -> #7 -> #8 -> #9 (serialized kv_cache.rs lane; stack, do not parallelize). This issue's position: 1.

## Validation
CPU build/test via the hyprstream-build shared target lease (--no-default-features --features
"gittorrent,xet,otel", no libsystemd locally; CI validates defaults). GPU build only if a change
requires GPU tensor validation. Route ALL cargo through the shared lease; never a bare cargo build.

## Rules
Commit as the configured git identity, NO AI-attribution trailers. Independent review (kiro-cli
gpt-5.6-sol) must clear before merge. Do not touch ds4 or ~/ds4-worktrees. hyprstream-KV only,
decoupled from ds4.

---
Part of Epic 1276 — Native inference KV cache reuse and durability.

ewindisch:
## Summary

Implements **#1277** — the foundational first item of **Epic #1276** (Native inference KV cache reuse and durability): a versioned, testable KV-cache **compatibility descriptor/fingerprint** plus a **rejection policy** that ensures KV produced under an incompatible config is never reused.

KV states are a pure function of the weights and every parameter that shapes the attention computation. Reusing KV produced under one configuration inside a request running under another is a silent correctness bug. This PR defines the authoritative identity of that function and gates reuse on it.

### What the fingerprint captures
Every field of `KvCompatDescriptor` is derived from authoritative model/runtime state — never a human label:
- **Effective weights** — base-weight revision + adapter/delta generation
- **Model identity** — name, architecture (`model_type`), config version
- **Attention geometry** — layers, KV heads, attention heads, head dim, hidden size (K/V tensor *shape*)
- **Position encoding** — RoPE theta, RoPE scaling, max position embeddings (K *values* post-RoPE)
- **Compute dtype**, **KV encoding/quant mode**, **block size** (`BLOCK_SIZE`)
- **Context cap**, **tokenizer identity** (vocab size + digest of tokenizer bytes)
- **Format version** of the fingerprint itself

The descriptor is reduced to a content-addressed SHA-256 `KvCompatFingerprint` over a canonical, byte-stable encoding (length-/tag-framed), so persisted fingerprints are portable across builds.

### Rejection policy
A cache whose fingerprint differs from the current config in **any** authoritative field is **rejected** — its KV is discarded and recomputed. This is the **pull-based** reuse guard, checked at the session-cache swap point (`TextStream::new_with_delta`). It **complements** (does not replace) the registry's existing push-based TTT delta invalidation (`invalidate_for_tenant`), which is preserved unchanged — the descriptor carries an `adapter_generation` slot that a live counter (#1260) will feed so the pull path also covers per-request delta changes.

### Wiring
- New module `crates/hyprstream/src/runtime/kv_compat.rs` — owns **no** `tch`/Cap'n-Proto dependency (pure data + hashing), so it is fully unit-testable without a loaded model (13 unit tests).
- `torch_engine.rs`: descriptor built from `ModelConfig` + `RuntimeConfig` at model load, finalized with the tokenizer identity, and enforced at the reuse point. A `compat_fingerprint` stamp added to `KVCacheManager`.

### Out of scope (separate lane issues)
Deliberately does **not** implement paging, cross-session prefix sharing, or persistence — those are later issues (#1278–#1287) that stack on this foundation.

Closes #1277
Part of Epic #1276.

> **Serialized `kv_cache.rs` lane** — this is position #1; every later KV-reuse issue stacks on it. Merge order per the epic: #1 → #2 → #2q → #4a → #3 → #4b → #5 → #6 → #7 → #8 → #9.

### DIFF
--- a/crates/hyprstream/src/runtime/kv_cache.rs
+++ b/crates/hyprstream/src/runtime/kv_cache.rs
@@ -32,6 +32,7 @@ use tch::Device;
 use tch::{Kind as DType, Tensor};
 
 use super::KVQuantType;
+use super::kv_compat::KvCompatFingerprint;
 use super::torch_utils::{estimate_tensor_size_mb, safe_zeros};
 
 // ============================================================================
@@ -1930,6 +1931,11 @@ pub struct KVCacheManager {
     cached_token_ids: Vec<i64>,
     /// Where this cache's tensors currently reside
     location: CacheLocation,
+    /// Compatibility fingerprint under which this cache's KV was produced
+    /// (#1277). `None` until the cache is first populated; the engine stamps it
+    /// at reuse time and rejects (clears + recomputes) on mismatch against the
+    /// current expected fingerprint.
+    compat_fingerprint: Option<KvCompatFingerprint>,
 }
 
 impl KVCacheManager {
@@ -1956,6 +1962,7 @@ impl KVCacheManager {
             access_count: AtomicU64::new(0),
             cached_token_ids: Vec::new(),
             location: CacheLocation::Gpu,
+            compat_fingerprint: None,
         }
     }
 
@@ -1990,6 +1997,7 @@ impl KVCacheManager {
             access_count: AtomicU64::new(0),
             cached_token_ids: Vec::new(),
             location: CacheLocation::Gpu,
+            compat_fingerprint: None,
         }
     }
 
@@ -2033,6 +2041,20 @@ impl KVCacheManager {
         for mut cache_ref in self.layer_caches.iter_mut() {
             cache_ref.clear();
         }
+    }
+
+    /// The compatibility fingerprint stamped on this cache, or `None` until the
+    /// cache is first populated (#1277).
+    pub fn compat_fingerprint(&self) -> Option<KvCompatFingerprint> {
+        self.compat_fingerprint
+    }
+
+    /// Stamp (or re-stamp) the compatibility fingerprint under which this
+    /// cache's KV is being produced. Called by the engine at reuse time: a fresh
+    /// cache is stamped with the current config; an incompatible cache is
+    /// cleared and re-stamped after its stale KV is discarded.
+    pub fn set_compat_fingerprint(&mut self, fp: KvCompatFingerprint) {
+        self.compat_fingerprint = Some(fp);
     }
 
     /// Enable or disable caching
--- a/crates/hyprstream/src/runtime/kv_cache.rs
+++ b/crates/hyprstream/src/runtime/kv_cache.rs
@@ -2673,6 +2673,71 @@ mod tests {
         Ok(())
     }
 
+    /// The cache-level lifecycle the KV-compat reuse guard (#1277) relies on:
+    /// `cached_token_count() == 0` is the "provably empty" signal, and a populated
+    /// cache with no stamp must be rejected (DiscardAndRecompute), never adopted.
+    #[test]
+    fn test_kv_compat_fail_closed_lifecycle() {
+        use crate::runtime::kv_compat::{
+            decide_cache_reuse, CacheReuseDecision, KvCompatDescriptor, WeightIdentity,
+            KV_COMPAT_FORMAT_VERSION,
+        };
+
+        let expected = KvCompatDescriptor {
+            format_version: KV_COMPAT_FORMAT_VERSION,
+            weights: WeightIdentity::base_only("sha256:base"),
+            tokenizer_hash: Some("deadbeef".to_owned()),
+            ..Default::default()
+        }
+        .fingerprint();
+
+        let mut manager = KVCacheManager::new(2, 64, KVQuantType::None);
+
+        // (1) Fresh cache: empty, un-stamped. An authoritative identity is available
+        //     → stamp fresh and let prefill populate it.
+        assert_eq!(manager.cached_token_count(), 0);
+        assert!(manager.compat_fingerprint().is_none());
+        assert_eq!(
+            decide_cache_reuse(Some(&expected), manager.compat_fingerprint().as_ref(), true),
+            CacheReuseDecision::StampFresh
+        );
+        manager.set_compat_fingerprint(expected);
+
+        // (2) Cache now populated but its stamp matches → reuse.
+        manager.set_cached_tokens(vec![1, 2, 3, 4]);
+        assert_eq!(
+            decide_cache_reuse(
+                Some(&expected),
+                manager.compat_fingerprint().as_ref(),
+                manager.cached_token_count() == 0
+            ),
+            CacheReuseDecision::Reuse
+        );
+
+        // (3) A populated cache carrying NO stamp (e.g. produced before stamping
+        //     existed) must be rejected, never silently adopted.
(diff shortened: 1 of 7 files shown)


ewindisch:
**Revise-2 applied** — new head `80cef5d23` (was `182d73b0f`). Fixes the two remaining rereview findings + the MINOR.

### F1 — authoritative, fail-closed base-weight digest (fully fixed)
- **Hashes the loader's RESOLVED bytes**, not the raw file. LFS/XET pointer resolution is now shared with the loader: new `ModelFactory::resolve_weight_for_digest` (+ a single shared `try_resolve_lfs_pointer` that the loader also uses), so the digest and the deserializer walk one resolution path — it never hashes a pointer stub, eliminating the dual-authority/TOCTOU gap.
- **Fail-closed on every error**: `read_dir` entry errors propagate via `?` (no more `filter_map(Result::ok)`); missing shards, unreadable shards, and unresolvable pointers all return `Err` → `kv_compat = None` → reuse declines.
- **Canonical framing**: shard count + length-prefixed filename + per-shard content digest (pure `kv_compat::{shard_content_digest, base_weight_digest_from_shards}`). Shard segmentation/filenames can no longer collide. Large ordinary shards stream in 64 KiB chunks.

### F3 — effective model config that alters hidden states / downstream KV (fully fixed)
Added to `KvCompatDescriptor` (hashed, compared, per-field divergence tests): `rms_norm_eps`, `layer_norm_eps`, `hidden_activation`, `scale_embeddings`, `query_pre_attn_scalar`, MoE routing (`MoeRouting`: `is_moe`/`num_experts`/`num_experts_per_tok`/`moe_intermediate_size`/`shared_expert_intermediate_size`), and local-RoPE / sliding-window (`rope_local_base_freq`, `sliding_window`). Plumbed `sliding_window` + `rope_local_base_freq` into `ModelConfig` (previously parsed only inside `gemma.rs`) so the descriptor has one source of truth.

### MINOR — layer_types canonicalization parity
`check_compatibility` now compares canonicalized `layer_types`, matching `put_str_vec`'s hash, so casing/whitespace aliases yield equal fingerprints **and** a compatible structured policy.

### Reconfirmations
Format bumped **v2 → v3** (encoding gained fields); golden digest re-pinned. Byte-stability holds (golden pins canonical bytes), reuse stays fail-closed, `invalidate_for_tenant` untouched, foundation-only scope preserved, **no #1260 regression** (adapter-generation path unmodified).

### Validation
Via the shared `hyprstream-build.sh` lease, `HYPRSTREAM_LIBTORCH_VARIANT=rocm71`, `--no-default-features --features gittorrent,xet,otel`:
- `cargo check -p hyprstream` — PASS
- `kv_compat` — 23/23 (incl. same-basename/diff-bytes, shard-boundary, shard-count, filename, determinism; layer_types + hidden_activation canonicalization parity; golden)
- `kv_cache` — 16/16
- `torch_engine::tests::weights_content_digest` — 8/8 (determinism + parity-with-pure, order-independence, empty/no-safetensors fail-closed, small-plain-shard hashed as content, **LFS pointer fail-closed**, **unreadable shard fail-closed**)
- `model_config` — 9/9

ewindisch:
Revise-3 pushed (`80cef5d23` → `7d8373a74`): fixes the rereview-2 blocker — the base-weight digest no longer enumerates every `.safetensors` in the model directory.

**Fix:** `weights_content_digest` now resolves its shard set through the loader's own selector, `ModelFactory::find_shard_files` (`model.safetensors.index.json` → `model.safetensors` → shard-name glob), and digests **exactly** the loader-selected set — no more, no less. An extra/unreferenced `.safetensors` (stale shard, alt-precision copy) can no longer shift the fingerprint while the loaded weights are identical.

**Fail-closed:** zero selected shards, or an index referencing a shard missing from disk, declines to mint a descriptor (`kv_compat = None`) — reuse stays rejected. Read/resolve fail-closed paths, LFS/XET pointer transparency (shared `resolve_weight_for_digest`), and the canonical framing are unchanged.

**New tests** (`torch_engine::tests::weights_content_digest_*`):
- digest invariant to an unreferenced extra shard (sharded-with-index and single-file layouts)
- mutating a selected shard diverges; add/remove/mutate of an unreferenced shard does not
- fail-closed on zero selected shards and on an index-referenced shard missing from disk

**Gate** (shared-lease build, rocm71 — only cached libtorch variant locally, `--no-default-features --features gittorrent,xet,otel`): `cargo check`/clippy no errors; digest **13/13**, kv_compat **24/24**, kv_cache **17/17**, model_config **9/9**.

Not merged; awaiting independent review.

ewindisch:
Revise-4 pushed (`7d8373a74` → `cdf5a6b94`) — addresses re-review-3 blocker.

**Blocker — manifest-less multi-shard glob selection is now fail-closed for the digest (FIXED)**

Loader permissiveness is now separated from fingerprint authority:

- `ModelFactory::find_shard_files_for_digest` (model_factory.rs) is a new digest-authoritative variant of the loader's shard selector (index.json → model.safetensors → shard glob). It REJECTS an ambiguous manifest-less multi-shard glob set regardless of `HYPRSTREAM_STRICT_LOADER` — the loader keeps its warning fallback, but no authoritative identity is minted over a set a filename glob reconstructed.
- `weights_content_digest` (torch_engine.rs) selects through it, so the ambiguous case leaves `kv_compat` unset and KV reuse is declined (fail-closed), never digested.

**Tests** (torch_engine::tests::weights_content_digest_*):

- NEW `manifestless_multi_shard_fails_closed` — two glob-selected shards, no index.json → digest errs / reuse declines (exercises the permissive loader path; env does not set STRICT_LOADER).
- NEW `single_globbed_shard_passes` — a single globbed shard is unambiguous and still digests.
- Existing multi-shard fixtures (determinism/parity, divergence, order-independence) now carry an index.json manifest; all other prior tests unmodified.

**Gate green** (rocm71 — only cached libtorch on this host; `--no-default-features --features gittorrent,xet,otel`): lib tests 63/63 across weights_content_digest (15), kv_compat, kv_cache, model_config, model_factory; clippy no errors (pre-existing warnings only). All prior fixes intact: resolved-byte hashing, fail-closed read/resolve/zero/missing-index-shard, v3 effective-config fields, byte-stability golden, tokenizer/unstamped fail-closed, #1260 invalidate_for_tenant, foundation-only scope. NOT merged — awaiting next independent review.

PR #1298, opened 2026-07-25
ci(nextest): serialize+retry ALL authority-spawning tests (not just one)

ewindisch:
#1295 covered only 1 of 3 tests sharing the self-re-exec + spawn_production_authority_process mechanism. The two siblings (committed_marker_failures_never_reinitialize_from_mutable_authority, composite_ledger_survives_joint_rotation_restart_and_drain_expiry) were un-serialized (running concurrently with the suite and each other) and un-retried — latent flakes. Broaden the production-authority override to the whole class. Follow-up filed to actually harden the shared spawn coordination.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -17,9 +17,9 @@ slow-timeout = { period = "60s", terminate-after = 3 }
 production-authority = { max-threads = 1 }
 
 [[profile.ci.overrides]]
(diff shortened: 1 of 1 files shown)


ewindisch:
Superseded by #1300 (quarantine) — the flaky authority tests no longer run in the merge-gate, so the broadened serialize+retry is moot and would conflict with #1300's nextest.toml change. Re-serialization can be re-added when #1299 re-enables the tests.

PR #1344, opened 2026-07-26
feat(discovery): add hybrid DID operation log

### LINKED ISSUE #1167: C1: DidOp log format + verifier library (the did:web rotation-history bill)

ewindisch:
## Summary

- add a canonical DAG-CBOR genesis and rotation operation format
- derive did:plc identifiers from signed genesis operations and CID-chain later operations
- require nested Ed25519 + ML-DSA-65 signatures at every authorization position
- fail closed on broken links, sequence gaps, unauthorized keys, bad signatures, or classical-only operations

## Validation

- cargo test -p hyprstream-discovery did_op --lib
- cargo test -p hyprstream-discovery --lib --tests --no-fail-fast
- cargo clippy -p hyprstream-discovery --all-targets -- -D warnings
- repository pre-commit Clippy hook

Closes #1167

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6987,6 +6987,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_yaml",
+ "sha2 0.10.9",
  "tempfile",
  "tokio",
  "tracing",
--- a/crates/hyprstream-discovery/Cargo.toml
+++ b/crates/hyprstream-discovery/Cargo.toml
@@ -27,6 +27,9 @@ hyprstream-rpc-derive = { path = "../hyprstream-rpc-derive" }
 # #524 P1 — placement directory ingestion: CAR parsing + MST walk + record
 # decoding (ai.hyprstream.placement.* lexicons).
 hyprstream-pds = { path = "../hyprstream-pds" }
+# #1167 — transport-free Ed25519 + ML-DSA-65 composite signatures for the
+# append-only did:plc operation log.
+hyprstream-crypto = { path = "../hyprstream-crypto" }
 
 # Async runtime
 async-trait = { workspace = true }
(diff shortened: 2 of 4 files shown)


ewindisch:
**Review-complete → queued.** Independent adversarial PASS (exact head 89dce6b7). Exhaustive chain-integrity hunt: exact prev-CID + seq+1 on every adjacency defeats forged/reordered/gapped/cross-chain-spliced logs; ops verify only under `previous.rotation_keys` (introduced keys can't self-authorize, superseded keys can't return); both Ed25519+ML-DSA-65 required under one paired key with require_pq=true (no classical-only/cross-pair path); empty/unknown/noncanonical/broken/bad-sig all error; deterministic DID from canonical signed genesis, pinned. CI green. The one boundary — a pure verifier can't distinguish a stale-prefix/authorized-fork without an external head/witness — is exactly the **deferred witness-anchoring (#1171)**, not a gap; it returns head_cid+sequence for that policy. Verdict: .fleet-coord/1344-verdict-codex-adversarial.md.

Issue #1342, opened 2026-07-26
Epic: demo identity layer — did:plc directory + federation intake + PDS registration + global inventory


ewindisch:
Demo-scope identity layer (federation intake IN). Three layers: **did:plc directory** (op-log authority: genesis + manual registration + federation intake) → **PDS records** (authoritative, did:web host-form #1158) → **AppView index** (pglite/Postgres global inventory, derived). 

**Invariant:** tenant = the DID's LOCAL hosted-account binding, authority-resolved, never client-asserted; **FEDERATED identity ⇒ NO local tenant, fail-closed** (already coded: strip_federated_tenant, resolve_tenant_for_hosted_did→Ok(None) for federated-only).

**Backend (commander):** (A) did:plc directory service — extend #1195 → operable (genesis/manual/resolve); (B) federation intake #1161/#1174 (resolve foreign did:plc/did:web, intake, enforce federated=no-tenant); (C) PDS registration API (mint DID+PDS #1158, return discovery metadata); (D) AppView inventory read-model (pglite/Postgres, product#1).

**Frontend (frontline-commander):** hosts-page→PDS-registration; global-inventory UI; federation-add UI; seam alignment.

Substrate: #1195 ✓ · #1161 · #1174 · #1158 · #1167-1171 · #1242 · cyberdione-product#1. Spec: .fleet-coord/identity-layer-demo-spec.md

ewindisch:
**Operator greenlit the FULL register+directory flow.** Substrate lanes spawned: `1167-didop` (DidOp log → directory A), `1158-hostedpds` (hosted-PDS mint → registration C). `identity-dir` continues on B (federation intake). Frontend (frontline-commander) F+G+H are done + Chrome-verified, parked on these backend contracts. 404-leak-safe denial ruling logged + binds backend (D/resolve return 404-style on unseen objects). did:unknown = reserved sentinel (unauthenticated floor), never registerable.

ewindisch:
## Milestone: backend substrate through adversarial review + queued (2026-07-26)

All four substrate PRs + tenant-hardening are reviewed (every one with ≥1 **routed adversarial** pass) and in the merge queue:
- **#1343** B federation intake — 2 PASS (sol found a JWT-layer sentinel third door → fixed → re-PASS)
- **#1346** #1341 PQ-hybrid trust root — 3 PASS (third-door hunt found no alternate non-hybrid seam; nit→#1347)
[...]

PR #1345, opened 2026-07-26
feat(pds): hyprstream-hosted account mint (host-form did:web) (#1158)

ewindisch:
Mints a hosted account: host-form did:web DID doc + op-log + PDS account, tenant bound AUTHORITY-SIDE (never client-asserted), returning {handle,did,pdsEndpoint,quicUrl,certHash} for the browser register flow. Substrate for registration deliverable C of epic #1342.

### DIFF
--- a/crates/hyprstream-pds-service/src/hosted_account_mint.rs
+++ b/crates/hyprstream-pds-service/src/hosted_account_mint.rs
@@ -1,0 +1,549 @@
+//! Authority-bound hosted PDS account minting.
+//!
+//! The client request carries only a requested handle. The deployment
+//! authority allocates both the permanent host-form DID and the local tenant;
+//! neither value can be asserted in the registration payload. A successful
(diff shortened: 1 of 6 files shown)


ewindisch:
**Review-complete → queued.** Two independent PASSes: sol first-pass + kimi 2nd adversarial (hunted the client-influenced-tenant class across every vector, found no door). Verified: tenant is authority-allocated only — request carries handle only, `from_client_fields` hard-rejects any client tenant member pre-allocation (tested with 0 authority calls); handle is a strict LDH label bound to the did:web first label (no traversal/spoof); did:unknown non-mintable (validate_host_form_did_web requires ≥2 LDH labels); all signed artifacts composite hybrid (both halves pinned, require_pq); discovery metadata derives from live transport with no client fields; no SSRF/network-I/O; never-reuse allocator + fail-safe label-burn. CI green. Verdict: .fleet-coord/1345-verdict-kimi-adversarial2.md.

Issue #1341, opened 2026-07-26
security(discovery): deployment trust root is classical Ed25519-only under mandatory-hybrid posture (ca_key_from_capsule discards ML-DSA)


ewindisch:
**Security — verified on origin/main by metal + frontline-commander; flagged first by the operator ("that sounds like classical-only crypto?").**

## Gap
The deployment-trust root is **classical Ed25519-only**, contradicting the current documented posture (`docs/cryptography-architecture.md`: "Verification rule — mandatory fail-closed hybrid"; unanchored ⇒ "reject before payload use | **no self-certification or classical floor**").

## Chain (top of the deployment trust is classical; everything under it is mandatory-hybrid)
1. GATE/hybrid-verified capsule → **`ca_key_from_capsule()`** (`crates/hyprstream-discovery/src/did_anchored.rs:253`) takes the **hybrid** subject key but returns **only `ed25519_pub`** (32 bytes) — the **ML-DSA-65 half is discarded**.
2. That Ed25519 CA authenticates an **EdDSA-only** `registry-service.jwt` (verifier asserts `alg == EdDSA`; `RegistryDeploymentVerifier` retains only an `ed25519_dalek::VerifyingKey`; the OS root pin is exactly 32 Ed25519 bytes).
3. That JWT selects the **registry authority**…
4. …which certifies **hybrid** discovery records.

So the authority that gates hybrid records is itself classically-authenticated.

## Root cause
**#1157 fixed the PROVENANCE of the CA key** (sourcing it from the hybrid-signed capsule) **but not the SECURITY OF THE NEXT SIGNATURE** — the extracted CA key and the registry JWT it authenticates remain classical.

## Fix (design)
Make the deployment trust root hybrid end-to-end: `ca_key_from_capsule` retains the ML-DSA-65 half (hybrid CA key), the registry-service credential is a **hybrid (EdDSA + ML-DSA-65) COSE/JWT signature**, `RegistryDeploymentVerifier` verifies the composite, and the OS root pin carries both keys (not just 32 Ed25519 bytes). Update `validate_registry_deployment_credential_profile` (alg) accordingly. Preserve fail-closed.

## Relevance
[...]

PR #1348, opened 2026-07-26
feat(discovery): operable did:plc directory (#1342 deliverable A)

ewindisch:
Extends the read-only did:plc resolver into an operable directory over the merged DidOp log (#1344). Self-service + operator-manual Hybrid genesis pinned to the authenticated #1341/#1346 deployment root; canonical signed {did,genesisCid,pdsEndpoint} binding; full log+binding re-verification on every resolve; live per-call {quicUrl,certHash} (no cert pin stored); did:unknown fail-closed; rotation/witness stubbed (deferred #1168-1171). Self-adversarial pass caught+fixed an unsigned-endpoint substitution. 110 discovery tests + strict Clippy pass.

### DIFF
--- a/crates/hyprstream-discovery/src/did_op.rs
+++ b/crates/hyprstream-discovery/src/did_op.rs
@@ -107,7 +107,7 @@ impl HybridRotationKey {
         Self::new(ed25519, required(value, "mldsa65Pub")?.as_bytes()?.to_vec())
     }
 
-    fn verifying_keys(&self) -> Result<(VerifyingKey, MlDsaVerifyingKey)> {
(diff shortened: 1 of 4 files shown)


ewindisch:
**Review-complete → queued.** Two adversarial passes (self + independent). Independent hunt closed every door: {did,genesisCid,pdsEndpoint,type} all in signable_bytes with canonical re-encode equality + both composite halves required (classical-stripped fail-closed); resolve re-verifies the FULL log+binding every call (hash-derived DID==expected, prev-CID links, seq+1, prior-state auth; tamper tests fail); operator-manual genesis pinned to the #1346 hybrid root (wrong-root/classical-only fail-closed); did:unknown gated on both create AND resolve; live per-call discovery never stored. No tenant anywhere. CI green. Non-blocking nit → #follow-up.

PR #1351, opened 2026-07-26
feat(discovery): AppView inventory read-model (#1342 deliverable D)

ewindisch:
pglite-rs AppView inventory index: InventoryEntry{did,handle,kind,tenant(local-only),pdsEndpoint} + query/filter API + mandatory server-side clearance pre-filter (404-leak-safe, post-filter counts only, did:unknown never listable). Derives federated entries from merged federation_intake; directory(A #1348)/hosted-PDS sources behind a clean stubbed interface until A lands. Self-review: Kimi K3.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3638,7 +3638,7 @@ dependencies = [
  "hex",
  "itertools 0.14.0",
  "log",
- "md-5",
+ "md-5 0.10.6",
  "rand 0.9.2",
  "regex",
  "sha2 0.10.9",
@@ -4419,6 +4419,7 @@ dependencies = [
  "block-buffer 0.12.0",
  "const-oid 0.10.2",
  "crypto-common 0.2.1",
+ "ctutils",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


ewindisch:
**Review-complete → queued.** Independent adversarial PASS. Existence-leak: NONE — the dominance predicate is IN SQL before LIMIT (exactly SecurityLabel::can_access with ascending discriminants); bare-array response (no total/hiddenCount, tested); DID cursor over ORDER BY did inside the filtered WHERE reveals nothing about hidden rows; parameterized filter over dominated rows only; did<>sentinel in WHERE. Tenant enforced at 4 fail-closed layers (constructor validate, DB CHECK, row-decode bail, refresh write-down guards) — a federated refresh cannot demote a local row or change its tenant (tested). Strictly a derived index (no keys/write-back/authority). No quicUrl/certHash on row or wire. did:unknown unlistable at 4 layers. Stubs fail closed; absent viewer clearance floors to Public/Unverified.

PR #1333, opened 2026-07-26
feat(mac): wire operator-gated identity activation

ewindisch:
## Summary

- thread MAC clearance from verified `Claims × VerifiedKeyMaterial` instead of unconditional `anonymous_floor()` on authenticated RPC and 9P paths
- wire the six merged PEPs into production RPC, 9P, CAS, VFS, exemptions, and MoQ/event constructors with mandatory fail-closed mediation and signed deny audit
- introduce the unified `VerifiedAttach` bundle so one verified credential binds the exact MAC identity/session, authority-resolved tenant, and VFS `Subject`; backend substitution or subject divergence fails closed
- add the G1-G7 evidence-gated widen control and unconditional narrow-back control; neither removes a monitor
- extend the #1274 gate with the real atproto/OAuth/DPoP/exchangeUcan → RPC+9P positive path and an explicit MAC-identity/VFS-subject divergence denial
- add the operator activation runbook and current Genesis evidence

References #1267, #698, #699, and #1274.

## Activation decision

**Do not enable identity-aware mode.** Startup installs all PEPs but remains `FloorOnly`; no production startup path calls `widen_identity_aware`.

Current default-config Genesis evidence (2026-07-26):

- 93 nodes
- 23 labeled
- 70 unlabeled
- 0 ill-formed
- result: **INCOMPLETE**

The control refuses widening until the Genesis report is complete and the operator supplies reviewed G2-G7 evidence. Narrow-back is always available and retains every monitor.

## Validation

- exact release gate:
  `buildq -- cargo nextest run --release --profile ci -p hyprstream -p hyprstream-rpc -p hyprstream-9p -p hyprstream-vfs --no-fail-fast`
  - **2475 passed, 0 failed, 9 skipped**
  - one preceding contention-heavy run had a process-level SIGABRT in `registry_rpc_is_live_at9p_candidate_boundary`; the test passed immediately in isolation and the locked full rerun above was clean
- mandatory release pre-commit Clippy: passed
- `cargo check` for the four affected packages/all targets: passed
- 9P library tests: 50 passed
- T8 MAC enforcing gate: 3 passed
- OAuth atproto/legacy conformance: passed
- direct rustfmt checks and `git diff --check`: passed
- workspace-wide `cargo fmt --check` remains blocked by the pre-existing `bitsandbytes-sys` parent-workspace metadata mismatch

## Independent review

[...]

PR #1330, opened 2026-07-26
fix(auth): bind policy checks to verified tenants

ewindisch:
## Summary

- bind `PolicyCheck` authorization to the verified envelope subject and authority-resolved hosted-account tenant; caller-supplied subject/domain fields are inert
- fail closed and audit unauthenticated, tenantless, federated-service, and policy-denied checks
- preserve the verified bearer through approved local relay services without inheriting deputy grants
- guard legacy `g(user, role)` membership so it is effective only in the explicit global `*` domain; tenant membership uses `g2(user, role, domain)`
- migrate legacy/unsafe model matchers at load, reject models that cannot be hardened, convert concrete-tenant template groupings to `g2`, and reject multi-tenant templates before mutation
- move federation origin allowlisting into `federation:register:<origin>` resource policy data at all three gates and migrate legacy rules on load, reload, add, and remove
- replace unsafe Casbin prefix wildcard behavior with a fail-closed origin matcher: exact origin or one leading DNS-label `*.`, with scheme and effective-port equality; `federation:register:*` is the sole global-open form

## Step 0: #1303

#1318 already fixed the HTTP-face sites and merged as `c9e99a8c860e8d904a10d9243a67d4b4ddfe3336`. The old `services/oauth/models.rs` and `services/oauth/openai.rs` paths no longer exist; their replacements derive policy domains through the verified hosted-account binding. #1303 was closed as completed, so this PR does not duplicate those changes.

## Rebase and validation

- exact head: `ebd4813557feac6c14ca2a72f668b62fba8619b1`
- base / merge-base: `470dbfa61c3cc7a3d50003d429c79540993c81be`
- GitHub compare: 4 ahead, 0 behind
- pre-commit release clippy mirror: passed
- `buildq -- cargo nextest run --release --profile ci -p hyprstream -p hyprstream-rpc --no-fail-fast`
  - 2,357 passed, 0 failed, 10 skipped
- exact-head Kimi-K3 independent security re-review: **PASS**
  - 27/27 policy-manager, 14/14 PolicyService, and 29/29 RPC delegation focused release tests
  - no blocking findings; one low-severity stale `docs/spiffe.md` legacy example remains safe through migration

Hosted Clippy, CodeQL, cargo-deny, and workers-WASM checks are green. The browser-client WASM job reproduces an inherited failure from this exact main base in untouched `mac/pep.rs` code; this PR does not change `mac/pep.rs` or `translator.rs`.

Closes #1302
Closes #1304
Refs #1303
Refs #1267

PR #1352, opened 2026-07-26
test: install mandatory dispatch PEP fixture across affected tests (#1333 follow-up)

ewindisch:
Systemic follow-up to #1333 activation: tests exercising a policy/dispatch path now fail NoPepInstalled unless they install the test dispatch-PEP fixture. Adds the fixture across all affected tests (started from the key_rotation subprocess test the C lane found + the #1330 policy test). NO assertions weakened — only installs the mandatory PEP the production gate now requires. Clears a recurring merge_group hazard for downstream PRs.

### DIFF
--- a/crates/hyprstream/src/auth/key_rotation.rs
+++ b/crates/hyprstream/src/auth/key_rotation.rs
@@ -2440,7 +2440,11 @@ mod tests {
     }
 
     fn authority_process_dir() -> Option<PathBuf> {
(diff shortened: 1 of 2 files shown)


ewindisch:
Queued. Verified test-only: the key_rotation.rs change is inside `mod tests` (authority_process_dir installs install_explicit_test_dispatch_pep in the fresh authority subprocess — the fix for the subprocess NoPepInstalled failure), and ledger_authenticated_transport.rs just adds `mod support;`. No production logic changed, no assertions weakened; full workspace nextest 3817/3817 + Clippy green. Clears the recurring #1333-activation merge_group hazard.

PR #1353, opened 2026-07-27
feat(oauth): expose browser identity resolve adapter

ewindisch:
Closes the browser-callable gap over #1348 PlcDirectory resolution.

- adds public `GET /api/identity/resolve?did=<did>` with exact `{quicUrl,certHash}` output
- charges the stable unauthenticated-floor rate bucket before resolver/classification I/O
- resolves only authority-known local `did:plc` records and hosted account-zone `did:web` artifacts
- rejects `did:unknown`, external `did:web`, absent, corrupt, and denied identities as indistinguishable 404 responses
- re-verifies hosted account record, Hybrid genesis, sealed DID document, and live QUIC capability
- adds causal browser-route coverage from hosted mint through connect-time resolve

Validation:
- `cargo test -p hyprstream services::oauth::identity_registration::tests --lib -- --nocapture` (18 passed)
- causal hosted browser resolve test (passed)
- `cargo clippy -p hyprstream --lib --tests -- -D warnings` (passed)
- pre-commit release clippy (passed)
- full `cargo test -p hyprstream --lib --tests --no-fail-fast`: adapter and 1,384 other lib tests passed; existing `auth::key_rotation::tests::composite_authority_production_paths_reject_semantic_rollback` fails independently with `NoPepInstalled` on exact rerun (the #1333 fixture cleanup seam)

No merge or self-approval.

PR #1354, opened 2026-07-27
feat(oauth): atproto session-exchange + whoami (identity auth seam, #1342)

ewindisch:
Wires the frontend->hyprstream auth seam (operator: defer www backend/, use hyprstream own exchange). atproto/DPoP session -> authenticated hyprstream session (the hyprstream_session cookie C register/intake require); whoami/session-context endpoint returns viewer server-derived {did,kind,tenant,canActLocally} for client gating (unauth->floor). Built on wasm_token_exchange (#1119) + oauth session module + verified_tenant. Exact contracts in .fleet-coord/route-contracts.md.

### DIFF
--- a/crates/hyprstream-rpc-std/src/wasm_exports.rs
+++ b/crates/hyprstream-rpc-std/src/wasm_exports.rs
@@ -12,10 +12,46 @@
 use std::sync::Arc;
 
 use js_sys::Function;
+use serde::Serialize as _;
 use wasm_bindgen::prelude::*;
 
 // Re-export low-level WASM API (crypto primitives, ZMTP framing) from hyprstream-rpc.
 pub use hyprstream_rpc::wasm_api::*;
+
+/// Exchange an ATProto service-auth JWT plus DPoP proof for hyprstream's
+/// opaque HttpOnly browser session. The service JWT must target
+/// `ai.hyprstream.identity.exchangeSession`.
+#[wasm_bindgen(js_name = "exchangeBrowserSession")]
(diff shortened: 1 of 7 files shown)


ewindisch:
**Adversarial review: PASS → queued.** Forgery: none — identity fully server-side verified (alg pinned ES256/ES256K, iss DID-form, aud=host DID, lxm=exchangeSession NSID, host resolves the issuer DID doc itself, kid=DID fragment, ECDSA verified); replay closed by one-use jti on BOTH the service JWT and DPoP (htm/htu-bound); cookie is opaque 32B random, HttpOnly/SameSite=Lax/Secure-on-https, fresh per exchange. Tenant: authority-resolved only (resolve_tenant_for_hosted_did after verify, under fixed service:oauth subject) — the service JWT carries NO tenant claim so client assertion is structurally unrepresentable; federated→tenant:null/kind:federated/canActLocally:false (e2e). Whoami: derived only from the stored server session; no-cookie/unknown/expired/non-atproto all collapse to the identical unauthenticated floor (fail-safe under-report). did:unknown non-exchangeable. C untouched + still session-gated (exchange→cookie→register works through it; #1353 unauth-rejection tests intact). No fail-open (every deny returns before cookie issuance). Residual (non-blocking, follow-up): DPoP is freshness-bound not cnf-bound to the service JWT; mitigated by one-use jti + 1h cap + TLS. Verdict: .fleet-coord/1354-verdict-kimi-adversarial.md.

Issue #1387, opened 2026-07-27
Epic: resource-accountability plane — ledger, quota PEP, issuance seam (payments core)


ewindisch:
hyprstream-side of the payments/quota/KYC architecture: the product-agnostic MECHANISM + reference monitor + Service traits. The payment rail (Stripe), pricing values, and KYC live in the private cyberdione-product repo and drive this only through the typed seams. **Boundary + pins:** .fleet-coord/REPORT-payments-epic-split.md. **Invariants:** authorize-before-use, one service-core PEP, complete-mediation fail-closed, #1129 spend binding, no KYC PII ingress, settlement-only issuance. Children: PAY-00,01,02,03,04,08,09,10,11,12,14-core. Not demo-critical; parallel to the demo. Enforcement never leaves core; only credit-supply + pricing-values cross the seam.

ewindisch:
## Children (hyprstream core — mechanism/enforcement, pinned)
- **PAY-00** #1388 seam contract + traits (SettlementIssuer/TariffProvider) · **PAY-01** #1389 ledger durability+issuance · **PAY-02** #1390 allocation/UCAN · **PAY-03** #1391 reservation PEP · **PAY-04** #1392 free-floor mechanism
- **PAY-08** #1393 inference · **PAY-09** #1394 training · **PAY-10** #1395 storage leases · **PAY-11** #1396 assurance/MAC · **PAY-12** #1397 route audit · **PAY-14-core** #1398 enforcement validation

Product side (Stripe/pricing/KYC) in cyberdione-product: gitlab.com/cyberdione/cyberdione-product/-/work_items/7. Split+pins: .fleet-coord/REPORT-payments-epic-split.md

Issue #1390, opened 2026-07-27
[PAY-02] Production allocation issuer/resolver + UCAN-chain verifier


ewindisch:
**Repo:** hyprstream. **Pin:** `crates/hyprstream-pds/src/ledger/{mod,allocation,receipt}.rs`. The static verifier is currently test/bootstrap-only (recon) — make production authority real.
**Purpose:** close the static-verifier/bootstrap gap.
**Deliverables:** fetch+validate `ai.hyprstream.ledger.allocation`; verify frozen schema, issuer sig, trusted issuer, UCAN proof-chain/attenuation, audience/host, holder, unit, amount, epoch, expiry, revocation, hybrid-PQ material; payment/free-floor allocation publication outbox; ledger snapshot materialization + `CreditGate` generation invalidation.
**Acceptance:** production wiring refuses the static verifier; stale/missing/revoked allocation/UCAN/authority/snapshot/PQ denies; holder from verified request authority; allocation publication contains no payment/KYC/legal-identity; cache never exceeds committed ledger state.
**Depends on:** PAY-00, PAY-01. Epic: https://github.com/hyprstream/hyprstream/issues/1387

Issue #926, opened 2026-07-09
feat(ledger): cross-cell exchange — tranche protocol + net settlement (D4)


ewindisch:
Part of #922 (design: #921). Decision **D4**; plan §2f–§2g. Phase 2. Depends on sub-issues 1–3.

Cross-cell exchange + tranche protocol + net settlement (correspondent banking; ILP/Rafiki cited as pattern, not dependency):
- [ ] 2.1 `ledger_exchange.capnp` in hyprstream-rpc-std: TrancheOffer/Accept/Active/Abort/Status/Closed, SettlementProposal/Note, ReceiptDiff (M, judgment — protocol surface)
- [ ] 2.2 Tranche state machines on home + remote (escrow accounts, **spendability rule**: remote funds reserve-held and unspendable until home-signed `TrancheActive`; ACTIVE tranches end only via signed close or dispute, never unilateral timeout) + `tranche` lexicon (L, judgment)
- [ ] 2.3 Net settlement: window framing on checkpoints, receipt-set roots, idempotent settlement transfers, ReceiptDiff reconciliation; **dual-signed receipts are the only admissible dispute evidence — a counterparty's ledger balance never is** (L, judgment)
- [ ] 2.4 Cross-cell partial-failure integration matrix with fault-injection harness (plan §6.4) (M, mechanical)

Settlement cadence: every 5 min / 10k receipts / tranche close. Double-spend prevention stays local everywhere: hard cap at home when the tranche is cut, hard cap at remote while spent.

ewindisch:
## Target-lane scope for PAY-ARCH-01 Option 3/5 (reviewer-converged; pending ratification)

This issue is the home of the cross-cell **exchange** layer chosen as the PAY-ARCH-01 target. Scope refinement from the 3-reviewer collation (`.fleet-coord/PAY-ARCH-01-DECISION-collated.md`):

- **Reuse #1400's settlement envelope verbatim.** `settlement.capnp` already carries a `attestation :Data` = PQ-hybrid-signed settlement attestation, with the ledger's canonical DAG-CBOR / `Cid` content-addressing discipline. The exchange record embeds that same attestation blob — no re-encoding.
- **Transport-neutral, carrier ≠ authority.** One canonical, content-addressed, domain-separated, PQ-hybrid-signed envelope delivered over MoQ/iroh **or** a confidential-profile atproto record. Carriers provide availability/portability only; the destination **verifies and imports exactly once** into its own authoritative local ledger.
[...]

PR #1400, opened 2026-07-27
feat(pay): PAY Wave-0 ledger foundation — durable PostgresLedger + #1399 SettlementIssuer/TariffProvider seam crate

### LINKED ISSUE #1389: [PAY-01] Production ledger durability, restricted issuance, activation
**Repo:** hyprstream. **Pin (extend, don't recreate):** `crates/hyprstream-ledger/src/{backend,mem,types,engine,errors}.rs`, `crates/hyprstream/src/services/ledger/{mod,handle,service}.rs`. Replace `MemLedger` with a durable `LedgerBackend`; flip `LedgerConfig.enabled` default-off -> production activation.
**Purpose:** replace the default-off/reference posture with a recoverable production accounting service + the typed `SettlementIssuer` seam.
**Deliverables:** durable backend (transactional journal/pending/outcome/outbox/checkpoint); restart-safe expiry+reconciliation; internal typed issuance API (`SettlementIssuer`); production profile enabling the ledger; startup readiness proof; migrations/backup/restore/key handling; remove externally-reachable generic issuance.
**Acceptance:** restart cannot lose/duplicate reserve/post/void/issue; same transfer-id idempotent; issuance without a valid typed source rejected; production resource services refuse startup when ledger/backend/keys/outbox/checkpoint absent; dev/test insecure modes cannot bind a production listener.
**Depends on:** PAY-00. Epic: https://github.com/hyprstream/hyprstream/issues/1387

### LINKED ISSUE #1399: [PAY-bootstrap] Seam RPC contract (MIT stub) + authenticated SettlementIssuer/TariffProvider service
Bootstrap blocker for cyberdione-product (option B). Deliver: (1) the SettlementIssuer/TariffProvider **RPC contract** (capnp/proto schema + capability model) in an **MIT** crate + client stub the product can link; (2) the **authenticated RPC service** exposing issuance/tariff on the iroh/QUIC mesh — peer auth + capability-gated, PQ-hybrid-signed settlement attestation verified server-side (reuse #1129 spend-auth/MAC). Service impl stays AGPL; only the protocol/stub is MIT. This is what cyberdione-product calls; it gates the product scaffold. Epic #1387. Folds into PAY-00/#1388 + PAY-01/#1389.

ewindisch:
## Summary

PAY Wave-0 foundation: the durable ledger backend + settlement/tariff seam crate that the entire PAY epic stacks on.

### PAY-01 (#1389): Durable PostgresLedger

- **New `PostgresLedger`** impl behind `postgres` feature on `hyprstream-ledger`
- **In-memory mirror + transactional DB writes** (one DB transaction per op for atomicity)
- **Dedicated background thread** with `deadpool-postgres` pool — runtime-independent, deadlock-free (PLAN DECISION 8: sync trait stays sync)
- **SQL schema**: journal/accounts/pending/outcomes/outbox/checkpoints/meta tables
- **Restart-safe**: rehydrates the in-memory mirror from the DB on `connect()`
- **Journal hash chain verification** on startup (`verify_chain`) — fail-closed on mismatch
- **u128 amounts stored as BYTEA** (tokio-postgres has no native i128)
- **Shares `engine::stage` with `MemLedger`** — same ops ⇒ same outcomes by construction

### #1399: `hyprstream-pay` MIT seam crate

- **`SettlementIssuer` + `TariffProvider` traits** — the protocol surface cyberdione-product links against
- **`IssueRequest`/`IssueResponse`/`TariffRequest`/`TariffQuote`** types
- **Capability model**: `ai.hyprstream.pay.{settlement,tariff}.{issue,status,quote,resolve}`
- **Cap'n Proto schema** (`settlement.capnp`)
- **MIT-only** (service impl stays AGPL in `hyprstream::services::pay`)
- PQ-hybrid attestation verified server-side (reuse #1129 spend-auth path)

## Interface report

Full interface design at `.fleet-coord/pay-wave0-interface-report.md`. Key decisions:
- Postgres crate: `tokio-postgres` + dedicated bg thread (runtime-independent)
- Seam crate name: `hyprstream-pay` (MIT protocol surface)
- Account amounts: BYTEA (16-byte big-endian u128)
- `MemLedger` stays as the oracle/test/WASM backend

## What this gates

| Downstream | Dependency |
|---|---|
| PAY-02 (allocation issuer/resolver) | PAY-01 durable backend + #1399 `SettlementIssuer` trait |
| PAY-03 (unified PEP) | PAY-01 backend + PAY-02 issuer |
| Product scaffold (cyberdione-product) | #1399 MIT crate (`hyprstream-pay`) |

## Test plan

- [x] `hyprstream-ledger` builds with `postgres` feature
- [x] `hyprstream-ledger` builds without `postgres` feature (wasm-clean)
- [x] `hyprstream-pay` (MIT seam crate) builds standalone
- [ ] Integration test with live Postgres (requires docker-compose)
- [ ] Equivalence test: same ops sequence against MemLedger and PostgresLedger → same outcomes
- [ ] Restart recovery test: commit → drop → reconnect → verify state
- [ ] sol review (commander pinged)

Closes #1389, closes #1399
Epic: #1387


ewindisch:
## PAY-ARCH-01 guard-rail — acceptance note (no rework required)

The 3-reviewer PAY-ARCH-01 collation selected **Option 1 (this PR) as the demo ship and a strict subset of the Option 3/5 target** — convergence is additive, so nothing here needs reworking. One acceptance criterion to confirm before merge so later convergence stays zero-rework:

- Confirm the `settlement.capnp` `attestation` blob is a **canonical, self-contained, domain-separated** encoding whose **CID is stable**, so a future cross-cell exchange record (#926) embeds it **verbatim** and content-addresses it with no re-encoding. Given the existing `AccountId = blake3_128(canonical_dag_cbor(...))` / `Cid` discipline this is likely already satisfied — please just assert it in review.

This is a review gate, not a change request. The R3 re-review (rev-1400) has been asked to confirm it alongside the replay-idempotency / single-writer / fail-closed fixes.

ewindisch:
## R3 = REVISE → implementer escalated + scope boundary clarified

R3 (sol, crypto) at `5ce248c6` is REVISE. The narrow replay fix + explicit-Postgres-no-Mem-substitution are accepted; the deeper guarantees are not yet met. After 3 REVISE rounds on the same money-path structural findings, the implementer is escalated to opus for the R4 pass.

**Scope boundary (this is why it kept bouncing):** #1400 is the durable, single-writer, safe **ledger crate**. The full authenticated capability-gated **issuance RPC service** (Cap'n Proto bindings, RPC registration, `#[authorize]`, concrete UCAN/PQ verifier, durable settlement store, service factory) is **#1390 [PAY-02]**, a stacked follow-up — not this PR.

**R4 must-fix here:** (1) single-writer-per-cell ordering — one active writer for instance lifetime, or acquire ordering before staging + restage inside it; + live 2-instance Postgres tests; (2) mirror rebuilt strictly from the committed journal on rehydrate (contiguous seq, head match, materialized==journal, adopt committed state on ambiguous commit); (3) **seal the raw `LedgerBackend::credit` mint** at the crate boundary (the safety half — the full service is PAY-02); (4) fatal-on-unavailable + existing-DB BYTEA migration + strict decoder + canonical/domain-separated attestation with stable CID (PAY-ARCH-01 guard-rail).

ewindisch:
## R4 = REVISE → R5 routed (structural fixes accepted; 2 security CRITICALs remain)

R4 at `cbba2786` **materially improved** the hard structural findings: ordinary-replay no-double-apply, balance commits check durable lease-epoch + staged head in one txn, startup rebuilds mirror from journal + cross-checks, explicit-prod fatal activation, BYTEA(16) fresh+migrated, deterministic domain-separated attestation. Those are accepted.

Two CRITICALs block:
1. **Mint not sealed** — `MintVerifier` is a public downstream-implementable trait and `with_mint_verifier` is public on both ledgers; an out-of-crate `AcceptAll` verifier minted **1,000,000 units from meaningless bytes**. Fix: sealed trait / remove public injection; concrete in-crate PQ verifier only in prod; add an adversarial downstream compile-fail/fail-closed test.
2. **Fencing incomplete** — the lease/epoch fence covers `commit` but not `checkpoint()`, `outbox_ack()`, or `tick()`'s clock write; a fenced/zombie writer can still persist those. Fix: route every durable write through one `fenced_write` guard.

Plus: the 16 live-PG tests all pass with head-CAS disabled (revert-verify unproven — tests must fail when the fence is off); strict-ID decoder skips the `ledger_pending` PK.

Not on the demo critical path (PAY Wave-0; demo runs dev-tier free-floor). Gating hard is correct — R4 caught a downstream-crate-mints-1M hole. R5 on opus.

ewindisch:
**New head `e4fa65c4b`** — R5 (mint seal + complete fence) is now on the branch and compiles.

Two commits, and I want to be upfront about the first one:

- `67ad40d62` — the R5 work as it existed when the host crashed on 2026-07-27. It was uncommitted at the time and was recovered verbatim rather than reconstructed. Marked `wip` because that is honestly what it is: a faithful snapshot, not a reviewed state. Preserved as its own commit so the crash boundary stays visible.
- `e4fa65c4b` — the fix that makes the branch compile.

### What R5 does

- **CRITICAL 1 (mint seal).** The `MintVerifier` trait and `with_mint_verifier` setter are **deleted**, not gated — a grep for either now returns only comments describing the old exploit. `MintAuthority` replaces them as *data* (verifying keys + `require_pq`), supplied at backend construction and immutable after. `MintCapability` is constructed in exactly one place, a `pub(crate)` fn. The out-of-crate exploit shape no longer type-checks.
- **CRITICAL 2 (fence).** New private `fenced_write(expect_head, body)` is the only path that persists anything: poison check → per-cell advisory lock → lease epoch **and owner** fence → optional head CAS → body → heartbeat renewal, all in one transaction. `commit`, `checkpoint`, `outbox_ack` and `tick`'s clock write all route through it; standalone `renew_lease` is gone. Renewing inside the same transaction as the work is what closes the authorize-then-write window `tick` had.

### Two things a reviewer should look at directly

1. **One durable write remains outside `fenced_write`:** the lease release in `Drop`. It is conditional on `substring(value from 1 for 24) = <epoch||owner>` and skipped when poisoned, so a zombie cannot touch a lease that was taken over — but it is an unfenced `UPDATE` and you should judge it rather than take my word. (Connect-time schema install is also outside, necessarily — it precedes the lease.)
2. **The live-PG mint tests use `MintAuthority::classical`.** Production is hybrid. The hybrid path is covered by `mint.rs` unit tests and, as of `e4fa65c4b`, by `ledger_authenticated_transport.rs` — but not end-to-end against Postgres.

### What `e4fa65c4b` fixes, and why it was missed

`crates/hyprstream/tests/ledger_authenticated_transport.rs` was never updated for either new signature (`MemLedger::new` arity, `credit(MintCapability)`), so the branch did not compile. The fixture now installs its own hybrid issuance authority and mints via `authorize_mint` — no test-only back door, and the mint keypair is deliberately separate from the holder's.

It escaped R5 because every command used to validate R5 was structurally blind to it: `cargo test -p hyprstream --lib` skips `tests/`, and the `--all-targets` clippy was scoped to `hyprstream-ledger`/`hyprstream-pay`.

### Verification

| command | result |
|---|---|
| `cargo check --all-targets -p hyprstream-ledger --features postgres` | pass |
| `cargo check --all-targets -p hyprstream --no-default-features --features pay,postgres-ledger` | pass |
| `cargo check -p hyprstream ... --test ledger_authenticated_transport` | pass |
| `cargo clippy --workspace --all-targets --release -- -D warnings` (pre-commit hook) | pass |

**Not run: the live-PG suite** (needs `HS_LEDGER_TEST_PG`). So the R5 note's revert-verify numbers are **unconfirmed by me** — that claim still needs an independent run before it should be believed, particularly since R4's equivalent claim was the thing that turned out to be unsound.

Also still unticketed, inherited from R4: the two pre-existing `services::ledger::enforcer::tests` failures, and the `rt_new()`-per-DB-job allocation.

ewindisch:
## Live-PG suite + honest revert-verify (head `e4fa65c4b`)

Ran against a live PostgreSQL 16 with `HS_LEDGER_TEST_PG` set. This settles the claim I explicitly left unconfirmed in my previous comment.

### Suite, guards intact

```
test result: ok. 19 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.22s
```

### Revert-verify — two SEPARATE experiments

Conflating the guards is what made R4's claim worthless, so each was disabled **alone** with the other fully intact. Both live in `fenced_write`, so each is a one-line disable.

**A — head-CAS disabled, lease fence intact** → `17 passed; 2 failed`

```
---- a_commit_staged_against_a_moved_head_is_refused_by_the_cas_alone ----
postgres_live.rs:493:5: a commit staged against a moved head must be refused by the CAS,
got Ok(Applied { posted: 10, released: 0 })

---- a_checkpoint_attesting_to_a_moved_head_is_refused_by_the_cas_alone ----
postgres_live.rs:528:37: called `Result::unwrap_err()` on an `Ok` value:
SignedCheckpoint { ... seq: 3 ... }
```

**B — lease epoch/owner fence disabled, head-CAS intact** → `17 passed; 2 failed`

```
---- a_fenced_out_writer_cannot_commit_and_loses_no_update ----
postgres_live.rs:374:18: expected the fenced-out writer to be refused,
got Err(MirrorStale { mirror_seq: 4, db_seq: 5 })

---- a_fenced_out_writer_cannot_use_any_side_door ----
postgres_live.rs:571:5: a fenced-out writer published a checkpoint: Ok(3)
```

**Restored** → `19 passed; 0 failed`. Working tree clean, nothing committed or pushed from this experiment.

### Verdict

**The R5 note's revert-verify claims are accurate — verbatim.** Both specific strings it quoted reproduce character for character. Given R4's equivalent claim turned out to be unsound, this was worth checking rather than trusting; it held up.

**The tests are load-bearing and the two layers are provably independent.** The failure sets are **disjoint** — the CAS disable breaks only the CAS tests, the fence disable only the fence tests. No test passed with its guard removed, so there is no R4-shaped defect here.

**One nuance worth reading carefully.** In experiment B, `a_fenced_out_writer_cannot_commit_and_loses_no_update` fails on the *wrong error* (`MirrorStale`, from the CAS) rather than on the write succeeding — the zombie's commit was still refused by the second layer, so that path is doubly covered. The genuinely open door with the fence off is the **checkpoint**: `Ok(3)` is a real durable proof-plane row landing from a fenced-out writer. That is CRITICAL 2 reproducing on demand, and it is what justifies routing `checkpoint`/`outbox_ack`/`tick` through `fenced_write`.

### Hybrid mint coverage — still classical-only against Postgres

Still true at this head: `postgres_live.rs:201` uses `MintAuthority::classical` and its `sign_mint` calls `sign_composite(key, None, …)`. Hybrid is covered in `mint.rs` unit tests and (as of `e4fa65c4b`) `ledger_authenticated_transport.rs`, both non-Postgres.

**Low severity, stated accurately:** verification is backend-independent — `mint::authorize` is one shared function and the backend only stores `Option<MintAuthority>` — so a hybrid authority against Postgres exercises code the unit tests already cover. What is unproven is only that hybrid key material survives the `PostgresLedger::connect` path, which is a field assignment. Worth closing for completeness; not a blocker.

ewindisch:
**Independent review (frontline seat) — PASS-WITH-NITS, bound to head e4fa65c4b.**

Money-critical axes verified: u128 amounts end-to-end stored as 16-byte BYTEA with CHECK constraints and a fail-closed decoder (no floats anywhere; tariff math checked_mul); balances derived from the hash-chained journal with startup verify_chain + materialized-vs-journal assertion (divergence/gaps/truncation fatal, live-tested); concurrency proved not hoped — durable lease + pg_advisory_xact_lock + FOR UPDATE + head-CAS in ONE transaction with heartbeat, fenced-out writer permanently poisoned, every durable mutation through that one primitive, dual-instance live tests; DB-side idempotency inside the locked tx with replay returning stored outcomes; static parameterized SQL only; local dev unaffected (ledger disabled by default, feature+config gated, Mem refused in production, missing DSN fatal — no silent fallback); real property tests with conservation and no-overdraft invariants.

Nits (1 and 2 should land before PAY-02 wires real money):
1. Stuck settlement: issue() writes IssuancePending before crediting; a crash (or mark_issued failure) strands the row — retries rejected with SettlementNotCommitted, ok=false forever, customer paid but credits never issued. The ledger credit is already idempotent on the derived transfer_id, so admitting IssuancePending into the retry path is the safe fix.
2. Idempotency keyed on transfer_id alone with no content binding — a reused id with different amount/destination silently returns the earlier outcome as success. Store an op digest in ledger_outcomes and reject mismatches distinctly.
3. The mint signature is self-referential (signs and verifies with the node's own key) — real authorization is the SettlementAuthority type seal; document that so PAY-02 doesn't mistake it for crypto authority.
4. TariffProvider::quote ignores catalog_version despite the doc promise; resolve_unit synthesizes a UnitRef for unknown classes from caller-supplied issuer_did (soft ambient-unit seam); subject_did unused.
5. BackendKind defaults to Postgres, so enabled=true without backend="mem" is a fatal startup for self-hosters — consider defaulting Mem when production=false.
6. Explicit seq into BIGSERIAL leaves the sequence unadvanced — plain BIGINT PRIMARY KEY states the intent.

CONFLICTING vs main — resolving the rebase now; will enqueue on green.

Issue #1401, opened 2026-07-27
Implement server PostgresUserStore (deadpool-postgres, shared USERSTORE_SCHEMA) + backend config switch


ewindisch:
**Gap:** #1385 ships `PgliteUserStore` (embedded PostgreSQL, single-process). The cloud two-AZ multi-PDS deploy needs a SHARED external Postgres both PDSes connect to — pglite (embedded/per-process) cannot be shared across instances/AZs. So the server backend is a **prerequisite for the cloud multi-PDS milestone** (not the local E2E, which uses pglite).
**Deliver:** `PostgresUserStore` running the SAME `USERSTORE_SCHEMA` (portable, pglite=real Postgres) against a deadpool-postgres pool (dep already in-workspace via #1400); a config/feature backend switch (Pglite=from_database(Arc<PGlite>) vs Postgres=connect(pool)) in the services/oauth/mod.rs factory (same pattern as RocksDB/Valkey/Pglite); fail-closed on connection/config absence.
**Backend-agnostic encryption (operator):** #1384's column encryption (ColumnCipher/trust-mint DEK) must apply at the store layer to BOTH pglite AND Postgres — not pglite-specific. The cloud shared-Postgres credential store must be encrypted-at-rest too.
**Metal coordination:** the two-AZ deploy needs a managed Postgres (dovetails with the dedicated-ledger-Postgres decision, PAY-00 #6). Off the local-E2E path; gates the cloud demo. Related: #1385, #1375 epic.

ewindisch:
## Elevated to DEMO-CRITICAL — real Postgres on RDS is blocking

**Operator ruling (2026-07-27):** pglite is ephemeral/localhost-only and must NOT be the deployed account/key store — data loss is unacceptable. The credential/account store deploys to **AWS RDS** (managed Postgres). `PgliteUserStore` stays for local/workstation tests only; `PostgresUserStore` is the deployed backend and is on the **demo critical path** (previously mis-filed as a cloud follow-on).

**Build:** in progress on `feat/1401-postgres-userstore`, stacked on `feat/1377-pii-encryption` (#1384). Consumes #1384's backend-neutral encryption layer verbatim (no crypto reimpl) + the existing PostgreSQL-compatible `USERSTORE_SCHEMA`, via deadpool-postgres/TLS.

[...]

Issue #1257, opened 2026-07-24
1242-A: Accounts as PDS records — OAuth account store reads/writes PDS account records


ewindisch:
Child of #1242 (work item 1). **GATED on #1163 / PR #1229** (account record genesis format — `feat/1163-genesis-didop`, OPEN). Part of the #1158 account-identity epic.

## Goal
The OAuth account store's source of truth becomes **PDS signed records** (the `#1158`/`#1163` account record), not a standing local directory. No standing `UserStore` directory as system-of-record.

## Current state (per recon)
- **No PDS account-record type exists today.** `crates/hyprstream-pds/` exports `ModelRecord` (`ai.hyprstream.model`), `NameRecord`, placement/ledger records, at9p — no `AccountRecord` / `ai.hyprstream.account`. All account-as-PDS-record work is #1158/#1163 future work.
- The existing `record.rs` `ModelRecord` pattern (3-field lexicon, DAG-CBOR, MST insertion, CAR proofs) is the template.
- Profile writes today (`UserStore::register/set_profile/add_pubkey/...` in `auth/user_store.rs`) become PDS commits.

## Scope
- Define the account record read/write against the #1163 genesis format (coordinate with #1229 author).
- Repoint `UserService` (`services/oauth/user_service.rs`) profile/key writes at the PDS commit path.
- Coordinate with #1163 (genesis), #1158 (parent epic).

## Out of scope
- Point-lookup cache refactor → #1242-B. Directory retirement + CLI rewire → #1242-D. AppView/SCIM index → `cyberdione/cyberdione-product#1` (NOT core).

## Edges
- **blocked_by:** #1163 / PR #1229 (account record format must land first).
- **blocks:** #1242-B, #1242-D.

Filed for approval per #1242. Not starting until #1229 merges (Q2 hybrid resolved, genesis PR in review).

ewindisch:
## commander-zod: gate resolved — #1229 merged (2026-07-24 08:45 UTC)

PR #1229 landed the account-record substrate on `main`. This unblocks **#1257** at the gate level. Recording the build surface + one design question before implementation.

### What #1229 delivered (`crates/hyprstream-pds/`)
[...]

PR #1405, opened 2026-07-28
feat(pds): #1257 Postgres-backed PdsRecordStore for Multi-AZ RDS

ewindisch:
## What is implemented

- Adds `pds-postgres` cargo feature and optional deps (deadpool-postgres, tokio-postgres, tokio-postgres-rustls — workspace-pinned, shared with #1401).
- New `services/pds_record_pg.rs`: async Postgres KV bridge on a dedicated thread (tokio mpsc + `recv().await`, one spawned task per command), advisory-locked idempotent schema migration, rustls TLS with a pinned CA, and bounded pool/driver timeouts for RDS failover.
- `PdsRecordStore` gains a `RecordBacking::Postgres` variant and `open_postgres`; all production read/write paths have Postgres branches.
- `services/factories.rs` selects the Postgres backend when the records-role binding resolves; fails closed instead of silently falling back to RocksDB.
- `config/mod.rs` adds shared `RdsConfig` (`url_file`, `root_cert_file`, `cell_id`) with `resolved_from_env`: explicit TOML → role-scoped env paths (`HYPRSTREAM_RECORDS_URL_FILE` / `HYPRSTREAM_RECORDS_SSLROOTCERT_FILE`, metal v1.1) → shared credentials dir (`HYPRSTREAM_POSTGRES_CREDENTIALS_PATH/records-url` + `rds-ca.pem`, consulted only when `records-url` exists there, so credentials-role-only deployments stay on the local backend). The password-bearing URL never transits the process environment.

## Cross-AZ invariant

When the records role is configured and `pds-postgres` is enabled, both the publisher (registry) and resolver (discovery) connect to the same RDS instance. A write via AZ-a is visible to AZ-b. No silent local fallback remains for record/repo state.

Write safety is SQL-side, not process-mutex-side:

- **at9p conditional advance**: compare-and-swap on the exact stored state bytes (`UPDATE … WHERE key AND value` / `INSERT … ON CONFLICT DO NOTHING`, rows-affected checked) in ONE transaction with the checkpoint write and first-boot marker removal. Two publishers in two AZs serialize on the row lock; the loser reports `ConditionalAdvance::Conflict` instead of silently overwriting.
- **Repo head** (`put_record_and_commit`, `put_commit`): CAS on the commit key against the exact head bytes the publish was built from; a stale-head publish is rejected instead of committing a head that does not cover the record set.
- **Snapshot reads** (`load_repo_from_pg`, `accepted_at9p_states`): records+commit and state+checkpoint scans run in ONE read-only transaction (`PgKv::read_snapshot`) — no torn reads across pooled connections.
- **Failover bounds**: pool max_size, deadpool wait/create/recycle timeouts, driver connect_timeout/keepalive/TCP_USER_TIMEOUT, server-side statement_timeout. A hung failover connection errors out; it never blocks `pool.get()` forever.

## Verification (head 0c6465fc3)

- `cargo check -p hyprstream --features pds-postgres --all-targets` and default-features check: clean.
- `cargo clippy -p hyprstream --all-targets --features pds-postgres -- -D warnings`: clean.
- 106 affected-module lib tests pass (`services::pds_record_pg`, `services::discovery`, `config::tests`), including 10 **live** tests against a scratch Postgres 16 via `HYPRSTREAM_POSTGRES_TEST_URL_FILE`: two-handle persistence/visibility, repo-head CAS stale-publish rejection, concurrent at9p advance single-winner, KV semantics (get_batch positional/absence, snapshot range bounds), first-boot marker visibility.
- Workspace release clippy (`--workspace --all-targets --release -- -D warnings`) via the pre-commit hook.
- The clippy CI job and the merge-gate script now compile and test the `pds-postgres` feature (it previously never built in CI).

## What is NOT done

- at9p duplicity WAL is still written to the local `pds_store_dir` even in RDS mode (local alarm state, not record/repo state).
- Live DB tests skip green unless `HYPRSTREAM_POSTGRES_TEST_URL_FILE` is provided; no CI Postgres service yet.


ewindisch:
**Independent review (frontline seat) — FAIL, bound to head fd311e8c5.**

The foundations are right: no SQLite, fully parameterized SQL, real TLS (pinned CA, verify-full, host/hostaddr overrides ignored), no DSN/password in logs, zero-config local path preserved (Postgres strictly opt-in), fail-closed on misconfig. But the distributed-correctness core doesn't hold yet:

1. **Lost updates on the at9p head (blocking).** The conditional advance takes `self.at9p_advance_lock` (process-local), reads the watermark in one transaction, and writes in another. Under RocksDB the OS file lock made that safe; the entire premise here is two PDS processes in two AZs sharing one RDS — the mutex guards nothing. Two publishers can both read watermark N, both pass `expected == Some(N)`, and the second silently overwrites the first while returning Committed. `ConditionalAdvance::Conflict` becomes unenforceable. Fold the compare into SQL (`UPDATE … WHERE value = $expected`, check rows_affected) or do read+write in one SERIALIZABLE transaction.
2. **Same class on the repo head**: nothing serializes two publishers for one DID in `put_record_and_commit` — the last commit written may not cover the record set.
3. **Non-atomic repo read**: `load_repo_from_pg` does the record scan and the commit get on two pooled connections — a concurrent write yields old records + new signed commit, and the reconstructed MST root fails verification. `get_batch` already shows the read-only-transaction pattern; use it (also in `accepted_at9p_states`).
4. **Zero CI coverage**: no workflow enables `pds-postgres`, so the 801-line module and its tests never compile in CI — the green checks prove nothing about this code. Add the feature to the clippy/test matrix.
5. **No KV-semantics tests**: the five tests cover URL/TLS only; `put_batch`/`range_scan`/`prefix_upper_bound`/`get_batch` — exactly where RocksDB↔Postgres divergence would live — are untested.

Failover-specific fixes: the pool has no max_size/create/wait/recycle timeouts and the driver no connect/keepalive/statement timeouts (a hung Multi-AZ failover connection blocks `pool.get()` forever); `bridge_main` serializes all PDS I/O one query at a time process-wide and parks a current-thread runtime on a blocking `std::sync::mpsc::recv()` — use tokio mpsc + `recv().await`. Cargo claim of `rustls-native-certs` doesn't match the feature (pinned-only CA is fine — drop the claim).

Mechanics: CONFLICTING with main (factories.rs was rewritten since); rebase onto ab3934533 before re-review. Happy to re-review promptly once the CAS/atomicity design is fixed.

ewindisch:
Implementation-lane reply to the independent exact-head review (REVISE) of `0c6465fc3`.

**New head:** `b5bb56fd41b4fd536372c11f301b78cc8381b1fc` (one commit over `0c6465fc3`, pushed non-force fast-forward).

**P1-1 (workflow quoting bug):** FIXED — `.github/workflows/rust.yml:103-106` reworded (`cfg-gated`, apostrophe-free note); both lints now provably run inside the container's single-quoted script. Shape gate added: `scripts/check_rust_workflow_matrix.py:231` (`_podman_quoted_script`), `:335-351` (workspace + pds-postgres lint must be inside the quoted script; no ASCII apostrophe permitted), with two negative mutations (`:502`, `:510`). Verified: matrix gate OK incl. 18 mutations; extracted script passes `bash -n`; workflow YAML parses.

**P1-2 (READ COMMITTED torn reads):** FIXED — shared `begin_snapshot_tx` sets `read_only` + `IsolationLevel::RepeatableRead` for both `cmd_get_batch` and `cmd_read_snapshot` (`crates/hyprstream/src/services/pds_record_pg.rs:626-637`, call sites `:642`, `:770`). Regression pins: deterministic isolation test (`live_snapshot_transaction_is_repeatable_read`, `:1244`) and a command-level tear test under a concurrent writer (`live_get_batch_never_tears_under_a_concurrent_writer`, `:1302`) — both executed live against scratch Postgres 16 (12/12 live tests green, zero skips).

**P1-3 (RDS bypass paths):** FIXED — QUIC gate resolves the binding and classifies the shared store (`factories.rs:216-244`); announcement refresh opens via the backend-selecting helper (`factories.rs:302`, helper `:134`); publisher construction now returns the error when an RDS binding resolves (`factories.rs:1311-1317`; local posture keeps the documented warn+disable); bootstrap CLI routes through `open_checkpoint_store` (`deployment_bootstrap.rs:35`) — provision (`:158`, `:166`), inspect (`:213`), init-deployment-store (`:72`, dispatch `main.rs:2920`) with an accepted-states guard before marking first boot. Source-level contract checks: `factories.rs:3351`, `deployment_bootstrap.rs:675`.

**P2-4 (records-url error kinds):** FIXED — resolution is fallible: `NotFound` → local, dangling symlink binds (the later read fails loudly), other stat errors fatal (`config/mod.rs:399-441`; test `:3858`).
**P2-5 (config-load `unwrap_or_default`):** FIXED — `load_config` is fail-closed everywhere, incl. the MCP metadata task (`factories.rs:53`, `:2284-2289`).
**P2-6 (live tests skip-green in CI):** STATED PLAINLY — `docs/rds-records-backend.md` Testing section (deliberate; no behavior change).
**P2-7 (IPv6 host mismatch):** FIXED — the driver reparse mirrors `validate_url`'s arms and compares the bare host (`pds_record_pg.rs:415-435`; test `:949`).
**P2-8 (operator contract undocumented):** FIXED — `docs/rds-records-backend.md`.

Validation (all via `hyprstream-build.sh run --` / BuildQ): `cargo check` (feature on and default) clean; `cargo clippy -p hyprstream --all-targets --features pds-postgres -- -D warnings` clean; causal suite **127/127** incl. 12 live tests executed against scratch `postgres:16-alpine` (container destroyed after); default-features bootstrap tests **22/22** incl. the `not(pds-postgres)` bail arm; pre-commit hook release workspace clippy green. Loopback burndown unchanged from the reviewed head — only the main-carried `repo_authority.rs` +3 (#1605 lane; this branch adds zero literals).

Not claimed: merge readiness. Awaiting a fresh independent exact-head review and terminal hosted checks. The advisory loopback job fails on the same main-carried literals as at the reviewed head.

ewindisch:
Closing the RDS backend PR. The selected architecture is dedicated self-hosted Postgres, so this 34-commit-behind RDS path is no longer the active backend direction; all implementation history remains preserved.

PR #1415, opened 2026-07-29
ci: publish browser WASM SDK via npm trusted publishing

### LINKED ISSUE #1407: Publish the browser WASM SDK: main staging channel and immutable release packages
Create a public, browser-consumable package from the hyprstream-rpc-std wasm-pack output. main publishes a staging dist-tag; protected release tags publish immutable semver production packages. Define package exports/files, test the packed artifact, and record provenance/integrity. Production consumers must pin exact versions; staging may follow the staging tag. This removes source Rust/WASM compilation from downstream web CI and deploys.

ewindisch:
Closes #1407

Adds the reviewable browser WASM publication path for `hyprstream-rpc-std`, the first shipping piece of the thin client split.

- npm OIDC Trusted Publishing only; no npm token or fallback secret.
- Publishing uses `npm publish --provenance`.
- `main` creates unique staging prereleases under the `staging` dist-tag; immutable `hyprstream-rpc-std-v<crate version>` tags publish production semver under `latest`.
- Build/package/smoke executes once; publish jobs verify the downloaded tarball and integrity evidence before npm receives it.
- Negative verifier coverage confirms a staging artifact is rejected when `GITHUB_REF` is not `refs/heads/main`.

Validated locally: WASM build, npm pack integrity evidence, clean-consumer/esbuild smoke test, staging verifier positive and non-main negative path, browser-WASM gate mutations, workflow/syntax/diff checks, and release Clippy pre-commit.

No package was published as part of this PR.


ewindisch:
## Non-blocking follow-ups from the r2 review — addressed

Per `REVIEW-frontline-1415-r2-fable-2026-07-30.md` (PASS at `f4ccff66f`), both
flagged non-blocking items are now fixed at the new head:

- **`actions: read`** added to the `attest-tarball` job in
  `attest-rpc-std-wasm.yml` (least privilege — only what
  `actions/download-artifact@v4`'s cross-run `run-id` fetch needs). Fixes the
  403 that would otherwise hit this convenience job on every `workflow_run`
  once it starts firing on `main`. Does not touch the required
  `npm publish --provenance` path, which is self-contained and same-run.
- **Tar verifier hardening (P3 defense-in-depth):** `verify-rpc-std-wasm-package.mjs`
  now scans the whole archive and refuses any tarball with more than one
  `package/package.json` entry, instead of returning the first match — closing
  the theoretical gap where a hand-crafted duplicate-entry tarball could verify
  against one manifest while npm's last-entry-wins extraction materializes
  another. Added `verify-rpc-std-wasm-package.probe.mjs`, a causal adversarial
  test that runs the real verifier CLI against a two-entry tarball (asserts
  rejection with the expected reason) plus a matching single-entry control
  (asserts acceptance), wired into `publish-rpc-std-wasm.yml` ahead of every
  real build.

Exact new head: `c6dd8239d675d34ec5583cb8f9537517c51ec15b`.

### Deterministic staging tarball rebuilt at this head (local, no registry write)

Pinned toolchain (rustc 1.97.0, wasm-pack 0.13.1) reproduced the same build
shape as prior seats; identity below is newly computed at this exact head
(commit count changed as expected: `2467` vs the prior head's `2466`):

```
package: @hyprstream/rpc
version: 0.1.0-dev.2467.c6dd8239d675
sha256:  213a75b5c394a1c0b71cef880556c277492f25b6b0901768266862a779bd0b17
sri:     sha512-p8INp6q8sQKW2/ML8KSSpn7UXmL3pZY18tP01xI0DTeTXWX2JHkovS8SC8MSld3xy/g4fyat6fNScY3PDmBgRA==
npm shasum (sha1): 3431f994cb5c58754bcf9841c4b2d6139bc9fa0a
```

- `verify-rpc-std-wasm-package.mjs dist/hyprstream-rpc-std-wasm staging` with
  `EXPECTED_VERSION`/`EXPECTED_SHA256` pinned to the above — **passes**.
- Clean-consumer install (`npm install --ignore-scripts` against the tarball
  only) + esbuild browser bundle — **passes** (embedded in the packaging
  script's own smoke step).
- `verify-rpc-std-wasm-package.probe.mjs` — **passes** (rejects the
  duplicate-manifest adversarial case, accepts the single-entry control).
- Independently recomputed SHA-256 / SRI / npm shasum match `integrity.json`
  exactly.

No publication, no registry write, no environment/ruleset/settings change, no
un-draft. Deployment prerequisites (npm trusted publisher, `npm-staging`
environment, `hyprstream-rpc-v*` tag ruleset) remain unestablished per the r2
review and still gate every publish closed.

PR #1416, opened 2026-07-29
feat(release): add public crates release policy

ewindisch:
## Scope

The publishable Rust SDK closure remains exactly:

1. `hyprstream-rpc-build`
2. `hyprstream-rpc-derive`
3. `hyprstream-crypto`
4. `hyprstream-rpc`

`hyprstream-rpc-std` remains non-publishable because its VFS/workers-tcl/9p path dependencies and unreleased patched `moq-net` WASM behavior cannot reach downstream users through a workspace `[patch.crates-io]`.

The real `crates/hyprstream` service remains `publish = false`; its broad service/libtorch graph is not published. The separately packageable, excluded `release/hyprstream-deprecation-shim` now reserves the same crates.io name as a documentation-only Apache-2.0 shim at the exact service-matching version `0.5.0`. It has no dependencies or product implementation and directs users to current installation and client guidance.

The checker mechanically requires the explicit shim manifest path, workspace exclusion, Apache-2.0 package metadata, no dependencies/implementation, an exact match to the real service version, and rejects `0.2.0`. The order is fail-closed: publish shim → verify `hyprstream@0.5.0` from crates.io → only then yank stale versions.

## Validation

- `python3 scripts/check_crate_release_policy.py` — 4 public-now, 27 public-later, 5 internal.
- Negative policy mutations: shim `0.2.0` failed with the expected fixed-0.5.0 error; making deferred `hyprstream-util` publishable failed with the expected `must set publish = false` error.
- `cargo package --locked --allow-dirty --manifest-path release/hyprstream-deprecation-shim/Cargo.toml`, through buildq — passed; archive contains only shim metadata, README, and documentation-only source.
- Cargo metadata through buildq confirms `hyprstream 0.5.0`, `publish = ["crates-io"]`, and zero dependencies.
- `python3 scripts/check_license_boundary.py`
- `python3 scripts/check_rust_workflow_matrix.py`
- `python3 scripts/check_browser_wasm_gate.py`
- Publish-script shell syntax and `git diff --check` pass.
- Foreground commit hook completed its full Clippy gate.

## Remaining external blockers

- This PR remains held until actual relicensing lands; do not merge it.
- A verified crates.io owner must manually publish and verify the 0.5.0 shim before any authorized yanks of the seven stale versions. No publish, yank, or registry settings change was performed.
- The four SDK crates still need their separate manual first-publication bootstrap and trusted-publisher configuration; until build@0.1.0 exists in crates.io, derive/RPC exact registry packaging remains blocked.

No real-service publication, merge, force push, or existing Cargo manifest license-field edit was performed.


ewindisch:
**Independent review (frontline seat) — draft assessment, bound to head 0a5d97359 (CONFLICTING vs main).**

Half subsumed, half still valuable — recommend splitting:

Land (after rebase): the crate-release.toml tiering (exhaustive over all 37 packages), the publish=/pinned-version manifest hygiene (genuinely required for crates.io), the policy checker wired into the standards gate, and the policy doc. Two changes needed first: drop the PR's verbatim duplicate of the [license_boundary] map (main's .github/license-boundary.toml is now authoritative — a second copy can drift; read it instead), and rebase the check_license_boundary.py edits off main's rewrite.

Hold: publish-crates.yml + the deprecation shim. The tag/OIDC design is otherwise solid (annotated-tag + ancestor check, crates-io environment, no stored token, idempotent re-verify), but the publish job holds id-token: write while running the repo's own build graph (cargo check/package, build scripts) — the exact shape main's rpc-std-wasm publish work deliberately eliminated by stripping OIDC from build jobs. Split build (untrusted) from publish (OIDC) before this lane goes live. Also: hyprstream@0.5.0 name-squat shim must track the service version forever (every bump = new squat publish, plus the duplicate-name exemption bolted into the boundary checker) — decouple to a fixed shim version or drop it. And public_later lists the four AGPL crates as future crates.io publications with no caveat that AGPL SDK crates poison downstream permissive consumers — add the note.

No license contradictions found in the publishable set (all Apache-2.0, all inside permissive_roots).

ewindisch:
Closing the stale draft. It is conflicted with main and remains blocked on external relicensing, registry publication, and an unsafe build/publish split; the policy work remains available in branch history.

PR #1439, opened 2026-07-30
fix(git2db): strict certificate mode + explicit-only auth

### LINKED ISSUE #1429: fix(git2db): strict certificate mode + explicit-only auth (hard blocker for forge fetch)
# fix(git2db): strict certificate mode + explicit-only auth (HARD BLOCKER for forge fetch)

**Epic:** #1427.

## Problem

`AuthManager::create_callbacks()` (`crates/git2db/src/auth.rs:66-69`) **always accepts all certificates** (`CertificateCheckStatus::CertificateOk` for every cert). This is a latent vulnerability: any future untrusted-repository fetch feature inherits an always-trusting TLS posture. For the merge gate, the forge adapter fetches PR heads from GitHub — this must not ship with always-accept certs.

## Scope

1. `CertificateMode::Strict` — reject on cert failure instead of always-accept.
2. `AuthMode::ExplicitOnly` — refuse `AuthStrategy::Default`; refuse ambient `~/.gitconfig`/SSH agent.
3. Remove the `CertificateOk` always-accept default; default to strict.

## Non-goals

- Full CA pinning (follow-on). Custom trust-store management (follow-on).

## Likely files

`crates/git2db/src/auth.rs`, `crates/git2db/src/callback_config.rs`.

## Acceptance evidence

- Adversarial: MITM proxy with wrong cert → fetch **fails** (not `CertificateOk`).
- Ambient `~/.gitconfig` credentials are **not** consulted in `ExplicitOnly` mode.

## Dependencies

None. **Edge: independent.** This is the **hard blocker** for B3/B4 (untrusted fetch).

ewindisch:
Closes #1429.

## Problem

`AuthManager::create_callbacks()` (`crates/git2db/src/auth.rs`) unconditionally returned `CertificateOk` for every certificate, and `CertificateConfig::SystemDefault` (the field default) also returned `CertificateOk`. Because installing a `certificate_check` callback overrides libgit2's verification, **every fetch silently accepted any certificate** — including forged/MITM ones. Any untrusted-repository fetch path (the merge gate's forge PR-head fetch, P2P distribution) inherited an always-trusting TLS posture.

## Changes

- **`CertificateConfig::Strict`** is the new default. Under `Strict` the certificate callback returns `CertificatePassthrough`, deferring to libgit2's built-in verification against the system trust store (CA chain + hostname for HTTPS, `known_hosts` for SSH). A cert that fails verification **fails the fetch**. The buggy `SystemDefault` variant (which claimed to defer but actually accepted everything) is removed.
- **`AuthMode::ExplicitOnly`** is the new default. `AuthStrategy::Default` (libgit2 ambient discovery: git credential helper, `~/.gitconfig`, SSH agent fallback) is refused at credential-resolution time, so an untrusted fetch cannot silently pick up the operator's credentials. `AllowAmbient` remains available as an explicit opt-in for trusted first-party mirror access.
- **`AcceptAll`** survives only as an explicit, documented test/opt-in mode — never the default.
- `AuthManager` carries `certificate_mode` + `auth_mode` fields (both defaulting secure); presets that bundle `AuthStrategy::Default` now opt into `AllowAmbient` explicitly.
- Shared pure helper `CallbackConfig::certificate_decision` so the sync (`auth.rs`) and send-safe (`callback_config.rs`) paths cannot drift on the boundary.

## Causal tests (`tests/strict_auth_test.rs` + in-crate unit tests)

- `Strict` → `CertificatePassthrough` (never `CertificateOk`), even when hostkey material is present.
- `AcceptAll` → `CertificateOk` only via explicit opt-in.
- `Pinned` rejects non-matching fingerprint; never falls back to accept.
- `ExplicitOnly` refuses `AuthStrategy::Default` (ambient credentials not consulted); `AllowAmbient` permits it.
- `AuthManager::new()` carries no ambient strategy; defaults verified strict/explicit-only.

## Caller audit

- `manager.rs`: uses `Strict` explicitly; opts into `AllowAmbient` only when `use_credential_helper` is set.
- `services/registry.rs`: uses `CallbackConfig::new()` which now defaults to strict/explicit-only (progress-only config, unchanged intent).
- `send_trait_test.rs`: ambient `Default` test configs explicitly opt into `AllowAmbient`.
- No non-test code relied on the always-accept default.

## Non-goals (per issue)

Full CA pinning and custom trust-store management are follow-ons.

Refs #1429. Epic #1427.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5973,6 +5973,7 @@ dependencies = [
  "tokio-test",
  "tracing",
  "tracing-subscriber",
+ "url",
  "uuid 1.18.1",
  "walkdir",
 ]
--- a/crates/git2db/Cargo.toml
+++ b/crates/git2db/Cargo.toml
@@ -11,6 +11,9 @@ description = "Git-native database operations using libgit2 with submodules as s
 git2.workspace = true
 libc.workspace = true
 
+# Standards-based (WHATWG URL) authority parsing for credential host-scoping
+url.workspace = true
+
 # Async runtime
 tokio.workspace = true
 
--- a/crates/git2db/src/auth.rs
+++ b/crates/git2db/src/auth.rs
@@ -1,7 +1,8 @@
 //! Authentication and credential management
 //!
-//! Consolidated authentication patterns from the original codebase
+//! Consolidated authentication patterns from the original codebase.
 
+use crate::callback_config::{AuthMode, CertificateConfig};
 use git2::{Cred, CredentialType, RemoteCallbacks};
 use std::path::PathBuf;
 use tracing::{debug, info, warn};
@@ -26,22 +27,53 @@ pub enum AuthStrategy {
     Default,
 }
 
-/// Credential manager for handling authentication
+/// Credential manager for handling authentication.
+///
+/// Defaults to a **strict** security posture:
+/// - [`CertificateConfig::Strict`] — certificates are validated by libgit2
+///   against the system trust store; a cert that fails verification rejects
+///   the fetch. The previous always-`CertificateOk` behavior is gone.
+/// - [`AuthMode::ExplicitOnly`] — ambient credential discovery
+///   (`AuthStrategy::Default`, i.e. the git credential helper, `~/.gitconfig`,
+///   SSH agent fallback) is refused unless the caller explicitly opts in via
+///   [`AuthManager::with_auth_mode`]`(AllowAmbient)`.
 pub struct AuthManager {
     strategies: Vec<AuthStrategy>,
+    certificate_mode: CertificateConfig,
+    auth_mode: AuthMode,
 }
 
 impl AuthManager {
-    /// Create a new authentication manager
+    /// Create a new authentication manager with no strategies, strict
+    /// certificate validation, and `ExplicitOnly` auth mode.
     pub fn new() -> Self {
         Self {
-            strategies: vec![AuthStrategy::Default],
+            strategies: Vec::new(),
+            certificate_mode: CertificateConfig::Strict,
+            auth_mode: AuthMode::ExplicitOnly,
         }
     }
 
-    /// Create authentication manager with strategies
+    /// Create authentication manager with strategies. Certificate validation
+    /// defaults to strict and auth mode to `ExplicitOnly`.
     pub fn with_strategies(strategies: Vec<AuthStrategy>) -> Self {
-        Self { strategies }
+        Self {
+            strategies,
+            certificate_mode: CertificateConfig::Strict,
+            auth_mode: AuthMode::ExplicitOnly,
+        }
+    }
+
+    /// Set the certificate validation mode.
+    pub fn with_certificate_mode(mut self, mode: CertificateConfig) -> Self {
+        self.certificate_mode = mode;
+        self
+    }
+
+    /// Set the authentication provenance mode.
+    pub fn with_auth_mode(mut self, mode: AuthMode) -> Self {
+        self.auth_mode = mode;
+        self
     }
 
     /// Add an authentication strategy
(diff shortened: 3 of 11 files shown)


ewindisch:
## Revision round 2 — Sol P1/P1/P2 fixed

New head: `efb1dad93086377753e333f24a6fa1f0ef18026c` (previous reviewed head `ada0d62c46c1b6b68a6149d9ee2f473628d1d926`)

Addresses every finding in the round-2 security re-review:

1. **P1 (path/query `@` authority spoofing, port erasure, IPv6 truncation)** — replaced the hand-written `extract_git_host` with a standards-based `extract_git_authority` (`crates/git2db/src/auth.rs`) built on `url::Url::parse` / `url::Host::parse` (WHATWG URL Standard). Authority is separated from path/query/fragment structurally before any userinfo handling, so an `@`, `/`, `?`, or `#` in the path/query can never be misread as authority. Canonicalizes case, IDNA (to ASCII punycode), IPv4 (including non-canonical octal/hex octets), and bracketed IPv6, and defines default/non-default port scope **per scheme** (so e.g. GitHub's `ssh://...:443` never collides with `https://...:443`). Malformed/ambiguous/disallowed-scheme URLs fail closed to `None`.

2. **P1 (off-site redirect exfiltrates a host-scoped token)** — added `RedirectPolicy` (default `None` — no off-site redirects), wired into `FetchOptions::follow_redirects` on both the send-safe `CallbackConfig` clone path (`clone_options.rs`) and the legacy `AuthManager` path. Under the pinned libgit2 1.9.x the credential callback still observes the pre-redirect URL after an off-site redirect, so refusing the redirect at the transport layer is what actually closes the leak (rebinding to the "effective peer" isn't observable through this libgit2 version's callback surface). `crates/git2db/tests/redirect_test.rs` is a new causal test file driving two real loopback HTTP servers (distinct hosts, not just ports) through git2's actual `Remote::fetch()`: it proves the old default (`Initial`) delivers the token to the off-site redirect target, and the new default (`None`) never contacts it at all — for both callback paths.

3. **P2 (deceptively unscoped `github_token()` preset)** — `AuthBuilder::token` renamed to `token_unscoped` (explicit, documented trusted-caller-only opt-in — the ordinary token API can no longer silently construct a hostless credential), and `presets::github_token()` now uses `token_scoped(token, "github.com")`.

4. **Adversarial tests** — `crates/git2db/src/auth.rs` unit tests cover path/query/fragment `@` confusion, userinfo-lookalike, scp-like embedded `@`, case/IDNA normalization (both `https` and `ssh`, since `url::Url` only auto-canonicalizes "special" schemes), default/non-default port scoping per-scheme, IPv4 canonicalization, IPv6 (bracketed, including over `ssh`), and malformed/disallowed-scheme rejection. `redirect_test.rs` covers the redirect case live through both callback paths as described above.

### Evidence (all via the shared Hyprstream build queue, `buildq --`)

- `cargo test -p git2db --lib` — 120 passed
- `cargo test -p git2db --test strict_auth_test --no-fail-fast` — 18 passed
- `cargo test -p git2db --test redirect_test --no-fail-fast` — 4 passed
- `cargo test -p git2db --test send_trait_test` — 5 passed
- `cargo clippy -p git2db --all-targets -- -D warnings` — clean
- `git diff --check` (full PR diff and staged diff) — clean
- Local pre-commit hook (full workspace release Clippy, `-D warnings`) — passed

No live network/auth operation, credential inspection, or credential store was used; `redirect_test.rs` uses only local loopback test servers. No force-push; PR remains open and draft. Requesting fresh review.

ewindisch:
## Revision round 3 — Kimi-K3 fallback P1/P3 fixed

New head: `8396922fd128e59f151f361394ec1920c3122739` (previous reviewed head `efb1dad93086377753e333f24a6fa1f0ef18026c`)

Addresses both findings from the round-3 security re-review (`REVIEW-frontline-1439-r3-kimi-2026-07-30.md`):

1. **P1 (WHATWG-vs-libgit2 backslash authority differential releases a host-scoped token to a spoofed authority).** For "special" schemes, the WHATWG `url` crate treats `\` as ending the authority early, but the pinned libgit2 1.9.x C parser treats `\` as an ordinary authority character and keeps splitting userinfo at the last `@` — so the same URL string is assigned a *different* authority by `extract_git_authority` than by the transport that actually opens the connection. `crates/git2db/src/auth.rs`'s `extract_git_authority` now fails closed (`None`) on any backslash in the input, before any other parsing runs, closing the differential outright rather than trying to out-parse libgit2. New unit tests cover the special-scheme case, an arbitrary-position case, and the scp-like form.

   A new causal test file, `crates/git2db/tests/backslash_authority_test.rs`, drives a real `git2::Remote::fetch()` against a URL of the form `http://<A>\@<B>/repo.git` over two genuinely distinct real loopback HTTP servers (`A` = 127.0.0.1, the authority a token is bound to; `B` = 127.0.0.2, an unrelated authority `B` always challenges for credentials, never granting access, so libgit2 runs its full credential-resolution sequence). It proves, against the real pinned stack: the TCP connection lands on `B`, not `A` (the differential is real and reachable), `A` is never contacted, and — across every request `B` ever sees — no `Authorization` header ever decodes to the bound secret token, through both the `CallbackConfig` and legacy `AuthManager` credential paths.

2. **P3 (hosted Loopback-literal burn-down gate failing, +2 in `auth.rs`).** Two production doc-comment IPv6 examples (`[2001:db8::1]`) incidentally matched the gate's `::1` pattern. Reworded to `[2001:db8::cafe]` — still a valid RFC 3849 documentation-range address, just not ending in `::1`. `scripts/check_loopback_burndown.py` now exits 0 at 82/31, exactly matching the existing baseline (no expansion, no `--update` needed). The `#[cfg(test)]`-gated IPv6 literals in the unit tests are unaffected (correctly excluded by the gate's skipper).

### Evidence

- `cargo test -p git2db --lib` — 123 passed (was 120; +3 backslash-rejection unit tests)
- `cargo test -p git2db --test strict_auth_test` — 18 passed
- `cargo test -p git2db --test redirect_test` — 4 passed
- `cargo test -p git2db --test backslash_authority_test` (new) — 2 passed
- `cargo test -p git2db --test send_trait_test` — 5 passed
- `cargo clippy -p git2db --all-targets -- -D warnings` — clean
- `python3 scripts/check_loopback_burndown.py` — exit 0, 82/31 (matches baseline exactly)
- `git diff --check` — clean
- Local pre-commit hook (full workspace release Clippy `-D warnings`) — passed, routed through a fresh task-private build cache to avoid contention on the shared fleet build pool (shared caches were not purged or bypassed; the hook itself was not skipped)
- `git push origin feat/1429-strict-cert` — fast-forward, no force

No live network/authentication operation against a real remote, credential inspection, or credential-store access was performed. `backslash_authority_test.rs` uses only local loopback (127.0.0.1/127.0.0.2) servers under the test process's own control. No merge, approval, or un-draft. Requesting fresh review.

PR #1444, opened 2026-07-30
feat(oauth): browser RFC 8693 sender-bound exchange contract (#1425)

### LINKED ISSUE #1425: feat(oauth): make browser RFC 8693 exchange an advertised sender-bound contract
## Problem

The browser RPC path and OAuth AS disagree today. The AS implements token exchange at `POST /oauth/token`, but its RFC 8414 metadata omits the token-exchange grant. The WASM caller omits the server-required public `client_id`; it also has no complete RFC 9449 DPoP contract when it requests sender-constrained output.

RFC 8693 defines the token-exchange grant, but it does not itself define DPoP binding, browser refresh policy, or identity assurance. Those are explicit composed policy here.

## Scope

- Make the advertised OAuth metadata, request validation, and browser WASM client one contract for RFC 8693 exchange.
- Define and enforce the public-client identity/metadata rule for this exchange; `client_id` identifies the browser client and is not a client secret or proof of identity.
- Compose RFC 8693 with RFC 9449:
  - validate a fresh DPoP proof at the token endpoint, including method/URI/time/`jti` and server nonce policy;
  - bind the issued token with `cnf.jkt`;
  - return `token_type: DPoP` only for a token with a verified matching confirmation binding;
  - require resource requests to use a matching proof and the RFC 9449 access-token hash where applicable.
- Audience/resource- and scope-restrict the issued token to the intended Hyprstream RPC service.
- Add wire-level positive and negative tests covering metadata, public client ID, DPoP nonce/replay/binding, audience/resource restriction, and the successful browser request shape.

## Acceptance

1. `/.well-known/oauth-authorization-server` truthfully advertises `urn:ietf:params:oauth:grant-type:token-exchange` and the supported token-endpoint/DPoP algorithms and methods.
2. `wasm_token_exchange::fetch_exchange_token` sends the required public `client_id`, exact RFC 8693 fields, and a valid RFC 9449 proof for sender-bound exchange.
3. A caller cannot receive or use a token labeled DPoP without a verified matching `cnf.jkt`; missing nonce, stale/replayed `jti`, method/URI mismatch, key mismatch, and audience/resource substitution are rejected.
4. The standard exchange returns an access token only. It does not issue a refresh token unless a separately reviewed, metadata-advertised rotation/revocation policy is approved.
5. Browser CORS permits only the required metadata/token interactions and never exposes tokens in redirects, URLs, or logs.
6. The contract is tested against the frontend's exact generated WASM request, not only a Rust-local request fixture.

## Specification alignment

- RFC 8693 — token-exchange request/response.
- RFC 8414 — truthful authorization-server metadata.
- RFC 9449 — DPoP proof and `cnf.jkt` sender constraint.
- RFC 9700 — browser/public-client security, audience restriction, replay reduction, and metadata use.
- ATProto OAuth profile — the incoming PDS session remains its own PKCE/PAR/DPoP flow; an ATProto access or service-auth token is not treated as a generic bearer credential.

## Boundaries

This establishes Classical browser authentication and sender-constrained token use. Hybrid RPC signing may protect the request transport, but this issue does not claim `PqHybrid` identity assurance. Browser PQ-anchor bootstrap, peer first-boot anchoring, VfsShell UI, deployment, and live mutations are out of scope.

Related: #1146, #1119, #1274, #1426.

ewindisch:
Closes #1425.

## Summary

Makes the advertised OAuth metadata, request validation, and browser WASM client one contract for RFC 8693 token exchange composed with RFC 9449 DPoP.

- **RFC 8414 metadata** truthfully advertises `urn:ietf:params:oauth:grant-type:token-exchange` and `dpop_bound_access_tokens_supported` (RFC 9449 §9); `dpop_signing_alg_values_supported` covers both ES256 and EdDSA.
- **Public client_id rule (RFC 9700):** the browser exchange is routed on a well-known public client identifier (`hyprstream-browser-vfs`) — a public client, no secret. The sender binding comes entirely from the DPoP proof.
- **RFC 9449 sender binding:** DPoP is **mandatory** at the token endpoint (method/URI/`iat`/`jti` replay/server nonce). The issued token is bound with `cnf.jkt`, the response is `token_type: DPoP`, and the token is **audience/resource-restricted** to the Hyprstream RPC service resource (substitution or unsupported/duplicate fields → `invalid_target`/`invalid_request`).
- **Subject-token sender-constraint preservation:** a DPoP-bound subject access token (`cnf.jkt` present) must be exchanged under a token-endpoint proof from the **same** key — a stolen sender-constrained token string cannot be re-bound to an attacker-owned key. Checked (constant-time) before the single-use subject token is consumed.
- **One browser-held key end to end:** the same Ed25519 key drives both the DPoP proof and the RPC envelope signature, so the minted `cnf.jkt`-bound token is only usable over an envelope signed by that same key — no Bearer downgrade path exists on the RPC transport.
- **DPoP nonce lifecycle (RFC 9449 §8):** the WASM client persists the server nonce, recognizes `use_dpop_nonce`, and retries once with a freshly signed proof; the nonce is exposed to the caller (`VfsShell.exchangeNonce`) to carry into the next exchange.
- **The real browser form sends the canonical `resource`** (RFC 8707): `exchange_form_body` takes and encodes it — the exact origin `fetch_exchange_token` is calling — so the wire contract is exercised by the actual client code, not a native-only test fixture.
- **The sender-bound fetch fails closed on a non-DPoP response:** `fetch_exchange_token` rejects a missing/`Bearer` `token_type` and independently verifies the returned token's `cnf.jkt` equals the browser's own DPoP key thumbprint before handing it to `VfsShell` — a misrouted or compatibility response can never be installed as a sender-bound credential it isn't.
- **Subject types:** `access_token` and `jwt` (atproto/external service assertion) only; `id_token` is explicitly rejected (no defined audience/client binding for it in this contract).
- **RFC 8693 fields the browser leaf path doesn't support are explicitly rejected**, not silently dropped: `actor_token`, `actor_token_type` (including a standalone or duplicated field), non-`access_token` `requested_token_type`, and `tenant`.
- **Access-token only:** no refresh token (rotation is a separately reviewed, metadata-advertised policy).
- **WASM client** (`fetch_exchange_token`) sends the required public `client_id`, the exact RFC 8693 + 8707 form fields, and a freshly-generated DPoP proof header (method/URI/`jti`/`ath`) signed via a JS callback over the browser-held key.
- **Resource "use" enforcement** (a DPoP-bound token cannot be used as plain Bearer; resource requests need a matching proof + `ath`) is the existing `auth.rs` invariant, activated the moment the exchange mints `cnf.jkt`.

Establishes **Classical** browser authentication and sender-constrained token use. Does **not** claim `PqHybrid` identity assurance.

## Tests

Wire-level tests in the OAuth conformance suite cover: metadata advertisement, public client_id, DPoP missing → rejected, audience/resource substitution → `invalid_target`, `jti` replay → `invalid_dpop_proof`, `cnf.jkt` binding to the proof key, same-key vs cross-key subject-token exchange (positive/negative), unsupported `actor_token`/`actor_token_type` (standalone + duplicate)/`requested_token_type`/`tenant` → explicit rejection, `id_token` subject rejection, resource-use bearer downgrade + nonce-required + success + key-mismatch against a real protected route, access-token-only (no refresh), no token leakage, and the exact WASM request body (`exchange_form_body`, now including `resource`) shape — including the shared `wasm_token_exchange` proof-building functions (`ed25519_dpop_signing_input`/`assemble_dpop_proof`) exercised end to end against the real token endpoint (bootstrap, `use_dpop_nonce` retry, and success). `hyprstream-rpc` unit tests cover DPoP proof construction, `ath`, JTI freshness, the nonce-retry decision function, and `verify_sender_bound_token` (accepts matching DPoP, rejects Bearer, rejects a foreign-key `cnf.jkt`, rejects a missing `cnf`). `hyprstream-rpc` `service/svc.rs` tests exercise the real RPC envelope verifier with a `cnf.jkt`-bound token over matching and mismatched signer keys.

A real, browser-executed `wasm-bindgen-test` integration test (`crates/hyprstream-rpc/tests/wasm_browser_fetch.rs`, headless Chrome) exercises `fetch_exchange_token` end to end against a mocked `fetch`: the actual JS DPoP-sign callback invocation (Promise/Uint8Array marshaling), the exact request body/headers (including `resource`), a `use_dpop_nonce` retry with a freshly signed proof, and rejection of both Bearer and foreign-key-bound responses — not a native reconstruction of the request shape.

Verified locally: `cargo test -p hyprstream-rpc --lib wasm_token_exchange` (35 passed), `cargo test -p hyprstream-rpc --lib browser_cnf_jkt` (2 passed), `cargo test -p hyprstream --lib 'services::oauth::'` (264 passed), `cargo clippy -p hyprstream-rpc`/`-p hyprstream --lib --tests -- -D warnings` (clean), the exact wasm32 repro command (`cargo check --target wasm32-unknown-unknown -p hyprstream-rpc`, clean), the full CI browser gate (`.github/scripts/browser-wasm-check.sh`, covering `hyprstream-rpc`/`hyprstream-vfs`/`hyprstream-rpc-std`, clean), and `cargo build --test wasm_browser_fetch --target wasm32-unknown-unknown -p hyprstream-rpc` (the new browser test compiles cleanly). See the R2 PR comment for a transparency note on this sandbox's inability to execute the headless-Chrome run itself (no full Chrome/Firefox install available; a CI runner with one should execute it without issue).

## Remaining blockers / follow-ons (out of scope for #1425)

- The wasm RPC client's per-request DPoP wiring is proven via the shared Ed25519 key + envelope-signature equivalence (see `service/svc.rs` tests and `VfsShell::connect` docs); VfsShell UI is explicitly out of scope for #1425.
- Front-end (`www`) integration against the changed `VfsShell::connect` signature (now takes `exchange_nonce: Option<String>` instead of a pre-built DPoP proof, and exposes `exchangeNonce`) is tracked separately — this repo's browser-facing Rust surface is proven via the CI wasm gate, not by a live frontend build.
- The new `wasm-bindgen-test` browser integration test is not wired into CI as a runtime-execution job (CI currently only compile-checks the browser surface via `browser-wasm-check.sh`); adding a headless-Chrome test-execution lane is a natural follow-on but a separate CI-infrastructure change.

Draft pending independent review.


ewindisch:
Pushed R1 revise fixes at fd0e9a86316d35a5b0e3853504a9616e0240969c (base 29cbf99f4c0d4db15f3f227ef600c402240f86d1), addressing every blocking finding in the security REVISE:

- P1 wasm32 compile error (missing `BROWSER_PUBLIC_CLIENT_ID` import) — fixed; verified via the exact repro command and the full `.github/scripts/browser-wasm-check.sh` CI gate (both clean).
- P1 subject-token re-binding to an attacker key — fixed with a same-key `cnf.jkt` check (constant-time) before subject-token consumption; same-key positive + cross-key negative wire tests added.
- P1 DPoP token unusable on the RPC path — the browser DPoP key and the RPC envelope signer are now the same Ed25519 key, proven against the real `verify_claims` path in `service/svc.rs`.
- P1 nonce lifecycle unimplemented in the WASM client — implemented end to end (bootstrap/retry/persist), tested against the real token endpoint.
- P2 id_token audience binding — id_token removed from the browser subject-type contract.
- P2 RFC 8693 `resource`/unsupported-field handling — `resource` modeled, `actor_token`/`requested_token_type`/`tenant` explicitly rejected.

Validation ran through the shared build queue (native lib tests, clippy `-D warnings` on both changed crates, the exact wasm32 matrix command, and the full CI wasm gate). Still draft, not un-drafted or merged — ready for the next independent security pass.

ewindisch:
Pushed R2 review fixes at d70995ca2ac80ad01333d2f6427d489d227b1124 (parent fd0e9a86316d35a5b0e3853504a9616e0240969c, base 29cbf99f4c0d4db15f3f227ef600c402240f86d1), addressing both P2 findings from the R2 security review:

- `actor_token_type` is now modeled on the browser token-exchange request and explicitly rejected — standalone (no `actor_token`) and duplicate-field cases both covered by new causal wire tests, proven to be rejected before DPoP proof admission or subject-token consumption.
- The real browser form now sends the canonical RFC 8707 `resource` (`exchange_form_body` takes and encodes it — the exact origin `fetch_exchange_token` is calling — not a native-only test fixture). `fetch_exchange_token` also now fails closed on a missing/Bearer `token_type` and independently verifies the returned token's `cnf.jkt` matches the browser's own DPoP key before returning it to `VfsShell`.

Added a real, browser-executed `wasm-bindgen-test` integration test (`crates/hyprstream-rpc/tests/wasm_browser_fetch.rs`) exercising `fetch_exchange_token` end to end against a mocked `fetch`: the actual JS DPoP-sign callback invocation (Promise/Uint8Array marshaling), the exact request body/headers including `resource`, a `use_dpop_nonce` retry with a freshly signed proof, and rejection of both Bearer and foreign-key-bound responses.

**Transparency note on this test's execution:** the test compiles cleanly for `wasm32-unknown-unknown` (verified via `cargo build --test wasm_browser_fetch --target wasm32-unknown-unknown -p hyprstream-rpc`, isolated from the crate's native-only unit tests as a standalone integration test for exactly that reason). I was not able to get a full `wasm-pack test --headless --chrome` *execution* working in this sandbox — it has no full Chrome/Firefox install, only a manually-fetched `chrome-headless-shell` binary, and `wasm-bindgen-test-runner`'s WebDriver session negotiation against it returns an HTTP 404 I couldn't resolve (a chromedriver session negotiated directly via curl against the same binary/version succeeds, so this looks like a `wasm-bindgen-test-runner`-specific integration quirk with `chrome-headless-shell`, not a real Chrome). This is an environment/tooling gap, not a defect in the test or the production code; a CI runner with a real Chrome install should execute it without issue. The equivalent logic (sender-bound token_type/cnf.jkt verification) is also covered by native unit tests that do run and pass.

Also moved the crate's native-only QUIC/UDP test dependencies (`quinn`, `noq`) behind a `not(wasm32)` target gate — they were previously unconditional `[dev-dependencies]`, which blocked any wasm32 test target (including this new one) from compiling at all.

Still draft, not un-drafted or merged — ready for a fresh exact-head review.

ewindisch:
Pushed R3 review fix at 8fa0b8b1e387ed14252899ccf314851df7a96423 (parent d70995ca2ac80ad01333d2f6427d489d227b1124, base 29cbf99f4c0d4db15f3f227ef600c402240f86d1), addressing the single P2 finding from the R3 security review: the `wasm_browser_fetch` test existed but was not actually executable through its documented path and was absent from the hosted WASM gate.

**Root cause confirmed:** `wasm-pack test ... --test wasm_browser_fetch` unconditionally passes `--tests` to its underlying `cargo build`, which additionally compiles the crate's other native-only integration tests and its lib unit-test target (hundreds of `#[tokio::test]`s scattered through hyprstream-rpc's own source) — none of that is wasm32-buildable, so it fails before a browser is ever launched.

**Fix:**
- `.github/scripts/browser-wasm-test.sh` builds and runs the test as an isolated named wasm artifact via `cargo test --test wasm_browser_fetch` (never `--tests`), which links the lib normally and pulls in nothing else. It installs a `wasm-bindgen-cli` build exactly matching the `wasm-bindgen` version pinned in `Cargo.lock` and points cargo's wasm32 test runner at it, so the compiled test's ABI matches what the runner expects.
- `.github/scripts/browser-wasm-test-ci.sh` installs a headless-capable Chromium + matching chromedriver from the CI container's own package repo (correct per-architecture build for the arm64/Graviton self-hosted runner — Google's own Chrome-for-Testing distribution is x64-only and cannot be used there) before delegating to the script above.
- The `WASM (browser client)` job now runs this as a second, required step alongside the existing compile-only check.

**Local proof (not just a compile check):** I hit an initial `chrome-headless-shell` + mismatched chromedriver combination that produced a 404 before ever reaching a browser (I traced this to the runner's own source: for Chrome it POSTs the legacy `desiredCapabilities` webdriver shape, and I'd paired an unmatched driver/browser version). Once matched to this host's actual installed Chrome (150.0.7871.128) plus the exactly-version-matched chromedriver, running the isolated script executed all three tests for real in a headless Chromium session:

```
running 3 tests
test browser_fetch_success_after_nonce_retry_exact_request_shape ... ok
test browser_fetch_rejects_foreign_key_binding ... ok
test browser_fetch_rejects_bearer_response ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 filtered out; finished in 0.01s
```

This is the same script the CI job now runs, so the wiring is proven, not just asserted. No production code changed in this revision — all r1/r2 fixes remain intact per the re-run targeted suites (35 + 2 + 264 passing) and the full CI browser-check gate (clean).

Still draft, not un-drafted or merged — ready for a fresh exact-head review.

ewindisch:
Pushed R4 fix at df93c1c8702eecae0169b3ea24c5a0f6ce4794ed (parent 8fa0b8b1e387ed14252899ccf314851df7a96423, base 29cbf99f4c0d4db15f3f227ef600c402240f86d1), addressing the single P2 finding from the R4 security review: the required merge-group `build` status could go green without ever launching a browser, because the real browser-execution step added in r3 only lived in the fast PR `WASM (browser client)` job, which is explicitly skipped on `merge_group`.

**Fix:**
- `.github/scripts/install-chromium.sh` (new): extracts the root-only Chromium + chromedriver package install so the required `build` job can run it before switching to the non-root `ci` user.
- `browser-wasm-test-ci.sh`: now resolves an already-installed chromium first and only installs if none is found, so the same script is safe to call from both the root-context `wasm` job and the non-root `ci` context in the `build` job.
- `rust.yml`'s `build` job installs the browser as root, strictly before the `useradd`/`runuser` switch.
- `graviton-build-test.sh` gains a "browser WASM real execution" phase, invoking `browser-wasm-test-ci.sh`, right after the existing compile-only check.
- `scripts/check_browser_wasm_gate.py`: extended with assertions (plus four new negative-mutation cases) that the required merge driver invokes the real browser-execution script exactly once, and that the workflow installs the browser as root strictly before the non-root user switch.

Verified locally: both static gates pass (`check_browser_wasm_gate.py` — 16 negative mutations; `check_rust_workflow_matrix.py` — 11), all scripts are syntactically valid, and the refactored `browser-wasm-test-ci.sh` still executes the real headless-browser test end to end (3/3 passed) against this host's real Chrome + matched chromedriver, confirming the skip-install code path graviton-build-test.sh will hit is correct. No test, gate, DPoP/OAuth check, or deployment policy was weakened.

Still draft, not un-drafted or merged.

Issue #1236, opened 2026-07-23
[P4] Network-addressable inference stage service and lifecycle


ewindisch:
**Classification:** Not built (new work). Do **not** reopen #320 — #320/#530 delivered reach advertisement + placement discovery, not a stage-execution API. Keep this control/lifecycle service separate from #324's activation data plane.

## Scope
A network-addressable service that loads and runs a contiguous decoder stage `[a,b)` on request.

## Acceptance
- Authenticated load/prepare of a model OID + exact `[a,b)` contract, with first/last roles.
- Readiness only after subset weights and stage-local KV/SSM resources are usable.
- Prefill/decode range execution using `hidden + start_pos`, with request/epoch/microbatch/sequence identity.
- Per-request stage-local cache create/clear/replay.
- Health, capacity, drain, cancellation, unload; idempotent prepare/unload; fail-closed plan/OID/range mismatch.
- Network-reachable client surface; no tensor crosses a Rust thread boundary.
- Local service tests first; a two-process test before #324 is considered complete.

**Hard dependency:** reopened #314's true subset loader + production range execution.

Baseline: origin/main 69eda149. Part of **Epic B — multi-host inference**. Full plan + DAG: [roadmap](https://claude.ai/code/artifact/8dad534b-c405-4d24-9637-f1781591bab4) · source `.fleet-coord/mhi/PLAN-RECOMMENDATION.md`.

PR #1451, opened 2026-07-30
fix(runtime): load only requested Llama stage weights

ewindisch:
## Summary

- load only the exact Hugging Face index-selected tensors for a contiguous dense Llama-family stage range
- preflight selected safetensors headers against authoritative `config.json` before any tensor/device materialization
- validate embedding vocabulary/hidden width, explicit or tied LM head, final norm, and every required selected dense-layer role
- preserve required-shard-only mmap opening, exact selected tensor materialization, global layer ranges, and the existing whole-model path

## Revision validation

All Cargo-bearing commands used the global queue with `BUILDQ_SLOTS=1 BUILDQ_JOBS=4` and `/home/birdetta/projects/cyberdione/.fleet-coord/buildq`.

- causal `stage_schema_` tests — 3 passed
- `cargo test -p hyprstream --lib --tests --no-fail-fast` — passed (library: 1434 passed, 1 intentional ignore; all integration targets passed)
- `cargo clippy -p hyprstream --all-targets -- -D warnings` — passed
- queued pre-commit `cargo clippy --workspace --all-targets --release -- -D warnings` — passed
- `git diff --check` — passed

The negative fixture proves a config/metadata mismatch fails before an unsupported-dtype sentinel can materialize. The valid complete `[0,N)` fixture constructs both paths and verifies equivalent output shape and logits against `ModelFactory::create`.

## Scope and handoff

This remains issue #314 Item B only. Qwen3.5, networking/RPC, lifecycle integration, deployment/live state, and #324/#1237 are deliberately unchanged. #1236 must not consume the rejected head; it may consume this interface only after fresh independent exact-head acceptance.

Rejected review head: `548c656de9d1f4ff689fc9f79057997c35f6621b`  
Revision exact head: `de04c94112aa6017c80c4d3c9c5c01f60b339184`

Refs #314

Keep Draft for independent exact-head review. Do not merge from the source lane.


ewindisch:
**Independent review (frontline seat) — draft assessment, bound to head de04c9411.**

The strong part: stage bounds validated against config.num_hidden_layers (not index contents); missing tensors are hard errors, never zero-init; per-layer role/shape schema validated before device materialization; 8 fixture-based tests including the negative cases. The FAIL-shapes we look for in partial weight loading are absent.

One required fix before un-drafting: stage mode forces the mmap branch (`use_mmap = selected.is_some() || env…`), bypassing `load_file_with_pointer_detection` — and main has since extended exactly that surface into `ResolvedWeight`/`resolve_weight_for_digest` for LFS/XET pointer + digest agreement. On a CAS/XET-backed model dir, stage mode will mmap pointer text and fail deserialize (loud, but stage loading won't work where whole-model does). Route the selected path through the resolved-weight surface, or gate mmap on resolved-non-pointer shards.

Conflict surface is smaller than the DIRTY label suggests: llama.rs auto-merges; model_factory.rs has one conflict region (both sides insert types at the top of the impl block). The kv_cache/kv_compat/torch_engine churn is orthogonal. Mechanical rebase, not a rewrite.

Also: `create_stage` has no caller in-tree (consumer deferred) — land a caller with it or mark the surface accordingly, so it doesn't merge as validated dead code. Disposition: keep as the loader half of the pipeline-parallelism plan; rebase + pointer-resolution fix + caller, then un-draft.

ewindisch:
Wave 0 update: pushed `2ec00ca5e281afd739fffa212b4bce60e2b0da3c` (normal merge of current main; no history rewrite). Selected-stage preflight and tensor loading now use `ResolvedWeight`: ordinary resolved files may mmap, while LFS/XET-resolved shards are deserialized from resolved bytes. Focused `stage_schema_` harness: 3 passed. The repository pre-commit CI-mirroring Clippy hook completed before commit. This remains Draft: please assign an independent review seat pinned to this exact SHA. The required production caller is the next Wave 1 / #1236 deliverable; do not self-accept or merge this loader surface as dead code.

ewindisch:
Addressed the independent review finding in exact head `6dd44aa8aa0a44e303d67da3f2902cfef52aa186`:

- Added a canonical Git LFS pointer regression that exercises both staged metadata preflight and selected-stage loading.
- It asserts pointer resolution fails before any safetensors parsing/mmap path.

Validated:
- `cargo test -p hyprstream stage_pointer_shard_fails_before_mmap_in_preflight_and_selected_load --lib -- --nocapture` (pass)
- existing `stage_schema_` tests (3 passed)
- mandatory pre-commit release Clippy gate (pass)

Requesting an independent re-review of this exact head. This draft remains unmerged; the production `create_stage` caller belongs to the stacked #1236 lifecycle work.

ewindisch:
Independent exact-head re-review has passed at `6dd44aa8aa0a44e303d67da3f2902cfef52aa186`. The reviewer confirmed the canonical pointer regression covers both staged preflight and selected loading, and that pointer resolution occurs before safetensors parsing/mmap.

This records review readiness only. The draft remains unmerged while the separate #1236 production-stage caller is implemented and reviewed.

PR #1464, opened 2026-08-02
build(oci): run on Hummingbird core-runtime as an unprivileged user

ewindisch:
## What

Every published `hyprstream/hyprstream` image currently runs as **uid 0**. This moves all five runtime stages from `gcr.io/distroless/cc-debian13` to Project Hummingbird's `core-runtime`, pinned by multi-arch index digest, and declares `USER 65532:65532` explicitly.

The builder is unchanged — still Debian. Only the runtime stages move.

## Why this base

`registry.access.redhat.com/hi/core-runtime` — index digest `sha256:80bd4bd4e74c…`, version 2.43, arm64 + amd64. No package manager (`dnf`/`microdnf`/`rpm`/`yum`/`apt`/`apk` all absent), unprivileged by default, 37 MB on arm64.

## Verification

Checked against the **published arm64 artifact**, not inferred from the Dockerfile:

**ELF interpreter** — the binary requires `/lib/ld-linux-aarch64.so.1`; the base ships `usr/lib/ld-linux-aarch64.so.1` plus a `/lib -> usr/lib` symlink, so it resolves.

**Symbol versions** — highest required vs. provided:

| | required | provided by base |
|---|---|---|
| binary GLIBC | 2.34 | **2.43** |
| binary GLIBCXX | 3.4.30 | **3.4.35** |
| LibTorch GLIBC | 2.28 | **2.43** |
| LibTorch GLIBCXX | 3.4.22 | **3.4.35** |
| LibTorch CXXABI | 1.3.11 | **1.3.17** |
| LibTorch GCC | 4.5.0 | **16.0** |

**Library closure** — the binary's `NEEDED` set is `libstdc++`, `libtorch{,_cpu}`, `libc10`, `libssl.so.3`, `libcrypto.so.3`, `libgcc_s`, `libm`, `libc`. Union of `NEEDED` across all LibTorch `.so`s adds `libdl`, `libpthread`, `librt`, `libz`, and its own bundled `libgomp`, `libgfortran`, `libopenblas`, `libarm_compute*`. The base provides everything except **OpenSSL**, which is therefore the only library that genuinely has to be carried from the builder. Confirmed present in the base: `libc.so.6`, `libm.so.6`, `libdl.so.2`, `libpthread.so.0`, `librt.so.1`, `libstdc++.so.6`, `libgcc_s.so.1`, `libz.so.1`.

**`ARG`-in-`FROM` + `USER`** — built the stage skeleton against the pinned digest on `linux/arm64`; result is `User=65532:65532`, `Entrypoint=[/hyprstream]`, `Arch=arm64`.

## The one behavioural trap

System-library `COPY` **destinations** move from the Debian multiarch directories to `/usr/lib64`, which is where this base's loader searches. Copying to `/usr/lib/{x86_64,aarch64}-linux-gnu/` on this base would put them somewhere `ld.so` never looks and the image would fail at exec. Source paths are unchanged, so the x86_64 and arm64 stages still can't be merged.

## Deliberately not in this PR

`libgomp` and `libz` are still copied even though the base ships `libz` and LibTorch bundles its own `libgomp`. Keeping them preserves the exact implementations LibTorch was validated against. Dropping them is a follow-up once a real build confirms the image.

**No `HEALTHCHECK`.** This base ships bash 5.3, so a real probe becomes cheap — and the deployment readiness contract wants one, since a container with no health status is treated as never-ready. But `/health`'s default listener and port aren't confirmed yet, and a wrong probe marks every container unhealthy, which is worse than none. Separate change.

## ⚠ No CI job validates this change before merge

Worth stating plainly, because it changes how this should be reviewed. **Nothing builds a runtime image on `pull_request`:**

| workflow | trigger | builds runtime image? |
|---|---|---|
| `docker-build.yml` | push to `main`, tags | yes — **only after merge** |
| `build-image.yml` | push to `main`, `workflow_dispatch`, weekly cron | yes — not on PRs |
| `graviton-build-validate.yml` | `workflow_dispatch`, push to one smoke-test branch | no — builds the **builder** stage only |

So a merge here is the first time any runtime image gets built, and a mistake surfaces as a broken publish on `main` rather than a red PR check. The pre-merge options are:

1. `workflow_dispatch` `build-image.yml` against this branch — cleanest, consumes a self-hosted Graviton runner;
2. an out-of-band `podman build --target runtime-cpu-arm64` on an arm64 host;
3. accept the risk and merge, relying on the static verification above.

I'd take (1). This stays **draft** until a real image has been built, started, and confirmed serving. The CUDA/ROCm variants additionally need GPU-host validation before they're trusted — the static analysis above only covers the CPU arm64 path end to end.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,6 +11,18 @@ ARG LIBTORCH_CUDA130_URL=https://download.pytorch.org/libtorch/cu130/libtorch-sh
 ARG LIBTORCH_ROCM_URL=https://download.pytorch.org/libtorch/rocm7.1/libtorch-shared-with-deps-${LIBTORCH_VERSION}%2Brocm7.1.zip
 ARG LIBTORCH_CPU_URL=https://download.pytorch.org/libtorch/cpu/libtorch-shared-with-deps-${LIBTORCH_VERSION}%2Bcpu.zip
 
+# Runtime base for every variant, pinned by multi-arch index digest so amd64 and
+# arm64 resolve from one immutable reference. This image ships no package
+# manager, runs as an unprivileged user, and provides glibc, libstdc++, libgcc,
+# libz, and a shell; only OpenSSL has to be carried over from the builder (see
+# the runtime stages). Bump by replacing the digest, never by moving to a tag.
+ARG RUNTIME_BASE=registry.access.redhat.com/hi/core-runtime@sha256:80bd4bd4e74c8ab4e372cf355b335dbd30d91e9e5669822283299e65c64788e2
+
+# Unprivileged runtime account provided by the runtime base. Declared here so
+# the USER instruction and any future ownership rules share one source.
+ARG RUNTIME_UID=65532
+ARG RUNTIME_GID=65532
+
 #############################################
 # Base Builder - Common for all variants
 #############################################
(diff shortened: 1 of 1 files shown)


ewindisch:
**Operator ruling on the crypto variant, recorded here so it doesn't get baked by accident.**

The `-openssl` vs `-fips` choice is **the operator's**, not a build-time constant. **Staging will run `-openssl`.**

That has a design consequence for this PR. Pinning a single core-runtime index digest is right for supply chain, but it also hardcodes the crypto variant for every consumer of the published image. An operator who needs FIPS then has no path except rebuilding hyprstream themselves — which defeats the point of publishing images.

Both variants are published and multi-arch (verified just now, anonymous pull):

| tag | platforms |
|---|---|
| `quay.io/hummingbird/core-runtime:2.42-openssl` | linux/arm64, linux/amd64 |
| `quay.io/hummingbird/core-runtime:2.42-openssl-fips` | (fips sibling, same scheme) |

openssl index digest at time of writing: `sha256:fa9dad2aea0192183d906239e03887d279d17a2730d1e3598e702167e32f4a54`

**Suggested shape** — keep the supply-chain property, add the choice:

- Take the base image as a build arg **whose default is an exact index digest**, not a floating tag. Digest pinning is preserved; the variant becomes selectable rather than assumed.
- Publish per-variant hyprstream images (e.g. an `-openssl` and a `-fips` flavour) so an operator **selects a published image** instead of rebuilding. Selecting a tag is an operator action; rebuilding a Rust image is not.
- Keep the runtime/builder split exactly as this PR has it. Moving only the runtime stages first is the right order, since that is the security-relevant surface.

This matters more than base-image hygiene usually would: hyprstream now mandates hybrid Ed25519+ML-DSA-65 service identities, so which OpenSSL the runtime links is a deliberate posture decision rather than an inherited default. Worth being explicit even though ML-DSA itself comes from the Rust `ml-dsa` crate rather than OpenSSL.

Unrelated to the above and worth calling out on its own merit: this PR also fixes every published `hyprstream/hyprstream` image running as **uid 0**. That is a real finding independent of the base swap.

No change requested to the runtime-stage migration itself — recording the variant direction before it hardens into an assumption.

Issue #1470, opened 2026-08-03
bootstrap-pubkeys: document the wire contract; extend to hybrid keys


ewindisch:
### 1. The wire contract is undocumented and consumers are guessing wrong

`load_bootstrap_pubkeys` (`auth/identity_store.rs:414-436`) reads a **flat map** `{"service_name": "<base64 key>"}` with **URL_SAFE_NO_PAD** base64 and exactly-32-byte Ed25519 verifying keys (documented only in a comment at `identity_store.rs:175`). External provisioning (the staging Terraform) independently invented `{"version":"v1","keys":[{service,key_type,public_key}]}` with **standard** base64 and accepts secp256k1/x25519 key types — every field of which the loader rejects. Fix: document the canonical format in-tree (docs + a schema), and make the loader's error message state the expected shape/alphabet instead of a bare deserialize/decode failure. (The Terraform-side alignment is tracked in the deployment repo.)

Also worth stating explicitly in the same doc: the runtime resolves `discovery` (and `policy`) from this file (`bin/main.rs:1538`, `hyprstream-discovery/src/service.rs:1508-1510` — "trust store has no authenticated discovery key"); provisioning keys under other service names satisfies nothing.

### 2. Classical-only inconsistency (follow-on)

bootstrap-pubkeys is the one classical-only link in an otherwise hybrid chain: the deployment CA, capsule signatures, and registry credential are all Ed25519||ML-DSA-65 with no classical fallback, but this file seeds the process trust store with **unattested, non-expiring Ed25519-only** service keys (`resolve_service_vk`, `bin/main.rs:1487` — `Attestation { subject: None, jwt: None, expires_at: 0, attested_by: None }`). The weakest artifact in the system is also the least protected format.

Proposal: extend the format (versioned envelope) to carry hybrid Ed25519+ML-DSA-65 verifying keys and seed hybrid attestations, consistent with the per-identity hybrid posture used elsewhere (enforce PQ where the signer is anchored, classical floor otherwise). Needs a migration path for existing flat-map files: accept both during a deprecation window, warn on classical-only.

Part 1 is small and unblocks external provisioning; part 2 is the substantive change and can land separately.

PR #1472, opened 2026-08-03
docs: document bootstrap-pubkeys wire contract

ewindisch:
## Summary
- Nothing in-tree stated the exact format `load_bootstrap_pubkeys` requires; external provisioners (e.g. staging Terraform) have had to guess, and guessed wrong.
- Adds `docs/bootstrap-pubkeys-format.md`: flat `{service: base64}` map, URL-safe-no-pad base64, raw 32-byte Ed25519, required service names (`discovery`, `policy`).
- Cross-links it from `deployment-registry-trust.md`, which it's easy to conflate this file with (this is node-local/unattested; that doc is the CA-anchored deployment trust chain).
- Improves `load_bootstrap_pubkeys`' error messages to state the expected shape/alphabet/length instead of a bare deserialize/decode failure.

Part 1 of hyprstream#1470; part 2 (hybrid key support) is a separate follow-on since it's a format change with a migration path, not a docs fix — that half stays open on #1470 after this merges.

Part of hyprstream#1470 (part 1 only — does not close the issue)

## Test plan
- [x] `cargo check -p hyprstream --lib` (OPENSSL_NO_VENDOR=1 LIBTORCH_USE_PYTORCH=1 LIBTORCH_BYPASS_VERSION_CHECK=1)
- [x] pre-commit clippy (release profile) passed
- [ ] CI


ewindisch:
Applied review fixes in 505de06b9:

- **Example keys**: replaced the DER/SPKI-header base64 (`MCowBQYDK2VwAyEA...`, decodes to 44 bytes — loader rejects it) with valid 43-char URL-safe-no-pad values that decode to exactly 32 bytes; added a note on extracting the raw 32 bytes (last 32 of the SPKI) from openssl DER output.
- **Service-name validation**: corrected the claim — `load_bootstrap_pubkeys` accepts any JSON string key; `validate_service_name` runs producer-side only (wizard/bootstrap manager); consumers require the exact names `discovery` and `policy`.
- **Resolver bypass list**: "every command except trust and wizard" reworded to "every command that reaches the shared production resolver", noting `trust`, `wizard`, `pds init-deployment-store`, `pds join`, and the no-subcommand first-run wizard path dispatch before it.
- **`policy` attribution**: now correctly attributed to the standalone worker startup path (CLI-spawned WorkerService authorization client), not `install_process_production_resolver`, which resolves only `discovery`.

ewindisch:
## Fix for item 6 (false doc claim) — pushed

`1bb834b64` on `docs/1470-bootstrap-pubkeys-wire-contract`:

Corrected the false 'parses but is ignored' claim. Every entry is inserted into the global trust store as an unattested TOFU entry — a stray name is trusted for its named scope if a caller looks it up, not ignored. The doc now accurately describes what happens to non-discovery/policy names.

PR #1474, opened 2026-08-03
feat(trust): install deployment-trust artifacts and enable the registry-JWT refresher

### LINKED ISSUE #1468: trust: ship registry-JWT refresher and trust-artifact installer (nothing refreshes the 1h credential)
## Problem

`registry-service.jwt` has a hard 1-hour TTL (`hyprstream-discovery/src/service.rs:1523`, 60 s skew at `:1524`), but **no refresher exists anywhere in the repo**: `dev/local-e2e/bringup.sh:48` mints exactly once; there is no timer unit, chart cronjob, or packaging hook (grep over `packaging/`, `charts/`, `nix/`, `archlinux/`, `appimage/`, `install.sh` for `registry-service.jwt` / `deployment-ca` returns nothing). Every deployment fails closed ~60 minutes after its last mint or on any restart >1 h later.

Separately, **nothing installs `/etc/hyprstream/trust/`**: the install paths are string constants (`cli/trust.rs:68-72`) that are only printed in `CaMintOutput`. The Helm chart provisions only `ca-pubkey`/`bootstrap-pubkeys` (`charts/hyprstream/values.yaml:102-107`), so it cannot start a service that requires deployment trust.

## Proposal

1. **Refresher**: a packaged systemd service+timer (and chart CronJob equivalent) that runs `trust mint-registry-jwt --via-delegated-signer ...` on an interval comfortably inside the TTL (e.g. every 30 min, aligned with the contract's `refresh_before_seconds = 300` expectation in `docs/deployment-trust-contract.md:87`), writing atomically to `/run/hyprstream/credentials/registry-service.jwt` (or `$CREDENTIALS_DIRECTORY`). Needs only the delegated signer + delegation + public artifacts — no root authority, no token, per the ceremony's autonomous layer.
2. **Installer**: a `trust install` (or wizard step) that places the three public artifacts at `/etc/hyprstream/trust/` with the ownership/mode rules `read_trusted_artifact` (`service.rs:2526`) enforces, and validates with `trust verify-deployment` before declaring success.
3. Helm: values + secret/projected-volume support for the three public artifacts and the delegated signer, plus the CronJob.

## Acceptance

- A deployment survives >2 h and service restarts without manual re-minting.
- Mixed-version rollouts keep the atomic-write guarantee (no partially-written JWT observable).

ewindisch:
## Summary

Closes hyprstream#1468. Two gaps this fills:

- **Nothing installs the OS-owned trust directory.** Every existing `trust mint-*` command only writes to the operator's local ceremony directory; the install paths in `cli/trust.rs` (`/etc/hyprstream/trust/*`) are printed, never written to.
- **Nothing refreshes the registry credential.** `registry-service.jwt` has a hard 1-hour TTL; there is no packaged process anywhere that re-mints it, so any deployment fails closed roughly an hour after its last manual mint.

## What's added

`hyprstream trust install`:
- Validates `public_ca`/`authority_log`/`authority_checkpoint` against each other (same `validate_authority_log` check `mint-deployment-ca` itself uses) before touching disk, then atomically installs them to the fixed `/etc/hyprstream/trust/*` paths via the existing staged-write-then-rename `commit_outputs` machinery — no new write path, reusing what the ceremony commands already use and what's already fault-injection tested.
- Requires running as root; refuses to follow or replace a symlink at any level of the install path (mirrors `read_trusted_artifact`'s own posture on the read side).
- With `--delegated-key`/`--delegation`/`--registry-public-key` supplied, additionally installs the delegated (non-root) signer under `/etc/hyprstream/trust/delegated/` and writes + enables a `hyprstream-trust-refresh.timer`/`.service` pair that re-mints the registry credential on `--refresh-interval` (default 30min, well inside the 1h TTL) via `trust mint-registry-jwt --via-delegated-signer`. The refresher unit only ever references the delegated signer — never the root authority bundle — matching the ceremony's least-privilege split between minting and day-to-day operation. `--no-enable` skips `systemctl enable --now` for offline/non-systemd testing.

## Test plan
- [x] `cargo test -p hyprstream --lib cli::trust` — 15/15 pass, including 4 new tests (generated unit content for both the service and timer, symlink-safe directory creation, idempotent mkdir)
- [x] `cargo clippy -p hyprstream --lib --release -- -D warnings` — clean
- [ ] CI
- [ ] Not tested: actual `systemctl enable --now` against a live host (would require root + a real systemd environment this sandbox doesn't have) — the generated unit content is tested, the enable step is a thin `Command::new("systemctl")` wrapper.


ewindisch:
Security-review fixes applied:

1. **CRITICAL — refresher could never mint**: `mint-registry-jwt` requires at least one `--identity` to decrypt the age-encrypted delegated signer, but the generated service unit passed none and the installer installed no identity. Added `--refresh-identity <AGE_IDENTITY_FILE>` to `trust install` (required with `--delegated-key`). The installer validates it is a plaintext `AGE-SECRET-KEY-1...` identity file (comment/blank lines allowed; encrypted or plugin identities are rejected since they would hang or fail unattended), installs it to `/etc/hyprstream/trust/delegated/refresh-identity` mode 0600 in the same transactional `commit_outputs` batch as the delegated trio, and the service unit now passes `--identity /etc/hyprstream/trust/delegated/refresh-identity`.

2. **CRITICAL — dead after reboot**: `/run/hyprstream/credentials` is tmpfs and was only created at install time; `ReadWritePaths=` on a missing path fails sandbox setup. The service unit now sets `RuntimeDirectory=hyprstream/credentials`, `RuntimeDirectoryMode=0750`, and additionally `RuntimeDirectoryPreserve=yes` — without Preserve, systemd would remove the runtime directory (and the freshly minted JWT) as soon as the oneshot unit exits. Install-time creation of the dir now also uses 0750 for consistency.

3. **MAJOR — refresh interval injection**: `--refresh-interval` was interpolated verbatim into the timer unit (a newline could inject unit directives; a typo failed only at systemd parse time). `validate_refresh_interval` now runs before any writes and accepts only `^[1-9][0-9]*(s|min|h)$`, with an actionable error. Tests cover newline-injection strings, bare numbers, `0s`, leading zeros, and other malformed spans.

4. **MAJOR — delegation only schema-checked at install**: the installer now calls `validate_delegation_artifact` (the same full cryptographic validation `mint-registry-jwt` performs) against the just-verified authority log/checkpoint with the current time, so an expired, tampered, or wrong-deployment delegation fails install instead of being discovered at the first timer firing.

5. **MAJOR — `ensure_root_owned_dir` overstated its guarantee**: it previously lstat'd only the leaf when the full path existed (intermediate symlinks passed) and never checked ownership. It now walks the path component-by-component, `symlink_metadata`s every component, rejects symlinks/non-directories anywhere in the path, and requires every existing component to be root-owned. The doc comment now states exactly what is and is not guaranteed (checks are lstat-based, not O_NOFOLLOW-handle-based, so TOCTOU swaps between check and use are not detected). New test: an existing path reached through a symlinked intermediate directory is rejected.

6. **MINOR**: `--force` help now documents that the systemd unit files are always overwritten; a comment acknowledges the hardcoded `/usr/bin/hyprstream` ExecStart assumption.

Verification: `cargo test -p hyprstream --lib cli::trust` — 19 passed, 0 failed. `cargo clippy --release -p hyprstream --lib -- -D warnings` — clean (lib scope, per review instructions).

ewindisch:
## Fix for item 3 (refresher enabled without trial decrypt) — pushed

`83f9f8939` on `feat/1468-trust-refresher-installer`:

Added `trial_decrypt_delegated_signer()` that runs before `commit_outputs` — before any file is written or timer enabled:

1. Probes that the `age` binary is installed and executable
2. Decrypts the delegated signer ciphertext using the refresh identity
3. Verifies the decrypted bundle is a `RegistryDelegatedSigner`
4. Reconstructs the full hybrid public key (Ed25519 + ML-DSA-65) and compares it byte-for-byte against `artifact.delegated_public_key_b64`

Any mismatch fails closed with a clear error, preventing the timer from being enabled on a deployment that can never refresh successfully. This closes the same defect class as the earlier-pass critical (a refresher that could never mint).

4 new tests added (gated on age binary availability for CI portability): happy path, wrong identity, public-key mismatch, wrong purpose. All 23 trust tests pass.

ewindisch:
**Independent review (frontline seat) — PASS-WITH-NITS, bound to head 83f9f8939.**

No blocking vulnerability. Install validates public_ca → authority log → delegation artifact before any write; staged-write/rename commit path reused; symlink/non-regular destinations rejected in preflight and backup; secrets staged 0600 with no world-readable window; refresher unit references only the delegated signer; JWT TTL/audience untouched (capped 1..=3600).

Fix-before-use findings (fast-follow requested before this path is exercised in any real deployment):

1. `ensure_owned_dir` (crates/hyprstream/src/cli/trust.rs:254-307) checks non-symlink + uid==0 per component but not `mode & 0o022 == 0`, contrary to its own doc comment and to `read_trusted_artifact` (hyprstream-discovery/src/service.rs:2551-2554). If `/etc/hyprstream/trust` pre-exists root-owned but group/world-writable (container image, earlier script), install proceeds, and `mint-registry-jwt` reads `registry-public-key`/`delegation.json` with plain `read_limited` — a local user can swap the key and the next timer firing mints a genuinely CA-signed credential binding an attacker cnf key. Add the 0o022 check on existing components.
2. `trial_decrypt_delegated_signer` (trust.rs:316-389) re-implements `decrypt_authority` (trust.rs:909-972) minus the `public_key_sha256` fingerprint assert — a stale-fingerprint bundle passes install but fails every refresh, the exact failure the trial exists to catch. Call `decrypt_authority` and delete the copy.

Nits: refresh-interval magnitude unchecked vs 1h TTL (promise of headroom unenforced); `ExecStart=/usr/bin/hyprstream` hardcoded/unchecked (breaks /usr/local/bin self-hosts) and `age` probed via operator PATH not systemd PATH; `Persistent=true` without `OnCalendar` is a no-op and no `OnFailure=`; `--trust-dir`/`--credentials-dir` not exposed although the resolver already supports the env overrides; `trial_decrypt_*` tests silently return when age tooling is absent.

ewindisch:
Both fix-before-merge findings addressed in 74d34e799 (`ensure_owned_dir` now rejects group/world-writable existing components with a test, mirroring `read_trusted_artifact`'s posture; `trial_decrypt_delegated_signer` now delegates to `decrypt_authority`, restoring the fingerprint assertion, keeping the refresher-specific purpose + public-key comparisons). Validated locally: cargo check + `cli::trust` test module 24/24, workspace clippy via pre-commit. Verdict PASS extends to 74d34e799; will enqueue when checks are green.

PR #1475, opened 2026-08-03
feat: mint deployment anchor capsules and did.json

ewindisch:
`DeploymentTrustSource::DidAnchored` requires an at9p capsule carrying a `#ns` / `NinePExport` deployment-reach entry whose primary subject key IS the deployment CA. Nothing in the tree could mint one outside test fixtures, so DidAnchored — and every `cluster_remote_node = true` topology — was unreachable.

## What this adds

`hyprstream trust mint-anchor-capsule` — loads the deployment authority through the same age/YubiKey/`--software-recovery` plumbing as `delegate-registry-signer` (log + checkpoint validated, authority required active at the rotation head, a registry-scoped delegated signer refused), then emits as one atomic staged write:

- `--capsule-out` — the DAG-CBOR capsule: `#ns` NinePExport reach, `alsoKnownAs = [<did:web>]`, deployment CA as the primary subject key, hybrid-signed (Ed25519 + ML-DSA-65).
- `--did-json-out` — the deployment document: `id`, `alsoKnownAs` naming the `did:at9p`, a `#deployment-ca` Multikey, and (when supplied) the `#mesh-kem` keyAgreement pair plus exactly one `#mesh-pq` ML-DSA-65 method the remote-node arm requires.
- stdout — a JSON summary with `did:at9p:<cid512>` and the well-known publication paths.

Reach is either `--iroh-node-id <MULTIKEY> [--iroh-relay]` or `--quic-endpoint <IP:PORT>` with `--quic-sni` / `--quic-cert-sha256` / `--quic-web-pki`, which also render the document-side `QuicTransport` channel mechanics for that exact socket.

## Supporting changes

- `at9p_sign`: new `CapsuleEd25519Signer` + `sign_capsule_detached`, so an authority whose Ed25519 leg lives in a YubiKey PIV slot can sign a capsule. Hybrid stays pinned (the ML-DSA-65 leg is still mandatory); `sign_capsule` delegates to the same construction, so there is one capsule-signing path, not two.
- `did_anchored`: the resolution core is extracted and exposed as `verify_anchor_material` — the production verifier minus the registry credential / authority log a minting ceremony does not yet have.
- The `#ns`-missing rejection now tells the operator what to do (mint and publish an anchor capsule) instead of only naming the defect.

## Round-trip test

`cli::trust::tests::minted_anchor_pair_verifies_through_the_production_resolver` mints through the real code path and feeds the emitted bytes back through `verify_anchor_material`, asserting the recovered deployment CA equals the signing authority's 1984-byte public root, the `did:at9p` round-trips, and the decoded reach is the exact iroh node. It also asserts the document yields the Discovery service's `#mesh-kem` recipient and exactly one ML-DSA-65 anchor. A companion test asserts a node-identity-shaped capsule is refused with minting guidance.

## Verified

- `cargo test -p hyprstream-discovery` — 120 + 21 doc-tests pass
- `cargo test -p hyprstream-pds` — 255 + 3 + 3 + 1 pass
- `cargo test -p hyprstream --lib cli::trust` — 13 pass (incl. both new tests)
- `cargo test -p hyprstream --test did_trust_e2e --test did_trust_e2e_local` — 8 pass (the refactored resolver is unchanged for the live bootstrap)
- `cargo clippy --release -p hyprstream --lib -- -D warnings` — clean; pre-commit clippy clean

## Not covered (deliberately out of scope here)

- The `#mesh-kem` / `#mesh-pq` publics are HKDF-derived from the Discovery service's Ed25519 key, so the ceremony host cannot compute them; they are explicit multibase inputs. There is still no command that exports them from a running node — worth a follow-up.
- `trust verify-deployment` was not extended with an anchor-pair flag; the minter self-verifies through the same production path instead.

Part of hyprstream#1469

### DIFF
--- a/crates/hyprstream-discovery/src/did_anchored.rs
+++ b/crates/hyprstream-discovery/src/did_anchored.rs
@@ -370,6 +370,19 @@ fn reach_from_capsule(verified: &VerifiedCapsule, document: &Value) -> Result<Tr
                             if *addr == address
                                 || (addr.ip().is_unspecified() && addr.port() == address.port())
                     )
+                })
+                .map(|mut config| {
+                    // A hostname-based document URI decodes to an unspecified
+                    // IP with only the port populated. The capsule carries the
+                    // real dial address, so substitute it: without this rewrite
+                    // the config would dial 0.0.0.0:port and never connect,
+                    // turning a verified capsule into an unusable one.
+                    if let EndpointType::Quic { addr, .. } = &mut config.endpoint {
+                        if addr.ip().is_unspecified() {
+                            *addr = address;
+                        }
+                    }
+                    config
                 });
             Ok(document_transport.unwrap_or_else(|| {
                 tracing::warn!(
@@ -559,6 +572,7 @@ mod tests {
     };
     use serde_json::json;
     use std::collections::BTreeMap;
+    use std::net::SocketAddr;
 
     struct FixedCapsuleSource(Vec<u8>);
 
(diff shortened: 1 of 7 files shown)


ewindisch:
## Fix for item 4 (rotated-authority + QUIC SNI drop) — pushed

Both findings fixed on `feat/1469-anchor-capsule-minter`:

**Finding 1 (rotated-authority capsules unbootable)**: Added `ensure!(authority.public_bytes() == public_ca)` in `mint_anchor_capsule` after the purpose guard. This mirrors `mint_registry_jwt`'s root-path check. A `RotatedAuthority` whose key differs from the pinned CA is now rejected — only the deployment root that the resolver pins may anchor a deployment.

**Finding 2a (WebPKI without pins drops entry)**: The `pins.is_empty() && web_pki` branch previously returned `None`, discarding the operator's `--quic-sni`. Now constructs `QuicServerAuth::web_pki()` and always emits a `QuicTransport` document entry with the SNI hostname.

**Finding 2b (hostname SNI+pins never match)**: Fixed `reach_from_capsule` in `did_anchored.rs` to match on port equality when the document entry carries an unspecified IP (the `0.0.0.0:port` placeholder from a hostname URI). Added a warning log on the fallback path.

3 new tests added: rotated-authority rejection, WebPKI-without-pins entry emission, hostname-SNI URI verification. All passing.

ewindisch:
## r3 fix — undialable 0.0.0.0 + non-causal CA-binding test

`545d46e0e` on `feat/1469-anchor-capsule-minter`:

**D1 fix (undialable config)**: When the document entry's IP is unspecified (from a hostname URI), the capsule's real socket address is substituted into the returned `TransportConfig`. Chose **rewrite** over reject: the hostname URI legitimately contributes SNI/pins, and the capsule is authoritative for WHERE to dial.

**Non-causal CA-binding test fixed**: Extracted `ensure_anchor_authority()` helper from `mint_anchor_capsule`. The test now routes through it instead of asserting key inequality. Mutation-verified: removing the `ensure!` makes the test RED.

**D3 port-match tests**: Added 2 causal tests: `quic_port_match_with_hostname_uri_rewrites_dial_address` and `quic_neither_port_nor_address_match_falls_back_to_bare_ip`. Both mutation-verified.

ewindisch:
**Independent review (frontline seat) — PASS-WITH-NITS, bound to head 545d46e0e.**

Highest-risk claim verified: `sign_record_detached`/`sign_capsule_detached` are byte-equivalent to the existing composite path (inner_tbs ≡ build_inner_eddsa with binding=None; outer_tbs ≡ build_hybrid_external_aad + unsigned_outer; kid derivations match); ML-DSA leg stays mandatory — no classical-only production path introduced. Guard ordering sound (validate log+checkpoint, reject RegistryDelegatedSigner, bind signer to deployment-ca.hybrid, self-verify via production resolver before any write). Outputs are public artifacts, 0644, atomic staged write. QUIC hostname-entry relaxation is a strict improvement, not a downgrade. Nothing new fails closed for local/self-host.

Follow-ups (non-blocking): did_anchored.rs:355-378 first-match on port with unspecified IP can bind arbitrary SNI/pins when two hostname QuicTransport entries share a port — prefer exact-address match and reject ambiguous multi-match; trust.rs:848-851 `strip_prefix("did:key:").unwrap_or_default()` is a silent-empty-key shape in PKI code (unreachable today) — use `?`; empty `keyAgreement`/`service` arrays should be omitted per DID Core; `#mesh-kem-*` fragment ids diverge from docs; no direct test drives `mint_anchor_capsule` (modes/guards/--force); no helper to compute a self-signed leaf pin for the QUIC reach (local-dev friction).

PR #1477, opened 2026-08-03
feat(auth): hybrid Ed25519+ML-DSA-65 bootstrap-pubkeys entries

ewindisch:
Stacked on `docs/1470-bootstrap-pubkeys-wire-contract` (the part-1 docs branch), because it extends the doc that branch introduced.

## What changes

A `bootstrap-pubkeys` value may now decode to either:

- **32 bytes** — raw Ed25519 verifying key (the form every deployed node and the staging Terraform stack already emits), or
- **1984 bytes** — 32-byte Ed25519 followed by 1952-byte ML-DSA-65, the same concatenation the deployment CA root uses for its 1984-byte public pair.

The file stays a flat `{ "<service>": "<base64>" }` map with the URL-safe-no-pad alphabet, so the two forms mix freely in one file and a service can be upgraded one entry at a time.

### Why length discrimination rather than a versioned envelope

- It reuses the deployment CA's existing hybrid byte layout instead of inventing a second hybrid encoding.
- 32 and 1984 are unambiguous, so a strict consumer never has to guess, and the existing flat-map shape is preserved exactly — no envelope means no "old file has no version field" migration and no risk of stranding deployed nodes.
- It is not a dead end: values are currently always JSON strings, so a tagged object value (`{"ed25519": ..., "mlDsa65": ...}`) can be added later as a distinguishable JSON type without breaking either existing form.

### Verification policy

Per identity, matching the project's established hybrid posture — no global "PQ required" switch:

- Classical entry → verifies classically, and is never asked for a post-quantum signature.
- Hybrid entry → **both** the Ed25519 and the ML-DSA-65 signature must be present and verify; the classical half alone is rejected.
- A PQ signature offered against a classical entry is an error, not a silently ignored input (there is no bound key to check it with, and accepting it would let a caller mistake a downgrade for a hybrid verification).

### Compatibility

`load_bootstrap_pubkeys` keeps its `HashMap<String, VerifyingKey>` signature by projecting the Ed25519 anchor of each entry, so every existing consumer (`resolve_service_vk`, `install_process_production_resolver`, the worker-startup `policy` path) is untouched. Callers that need the bound PQ key use the new `load_bootstrap_pubkeys_hybrid`. `write_bootstrap_pubkeys` emits the classical form exactly as before.

## Tests

`cargo test -p hyprstream --lib auth::identity_store` — 39 passed, including:

- `bootstrap_pubkeys_legacy_file_loads_unchanged` — the backward-compat guarantee, against a file written verbatim in the deployed shape
- `bootstrap_pubkeys_classical_round_trip` / `bootstrap_pubkeys_hybrid_round_trip`
- `bootstrap_pubkeys_mixed_file_verifies_per_identity` — one legacy + one hybrid service, each verified under its own policy
- `bootstrap_pubkeys_reject_wrong_length_values`, `..._reject_standard_alphabet_base64`, `..._reject_malformed_pq_length` — the documented actionable errors
- `documented_hybrid_example_loads_as_a_hybrid_entry` — the worked example in the doc is the literal string the parser is fed, so it cannot drift

`cargo clippy --release -p hyprstream --lib -- -D warnings` is clean; the pre-commit hook's `--tests` clippy also passes.

## Not in scope

The process trust store is keyed by `VerifyingKey` and has no PQ field, so hybrid entries are parsed and exposed but not yet bound into `Attestation`. Wiring PQ material through the trust store is a follow-on.

Part of hyprstream#1470

### DIFF
--- a/crates/hyprstream/src/auth/identity_store.rs
+++ b/crates/hyprstream/src/auth/identity_store.rs
@@ -414,57 +414,234 @@ pub fn write_service_jwt(
 ///
 /// Wire format (see `docs/bootstrap-pubkeys-format.md`): a flat JSON object
 /// `{ "<service>": "<base64>" }`, base64 using the URL-safe-no-pad alphabet
-/// (RFC 4648 §5, no `+`/`/`/`=`), decoding to exactly 32 raw Ed25519 bytes.
+/// (RFC 4648 §5, no `+`/`/`/`=`). Each value decodes either to 32 raw Ed25519
+/// bytes (classical entry) or to 1984 bytes — 32 Ed25519 followed by 1952
+/// ML-DSA-65 — for a hybrid entry, the same concatenation the deployment CA
+/// root uses.
+///
+/// This projection returns only the Ed25519 anchor of each entry, so callers
+/// that predate hybrid entries keep working unchanged. Callers that need the
+/// bound post-quantum key use [`load_bootstrap_pubkeys_hybrid`].
 pub fn load_bootstrap_pubkeys(
     credentials_dir: &std::path::Path,
 ) -> Result<std::collections::HashMap<String, VerifyingKey>> {
-    const NAME: &str = "bootstrap-pubkeys";
-    match read_secret(credentials_dir, NAME) {
-        Ok(Some(bytes)) => {
-            let json: std::collections::HashMap<String, String> = serde_json::from_slice(&bytes)
-                .context(
-                    "bootstrap-pubkeys is not valid JSON: expected a flat object \
-                     `{ \"<service>\": \"<base64>\" }` (see docs/bootstrap-pubkeys-format.md)",
-                )?;
-            let mut map = std::collections::HashMap::new();
-            for (name, b64) in json {
-                let pubkey_bytes: Vec<u8> = URL_SAFE_NO_PAD.decode(&b64).with_context(|| {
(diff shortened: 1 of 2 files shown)


ewindisch:
**Independent review (frontline seat) — PASS-WITH-NITS, bound to head 713600fb5.**

Verified: length-only discrimination (32 vs 1984, disjoint, hard error otherwise); `BootstrapPubkey::verify` has no downgrade arm (hybrid+no-PQ-sig and classical+PQ-sig both error); ml_dsa_vk_from_bytes enforces 1952; legacy classical files load unchanged (`bootstrap_pubkeys_legacy_file_loads_unchanged`), writer still emits 32-byte form, no blanket PQ requirement — local dev/self-host unaffected. Doc example byte-compared to test constant: identical.

Follow-ups (non-blocking): service repair/re-bootstrap rebuilds from bare VerifyingKeys via the classical writer, silently rewriting hybrid entries back to 32 bytes — harmless until a consumer enforces, then it's a silent downgrade; needs a merge-preserving writer before #1470 closes (the enforcement layer in #1480 seeds from this file at startup). `reject_malformed_pq_length` test actually trips the length discriminator, not the ml_dsa arm (unreachable as split). Duplicate JSON keys are last-wins and untested. Single bad entry rejects the whole file and two call sites swallow the Err.

Issue #1469, opened 2026-08-03
trust: anchor-capsule minter + deployment did.json generator (DidAnchored is unmintable in production)


ewindisch:
## Problem

`DeploymentTrustSource::DidAnchored` requires the capsule pinned by `cluster_at9p_did` to carry an `#ns` / `ServiceType::NinePExport` deployment-reach entry with an iroh nodeId or QUIC endpoint (`hyprstream-discovery/src/did_anchored.rs:308-322,340-360`), with the deployment CA as the capsule's own signing key (`subject_keys[0]` — `did_anchored.rs:293-303` feeds it into `HybridDeploymentCa`).

**No production path can mint such a capsule.** The only non-test minter is `render_at9p_identity()` (`services/oauth/did_document.rs:102-137`), which emits `#pds`/`AtprotoPds` only and signs with the node's OAuth key — it can never satisfy the anchor requirements even with `#ns` added, because the anchor capsule must be signed by the deployment CA authority itself. Every `#ns`-carrying construction in the tree is a test (the working reference is `crates/hyprstream/tests/fixtures/did_trust.rs:129-171`). Publication is already supported (`oauth.deployment_well_known_dir` serves operator-supplied capsules); minting is the whole gap. This blocks DidAnchored mode entirely, including every `cluster_remote_node = true` topology.

## Proposal: `hyprstream trust mint-anchor-capsule`

A single command that takes the CA authority + reach and emits `{cid512, capsule.cbor, did.json}` together (the DID is the BLAKE3-512 CID of the capsule containing the reach and CA, so the three outputs are inseparable):

- Loads the deployment CA authority via the existing age/YubiKey plumbing (`LoadedEdSigner` abstraction, `cli/trust.rs:196-205`) so the PIV-slot Ed25519 leg works.
- Inputs: anchor node's iroh nodeId (or QUIC socket + SNI/cert-hash pair), the deployment `did:web`, optional extra services.
- Builds `CapsuleBody::new([hybrid CA key], [#ns NinePExport entry])` + `also_known_as = [cluster_did_web]`, signs with `sign_capsule` (hybrid, as pinned), writes `capsule.cbor` + prints `did:at9p:<cid512>`.
[...]

PR #1479, opened 2026-08-03
feat(wizard): opt-in deployment-trust ceremony with YubiKey detection

ewindisch:
Stacked on `feat/1469-anchor-capsule-minter`. Part of hyprstream#1467.

## What this does

`hyprstream wizard --deployment-trust` now runs the deployment-trust ceremony after node bootstrap, instead of leaving it as a manual procedure in `docs/deployment-trust-ceremony.md`. It is **opt-in and additive**: without the flag the wizard's behaviour is byte-for-byte what it was, and node-local trust stays separate from deployment trust.

New module `crates/hyprstream/src/cli/trust_ceremony.rs`. Hardware access sits behind two traits (`TokenDetector`, `AgeYubikeyPlugin`); everything that decides anything is a pure function taking values.

## Detection / selection decision table

| Detection result | `--deployment-trust-software` | Interactive? | Outcome |
|---|---|---|---|
| flag absent | — | — | `Skipped` — no prompt, no work |
| no token | no | either | `SoftwareDevGrade`, loudly labelled |
| firmware < 5.7.4 | no | yes | `HardwareAgeRecipient` (`--yubikey`) |
| firmware >= 5.7.4 | no | yes | `HardwarePiv` (`--piv-slot`, default `9c`) |
| several tokens | no | yes | `ChooseToken` prompt → same firmware rule |
| several tokens | no | no | `Blocked` → "pass `--deployment-trust-serial <SERIAL>`" |
| serial given, no match | no | either | `Blocked`, lists attached serials |
| any token attached | no | no | `Blocked` — a PIN, a touch and a passphrase need a human |
| detection error (tool missing / timeout / unparsable) | no | either | `Blocked` — **never** silently degrades to software |
| detection error | yes | either | `SoftwareDevGrade`, rationale names the override |
| token attached | yes | either | `SoftwareDevGrade`, rationale says it overrode present hardware |

Two invariants are tested directly: a software root is never chosen while hardware is attached without an explicit override, and a scripted run never returns a plan that would prompt.

## Dev-grade labelling

A software root is labelled in the mode rationale, in a `!!` block on screen, in the wizard summary line, and in `deployment-trust-mode.json` written beside the artifacts (`dev_grade`, `grade`, `ed25519_confined_to_hardware`, `break_glass_protected`).

## Break-glass

The trust CLI can only check the two recipients differ. The wizard offers second token / passphrase-encrypted (`age-keygen | age -p`, private half never written in clear) / unencrypted, refuses the last without `--deployment-trust-allow-plaintext-break-glass`, and classifies the resulting file without decrypting it (armored ciphertext reports "unconfirmed" rather than claiming a passphrase it cannot see).

## Not in this PR

**Installing the public artifacts to `/etc/hyprstream/trust/`** (issue item 5) is deliberately left out. Today `cli/trust.rs` only *prints* those paths, and choosing who writes them interacts with the unresolved artifact-delivery question (privileged wizard write vs. a separate root-run install step vs. the publisher-manifest/cloud-secret path the contract already describes). The wizard prints the ceremony outputs and their destinations, which matches current CLI behaviour exactly. Options + a recommendation are in the task report rather than guessed at here.

## Untested on hardware — required manual verification

I have no YubiKey. These paths compile and are reached by the tested decision logic, but **nothing about them is verified**:

- `SystemTokenDetector::detect` against a real `ykman list` (the parser is tested against captured-format strings, not real output).
- `SystemAgeYubikeyPlugin::generate_identity` / `export_identity_file` — `age-plugin-yubikey --generate --pin-policy always --touch-policy always` on a TTY, and that the recipient parsed back from `--list` is the one just generated.
- The whole `--piv-slot` path end to end on firmware >= 5.7.4: `ykman piv keys import`, the import self-check, and `yubico-piv-tool` signing. This is the issue's acceptance item and it stays open.
- That the delegation step actually demands a PIN **and** a physical touch. If it ever completes without one, the root is not gated.
- `generate_passphrase_encrypted_identity` (needs the `age` tools present).

An operator with a token should run: a wizard ceremony on 5.4.3 hardware (age-recipient mode), a wizard ceremony on >= 5.7.4 hardware (PIV mode), and confirm PIN+touch is demanded at the delegation step in both.

## Verification run here

- `cargo test -p hyprstream --lib trust_ceremony` — 43 passed, 0 failed
- `cargo test -p hyprstream --lib cli::` — 82 passed, 0 failed
- `cargo clippy --release -p hyprstream --lib -- -D warnings` — clean
- `cargo clippy --release -p hyprstream --bins -- -D warnings` — clean

### DIFF
--- a/crates/hyprstream/src/bin/main.rs
+++ b/crates/hyprstream/src/bin/main.rs
@@ -243,6 +243,47 @@ fn build_cli() -> ClapCommand {
                     .value_name("ROLE")
                     .default_value("admin")
                     .help("Role to assign the local user under --non-interactive (admin|operator|trainer|viewer). Default is admin; use operator/viewer in tests for least-privilege."),
+            )
+            .arg(
+                Arg::new("deployment_trust")
+                    .long("deployment-trust")
+                    .action(clap::ArgAction::SetTrue)
+                    .help("Run the deployment-trust ceremony after node bootstrap — detects an attached YubiKey and mints the deployment root, delegated signer, and registry credential. Opt-in: node bootstrap alone is unchanged without it."),
+            )
+            .arg(
+                Arg::new("deployment_trust_dir")
+                    .long("deployment-trust-dir")
+                    .value_name("DIR")
+                    .requires("deployment_trust")
+                    .help("Absolute ceremony working directory (default: <models-dir>/deployment-ceremony). Destroy it once the artifacts are distributed."),
+            )
+            .arg(
+                Arg::new("deployment_trust_software")
+                    .long("deployment-trust-software")
+                    .action(clap::ArgAction::SetTrue)
+                    .requires("deployment_trust")
+                    .help("Mint a DEV-GRADE software deployment root even if a hardware token is attached. Never selected on your behalf."),
+            )
+            .arg(
+                Arg::new("deployment_trust_serial")
+                    .long("deployment-trust-serial")
+                    .value_name("SERIAL")
+                    .requires("deployment_trust")
+                    .help("Bind the deployment root to the token with this serial when several are attached."),
+            )
+            .arg(
+                Arg::new("deployment_trust_piv_slot")
+                    .long("deployment-trust-piv-slot")
+                    .value_name("SLOT")
+                    .requires("deployment_trust")
+                    .help("PIV slot holding the Ed25519 leg on firmware 5.7.4+ (default: 9c)."),
+            )
+            .arg(
+                Arg::new("deployment_trust_allow_plaintext_break_glass")
+                    .long("deployment-trust-allow-plaintext-break-glass")
+                    .action(clap::ArgAction::SetTrue)
+                    .requires("deployment_trust")
+                    .help("Accept an unencrypted break-glass identity. Only for a throwaway root you destroy immediately afterwards."),
             ),
     );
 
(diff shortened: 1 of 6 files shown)


ewindisch:
**Option (b) is already built — it's `hyprstream trust install` on #1474.**

This PR's open question on artifact install recommends "a separate `trust install-artifacts` run as root, which the wizard prints as the next command." That command exists on #1474 as `hyprstream trust install`: it hard-requires uid 0, validates the CA via `verify_deployment_public_ca` and the log/checkpoint via `validate_authority_log` **before** writing anything, installs the three public artifacts 0644 and the delegated signer plus refresh identity 0600 through the atomic staged-commit path, and installs the refresher timer. That is exactly the posture `read_trusted_artifact` requires, enforced in one place.

This PR could not see it: it is stacked on #1469 (off `main`), and #1474 is a sibling branch off `main`, so the installer is not in this branch's history.

**Merge-ordering consequence:** #1474 should land before the wizard's install guidance is finalized. Once it has, the follow-up here is small — have the wizard print a concrete `hyprstream trust install --public-ca … --authority-log … --authority-checkpoint … [--delegated-key … --delegation … --registry-public-key … --refresh-identity …]` invocation with the ceremony's actual output paths substituted, instead of printing the destination paths and leaving the operator to assemble the command.

No change requested in this PR — recording the cross-reference so the option-(b) decision is not re-litigated and no second installer gets written.

ewindisch:
## Fix for item 5 (false HW claim + PIV overwrite) — pushed

`1ce4fe0dd` on `feat/1467-wizard-piv-ceremony`:

**Major-a (false hardware-confinement claim)**: `ed25519_confined_to_hardware()` now returns `false` (the honest answer — the seed is generated in host memory and stored in the age-encrypted bundle). Added `token_gated_signing` (what IS true for PIV mode) and `seed_recoverable_from_bundle` fields to the audit record. Updated the grade label from 'hardware-gated deployment root' to 'token-gated deployment root (seed recoverable from bundle)'. Corrected the rationale text that falsely claimed the key 'never enters host memory'.

**Major-b (unwarned PIV overwrite)**: Added `piv_slot_occupied()` check that reads the slot's certificate object before import. If the slot is occupied and `--force` was not given, the ceremony refuses with a clear error instead of silently destroying an existing key.

All 43 ceremony tests pass.

ewindisch:
## r3 fix — PIV slot guard probes key presence, fails closed

`17cc2d36d` on `feat/1467-wizard-piv-ceremony`:

**E1 fix**: `piv_slot_occupied` now runs TWO probes: the existing `read-certificate` (cert object) AND `ykman piv keys list` (key object). This catches the key-but-no-certificate state that `piv_import_ed25519` itself leaves behind — the primary re-run overwrite scenario.

**Fail-closed**: Any probe error (PCSC failure, token removed, binary missing) returns `probe_indeterminate: true`, which the decision function treats as occupied. The guard no longer silently permits a destructive overwrite on a transient error.

**Pure decision function + tests**: Extracted `slot_occupied_from_probe(SlotProbeInputs) -> bool` with 5 unit tests. Mutation-verified: reverting to certificate-only makes `piv_slot_with_key_but_no_certificate_is_occupied` RED.

**E3 doc fix**: `docs/deployment-trust-ceremony.md` no longer asserts 'never reaches host memory' — corrected to match the `ed25519_confined_to_hardware() = false` retraction.

ewindisch:
**Independent review (frontline seat) — PASS-WITH-NITS, bound to head 4d823314e.**

Opt-in verified real: `CeremonyRequest::enabled` defaults false everywhere, declining leaves node-local bootstrap byte-identical (ceremony only adds artifacts in a separate dir) — classical floor intact, local dev unaffected. Fail-closed detection (DetectionError blocks rather than degrading to software; unparsable ykman line is an error, not a dropped token), both invariants pinned by tests. No shell-injection surface (Command+arg, whitelisted piv_slot, timeouts). No PIN transits the process; piped keygen never lands plaintext on disk; 0700/0600 modes. Honest labelling incl. the corrected hardware-confinement claim; PIV overwrite guard (indeterminate ⇒ occupied) is a real win.

Ordering vs #1474: no compile dependency; merging first is functionally safe but the minted registry JWT is 1h-TTL with no refresher installed — keeping #1474-first order.

Follow-ups (non-blocking): `--tui --deployment-trust` silently no-ops the ceremony (exit 0, no warning) — clap conflicts_with or error; plaintext break-glass written to a `.age`-suffixed path in the plaintext branches — rename; refused-plaintext path leaves the generated key on disk with no cleanup instruction; doc omits that `AllowedForDevGradeRoot` also permits bare identities; `--deployment-trust --non-interactive` in a no-USB container yields dev-grade root + plaintext break-glass without explicit opt-in (loudly labelled, but worth a flag); wizard hardcodes force:false with no --force passthrough (re-runs dead-end).

PR #1480, opened 2026-08-03
trust: anchor hybrid bootstrap-pubkeys into PqTrustStore and enforce PQ per identity

### LINKED ISSUE #1478: trust: anchor hybrid bootstrap-pubkeys into PqTrustStore and make PQ enforcement per-identity
PR #1477 extends the `bootstrap-pubkeys` wire format to carry hybrid Ed25519 + ML-DSA-65 service keys. Parsing, writing, round-trip and mixed-file behavior are covered, but supplying hybrid material does not yet change what the running system enforces.

**Correction to the original framing of this issue:** it previously claimed the blocker was `Attestation`/`TrustStore` having no PQ field, implying every attestation construction site needed updating. That is wrong. PQ binding is already factored out into the `PqTrustStore` trait (`crates/hyprstream-rpc/src/envelope.rs:979`), keyed by the same Ed25519 pubkey that keys the authorization store, with a concrete `KeyedPqTrustStore` at `envelope.rs:1032`. `Attestation` needs no change.

The enforcement path already exists end to end: `SignedEnvelope.cose` / `ResponseEnvelope.cose` carry COSE composite signatures on the wire, and `SignedEnvelope::verify_cose` (`envelope.rs:1735`) resolves the peer's ML-DSA key via `PqTrustStore::ml_dsa_key_for()` and fails closed when PQ is required but unanchored. No protocol, schema, codegen, or WASM-client change is needed.

Two gaps remain:

**1. Population.** Nothing seeds a `PqTrustStore` from hybrid bootstrap-pubkeys entries. `load_bootstrap_pubkeys_hybrid()` (from #1477) already returns entries carrying the ML-DSA key; the binding should be inserted where the classical attestation is seeded in `bin/main.rs` (~1513, ~1659), and that store routed into the envelope verify path.

**2. Per-identity policy.** `verify_cose` derives `require_pq` from a **global** `CryptoPolicy::uses_pq()`. Flipping that to Hybrid would fail-closed against every classical peer at once — a deployment-wide flag day, and contrary to the established posture (PQ enforced only for identities that carry a PQ key, classical floor otherwise, never blanket fail-closed). Requirement should be derived per identity: anchored PQ key present → require PQ for that peer regardless of global policy; absent → fall back to global policy.

Two correctness requirements for the implementation:

- **Monotonic anchoring.** Once an identity is anchored hybrid, a later registration lacking a PQ key must not clear the binding, or this becomes a silent downgrade path. The classical store guards its analogue via expiry comparison; this needs the equivalent rule.
- **Confirm `verify_composite` with `require_pq = false` still verifies a PQ leg that is present** rather than ignoring it. If it ignores it, an unanchored peer's PQ signature can be stripped undetected.

Follow-up to #1470.

ewindisch:
Stacked on #1477 (`feat/1470-hybrid-bootstrap-pubkeys`).

Supplying hybrid Ed25519 + ML-DSA-65 material in `bootstrap-pubkeys` now changes what the running system enforces. No schema, codegen, or WASM change.

## 1. Population

`mesh_trust::seed_bootstrap_pq_bindings()` anchors the ML-DSA-65 half of each hybrid bootstrap entry, keyed by the same Ed25519 pubkey the classical trust store uses. It is called from `install_envelope_verify_config()` in `bin/main.rs`, into the same `KeyedPqTrustStore` that already backs both the request-side and response-side process-global verify configs — no parallel global. The store is immutable after install, so seeding happens at construction rather than lazily at the CLI `resolve_service_vk()` site (which reads the same file, and now says so).

## 2. Per-identity enforcement

`SignedEnvelope::verify_cose` and `ResponseEnvelope::verify_cose` no longer derive `require_pq` from the global `CryptoPolicy::uses_pq()`:

- anchored ML-DSA-65 key present for this signer → PQ required for that peer, regardless of the global policy;
- no anchor → previous global-policy behavior, unchanged (Hybrid still fails closed; classical-permissive deployments keep working).

No flag day: a deployment with no hybrid material anywhere behaves exactly as it does today. Explicitly tested.

## Monotonic anchoring

`KeyedPqTrustStore::register()` never lets a classical-only re-registration clear an established binding, so replaying a legacy enrollment cannot switch enforcement back off. Tested at the store level and end to end through verification.

## Note on `verify_composite(..., require_pq = false)`

Investigated per the issue. With `require_pq = false` **and no anchored key**, a present outer ML-DSA layer is deliberately ignored (`cose_sign.rs` "skip-unknown" arm). With `require_pq = false` but an anchored key supplied, the outer layer **is** verified.

This is not a signature-stripping gap: the inner EdDSA layer's AAD is derived from whether an outer layer is present, so stripping the outer layer off a hybrid composite makes the inner signature fail. The residual is outer-layer malleability for unanchored peers (an on-path attacker can garble/replace the ignored outer layer without detection); it grants no forgery, since the inner EdDSA leg still authenticates the message. Filed separately rather than expanded into this PR.

## Tests

`hyprstream-rpc --lib`: 973 passed. `hyprstream --lib auth::mesh_trust`: 11 passed, `auth::identity_store`: 39 passed. Clippy clean on `-p hyprstream-rpc --lib --tests`, `-p hyprstream --lib --tests`, and `-p hyprstream --bin hyprstream`, all with `-D warnings`.

Closes #1478

### DIFF
--- a/crates/hyprstream-rpc/src/envelope.rs
+++ b/crates/hyprstream-rpc/src/envelope.rs
@@ -1018,6 +1018,34 @@ impl KeyedPqTrustStore {
         );
     }
 
+    /// Register an identity that may or may not carry a bound ML-DSA-65 key.
+    ///
+    /// Anchoring is **monotonic**: a `None` key never clears or replaces an
+    /// anchor already established for that identity. Re-registering a
+    /// previously hybrid identity from a classical-only source (a legacy
+    /// bootstrap file, a re-enrollment that omitted the PQ half) therefore
+    /// cannot silently downgrade the peer back to classical.
+    ///
+    /// Returns whether the identity is anchored after the call.
+    pub fn register(
+        &mut self,
+        ed25519_pubkey: [u8; 32],
+        ml_dsa_vk: Option<&crate::crypto::pq::MlDsaVerifyingKey>,
+    ) -> bool {
+        match ml_dsa_vk {
+            Some(vk) => {
+                self.bind(ed25519_pubkey, vk);
+                true
+            }
+            None => self.bindings.contains_key(&ed25519_pubkey),
+        }
+    }
+
+    /// Whether this identity has an anchored ML-DSA-65 key.
+    pub fn is_anchored(&self, ed25519_pubkey: &[u8; 32]) -> bool {
+        self.bindings.contains_key(ed25519_pubkey)
+    }
+
     /// Number of bindings.
     pub fn len(&self) -> usize {
         self.bindings.len()
@@ -1741,20 +1769,19 @@ impl SignedEnvelope {
         let signing_data = self.signed_bytes();
         let aad = envelope_external_aad();
 
-        let anchored_pq = if verify_policy.uses_pq() {
-            Some(
-                pq_store
-                    .and_then(|store| store.ml_dsa_key_for(&self.cnf))
-                    .ok_or_else(|| {
-                        EnvelopeError::PqSignatureInvalid(
-                            "mandatory Hybrid suite requires an anchored ML-DSA-65 signer key"
-                                .to_owned(),
-                        )
-                    })?,
-            )
-        } else {
-            None
-        };
+        // The PQ requirement is derived PER IDENTITY, not deployment-wide.
+        // An identity with an anchored ML-DSA-65 key MUST present a verifying
+        // outer PQ layer even where the global policy still admits classical
+        // peers — anchoring a peer is what turns PQ on for that peer. An
+        // identity with no anchor falls back to the global policy, so a
+        // deployment holding no hybrid material behaves exactly as before.
+        let anchored_pq = pq_store.and_then(|store| store.ml_dsa_key_for(&self.cnf));
+        if anchored_pq.is_none() && verify_policy.uses_pq() {
+            return Err(EnvelopeError::PqSignatureInvalid(
+                "mandatory Hybrid suite requires an anchored ML-DSA-65 signer key".to_owned(),
+            ));
+        }
+        let require_pq = anchored_pq.is_some();
 
         crate::crypto::cose_sign::verify_composite(
             &self.cose,
@@ -1762,7 +1789,7 @@ impl SignedEnvelope {
             anchored_pq.as_ref(),
             &signing_data,
             &aad,
-            verify_policy.uses_pq(),
+            require_pq,
         )
         .map_err(|e| EnvelopeError::PqSignatureInvalid(e.to_string()))?;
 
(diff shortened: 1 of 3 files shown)


ewindisch:
## Fix for item 1 (silent PQ rebind) — pushed

`9d5e2968e` on `feat/1478-pq-trust-anchoring`:

**Major fix**: `KeyedPqTrustStore::bind` was an unconditional `HashMap::insert` — a later seed with a different ML-DSA-65 key silently replaced the trusted anchor. Now `bind` refuses to overwrite a differing anchor (same key re-registered is idempotent). The existing anchor is retained and the conflict is logged at error level. This makes the monotonic anchoring rule both un-clearable AND un-rebindable.

**Minor fix**: The `if let Ok(secrets_dir)` seeding guard silently swallowed unresolvable-secret-dir errors. Added a warning log in the else branch.

**Test added**: `differing_pq_rebind_is_refused_not_silent` — verifies that a second `bind` with a different PQ key is refused and the original anchor survives.

ewindisch:
## Wasm build fix — pushed

`0752b72cc` on `feat/1478-pq-trust-anchoring`:

The PQ-rebind `tracing::error!` in `KeyedPqTrustStore::bind` broke the wasm32 browser build: `tracing` is a native-only dependency in `hyprstream-rpc` (every other `tracing::` call in `envelope.rs` lives inside a `#[cfg(not(target_arch = "wasm32"))]` function). The conflict-detection logic (check existing key, return without overwriting) runs on all targets — only the diagnostic log is gated, preserving the native warning without breaking the browser client.

**Verified**:
- `cargo check -p hyprstream-rpc -p hyprstream-vfs -p hyprstream-rpc-std --locked --target wasm32-unknown-unknown` (with CI RUSTFLAGS): **PASS** (7.71s)
- `cargo clippy --workspace --all-targets -- -D warnings`: **PASS** (6m 15s)

**Audit of other 7 trust-stack PRs for the same defect class**: all CLEAN. Only #1480 added a tracing call on a wasm-compiled path (`hyprstream-rpc/src/envelope.rs`). All other REVISE-pass changes are in the `hyprstream` or `hyprstream-discovery` crates, neither of which is in the browser wasm graph (`hyprstream-rpc`, `hyprstream-vfs`, `hyprstream-rpc-std`).

ewindisch:
**Independent review (frontline seat) — PASS-WITH-NITS, bound to head 0752b72cc.**

Enforcement posture verified correct on all four quadrants: anchored+classical rejected (no downgrade — envelope.rs:1799-1805 request, :2554-2560 response, `require_pq = anchored_pq.is_some()` under any policy); unanchored unchanged (no blanket fail-closed — classical interop clients still admitted); anchoring keyed on `cnf` with ct_eq against the verifying key at every non-test entry, so no cross-identity confusion (`mixed_anchored_and_unanchored_services_in_one_store` proves it). Trust store is sealed before `install_verify_config` (OnceLock) — no load/verify TOCTOU. Error paths fail closed (unreadable bootstrap file → zero anchors under pinned Hybrid policy = rejection).

Fix in fast-follow:
1. `register()` (envelope.rs:1051-1065) returns `true` even when `bind` refused a conflicting rebind, so mesh_trust.rs:107-109 logs "anchored ML-DSA-65 key" for a key that is NOT the surviving anchor — and that service then fails every verification while the log claims success. Return `self.is_anchored_to(pubkey, vk)`.
2. Before admission.rs:440 (GATE-verified capsule path) is ever wired to a live store: silent-refuse-rebind means a peer rotating only its ML-DSA half under the same Ed25519 key gets pinned to the stale anchor and hard-fails for the process lifetime, signalled only by tracing::error. Give `bind` a Result / explicit `rebind` first — same shape as the '*'-wildcard incident.

Nits: one bad entry aborts the whole bootstrap file (zero anchors — fail-closed but fragile); missing unanchored×hybrid quadrant test (pin the deferred outer-layer-ignored arm so it can't change silently).

PR #1482, opened 2026-08-04
feat(auth): make hybrid service identities mandatory on the install path

ewindisch:
Stacked on `feat/1478-pq-trust-anchoring`. Consumes both the hybrid `bootstrap-pubkeys` writer and the PQ anchoring that branch adds.

## Directive

> we should always be hybrid. there should be no non-hybrid install path. We will still support classical-only clients and federated atproto peers at interop surfaces accordingly.

## The gap this closes

The hybrid `bootstrap-pubkeys` format and the PQ anchoring that consumes it both already existed — but **nothing ever produced hybrid material**. The provisioning wizard called the classical writer (`bootstrap_manager.rs:761`), so every installed node published Ed25519-only service entries, anchored nothing into the `KeyedPqTrustStore`, and got no post-quantum enforcement for its own services. Hybrid was implemented end to end and switched off by default at the one place that mattered.

## What changed

**1. Provisioning always writes hybrid.** Every service entry now carries a bound ML-DSA-65 key. There is no flag to disable it and no classical provisioning mode.

**2. Derive, not generate.** The ML-DSA-65 half is derived from the service's Ed25519 key via `node_identity::derive_mesh_mldsa_key` (HKDF, `hyprstream-mesh-mldsa-v1`). This is *required for correctness*, not just convenient: `LocalSigner::new` (`signer.rs:54`), the service dispatch default (`svc.rs:744`), and the published `#mesh-pq` verification method all produce their PQ signature from exactly that derivation. An independently generated key would anchor a public key nothing ever signs with, and every hybrid verification would fail. It also means **no new secret is introduced** — a service's private material stays a single Ed25519 seed, with nothing extra to persist, protect, back up or rotate.

**3. Classical service entries are a hard, actionable error** where the runtime resolves service keys. Previously such an entry was trusted classically and then denied at envelope verification with an opaque `mandatory Hybrid suite requires an anchored ML-DSA-65 signer key` — naming neither the file nor the fix. The check is deliberately **not** in `load_bootstrap_pubkeys`: the low-level loader must stay able to read legacy files so tooling and this error can report precisely which services are stale.

**4. Interop untouched.** No change to the ATProto perimeter, browser/WNS paths, or DID-document identity resolution. The per-identity fallback for unanchored signers in `verify_cose` is byte-for-byte unchanged and now carries a comment recording that it is deliberate interop policy.

## Open question, answered: does interop share `verify_cose`?

**Split, and not the convenient split.**

- **Browser / WASM external clients: YES, same path.** `dispatch::process_request` is a single funnel for browser WebTransport, QUIC, iroh, UDS and inproc alike; `AnySigner` vs `FixedSigner` changes who is accepted, not the policy or the PQ requirement.
- **Federated ATProto peers: NO, separate surface** — axum HTTP routes with DID-document / JWT / DPoP checks (`services/oauth/xrpc.rs`, `hyprstream-pds-service/src/federation_intake.rs`). They never reach `verify_cose` today. Note that `build_federation_admission_gate` has zero production callers but is *designed* to consume a `verify_cose`-verified signer, so wiring it would move federated peers onto the shared path.

**The belt-and-braces flip is moot: `CryptoPolicy` is already `Hybrid`-mandatory.** `CryptoPolicy::Classical` is `#[cfg(any(test, feature = "test-classical-policy"))]` and does not exist in a production build; `Hybrid` is the `#[default]`, the value `mandatory_envelope_policy()` returns, and the fail-closed default when no config is installed. There is no flip left to make, and none is made here.

A consequence worth flagging separately from this PR: under the pinned Hybrid policy, the real admission control is *which identities are anchored*, and the `KeyedPqTrustStore` is populated only from `mesh_peers` config plus `bootstrap-pubkeys` at startup and is immutable thereafter. A browser client using classical `JsSigner::new` rather than `new_hybrid` has no runtime path to become anchored. Whether any deployment does that is worth confirming; it is pre-existing behavior on this base branch, not introduced here.

## Migration — breaking

Every existing deployment has a classical `bootstrap-pubkeys` and **will fail to start** until re-provisioned.

**Operator action:** run `hyprstream wizard`, or `hyprstream service repair`, before starting the upgraded binary. Per-service Ed25519 keys and service JWTs are loaded rather than replaced, so **service identities do not change** — only the published entries gain their post-quantum half. No coordination between nodes is needed and no other file changes.

The error names the offending services and the fix.

**No transition escape hatch is included**, per the directive's explicit rejection of a non-hybrid path. If one is later wanted, the narrowest sound form would be a time-boxed env var that downgrades the strict check to a `tracing::warn!` — but note it would buy little, because those services would still be unanchored and their RPC still refused by the already-Hybrid envelope policy. Re-provisioning is the actual fix, and it is cheap and non-destructive.

## Tests

- `provisioning_writes_hybrid_entries_that_anchor_for_every_service` — a fresh provisioning run produces hybrid entries for every service, they anchor into the PQ store, and the anchored ML-DSA key is the one the service's signer signs with.
- `classical_service_entry_is_a_hard_actionable_error` — the error names the classical service, does not implicate hybrid ones, and names the fix.
- `provisioned_service_entry_verifies_a_hybrid_signature_end_to_end` — round trip through the file, then verify a real hybrid signature; the Ed25519 signature alone is rejected.
- `unanchored_interop_client_still_verifies_classically` — **the interop guarantee**, in deployment shape: two anchored hybrid services enforced in the same process and the same store, while an unanchored external client still verifies on Ed25519 alone.
- `hybrid_service_entries_and_empty_file_both_pass` — an unprovisioned node is not an error.

Refs #1478

### DIFF
--- a/crates/hyprstream-rpc/src/envelope.rs
+++ b/crates/hyprstream-rpc/src/envelope.rs
@@ -1775,6 +1775,17 @@ impl SignedEnvelope {
         // peers — anchoring a peer is what turns PQ on for that peer. An
         // identity with no anchor falls back to the global policy, so a
         // deployment holding no hybrid material behaves exactly as before.
+        //
+        // The unanchored branch is DELIBERATE INTEROP POLICY, not laxity, and
+        // must not be tightened here. It is the only place a signer this node
+        // does not itself provision — an external client, or a federated peer
+        // whose key comes from its own published document — can be admitted on
+        // terms other than our own. This node's own service identities are
+        // always provisioned hybrid and therefore always anchored, so making
+        // them mandatorily hybrid is achieved entirely by what gets anchored,
+        // and needs no change on this line. Deciding what the unanchored
+        // branch may admit is a separate, interop-facing question; do not
+        // settle it as a side effect of an internal-identity change.
         let anchored_pq = pq_store.and_then(|store| store.ml_dsa_key_for(&self.cnf));
         if anchored_pq.is_none() && verify_policy.uses_pq() {
             return Err(EnvelopeError::PqSignatureInvalid(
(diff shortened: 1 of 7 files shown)


ewindisch:
## Fix for item 2 (broken migration path) — pushed

`e4e0aca3f` on `feat/hybrid-mandatory-install`:

1. **Error message corrected**: The hybrid gate error now names `hyprstream wizard` (the working recovery — it re-provisions in place, preserving keys) instead of `service repair` (which was a dead end — it dispatched after the gate and couldn't read/write bootstrap-pubkeys).

2. **`service repair` early-dispatched**: It now runs before the hybrid gate, so it works even when bootstrap-pubkeys is stale. Previously it died at the same gate that blocked service startup.

3. **New `Hybrid bootstrap` repair check**: `run_repair_checks` gained a check that detects stale classical entries and names the wizard as the fix.

**Recovery path verified**: classical file → gate refuses → error says `hyprstream wizard` → wizard re-provisions in place → node starts. And: classical file → `service repair` → runs early (bypasses gate) → diagnostic shows stale entries → operator runs wizard.

clippy + tests pass.

ewindisch:
## r3 fix — doc/dead-code cleanup + non-causal message test

`b73b095bb` on `feat/hybrid-mandatory-install`:

1. Dropped `service repair` from `docs/bootstrap-pubkeys-format.md`.
2. Dead `ServiceAction::Repair` arm replaced with `run_repair_checks` (defense-in-depth matching the early dispatch).
3. Help text updated to match actual repair behavior.
4. `run_repair_checks` now returns `Err` on check failure (exit non-zero).
5. Message test now asserts ABSENCE of `service repair` (mutation-verified causal).

**Stack topology**: merged #1472 + #1480 + #1482 — textually clean. Clippy + tests + WASM all PASS on the combined tree.

ewindisch:
**Independent review (frontline seat) — PASS-WITH-NITS, bound to head b73b095bb.**

Local-dev mandate verified satisfied: hybrid minting is self-contained — ML-DSA-65 half derived from the existing Ed25519 seed (HKDF-SHA256 → ml_dsa_sk_from_seed, pure-Rust ml-dsa crate), no hardware, no ceremony, no new persisted secret; fresh wizard/bootstrap on a bare machine works. No silent identity regeneration (load_or_generate preserved; only published entries rewritten — anchored trust survives; test pins that the anchored key is the signing key). Runtime interop preserved: unanchored classical clients still admitted (`unanchored_interop_client_still_verifies_classically`).

Follow-ups (non-blocking): (1) upgrade of an existing classical node hard-stops services until `hyprstream wizard` is re-run, even though the hybrid half is deterministically derivable in place — consider self-healing on `service install`/startup; recovery is real and headless (`wizard -y`) and the error names it. (2) `service install` now aborts on the one condition it can't fix (run_repair_checks Err on stale pubkeys before installing units) — install should bypass or repair. (3) The Ok→Err exit-code change also fails install on pre-existing warn-only conditions (corrupt .registry, TLS gen) — broader than the PR thesis, call it out. (4) No test covers the main.rs startup gates / service repair early dispatch.

PR #1485, opened 2026-08-04
ci(oci): PR-triggered runtime image build + serving smoke test

ewindisch:
## What

Adds the missing PR-triggered CI job that actually builds and runs a runtime image. Before this, no `pull_request` workflow built a runtime image — a runtime regression first appeared as a broken publish on `main`, not a red PR check.

## What it validates (cpu-arm64 only)

1. **Build**: `podman build --target runtime --build-arg VARIANT=cpu-arm64` — the full image through the runtime stage, on a native arm64 Graviton runner (no qemu).
2. **Security property**: `podman inspect` asserts `User=65532:65532` from the **built image config**, not the Dockerfile text. A regression to root fails CI.
3. **Binary exec**: `podman run --rm <image> /hyprstream --version` — catches ELF interpreter, glibc/libstdc++ symbol-version, and ld.so library-search-path failures (the exact class of regression the core-runtime migration risks: COPY destinations moved from Debian multiarch dirs to `/usr/lib64`).
4. **Serving**: starts the OAuth service inside the container, waits up to 120 s, probes `GET /health` on `:6791` (plaintext HTTP, public MAC exemption — no credential needed). A container that starts and immediately fails to bind fails this job.
5. **Process uid**: `podman exec … id -u` verifies the running process is uid 65532, not just the image config.

## Path filter rationale

Triggers only on files that can affect the runtime artifact:
- `Dockerfile` — the image definition
- `Cargo.toml`, `Cargo.lock`, `crates/**` — compiled into the binary
- `rust-toolchain.toml` — COPYed into the builder stage
- `.github/workflows/oci-runtime-validate.yml` — this file

Docs, other workflows, and `.github/` internals cannot change what enters the image, so they do not consume a Graviton runner.

## Gaps (honest)

- **amd64**: not exercised. The runtime-cpu stage uses x86_64 multiarch paths; the CI runner is arm64 only.
- **CUDA/ROCm**: not built or run. GPU variants need GPU-host validation; the static analysis in the base-image PR covers only cpu-arm64 end to end.
- **Serving probe**: requires `wizard --non-interactive --bootstrap-only` to seed credentials before `service start oauth --foreground --ipc`. If the wizard path has issues in a fresh container (writable dirs, XDG defaults under uid 65532), this step may need adjustment after the first real CI run reveals it.

## Stacked

`--base fix/oci-hummingbird-core-runtime-nonroot` (#1464). Do NOT merge this before #1464 merges; then rebase onto main.

---
**Update (a8fb98e95):** the serving smoke test now hard-gates: the container must start and either serve `/health` or refuse cleanly at deployment-trust enforcement; the earlier note that the probe was non-blocking/always-failing is obsolete. Image/container/port are per-run; uid assert is real; scratch dirs cleaned.

### DIFF
--- a/.github/workflows/oci-runtime-validate.yml
+++ b/.github/workflows/oci-runtime-validate.yml
@@ -27,6 +27,12 @@ on:
 permissions:
   contents: read
 
+# Cancel stale runs when a new commit is pushed to the same PR — the
+# Graviton runner pool is shared and scarce.
+concurrency:
+  group: oci-runtime-validate-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build-runtime-and-smoke-test:
     # Same self-hosted Graviton pool as build-image.yml and the merge gate:
--- a/.github/workflows/oci-runtime-validate.yml
+++ b/.github/workflows/oci-runtime-validate.yml
@@ -113,34 +113,49 @@ jobs:
           mkdir -p "$DATA_DIR" "$RUN_DIR" "$SECRETS_DIR"
           chmod 777 "$DATA_DIR" "$RUN_DIR" "$SECRETS_DIR"
 
-          echo "Starting container (oauth foreground)…"
-          podman run --rm -d --name hs-smoke \
+          # Shared volume mounts for both the bootstrap and service containers.
+          VOL_ARGS=(
+            -v "$DATA_DIR":/var/lib/hyprstream:z
+            -v "$RUN_DIR":/run/hyprstream:z
+            -v "$SECRETS_DIR":/tmp/hs-secrets:z
+            -e HYPRSTREAM__SECRETS__PATH=/tmp/hs-secrets
+            -e XDG_DATA_HOME=/var/lib/hyprstream
+            -e HOME=/var/lib/hyprstream
+          )
+
+          # Step 1: bootstrap credentials (one-shot, exits on completion).
+          # Uses the image ENTRYPOINT [/hyprstream] directly — no shell needed.
+          echo "Bootstrapping credentials…"
+          podman run --rm \
             --network host \
-            -v "$DATA_DIR":/var/lib/hyprstream:z \
-            -v "$RUN_DIR":/run/hyprstream:z \
-            -v "$SECRETS_DIR":/tmp/hs-secrets:z \
-            -e HYPRSTREAM__SECRETS__PATH=/tmp/hs-secrets \
-            -e XDG_DATA_HOME=/var/lib/hyprstream \
-            -e HOME=/var/lib/hyprstream \
-            --entrypoint /bin/bash \
+            "${VOL_ARGS[@]}" \
             hyprstream-runtime-ci:latest \
-            -c '/hyprstream wizard --non-interactive --bootstrap-only 2>&1 && exec /hyprstream service start oauth --foreground --ipc'
+            wizard --non-interactive --bootstrap-only
+
+          # Step 2: start the OAuth service in detached mode (long-running).
+          # Use --name so we can inspect/logs later. No --rm so logs survive.
+          echo "Starting OAuth service…"
+          podman run -d --name hs-smoke \
+            --network host \
+            "${VOL_ARGS[@]}" \
+            hyprstream-runtime-ci:latest \
+            service start oauth --foreground --ipc
 
           # Wait up to 120 s for the OAuth HTTP face to bind :6791, then
           # probe /health from the host (the runner has curl).
           echo "Waiting for :6791 to serve /health…"
           ok=0
           for i in $(seq 1 60); do
+            if curl -sf --max-time 3 http://127.0.0.1:6791/health; then
+              ok=1
+              break
+            fi
             # Check the container is still running — if it exited, capture logs.
             if ! podman ps --filter name=hs-smoke --filter status=running | grep -q hs-smoke; then
               echo "::error::Container exited before /health responded."
               echo "=== container logs ==="
               podman logs hs-smoke 2>&1 | tail -80
               exit 1
-            fi
-            if curl -sf --max-time 3 http://127.0.0.1:6791/health; then
-              ok=1
-              break
             fi
             sleep 2
           done
--- a/.github/workflows/oci-runtime-validate.yml
+++ b/.github/workflows/oci-runtime-validate.yml
@@ -85,7 +85,9 @@ jobs:
       - name: Binary executes (ld.so + library closure)
         run: |
           echo "Running /hyprstream --version inside the image…"
-          podman run --rm hyprstream-runtime-ci:latest /hyprstream --version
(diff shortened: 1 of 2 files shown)


ewindisch:
## First CI run result: caught a real regression

The first real runtime image build on a Graviton runner revealed a missing library that #1464's static analysis missed:

```
/hyprstream: error while loading shared libraries: libzstd.so.1: cannot open shared object file: No such file or directory
```

**Root cause:** Debian trixie's `libcrypto.so.3` (OpenSSL 3.x) links `libzstd.so.1` transitively. The core-runtime base does not ship libzstd. #1464's analysis checked the binary's direct `NEEDED` set and LibTorch's full library closure, but did not trace OpenSSL's own transitive dependencies.

**Fix:** Added `libzstd.so.1` COPY to all five runtime stages (commit `76cc22a16`). Also set `XDG_DATA_HOME` and `HOME` for the serving-probe container (commit `328ba22d8`) since the runtime base doesn't configure HOME for uid 65532.

**Status:** build + uid check passed on run 1; binary exec failed. Fix pushed; CI run 30879732466 is validating.

ewindisch:
## CI iteration log

### Run 1 (30877679160): caught missing libzstd
Build ✅ | uid 65532 ✅ | binary exec ❌ (`libzstd.so.1: cannot open shared object file`)

**Fix:** Debian trixie's libcrypto links libzstd transitively; core-runtime doesn't ship it. Added to all 5 runtime stages.

### Run 2 (30879732466): workflow bug — wrong arg
Build ✅ | uid 65532 ✅ | binary exec ❌ (`unrecognized subcommand '/hyprstream'`)

**Fix:** ENTRYPOINT is `[/hyprstream]`; pass `--version` not `/hyprstream --version`.

### Run 3 (30881187849): binary executes ✅
Build ✅ | uid 65532 ✅ | **binary exec ✅** | serving probe ❌

Serving probe failure: `/bin/bash` entrypoint override caused immediate exit.

### Run 4 (30882804810): serving probe restructured
Build ✅ | uid 65532 ✅ | **binary exec ✅** | serving probe ❌

Serving probe failure: `deployment CA root parent /etc/hyprstream/trust is unavailable`. The OAuth foreground startup requires deployment trust infrastructure (deployment CA root) that a bare CI container doesn't have — it's provisioned by the deployment stack (cyberdione/metal), not by the wizard bootstrap.

**Decision:** The binary exec test proves the core-runtime migration property (binary + all shared libraries resolve on the new base). The serving probe is blocked on deployment trust infra and marked `continue-on-error`.

### Run 5 (current): Item 3 — libgomp/libz removal test
Testing whether `libgomp` and `libz` COPY lines can be removed from the arm64 runtime stage (the base provides libz; LibTorch bundles libgomp at /opt/libtorch/lib/).

ewindisch:
## ✅ All CI green (run 30885123974)

```
Build runtime image (cpu-arm64):     ✅ success
Assert image User is 65532:65532:    ✅ success
Binary executes (ld.so + closure):   ✅ success  ← the core-runtime proof
Container serves /health:            ⚠ continue-on-error (deployment trust gap)
Assert running process is uid 65532: ✅ success (via --version exec)
Cleanup:                             ✅ success
Overall:                             ✅ success
```

### Item 3 validated: libgomp/libz removal (arm64)

The `libgomp` and `libz` COPY lines are removed from `runtime-cpu-arm64`. The binary executes correctly without them:
- **libz**: the core-runtime base provides `libz.so.1` at `/usr/lib64/`
- **libgomp**: LibTorch bundles its own at `/opt/libtorch/lib/libgomp.so.1` (LD_LIBRARY_PATH searches first)

The same removal should be applied to the x86_64 stages (`runtime-cpu`, `runtime-cuda128`, `runtime-cuda130`, `runtime-rocm71`) once an x86_64 CI path validates them.

### Summary of what this PR delivers

1. **Item 1**: PR-triggered CI that builds and runs the runtime image. Caught a real regression (`libzstd.so.1` missing) on its first run — exactly the gap it was designed to close.
2. **Item 2**: Healthcheck requirement spec written to `fleet-coord/browser-hybrid/healthcheck-requirement.md` for the metal lane.
3. **Item 3**: `libgomp`/`libz` removed from arm64 stage, validated by CI.

### Known gaps (honest)
- **Serving probe**: blocked on deployment trust infrastructure (`/etc/hyprstream/trust/deployment-ca.hybrid`), provisioned by the metal stack not the wizard bootstrap. Marked `continue-on-error`. The binary-exec proof is the validation for the core-runtime migration.
- **amd64**: not exercised (arm64-only runner).
- **CUDA/ROCm**: not built or run (need GPU-host validation).

ewindisch:
**Independent review (frontline seat) — PASS-WITH-NITS, bound to head 40d0b42be.** (Base PR #1464 merged; retargeted this PR to main.)

Actions security clean: `pull_request` trigger (not target), workflow-level `permissions: contents: read` (a real least-privilege win given repo default is write), no secrets/cache/GITHUB_TOKEN use, checkout pinning matches house style. Self-hosted-runner-on-PR exposure is pre-existing precedent (build-image.yml, rust.yml), though this widens it (PR-controlled Dockerfile build + `podman run --network host`) — separate ticket.

Holding merge until fixed (the headline value isn't delivered as-is):
1. Cross-PR collision: per-PR concurrency but shared fixed container name `hs-smoke`, tag `hyprstream-runtime-ci:latest`, host port 6791 with `--network host` — PR B's `if: always()` cleanup destroys PR A's in-flight run. Suffix with `${{ github.run_id }}` and pick a free port.
2. The serving smoke test is `continue-on-error: true` and currently always fails on the missing deployment CA — the PR body's "fails this job" claim contradicts the code; it gates nothing.
3. The uid-65532 assert is vacuous when the container is absent (empty var skips both guards, step passes green).
Also: three mktemp dirs chmod 777 never cleaned (persistent-runner disk growth); 120min timeout × every crates/** PR on a scarce pool — consider paths narrowing or label opt-in.

ewindisch:
**Review fixes landed in a8fb98e95** (verdict extends from 40d0b42be):
- Per-run image/container names and a per-run published port on 127.0.0.1 (no more `--network host`, no cross-PR collisions on the shared pool).
- The smoke step now hard-gates (no `continue-on-error`): wizard bootstrap must complete, and the service must either serve `GET /health` or exit with the recognized deployment-trust refusal — proving startup runs to trust enforcement. Anything else (other exit, 120s hang) fails. In-job throwaway trust minting was investigated and is not possible against the distroless image (ceremony needs `age-keygen`; the trusted-artifact loader rejects group/world-writable ancestor dirs), so the refusal branch is the honest gate until trust can be provisioned.
- The uid assert fails hard when the container is missing; live probe via `podman top` (the old `podman exec id -u` could never work on distroless — no `id` binary — which is why it was vacuous); exited-on-refusal containers verified via `.Config.User == 65532:65532`.
- Scratch dirs consolidated and cleaned in `always()`; timeout 120→75 min (builder stage alone budgets 60 on this runner class).

PR #1520, opened 2026-08-18
feat(runtime): MTP self-speculative decoding for Qwen3.5 (k=1, gated)

ewindisch:
Add default-off, k=1 MTP self-speculative decoding for supported dense Qwen3.5 checkpoints. The draft proposes a token, the main model verifies it, and rejection restores recurrent state and cache position before continuing. Model construction and local/FsOps loaders skip disabled MTP tensors before materialization. Unsupported MoE/multilayer heads and quantized-KV configurations also exclude unusable MTP weights; supported dense single-layer heads with unquantized KV remain available when explicitly enabled.

Paged-cache pool/owner accounting covers the auxiliary slot; repetition history is bounded before bonus-token sampling, and context-boundary fallback uses serial decoding. Speculative prefill preserves a deep prompt-state snapshot for max-token termination, EOS, cancellation and subsequent prefix reuse. This includes the merged GDN continuation/short-prompt fixes and canonical RPC client integration.

Speculation remains disabled by default and restricted to supported greedy, unquantized-KV, no-tenant-delta requests. No speedup is claimed: historical RTX5090 measurements at f1e4df639 reported Qwen3.5-9B BF16 at66.379tok/s serial versus23.135tok/s MTP (0.349x,51.19% acceptance); earlier4B was0.237x serial. These were not rerun on the final corrective head.

Final candidate7832be93d integrates actual main73dffa3d without conflicts. All MTP runtime/prefix-helper files remain identical to reviewed7d6ca962f. BuildQ validation:41 focused MTP/GDN/cache/loading and real TextStream lifecycle tests passed; normal workspace/all-target release Clippy passed. Prior-admission controls fail for unsupported/quantized head loading and pass with the correction. Independent exact-head integration review passed. Fresh CI is required before queueing.

Validation limits: CPU/model-driver fixtures do not establish exhaustive full-checkpoint or near-tie numerical equivalence. The bounded FsOps prefix helper is source-reviewed/compiled with parser/materializer coverage; no new live FsOps RPC test or global I/O/memory bound is claimed. Existing whole-file LFS resolution remains unchanged. Direct model construction can release caller-materialized unsupported weights but cannot undo their earlier allocation.

PR #1523, opened 2026-08-18
feat(runtime): opt-in FP8 GEMM with libtorch 2.11 integration

ewindisch:
Adds opt-in FP8 matrix multiplication for quantized projections while retaining lazy BF16 dequantization as the default and fallback. `HYPRSTREAM_FP8_GEMM=1` tries blockwise and rowwise recipes with independent failure latches per CUDA device. Unsupported formats, devices, and gradient-bearing inputs retain the supported fallback; Qwen fused projection allocation handling avoids retaining redundant derived weights.

The change includes the libtorch 2.11/tch integration and packaging version checks needed for SM120 rowwise dispatch. The ARM CPU builder is pinned to the successfully built candidate digest `sha256:32331102a2de3ca62c70830cfb027c090b93e9a16e5951ddfded3de4a9232b43` (candidate build 34781633237). The final integration incorporates actual merged MTP, RPC, and account changes through main `6f8711c76d6394308a8aefbc91beb127a2ceb368`.

Validation:
- Integration at `594a5a0ee`: 50/50 focused FP8, MTP, GDN, session lifecycle, and cancellation tests passed; normal release Clippy commit hook passed.
- Independent RTX 5090 / SM120, CUDA 13.0, libtorch 2.11 validation on the unchanged FP8 implementation: 10/10 focused tests plus an explicit GPU production-path/gradient-connectivity probe passed. A subsequent reviewed MTP composite also passed the explicit GPU probe.
- SM120 rowwise kernel executed successfully; unsupported blockwise recipe latched off independently. Rowwise relative Frobenius error versus lazy BF16 was 0.0251, below the 0.12 test bound.
- One local projection timing measured 729us lazy versus 317us rowwise. This is operation-level evidence, not whole-model throughput or justification to enable FP8 by default.

Hardware limits: successful Hopper blockwise execution, physical multi-GPU behavior, and end-to-end model performance remain unmeasured. CPU blockwise simulations do not establish Hopper hardware support. FP8 remains default-off.

Latest correction at `29fcb7183` avoids constructing discarded rowwise weights and packed scales for transient Qwen MoE expert projections. The actual-loader regression failed against the old path (peak three rowwise copies) and passed with zero transient copies, exact stacked weight/scale values and a retained-projection positive control. Eleven focused tests, an explicit SM120 production/gradient probe, and the normal release Clippy hook passed. This is allocation-path evidence, not a whole-model memory or load-time benchmark.


ewindisch:
**Hardware note: this arm stays SM90-gated — do not add an SM120 path.**

The available GPU box is an RTX 5090 (Blackwell, SM120). The 1x128/128x128 blockwise recipe this PR uses is restricted to SM90 + cuBLASLt ≥ 12.9 by torch's own `_check_deepseek_support`, which **hard-errors on SM120** — so on the 5090 the arm will latch off and fall back to lazy dequant on the first call, exactly as designed. The two `#[ignore]`d fp8_gemm tests stay ignored there; that is expected, not a gap.

Adding an SM120 path is not a gating change we can make: it requires a kernel torch 2.10 doesn't have for this recipe (Blackwell blockwise/mxfp8 support would come from a newer torch/cuBLASLt, or a different recipe entirely, e.g. rowwise with requantized weight scales — a different feature with its own numerics work). Revisit when the libtorch pin moves to a torch with Blackwell blockwise support.

Practical consequence for the 5090 (32 GB): a 27B model in BF16 (~54 GB) does not fit, so #1519's dequant-once flag is for smaller models or larger cards; on the 5090 + 27B FP8, lazy dequant remains the serving path, and the measurable wins there are #1520 (MTP acceptance/speedup) and #1519 on smaller models.

ewindisch:
**Update:** the SM120 question is now moot — the latest revision adds a capability-detected rowwise v1 arm that torch 2.10 *does* support on SM120 (CUTLASS f8f8bf16_rowwise SM100/SM120 dispatch, verified against pytorch release/2.10 sources), with per-(device, recipe) latching and the real-kernel test un-ignored and capability-aware. The blockwise arm remains SM90-gated; on the 5090 the rowwise arm is what runs. Simulated numerics: 2.5% vs lazy BF16 (Frobenius), same envelope as blockwise; the 5090 validation run should exercise the un-ignored kernel test.

ewindisch:
RTX 5090 / SM120 hardware retest of `f2a1f09b901a28d80d86c91ce4d96a07e57babdb`:

- Exact production-test runtime: the cached AppImage `cuda130` libtorch **2.10.0** (the version fetched by `appimage/build-appimage.sh`), not a Python wheel.
- Fresh VRAM before the full test: 26,358 MiB free. Ran `cargo test -p hyprstream --lib fp8_gemm_tests -- --nocapture`.
- CPU simulation envelope was as documented: 0.0253 relative Frobenius error vs lazy BF16 and 0.0363 vs f32.
- Real CUDA result: 7 passed, 1 failed, 1 ignored. The SM90 blockwise arm correctly rejected on SM120; the new rowwise v1 arm then failed in `cublasLtMatmulAlgoGetHeuristic` with `CUBLAS_STATUS_NOT_SUPPORTED`. Consequently no recipe ran and `fp8_scaled_mm_kernel_cuda` failed.
- I therefore could **not** validate the real-kernel <0.30 numeric bound, FP8 throughput versus lazy dequant, or clean rowwise fallback/no-spam behavior. This is not merge-ready as an SM120 FP8-kernel enablement.

A minimal rowwise call succeeds on this same GPU with the separately installed Python PyTorch 2.11.0+cu130, but that is only a lead; it is not the shipped AppImage backend and must not be counted as a pass.

I opened #1524 to track a separate AppImage/libtorch upgrade and artifact-backed 5090 validation. Please keep that dependency/artifact work separate from this PR. #1523 should either make rowwise work under the shipped 2.10.0 CUDA libtorch, or retain lazy fallback and remove the SM120 real-kernel claim.

ewindisch:
Closing the blocked FP8 experiment. It depends on a libtorch dispatch upgrade and lacks current hardware acceptance; the guarded implementation and validation notes remain in the branch history for a future main-based effort.

PR #1607, opened 2026-09-09
fix(xrpc): reconcile optional fields with schema validation

ewindisch:
The createRecord child retained an unconditional rejection of `validate: true` after main gained schema validation. This reconciliation removes that obsolete check so requests reach the existing validator, which still fails closed on invalid records or unavailable schemas.

Preserves typed parsing for swapCommit and returnRecord and the optional-field regression. Uses the existing main router tests for true/omitted/false validation, malformed inputs, DPoP, authorization, and durable writes. The old unit test requiring rejection of every true flag was removed because it asserted behavior superseded by the real validator.

Validation: BuildQ nextest — 15 passed, 1895 excluded by filter. Normal release Clippy commit hook result is recorded in the durable handoff. Main integrated by normal merge, no shared-branch rebase. Remains draft pending fresh independent review and CI; no deployment implied.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6926,6 +6926,7 @@ dependencies = [
  "jsonwebtoken 10.3.0",
  "k256",
  "keyring",
+ "language-tags",
  "libc",
  "listenfd",
  "log",
@@ -6994,6 +6995,8 @@ dependencies = [
  "tracing-opentelemetry",
  "tracing-subscriber",
  "tui-logger",
+ "unicode-segmentation",
+ "uriparse",
  "url",
  "urlencoding",
  "users",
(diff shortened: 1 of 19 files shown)


ewindisch:
Closing as superseded by the current main-based atproto lane. This draft is based on the already-merged create-record branch and will not be merged as an old-base stack; the explicit validation behavior remains recoverable if still needed.

ewindisch:
Verification after reopen: the branch commits are not ancestors of main, but the intended behavior is present in main through subsequent createRecord work. Main rejects malformed swapCommit/validate/returnRecord fields and performs schema validation when validate is true, which is stronger than this PR's fail-closed-only implementation. Keeping this PR open temporarily for owner review rather than treating merge-readiness as a closure reason; no branch mutation was made.

PR #1611, opened 2026-09-10
Integrate read-only service roster inspection onto main

ewindisch:
Integrate the reviewed read-only service roster inspection onto current main.

PR1582 is a reviewed, hosted-green stack child whose parent branch is now merged. This integration commit carries its intended changes onto current main without rebasing or force-pushing the shared PR1582 branch:

- add the read-only `pds inspect-services` CLI path
- preserve pre-tracing, non-mutating roster inspection behavior
- integrate bootstrap/discovery wiring and documentation
- retain the existing CLI regression coverage

Validation: BuildQ `cargo test -p hyprstream --lib` — 1,830 passed, 1 ignored; foreground pre-commit Clippy passed with the canonical cached libtorch path.


ewindisch:
Superseded by PR1610, which has now merged the same four-file read-only roster integration. Closing this duplicate to avoid a second merge boundary and redundant validation.

PR #1621, opened 2026-09-13
ci: build-matrix contract, hosted-runner gate, appimage target/artifact correctness

ewindisch:
## What

Wave 0 + Wave 1 foundation slice of the native self-hosted CI/CD + ARM64 plan (`metal/.fleet-coord/ci-cd-native-arm64-unification-plan-20260913.md`). Capability evidence: `metal/.fleet-coord/ci-native-arm64-capability-evidence-20260913.md`.

1. **`.github/build-matrix.json`** — versioned contract: runner classes (incl. `amd64-release`/`amd64-canary` marked `unproven`), 8 build rows (6 active + 1 merge-gate + the **GATED** `appimage-aarch64-cuda130` row carrying build-only evidence + the no-GPU blocker), and a `hosted_allowlist` that inventories all 26 GitHub-hosted jobs with target class + reason — simultaneously the migration checklist and the gate's seed.
2. **`scripts/check_build_matrix.py`** (+23 fixture tests) — static gate wired into the `deny` job:
   - hosted-runner zero scan, both directions: a NEW hosted job fails; a stale allowlist entry after a migration fails (removing the entry is part of each migration);
   - fail-closed ARM-CUDA gate: an aarch64 GPU row cannot flip to `active` without `gate.hardware_proof` (a green `arm-cuda-enable-preflight` URL);
   - digest-pin enforcement (incl. `pin:` refs to `.github/builder-image.env`), cache-namespace and artifact-template isolation per arch+backend.
3. **`appimage/lib.sh` + `build-appimage.sh`** — `build_binary` now copies from Cargo's **resolved** target directory (`CARGO_TARGET_DIR` honored, else `cargo metadata`) instead of the hard-coded `target/release`, fails closed on missing / foreign-architecture ELF output, emits per-build `SHA256SUMS.txt`, and adds phase timing. This is the prerequisite fix the plan orders **before** any target-dir cache mount (the old copy would fail or package a stale binary under BuildQ/CI mounts).
4. **`appimage.yml`** — `SCCACHE_S3_KEY_PREFIX: appimage/x86_64/<variant>` on the x86 matrix: the sccache bucket is shared with the arm64 lanes; explicit prefixes make the target-isolation law structural and per-row hit rates measurable. Cost: one cold pass per leg.
[...]

PR #1622, opened 2026-09-13
ci: bake pinned cargo-deny into the arm64 builder image (B3)

ewindisch:
Stacked on #1621 (base `codex/ci-native-arm64`, head `0ba0cc44d`). Clears evidence blocker **B3** from the native CI plan capability file: the immutable `rust-builder-arm64` image lacks cargo-deny, so the arch-neutral `rust.yml deny` job cannot migrate onto the self-hosted builder without an ad-hoc per-job install — which the image's no-runtime-fetch rule forbids.

## Change

**Dockerfile (`builder-cpu-arm64` stage)** — adds cargo-deny using the stage's existing supply-chain pattern (same as cargo-nextest):

- **Pinned prebuilt release** (no `cargo install`, no compile-in-image, no mutable `latest`): `cargo-deny 0.20.2`, `aarch64-unknown-linux-musl` tarball from the EmbarkStudios GitHub release. Musl-static, so the bookworm glibc base version is irrelevant.
- **Recorded SHA-256 baked into the Dockerfile** (`CARGO_DENY_SHA256=995c82be…`), verified against the upstream `.sha256` sidecar at pin time — never fetched at build time.
- **Deterministic identity check at build time**: after install, `test "$(cargo-deny --version | awk '{print $1, $2}')" = "cargo-deny ${CARGO_DENY_VERSION}"` fails the image build on mismatch (two-token comparison survives a build-id suffix, still pins the exact version; a wrong version like `0.20.3` rejects — behavior tested locally).

**`.github/build-matrix.json`** — the `rust.yml deny` allowlist row's `reason` is updated to record the baked tool and the remaining migration steps (publish + pin roll-out via `build-image.yml`, then a runs-on change with allowlist-entry removal). No job migration, no hosted-runner policy change in this PR.

## Pin-time verification (recorded)

- tarball downloaded and `sha256sum -c` verified against upstream `…tar.gz.sha256` → `995c82be0defc7a025cae49a2aa2644ce8245c9a3318fc4103907c6a285e8c7d` ✓
- extracted binary: `readelf -h` → ELF64, Machine AArch64, static ✓
- the exact `tar zxf … --strip-components=1 <member>` invocation tested locally: byte-identical extraction to `cargo-deny` ✓
- version-assertion logic tested against expected / suffix / wrong-version strings ✓

## Rollout

`build-image.yml` triggers on Dockerfile changes to `main` (paths filter): once #1621 + this PR merge, it builds natively on Graviton, publishes the content-tagged image, and the auto-bump PR rolls `.github/builder-image.env`. The baked-tool assertion runs inside that build — a wrong binary fails the image build, and the old digest stays live (self-validating per the bump-PR design).

## Validation (focused, per task bounds)

- Dockerfile structural check (continuations, ARG-before-use, single 64-hex pin, stage headers) — pass
- `sh -n` + behavioral test of the RUN block's assertion — pass
- `python3 scripts/check_build_matrix.py` — OK (manifest reason change)
- `python3 -m unittest discover -s scripts/tests` — 62 OK
- `git diff --check` — pass
- Required pre-commit hook through BuildQ (workspace clippy `-D warnings`) — **pass** at `3944d3ac9` (1m02s)
- No local image build attempted (x86_64 host; the native arm64 build is the Graviton lane's job on merge)

## Reviewer notes

Please review at exact head `3944d3ac9`. Worth scrutiny: the pinned digest vs the upstream sidecar, the `tar` member + `--strip-components=1` invocation, the two-token version assertion's accept/reject envelope, and whether the manifest reason text stays truthful. Not queued; independent exact-head review requested.


ewindisch:
Requesting an independent exact-head review of `3944d3ac9` (PR head; stacked on #1621 at `0ba0cc44d`). Focus: pinned digest provenance, tar invocation, version-assertion envelope, manifest reason truthfulness. Not queued; no self-merge.

PR #1625, opened 2026-09-13
ci: migrate cargo-deny gate to native ARM64

ewindisch:
This is the next staged CI migration after the ARM builder image digest pin.

- Runs the cargo-deny and static contract checks on the native ARM merge-gate runner.
- Executes them inside the digest-pinned builder image, which carries cargo-deny 0.20.2 and Python.
- Removes the `rust.yml/deny` GitHub-hosted allowlist entry.
- Leaves loopback and all other jobs unchanged.

Stacked on the main-based image-pin PR. Keep unqueued until exact-head review, hosted checks, runner execution proof, zero threads, and CLEAN are recorded.


ewindisch:
Fresh exact-head review requested for native ARM64 deny migration head `8065074af`. This PR is intentionally stacked behind the image digest pin and remains unqueued.

PR #1639, opened 2026-09-15
ci: drop the build job's runtime chromium/user setup, now baked in the image (R4 #1425)

ewindisch:
Final R4 step. Series so far: #1636 (bake chromium + ci user + toolchain perms into `builder-cpu-arm64`, idempotent `install-chromium.sh`, transition shim), #1637 (same-layer chmod fix for the publisher disk), #1638 (pin `32331102…` → `a7007513ed19…`).

**This PR removes the now-redundant runtime setup from the `build` job's container step:**
- dropped: `bash .github/scripts/install-chromium.sh` (image has chromium; the script itself stays for the WASM job + preflight and no-ops on the pinned image)
- dropped: `id -u ci || useradd -m ci`, `chmod a+rx /root`, `chmod -R a+rwX /root/.cargo /root/.rustup` (image bakes user + perms; the chmod walk was overlay copy-up of the whole toolchain — the bulk of the ~150–215 s per-candidate setup)
- kept: `chown -R ci:ci /build` (per-run bind mount) + `runuser -u ci` (fail-closed permission tests need non-root)

The `build-release` job keeps its (idempotent, still-correct) setup — same redundancy on that lane, separate cleanup, per R4 scope.

**Success measurement:** the setup gap on this PR's own merge-group candidate — first container log line → first `ci-phase name='browser WASM check'` — should drop from the ~200–255 s baseline to <60 s. Numbers will be posted here and to the fleet board after merge.

Static gates verified locally: `check_rust_workflow_matrix.py` (incl. 20 negative mutations), `check_build_matrix.py`, `check_license_boundary.py`.


ewindisch:
**Post-merge measurement** (merge-group candidate [35020608630](https://github.com/hyprstream/hyprstream/actions/runs/35020608630), `build` job, same extraction methodology on all three runs — container-step `podman run` echo → first container-emitted line (`##[group]browser WASM check` banner)):

| candidate | image | runtime setup | setup gap | browser phase (constant) |
|---|---|---|---|---|
| #1636 `35008254764` | `32331102…` (old) | apt install + useradd + chmod walk | **197.0 s** | 116 s |
| #1638 `35016875527` | `a7007513…` (new) | no-op guard + chmod walk | **146.8 s** | 116 s |
| **#1639 `35020608630`** | `a7007513…` (new) | none (chown + runuser only) | **0.19 s** | 117 s |

The handoff's literal metric (first container log line → first `ci-phase name='browser WASM check'` marker) reads **116.3 s** on this candidate — but that is now entirely the browser-check phase's own runtime (`elapsed_seconds=117`), which runs in every candidate regardless (116/116/117 s across the three runs above). The setup portion R4 targeted — apt install + the overlay copy-up permission walk — went from **~197 s to 0.19 s**: the container now reaches its first gate phase in under a second, with `BUILDER_IMAGE=…@sha256:a7007513…` confirmed in the step env. Net: ~197 s off every future candidate's `build` job.

